PR #57, opened 2022-06-21
feat: add support for `job_options` argument

bautrey37:
In Que [v1.2](https://github.com/que-rb/que/pull/336), it was deprecated to pass top level keyword args in the
`enqueue` method. This is to prepare for Que v2. This causes warnings
such as:

```
Passing job options like (tags) to `JobClass.enqueue` as top level keyword args has been deprecated and will be removed in version 2.0. Please wrap job options in an explicit `job_options` keyword arg instead.
```

These warning logs can cause on a lot of noise on monitoring tools.

The keyword arguments that must be wrapped around  `job_options` are:
`queue`, `priority`, `run_at`, `tags`, and `job_class`.


bautrey37:
@ahayworth
Yes it is. Sorry for the delay. The tests weren't working locally, but I got them to work in Docker. 
I tested in the example app and my work app that the changes are as desired.

bautrey37:
delayed_job [error](https://github.com/open-telemetry/opentelemetry-ruby-contrib/actions/runs/3021796963/jobs/4893264369) from test:
`LoadError: Error loading the 'sqlite3' Active Record adapter. Missing a gem it depends on? sqlite3 is not part of the bundle. Add it to your delayed_job_4.1.gemfile.`

The `sqllite3` gem does exist in the gemfile. 


activerecord [error](https://github.com/open-telemetry/opentelemetry-ruby-contrib/actions/runs/3021796963/jobs/4893262458) from test:
`NoMethodError: undefined method `full_name' for nil:NilClass`

This seems like its supposed to be [fixed](https://github.com/rubygems/rubygems/issues/5422).


No idea why these failures are happening on my PR.

arielvalentin:
@bautrey37 Thank you again for following up. The gem install errors for sqlite3 and nokogiry is happening on our main release branches as well. It's something I have on my list to look into today. 

I will update your branch once I have a fix in.

bautrey37:
All of the CI checks pass on this PR. 

Can I get a review from one of the maintainers?

instrumentation/que/lib/opentelemetry/instrumentation/que/instrumentation.rb
```diff
@@ -10,7 +10,7 @@ module Que
       # The Instrumentation class contains logic to detect and install the Que
       # instrumentation
       class Instrumentation < OpenTelemetry::Instrumentation::Base
-        MINIMUM_VERSION = Gem::Version.new('1.0.0.beta4')
+        MINIMUM_VERSION = Gem::Version.new('1.2.0')
```
arielvalentin, line 13: So that I'm clear here, any users who were using this instrumentation with versions earlier than this will no longer be able to use this instrumentation. I understand the need to upgrade from a beta version but this now skips past 1.0 and 1.1, is that ok?

bautrey37: Yes, so users would need to be on at least que `v1.2`. The changlogs for 1.1 and 1.0.0.beta5 only has bug fixes and improvements, nothing that would seem to break changes between version jumps. 

[v1.0.0.beta5 changelog](https://github.com/que-rb/que/blob/master/CHANGELOG.md#100beta5-2021-12-23)
[v1.1 changelog](https://github.com/que-rb/que/blob/master/CHANGELOG.md#110-2022-02-21)

arielvalentin (approved):

bautrey37:
Thanks for approving @arielvalentin ! :) 

Since the merge button was not available, I selected `Update Branch`, which now it needs to run CI again. Hopefully, the merge button will appear after that.

PR #59, opened 2022-06-21
feat: add truffleruby support

ahayworth:
...and also add back tests for propagators :scream:

But this is essentially the instrumentation component of https://github.com/open-telemetry/opentelemetry-ruby/pull/1295 but ported to this repo.

PR #60, opened 2022-06-22
Fix: make sidekiq instrumentation rake task compatible with TruffleRuby

alexcameron89:
## Context

This PR starts to fix issues related to supporting TruffleRuby (added in https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/59) by fixing the Sidekiq implementation test issues.

This should fix https://github.com/open-telemetry/opentelemetry-ruby-contrib/issues/46

### Fixing Sidekiq Instrumentation issues in testing

This commit fixes the following error from the Rakefile due to using `fork` for starting
redis:

```
NotImplementedError: fork is not available
```

According to the TruffleRuby compatibility docs, `fork` is not compatible in TruffleRuby:

> You cannot fork the TruffleRuby interpreter. The feature is unlikely to ever be supported when running on the JVM but could be supported in the future in the native configuration. 

After fixing that issue, I was able to successfully run a passing test suite for the Sidekiq implementation.


alexcameron89:
It looks like the TruffleRuby tests are also passing in the CI workflow, so PR is ready for review! 👍

arielvalentin (approved):
Thanks!

PR #64, opened 2022-06-24
fix: suppress invalid span attribute value type warning in aws-sdk instrumentation

### LINKED ISSUE #63: OpenTelemetry warning: invalid span attribute value type NilClass for key 'messaging.url' on span 'SQS.GetQueueUrl'
**Description of the bug**

Observing the following warning:

> OpenTelemetry warning: invalid span attribute value type NilClass for key 'messaging.url' on span 'SQS.GetQueueUrl'

Looks like the `queue_url` param is not passed to the `get_queue_url` method:
https://docs.aws.amazon.com/sdk-for-ruby/v3/api/Aws/SQS/Client.html#get_queue_url-instance_method

Probably this line needs to be changed:
https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/f4cd86c44aa10550923500ac3264f8485946e59c/instrumentation/aws_sdk/lib/opentelemetry/instrumentation/aws_sdk/messaging_helper.rb#L35 

to something like:
```rb
 attributes[SemanticConventions::Trace::MESSAGING_URL] = context.params[:queue_url] if context.params[:queue_url] 
```

yoheyk:
fixes #63


plantfansam (commented):
Thanks for this @yoheyk! Will merge if specs come back green.

arielvalentin (approved):

Issue #42, opened 2022-06-06
Make actionpack instrumentation compatible with TruffleRuby


plantfansam:
We support TruffleRuby on a best effort basis (see https://github.com/open-telemetry/opentelemetry-ruby/pull/1295). Our actionpack instrumentation does not work properly with TruffleRuby, however.

This ticket will be complete when our CI build successfully runs specs `opentelemetry-instrumentation-action_pack`.

To run these specs in CI, delete the gem from the TruffleRuby filter stanzas ([1](https://github.com/open-telemetry/opentelemetry-ruby/pull/1295/files#diff-b803fcb7f17ed9235f1e5cb1fcd2f5d3b2838429d4368ae4c57ce4436577f03fR113), [2](https://github.com/open-telemetry/opentelemetry-ruby/pull/1295/files#diff-b803fcb7f17ed9235f1e5cb1fcd2f5d3b2838429d4368ae4c57ce4436577f03fR275)) in `.github/workflows/ci.yml`.

alexcameron89:
The issue looks to be with Bundler, and here's the failure [from a CI run](https://github.com/open-telemetry/opentelemetry-ruby-contrib/runs/7050619525?check_suite_focus=true) when running `bundle exec appraisal install`:
```
NoMethodError: undefined method `full_name' for nil:NilClass
  /home/runner/.rubies/truffleruby-22.1.0/lib/gems/gems/bundler-2.3.16/lib/bundler/installer/parallel_installer.rb:124:in `block (2 levels) in check_for_unmet_dependencies'
  /home/runner/.rubies/truffleruby-22.1.0/lib/gems/gems/bundler-2.3.16/lib/bundler/installer/parallel_installer.rb:123:in `each'
  /home/runner/.rubies/truffleruby-22.1.0/lib/gems/gems/bundler-2.3.16/lib/bundler/installer/parallel_installer.rb:123:in `block in check_for_unmet_dependencies'
  /home/runner/.rubies/truffleruby-22.1.0/lib/gems/gems/bundler-2.3.16/lib/bundler/installer/parallel_installer.rb:122:in `each'
  /home/runner/.rubies/truffleruby-22.1.0/lib/gems/gems/bundler-2.3.16/lib/bundler/installer/parallel_installer.rb:122:in `check_for_unmet_dependencies'
[...]

PR #67, opened 2022-06-25
Fix additional incompatibilities with TruffleRuby

### LINKED ISSUE #44: Make redis instrumentation compatible with TruffleRuby
We support Truffleruby on a best effort basis (see https://github.com/open-telemetry/opentelemetry-ruby/pull/1295). Our redis instrumentation does not work properly with truffleruby, however.

This ticket will be complete when our CI build successfully runs specs for `opentelemetry-instrumentation-redis`.

To run these specs in CI, delete the gem from the TruffleRuby filter stanzas ([1](https://github.com/open-telemetry/opentelemetry-ruby/pull/1295/files#diff-b803fcb7f17ed9235f1e5cb1fcd2f5d3b2838429d4368ae4c57ce4436577f03fR113), [2](https://github.com/open-telemetry/opentelemetry-ruby/pull/1295/files#diff-b803fcb7f17ed9235f1e5cb1fcd2f5d3b2838429d4368ae4c57ce4436577f03fR275)) in `.github/workflows/ci.yml`.

alexcameron89:
## Context
This is similar to changes in https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/60 that replace usage of `fork` with usage of `spawn`. According to the TruffleRuby docs:

> **You cannot fork the TruffleRuby interpreter.** The feature is unlikely to ever be supported when running on the JVM but could be supported in the future in the native configuration.

This PR fixes #44 and possibly #42 (I can't reproduce issues with the rails gems, so I want to see if errors arise in CI).

### Question: Is it better to open separate PR's for each fix or group them?
Would you prefer that I fix each gem in its own PR or is it okay to group them into one PR like this?


ahayworth:
> Would you prefer that I fix each gem in its own PR or is it okay to group them into one PR like this?

I think for circumstances like this, where the fixes are relatively trivial and small, grouping in one PR is fine. 😄 

Also, thank you!

ahayworth (approved):
LGTM! Thank you so much!

PR #69, opened 2022-06-26
fix: Ensure untraced blocks suppress spans

arielvalentin:
Prior to this change, the Net::HTTP instrumentation would defer to the sampling strategy
to determine if it should generate a span, however the sampling strategy
is not aware of the `untraced` block and it will always generate a span
even though the parent span is untraced.

The question here is... should the `untraced` block override the logic in the sampling strategy?

Also... a side effect of checking that the `current_span.recording?` is that the instrumentation
will never be allowed to generate a root span.

Related to https://github.com/open-telemetry/opentelemetry-ruby/discussions/1333

### DIFF
--- a/instrumentation/net_http/lib/opentelemetry/instrumentation/net/http/patches/instrumentation.rb
+++ b/instrumentation/net_http/lib/opentelemetry/instrumentation/net/http/patches/instrumentation.rb
@@ -16,8 +16,7 @@ module OpenTelemetry
 
             def request(req, body = nil, &block) # rubocop:disable Metrics/AbcSize
               # Do not trace recursive call for starting the connection
-              return super(req, body, &block) unless started?
-              return super(req, body, &block) unless OpenTelemetry::Trace.current_span.recording?
+              return super(req, body, &block) unless started? && OpenTelemetry::Trace.current_span.recording?
 
               attributes = {
                 OpenTelemetry::SemanticConventions::Trace::HTTP_METHOD => req.method,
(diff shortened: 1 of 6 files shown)


instrumentation/net_http/lib/opentelemetry/instrumentation/net/http/patches/instrumentation.rb
```diff
@@ -17,6 +17,7 @@ module Instrumentation
             def request(req, body = nil, &block)
               # Do not trace recursive call for starting the connection
               return super(req, body, &block) unless started?
+              return super(req, body, &block) unless OpenTelemetry::Trace.current_span.recording?
```
ahayworth, line 20: Should this be `unless started? && OpenTelemetry::Trace.current_span.recording?` ?

arielvalentin: well shoot... I think I got it even more wrong... `unless started? && !OpenTelemetry::Trace.current_span.recording?`

arielvalentin: Demorgans law, a fried brain, and inverted logic are killing me here:

```ruby
if started? && recording?
  # trace
else 
  # don't trace
end
```

arielvalentin: Ok, so now this code relies on the fact that there is an active recording span. That means it cannot be used in a standalone instrumentation _without_ wrapping it in  parent recording span. 

Run the example... and there are no exported spans.

```console
@arielvalentin ➜ .../opentelemetry-ruby-contrib/instrumentation/net_http/example (untraced-net-http) $ bundle exec ruby net_http.rb 
I, [2022-06-28T01:43:00.394156 #6728]  INFO -- : Instrumentation: OpenTelemetry::Instrumentation::Net::HTTP was successfully installed with the following options {}
```

PR #72, opened 2022-06-28
feat: instrumentation for racecar

chrisholmes:
This PR implements https://github.com/open-telemetry/opentelemetry-ruby-contrib/issues/15

Racecar is built on top of rdkafka, but it doesn't use the API that is implemented by the rdkafka instrumentation so an racecar specific adapter is required. 

Racecar's [main loop](https://github.com/zendesk/racecar/blob/73f424f36324f4161f820f6558097a704cbb3316/lib/racecar/runner.rb#L66-L80) repeatedly calls a rdkafka consumer's `#poll` method and then sends the returned messages to its own consumer classes (aka processors).

This instrumentation works by patching the Runner class and then its processor (Racecar::Consumer objects) at initialization.


plantfansam:
Thanks @chrisholmes! Are you set up with a tracing UI locally? If so, would it be possible to include a screenshot of a `racecar` trace? This isn't part of our PR template (yet!), but it might be nice to see what this looks like!

chrisholmes:
@plantfansam yep, makes sense. Here's the view in jaeger:

image

arielvalentin (commented):
Thanks again for this submission! 

I am curious about the choice to mixin the instrumentation separately of the processor object vs mixing into the built in instrumenter[^1] or decorating `process`, and `process_batch`.

Would you mind sharing your thought process behind the overall design? 

[^1]: https://github.com/zendesk/racecar/blob/master/lib/racecar/instrumenter.rb

chrisholmes:
> Thanks again for this submission!
> 
> I am curious about the choice to mixin the instrumentation separately of the processor object vs mixing into the built in instrumenter[1](#user-content-fn-1-5f7e614a36a1a95ce99b518b507ebc18) or decorating `process`, and `process_batch`.
> 
> Would you mind sharing your thought process behind the overall design?
> 
> ## Footnotes
> 1. https://github.com/zendesk/racecar/blob/master/lib/racecar/instrumenter.rb [↩](#user-content-fnref-1-5f7e614a36a1a95ce99b518b507ebc18)

I've tried a few different ways, but this form (though it is quite ugly) is the only format I can get the tests to pass so its spans are consistent to the other kafka instrumentations. 

This is my understanding of the impact of the methods you've mentioned:

When patching `process` and `process_batch` on `Racecar::Runner` the tests will pass except that exceptions are not propagated up to the Otel layer when racecar has its `pause` functionality enabled. Frustratingly, [`with_pause`](https://github.com/zendesk/racecar/blob/7a4bbcfbd173d4ce0323e3aae7acaca58ab4ed29/lib/racecar/runner.rb#L191) doesn't propagate the exception it rescues.

As far as I can tell, it is not fully possible to adapt Racecar's instrumentation interface as the `process_batch` payload doesn't have access to the batch of messages or their headers so it is not possible to add links to the traces. The `process_message` payload does have access to the message headers so we could use it there.

What do you think about this? Is it worth omitting batch message links or errors for a simpler looking instrumentation?

instrumentation/racecar/lib/opentelemetry/instrumentation/racecar/instrumentation.rb
```diff
@@ -0,0 +1,42 @@
...
+
+module OpenTelemetry
+  module Instrumentation
+    module Racecar
+      # The Instrumentation class contains logic to detect and install the Racecar instrumentation
+      class Instrumentation < OpenTelemetry::Instrumentation::Base
+        MINIMUM_VERSION = Gem::Version.new('2.0')
+
+        compatible do
+          Gem.loaded_specs['racecar'].version >= MINIMUM_VERSION
```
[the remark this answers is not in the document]
chrisholmes: sure, let me know if I'd done what you'd expect.

ericmustin:
assigned myself to this so i can remember to review while ariel is out, apols on the delay here chris.

instrumentation/all/Gemfile
```diff
@@ -7,3 +7,5 @@
 source 'https://rubygems.org'
 
 gemspec
+
+gem 'opentelemetry-instrumentation-racecar', path: '../racecar'
```
fbogsany, line 11: Why is this added here? It isn't necessary for the other instrumentation gems.

instrumentation/racecar/example/Gemfile
```diff
@@ -0,0 +1,17 @@
...
+# SPDX-License-Identifier: Apache-2.0
+
+source 'https://rubygems.org'
+
+# DO NOT ADD DEPENDENCIES HERE!
+# Please declare a minimum development dependency in the gemspec,
+# then target specific versions in the Appraisals file.
+
+group :development, :test do
+  gem 'opentelemetry-instrumentation-base', path: '../../base'
```
fbogsany, line 14: Is this line necessary for the example?

instrumentation/racecar/example/tracing.rb
```diff
@@ -0,0 +1,34 @@
+# frozen_string_literal: true
+
+require 'opentelemetry/sdk'
+require 'opentelemetry-instrumentation-rdkafka'
```
fbogsany, line 4: Do we need this line?

instrumentation/all/Gemfile
chrisholmes: actually, I'm wrong. This line is required so that the we can add the new instrumentation to all before it has been released. If we remove that line there are build errors as the all gem doesn't have a version of this gem to reference.

instrumentation/racecar/example/tracing.rb
chrisholmes: you're right, we don't need it

instrumentation/racecar/example/Gemfile
chrisholmes: no

instrumentation/racecar/Appraisals
```diff
@@ -0,0 +1,15 @@
...
+
+## TODO: Include the supported version to be tested here.
+## Example:
+# appraise 'rack-2.1' do
+#   gem 'rack', '~> 2.1.2'
+# end
+
+# appraise 'rack-2.0' do
+#   gem 'rack', '2.0.8'
+# end
```
ericmustin, line 15: As noted, this needs to be updated to run appraisals for racecar

instrumentation/racecar/lib/opentelemetry/instrumentation/racecar/instrumentation.rb
```diff
@@ -0,0 +1,42 @@
...
+
+module OpenTelemetry
+  module Instrumentation
+    module Racecar
+      # The Instrumentation class contains logic to detect and install the Racecar instrumentation
+      class Instrumentation < OpenTelemetry::Instrumentation::Base
+        MINIMUM_VERSION = Gem::Version.new('2.0')
+
+        compatible do
+          Gem::Version.new(::Racecar::VERSION) >= MINIMUM_VERSION
```
ericmustin, line 17: Nit but would you mind abstracting this gem version check into a private method `gem_version`, which `compatible` then calls ? This follows the style we've tried to implement in this pr https://github.com/open-telemetry/opentelemetry-ruby-contrib/commit/3b03ff7ea66b69c85ba205a369b85c2c33b712fe

see example https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/26a9b3b7cc6e37f71b29087ee06639aad6cd459a/instrumentation/pg/lib/opentelemetry/instrumentation/pg/instrumentation.rb#L50-L52

instrumentation/racecar/lib/opentelemetry/instrumentation/racecar/instrumentation.rb
```diff
@@ -0,0 +1,42 @@
...
+        end
+
+        install do |_config|
+          OpenTelemetry::Instrumentation::Rdkafka::Instrumentation.instance.install({})
```
ericmustin, line 21: I'm a bit confused here. So is the idea that the racecar instrumentation depends on the rdkafka instrumentation?

1. Is it possible to check if the instrumentation is already installed before attempting to re-install it?
2. What's the desired behavior if the rdkafka instrumentation has been intentionally disabled by the user via env var? Are we essentially overwriting that choice.
2a. Hypothetical since there's no config options at the moment, but won't we automatically overwriting any configuration the user has supplied for their `rdkafka` instrumentation should it be aded in the future?

I guess i have to think a bit more closely about what the "right" behavior should be, but imo we should be respecting if this instrumentation has been disabled and, imo, its not clear to me why we need to take rdkafka as a dependancy.

instrumentation/racecar/opentelemetry-instrumentation-racecar.gemspec
```diff
@@ -0,0 +1,52 @@
...
+
+  spec.files = ::Dir.glob('lib/**/*.rb') +
+               ::Dir.glob('*.md') +
+               ['LICENSE', '.yardopts']
+  spec.require_paths = ['lib']
+  spec.required_ruby_version = '>= 2.6.0'
+
+  spec.add_dependency 'opentelemetry-api', '~> 1.0'
+  spec.add_dependency 'opentelemetry-instrumentation-base', '~> 0.21.0'
+  spec.add_dependency 'opentelemetry-instrumentation-rdkafka', '~> 0.2.0'
```
ericmustin, line 30: why `~> 0.2.0`? This is lower than the min version of our rdkafka instrumentation (0.10.0) https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/26a9b3b7cc6e37f71b29087ee06639aad6cd459a/instrumentation/rdkafka/lib/opentelemetry/instrumentation/rdkafka/instrumentation.rb#L12

Should we be making the rdkafka instrumentation more permissive? As mentioned before, currently the coupling between these two instrumentations feels unclear and maybe a little bit unnecessary

ericmustin (changes requested):
Overall this looks good.

There are a few TODOs you have that still need to be completed,

The only real blocker to me is it's not clear to me why we need to take `rdkafka` as a hard dependency and install the instrumentation, I'd like to understand this better and whether it would be possible to decouple them. If not, I'd like to make sure we respect any env var's that disable the rdkafka instrumentation as well as make sure there's no edge cases in version requirements. 

Nice work as always, apols on the delay on review

instrumentation/racecar/lib/opentelemetry/instrumentation/racecar/instrumentation.rb
chrisholmes: Yes, my intention here was to rely on the Rdkafka implementation of producer spans as adding a Racecar implementation would duplicate them like this:

```
racecar send span:
  rdkafka send span:
    consumer span
```

There is a similar dependency between the Rack instrumentation with Rails and Sinatra so I did assume it would be alright to follow that pattern, but it is being used slightly different here and less extensively.

> 1. Is it possible to check if the instrumentation is already installed before attempting to re-install it?

Yes, there is an interface we could use.

> 2. What's the desired behavior if the rdkafka instrumentation has been intentionally disabled by the user via env var? Are we essentially overwriting that choice.
>    2a. Hypothetical since there's no config options at the moment, but won't we automatically overwriting any configuration the user has supplied for their `rdkafka` instrumentation should it be aded in the future?

Yes, I agree, there probably needs to be some more consideration here. If we don't want to load rdkafka's instrumentation then we should probably load some instrumentation here for producer spans.

As alternatives to this, I can think of the following options:
- Duplicate the spans (there is precedence for this with Faraday and it wrapping of other clients like net/http)
- Have the racecar producer instrumentation used `untraced` so that the rdkafka producer spans are dropped

What do you think?

instrumentation/racecar/lib/opentelemetry/instrumentation/racecar/instrumentation.rb
chrisholmes: done

instrumentation/racecar/opentelemetry-instrumentation-racecar.gemspec
chrisholmes: I'm not sure why it's like this actually. I'll update it depending on the conversation above.

ericmustin:
hey @chrisholmes just ack'ing your followups, plan to catch up here otw. hectic week for me there's been some internal stuff going on at work.

one thing we discussed in the SIG mtg on tuesday was moving this approach to use ASN notifications. I think everyone agreed that if it could be accomplished, it would be preferrable. I'm going to try to see if i can't fix the issues you'd mentioned when you tried the approach, if it is manageable to fix, this would probably be the preferred way forward.

chrisholmes:
> hey @chrisholmes just ack'ing your followups, plan to catch up here otw. hectic week for me there's been some internal stuff going on at work.

thanks @ericmustin!

> one thing we discussed in the SIG mtg on tuesday was moving this approach to use ASN notifications. I think everyone agreed that if it could be accomplished, it would be preferrable. I'm going to try to see if i can't fix the issues you'd mentioned when you tried the approach, if it is manageable to fix, this would probably be the preferred way forward.

Sure, I'm happy for you to try. I should have a working copy somewhere with my experiment too that I can push to a branch though I did it by implementing a customer Racer::Instrumenter rather than hoooking into ASN. For reference, the only real issue I had was with the Racecar's runner not providing [batch message headers to the instrumentation layer](https://github.com/zendesk/racecar/blob/master/lib/racecar/runner.rb#L209). If we're happy to omit links for batch messages then it should be a simply implementation.

arielvalentin:
@chrisholmes Thank you again for your patience and for being willing to try a different approach here. 

How would you feel about submitting an upstream PR to see if @dschierbeck or @bquorning would be willing to include the batch message headers so we are able to use links?

chrisholmes:
@arielvalentin (and @ericmustin), yes, I'm happy to raise a request for this. 

I've been exploring this further and I noticed that using ASN for batch messages runs into the similar issue as described above where the `with_pause` method swallows exceptions before they can get to the instrumentation layer. I expect we will need an additional change to workaround this.

As this might take a while to source out, would you be fine with merging this PR without support for instrumentation of batch message processing? I will push a version with ASN support for single message processing in the meantime.

ericmustin:
@chrisholmes i'd be ok with that. Did not catch up otw, as these things go. will review this evening.

instrumentation/racecar/opentelemetry-instrumentation-racecar.gemspec
```diff
@@ -0,0 +1,52 @@
...
+  spec.homepage    = 'https://github.com/open-telemetry/opentelemetry-ruby-contrib'
+  spec.license     = 'Apache-2.0'
+
+  spec.files = ::Dir.glob('lib/**/*.rb') +
+               ::Dir.glob('*.md') +
+               ['LICENSE', '.yardopts']
+  spec.require_paths = ['lib']
+  spec.required_ruby_version = '>= 2.6.0'
+
+  spec.add_dependency 'activesupport'
```
ericmustin, line 28: nit because it feels like quite the edge case, but i guess i have some misgivings about requiring activesupport explicitly as a dependancy. I'd almost rather just fail the compatibility check if the appropriate active support classes/modules aren't present, ala what active model serializer instrumentation does

https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/5d92b37671c59d8aff86d17e8661c3fc25da6a21/instrumentation/active_model_serializers/lib/opentelemetry/instrumentation/active_model_serializers/instrumentation.rb#L23-L25

ericmustin (approved):
I think this is incredibly close, i'm sorry for all the back and forth here. I had one nit that i'd like to get clarification on (I will ask about this in today's sig mtg and follow up), regarding whats the appropriate way to manage the activesupport dependancy. the prior art in `-contrib` it seems is to not require it as an explicit dependency if possible, and simply include checking for it's existence in a compatibility check. I'm not sure if that's _correct_ though so i'm going to get clarity on that today and follow up. I'm approving this though, with that one caveat.

instrumentation/racecar/lib/opentelemetry/instrumentation/racecar/instrumentation.rb
```diff
@@ -0,0 +1,52 @@
...
+
+        def require_patches
+          require_relative 'patches/consumer'
+        end
+
+        def add_subscribers
+          require 'active_support'
+          require_relative 'process_message_subscriber'
+          subscriber = ProcessMessageSubscriber.new
+          ::ActiveSupport::Notifications.subscribe('process_message.racecar', subscriber)
```
arielvalentin, line 38: @robertlaurin Should this use the OTel Instrumentation Package instead?

```ruby
tracer = :OpenTelemetry::Instrumentation::Racecar.tracer
::OpenTelemetry::Instrumentation::ActiveSupport.subscribe(tracer, process_message.racecar', notification_payload_transform: ->(payload) { #...some logic here 
}) 
```

instrumentation/racecar/opentelemetry-instrumentation-racecar.gemspec
chrisholmes: yes, I was expecting feedback along these lines.

I can't make the SIG, but I will adjust based on the decision there.

chrisholmes: 👍 I'll get this pushed later

chrisholmes: right, should be ready now

ericmustin (approved):

instrumentation/racecar/test/opentelemetry/instrumentation/racecar_test.rb
```diff
@@ -0,0 +1,229 @@
...
+
+  def stop_racecar(thread)
+    thread[:racecar].stop
+    thread.join(60)
+  end
+
+  def wait_for_messages_seen_by_consumer(count)
+    Timeout.timeout(60) do
+      sleep 0.1 until consumer_class.messages_seen.size >= count
+    end
```
arielvalentin, line 64: @open-telemetry/ruby-contrib-approvers, we should keep an eye on this to make sure the tests so we can avoid non-deterministic tests. This may be a good use case for a CountDownLatch.

chrisholmes: Reading this again, I'd like to tidy this up a little too. I'll introduce CountDownLatch and remove some of the duplication.

arielvalentin (approved):

instrumentation/racecar/lib/opentelemetry/instrumentation/racecar/instrumentation.rb
```diff
@@ -0,0 +1,51 @@
...
+        private
+
+        def require_patches
+          require_relative './patches/consumer'
+        end
+
+        def add_subscribers
+          require_relative 'process_message_subscriber'
+          subscriber = ProcessMessageSubscriber.new
+          ::ActiveSupport::Notifications.subscribe('process_message.racecar', subscriber)
```
plantfansam, line 37: Can we link to what emits this?

instrumentation/racecar/lib/opentelemetry/instrumentation/racecar/process_message_subscriber.rb
```diff
@@ -0,0 +1,59 @@
...
+
+module OpenTelemetry
+  module Instrumentation
+    # This class contains the ASN subsciber that instruments message processing
+    class ProcessMessageSubscriber
+      def tracer
+        Racecar::Instrumentation.instance.tracer
+      end
+
+      def start(_name, _id, payload)
```
plantfansam, line 11: I believe we're satisfying the `ActiveSupport::Subscriber` [interface](https://api.rubyonrails.org/classes/ActiveSupport/Subscriber.html) here. If so, can we explicitly say so in a comment? IDK if it makes sense to subclass the `Subscriber` class, but that would perhaps be another way of achieving this.

instrumentation/racecar/lib/opentelemetry/instrumentation/racecar/process_message_subscriber.rb
```diff
@@ -0,0 +1,59 @@
...
+          'messaging.destination_kind' => 'topic',
+          'messaging.kafka.partition' => payload[:partition],
+          'messaging.kafka.offset' => payload[:offset]
+        }
+
+        attributes['messaging.kafka.message_key'] = payload[:key] if payload[:key]
+        attributes
+      end
+
+      def finish(name, id, payload)
```
plantfansam, line 42: I think this instrumentation is adapted from the `ActiveSupport` instrumentation's `SpanSubscriber` ([link](https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/22796cb8141b3100c6ce94c25a15e31df3313464/instrumentation/active_support/lib/opentelemetry/instrumentation/active_support/span_subscriber.rb)), so maybe these questions were answered over there. At the risk of being told to RTFM, are we sure that:

1. `start` and `finish` will be called sequentially, with all context.attach calls matched with a context.detach call.
2. `start` and `finish` will be invoked by the same `Thread`.

If user code attaches to a new context between `start` and `finish`, but neglects to detach from that context, then [`detach`](https://github.com/open-telemetry/opentelemetry-ruby/blob/main/api/lib/opentelemetry/context.rb#L55-L62) will not work as expected. Of course, they'd see an error in their logs, so ... that'll probably help them fix it.

Similarly, if `finish` is called in a different thread than `start`, it'll be using the wrong Context stack, since that's stored on [`Thread.current`](https://github.com/open-telemetry/opentelemetry-ruby/blob/main/api/lib/opentelemetry/context.rb#L127).

I don't know much about ActiveSupport notifications (or Rdkafka! or Racecar!), so I'm most likely being paranoid here, but thought it was worth asking.

plantfansam: @ahayworth might have insight, since he worked on a [similar thing](https://github.com/open-telemetry/opentelemetry-ruby/pull/796).

instrumentation/racecar/lib/opentelemetry/instrumentation/racecar/process_message_subscriber.rb
```diff
@@ -0,0 +1,59 @@
...
+module OpenTelemetry
+  module Instrumentation
+    # This class contains the ASN subsciber that instruments message processing
+    class ProcessMessageSubscriber
+      def tracer
+        Racecar::Instrumentation.instance.tracer
+      end
+
+      def start(_name, _id, payload)
+        attrs = attributes(payload)
```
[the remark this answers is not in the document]
chrisholmes: Unfortunately, this introduces a ABC rubocop failure

instrumentation/racecar/lib/opentelemetry/instrumentation/racecar/process_message_subscriber.rb
chrisholmes: TBH, it's not 100% in manual, but `ActiveSupport::Notifications` will call in that order and in the same thread (they leave to subscribers to decide if async is needed). The code isn't terribly complicated [so you could follow on the source](https://github.com/rails/rails/blob/main/activesupport/lib/active_support/notifications.rb)

instrumentation/racecar/lib/opentelemetry/instrumentation/racecar/process_message_subscriber.rb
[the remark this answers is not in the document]
plantfansam: OK. We can always enable/disable the cop but I don't think it's worth the time 😄

instrumentation/all/opentelemetry-instrumentation-all.gemspec
```diff
@@ -50,6 +50,7 @@ Gem::Specification.new do |spec|
   spec.add_dependency 'opentelemetry-instrumentation-net_http', '~> 0.21.0'
   spec.add_dependency 'opentelemetry-instrumentation-pg', '~> 0.22.0'
   spec.add_dependency 'opentelemetry-instrumentation-que', '~> 0.4.0'
+  spec.add_dependency 'opentelemetry-instrumentation-racecar', '~> 0.1.0'
```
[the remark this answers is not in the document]
chrisholmes: I was considering something like that. Has this happened before?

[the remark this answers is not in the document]
plantfansam: Not sure, but maybe @open-telemetry/ruby-contrib-maintainers might know? I don't think there's any problem with it.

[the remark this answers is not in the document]
arielvalentin: Personally speaking I'm ok with adding it. If folks that use all don't want it they should disable it.

I minor bump with the change log/Release notes should be enough of a signal

plantfansam (approved):
I'd still like to see my open comments resolved at some point, but I don't think it's worth blocking merge. Happy to put this in the wild and see how it fares.

chrisholmes:
@arielvalentin @plantfansam: I've resolved some conflicts and I _think_ that I've switched over to the new CI structure. Please let me know if I've missed anything.

arielvalentin (approved):

PR #73, opened 2022-06-29
fix: Use semantic graphql attribute names

kroehre:
https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/instrumentation/graphql.md


arielvalentin (approved):
🎉 thank you for submitting this contribution!

It looks like the semconv gem is behind. 
https://github.com/open-telemetry/opentelemetry-ruby/tree/main/semantic_conventions

cc: @ahayworth

PR #76, opened 2022-07-02
feat: Add simple opentelemetry-collector config for use in dev

### LINKED ISSUE #74: chore: OTel collector for exploratory testing
Reviewing and and performing exploratory testing PRs would be made easier by being able to visualize the results in a tracing backend. 

This repository could provide contributors with an OTel collector service configured in the [main `docker-compose` file](https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/main/docker-compose.yml) like we do for other services.  

Inspired by: https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/72#issuecomment-1170077416

ahayworth:
Mostly this is borrowed from open-telemetry/opentelemetry-demo-webstore.

This commit adds the opentelemetry-collector and the jaeger all-in-one
image to the docker-compose file. This allows us to export traces in
development to a collector and then visualize them in the Jaeger UI.

Fixes #74

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -109,7 +109,7 @@ To test using Docker:
      2. Run the tests for the sdk
          *  `docker-compose run sdk bundle exec rake test`
 
-## OpenTelemetry Collector
+## Processing and visualizing traces locally
 
 You may wish to test your changes against an [OpenTelemetry collector](https://github.com/open-telemetry/opentelemetry-collector). To facilitate this, we provide configuration in `docker-compose.yml` that should be sufficient.
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -110,6 +110,15 @@ To test using Docker:
      2. Run the tests for the sdk
          *  `docker-compose run sdk bundle exec rake test`
 
+## OpenTelemetry Collector
+
+You may wish to test your changes against an [OpenTelemetry collector](https://github.com/open-telemetry/opentelemetry-collector). To facilitate this, we provide configuration in `docker-compose.yml` that should be sufficient.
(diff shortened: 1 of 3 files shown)


arielvalentin (approved):

CONTRIBUTING.md
```diff
@@ -109,6 +109,15 @@ To test using Docker:
      2. Run the tests for the sdk
          *  `docker-compose run sdk bundle exec rake test`
 
+## OpenTelemetry Collector
```
plantfansam, line 112:
```suggestion
## Processing and visualizing traces locally
```

otelcol-config.yml
```diff
@@ -0,0 +1,21 @@
+receivers:
+  otlp:
+    protocols:
+      grpc:
+      http:
+exporters:
+  jaeger:
+    endpoint: "jaeger:14250"
```
plantfansam, line 8: Is this the correct point? I think I see Jaeger listening on `16686` above, but I haven't worked much in this area...

plantfansam (approved):
Couple nits, but I'm sure I'm missing something. Happy to merge and sort out any confusion in follow-on PRs, as this is just for dev.

otelcol-config.yml
ahayworth: 16686 is the jaeger UI port - this port is for the collector to export to the jaeger storage backend itself.

ahayworth: Oh, also I suppose I should note that the explicit mention of 16686 up there doesn't mean that jaeger is only listening on that port - it's a docker-compose directive saying "please map port 16686 from the container network into the host network." :)

PR #77, opened 2022-07-03
add rack hooks

nozik:
Add support for request_hook and response_hook that allow adding user-customizable attributes to the span. This capability is common across http server instrumentations in other languages (e.g., [FastAPI in Python](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/ff412c6d40837c2332bd93991d207c433076dd22/instrumentation/opentelemetry-instrumentation-fastapi/src/opentelemetry/instrumentation/fastapi/__init__.py#L54)).

### DIFF
--- a/instrumentation/rack/README.md
+++ b/instrumentation/rack/README.md
@@ -49,6 +49,23 @@ OpenTelemetry::SDK.configure do |c|
 end
 ```
 
+### Request/response hooks
+
+You can pass hooks to extract additional attributes from the requests/responses.
+
+```ruby
+OpenTelemetry::SDK.configure do |c|
+  c.use 'OpenTelemetry::Instrumentation::Rack', {
+    request_hook: lambda { |span, env|
+      # Extract custom attributes from request
+    },
+    response_hook: lambda { |span, response|
+      # Extract custom attributes from response
+    }
+  }
+end
+```
+
 ## Examples
 
 Example usage can be seen in the `./example/trace_demonstration.rb` file [here](https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/main/instrumentation/rack/example/trace_demonstration.rb)
--- a/instrumentation/rack/lib/opentelemetry/instrumentation/rack/middlewares/tracer_middleware.rb
+++ b/instrumentation/rack/lib/opentelemetry/instrumentation/rack/middlewares/tracer_middleware.rb
@@ -79,6 +79,7 @@ module OpenTelemetry
               tracer.in_span(request_span_name,
                              attributes: request_span_attributes(env: env),
                              kind: request_span_kind) do |request_span|
+                safe_execute_hook(config[:request_hook], request_span, original_env) unless config[:request_hook].nil?
                 OpenTelemetry::Instrumentation::Rack.with_span(request_span) do
                   @app.call(env).tap do |status, headers, response|
                     set_attributes_after_request(request_span, status, headers, response)
@@ -91,6 +92,12 @@ module OpenTelemetry
           end
 
           private
+
+          def safe_execute_hook(hook, *args)
+            hook.call(*args)
+          rescue StandardError => e
+            OpenTelemetry.handle_error(exception: e)
+          end
 
           def untraced_request?(env)
             return true if @untraced_endpoints.include?(env['PATH_INFO'])
@@ -150,7 +157,7 @@ module OpenTelemetry
             end
           end
 
-          def set_attributes_after_request(span, status, headers, _response)
+          def set_attributes_after_request(span, status, headers, response)
             span.status = OpenTelemetry::Trace::Status.error unless (100..499).include?(status.to_i)
             span.set_attribute('http.status_code', status)
 
@@ -159,6 +166,7 @@ module OpenTelemetry
             # e.g., "/users/:userID?
 
             allowed_response_headers(headers).each { |k, v| span.set_attribute(k, v) }
+            safe_execute_hook(config[:response_hook], span, response) unless config[:response_hook].nil?
           end
 
           def allowed_request_headers(env)
--- a/instrumentation/rack/test/opentelemetry/instrumentation/rack/middlewares/tracer_middleware_test.rb
+++ b/instrumentation/rack/test/opentelemetry/instrumentation/rack/middlewares/tracer_middleware_test.rb
@@ -84,6 +84,91 @@ describe OpenTelemetry::Instrumentation::Rack::Middlewares::TracerMiddleware do
       end
     end
 
+    describe 'config[:request_hook] / config[:response_hook]' do
+      let(:query_string_attribute) { 'query_string' }
+      let(:response_body_attribute) { 'response_body' }
+      let(:query_string) { 'abcd=1234' }
+
+      describe 'when hooks are valid' do
+        let(:config) do
+          {
+            request_hook: lambda do |span, request|
+              span.set_attribute(query_string_attribute, request['QUERY_STRING'])
+            end,
+            response_hook: lambda do |span, response|
+              span.set_attribute(response_body_attribute, response[0])
+            end
+          }
+        end
+
+        it 'should collect data from hooks' do
+          Rack::MockRequest.new(rack_builder).post("/post?#{query_string}", env)
+          post_span = finished_spans.find { |s| s.attributes['http.target'] == "/post?#{query_string}" }
+          _(post_span).wont_be_nil
+          _(post_span.attributes[query_string_attribute]).must_equal(query_string)
+          _(post_span.attributes[response_body_attribute]).must_equal('OK')
+        end
+      end
+
+      describe 'when hooks are configured with incorrect number of args' do
+        let(:received_exceptions) { [] }
+        let(:config) do
+          {
+            request_hook: ->(_span) { nil },
+            response_hook: ->(_span) { nil }
+          }
+        end
+
+        before do
+          OpenTelemetry.error_handler = lambda do |exception: nil, message: nil| # rubocop:disable Lint/UnusedBlockArgument
+            received_exceptions << exception
+          end
+        end
+
+        after do
+          OpenTelemetry.error_handler = nil
+        end
+
+        it 'should not fail the instrumentation' do
+          Rack::MockRequest.new(rack_builder).post("/post?#{query_string}", env)
+          post_span = finished_spans.find { |s| s.attributes['http.target'] == "/post?#{query_string}" }
+          _(post_span).wont_be_nil
+          error_messages = received_exceptions.map(&:message)
+          _(error_messages.all? { |em| em.start_with?('wrong number of arguments') }).must_equal true
+        end
+      end
+
+      describe 'when exceptions are thrown in hooks' do
+        let(:error1) { 'err1' }
+        let(:error2) { 'err2' }
+        let(:received_exceptions) { [] }
+        let(:config) do
+          {
+            request_hook: ->(_span, _request) { raise StandardError, error1 },
+            response_hook: ->(_span, _response) { raise StandardError, error2 }
+          }
+        end
+
+        before do
+          OpenTelemetry.error_handler = lambda do |exception: nil, message: nil| # rubocop:disable Lint/UnusedBlockArgument
+            received_exceptions << exception
+          end
+        end
+
+        after do
+          OpenTelemetry.error_handler = nil
+        end
+
+        it 'should not fail despite wrong hook config' do
+          Rack::MockRequest.new(rack_builder).post("/post?#{query_string}", env)
+          post_span = finished_spans.find { |s| s.attributes['http.target'] == "/post?#{query_string}" }
+          _(post_span).wont_be_nil
+          error_messages = received_exceptions.map(&:message)
+          _(error_messages).must_equal([error1, error2])
+        end
+      end
+    end
+
     describe 'config[:untraced_endpoints]' do
       describe 'when an array is passed in' do
         let(:config) { { untraced_endpoints: ['/ping'] } }
--- a/instrumentation/rack/lib/opentelemetry/instrumentation/rack/instrumentation.rb
+++ b/instrumentation/rack/lib/opentelemetry/instrumentation/rack/instrumentation.rb
@@ -30,6 +30,8 @@ module OpenTelemetry
         option :untraced_endpoints,       default: [],    validate: :array
         option :url_quantization,         default: nil,   validate: :callable
         option :untraced_requests,        default: nil,   validate: :callable
+        option :request_hook, default: nil, validate: :callable
+        option :response_hook, default: nil, validate: :callable
 
         private
 


ahayworth:
@nozik thanks for this! I took an initial look and I’m not opposed to it, but I was wondering if you could detail some concrete use-cases you have for it?

Basically what I’m wondering is if most of these use cases could be handled without running a proc? So like, for example - if we are interested in adding header values as attributes, that could be a list of header names. We can make sure that everyone who wants that functionality can get it without needing to write a proc to get it, and we can test for correctness and performance, etc.

I’m sure there are other use cases for this, and I think it would be helpful to discuss them a little here. Maybe all of them could be just config options without a proc; maybe some of them can’t be. But I’m not sure!

nozik:
Hi @ahayworth, happy to share more details. First, I'd like to emphasize that this is a common mechanism across several languages that are supported in OTEL (and soon, I believe, all of them). Moreover, the potential use cases are virtually endless: collecting payloads (automated logging), extracting data from headers, detecting security signals, and more. Supported only a limited set of configurations will add more complexity to the instrumentation code, and provide less power to the end user

ahayworth:
Thanks for that @nozik - what other instrumentations other than FastAPI currently implement something like this?

Regarding complexity in the instrumentation: you have a good point, however I personally feel we *should* take that complexity on in the instrumentation rather than requiring our users to do it over and over. It's hard to defend precisely *why* that is, but to me it doesn't feel incredibly ruby-like to place the work for such common cases upon the end user. I'm curious what the other maintainers think here.

ahayworth:
Thanks for that @nozik - what other instrumentations other than FastAPI currently implement something like this?

Regarding complexity in the instrumentation: you have a good point, however I personally feel we *should* take that complexity on in the instrumentation rather than requiring our users to do it over and over. It's hard to defend precisely *why* that is, but to me it doesn't feel incredibly ruby-like to place the work for such common cases upon the end user. I'm curious what the other maintainers think here.

ericmustin:
mmm, I generally think it's ok but definitely a bit of a pandoras box. 

JS does this with it's `http` instrumentation and, anecdotally speaking  it's a hook we've explored using internally for our koa apps for special treatment of certain requests. https://github.com/open-telemetry/opentelemetry-js/tree/main/experimental/packages/opentelemetry-instrumentation-http#http-instrumentation-options

In practice the issues i've seen with this sort of feature is that users accidentally cardinality bomb themselves by adding a hook that modify's the span name dynamically.

It's really a question of responsiveness on our part. if we don't want this feature as maintainers, are we willing to quickly respond to all future feature requests for specific new config options, and implement and release them in a timely way.

Also, by leaning on a feature like this we're really straying from the benefits of env vars. it would mean users have to add non-standard env vars into their hooks.

I'll try to review and think about this more closely in the coming days, but just providing my quick thoughts for @ahayworth

nozik:
@ahayworth Similar hooks are supported by all Python http frameworks (flask, django, tornado, starlette, etc.) - same for node.js - so this is really quite standard. Generally speaking I believe the number of use cases for OpenTelemetry is huge, way beyond how it's commonly used today, and I think it's better to let the end users be creative :)

nozik:
@ericmustin Let me give you my perspective on this - I think that features like this are important enablers for the creation of meta-packages that collect different pieces of data for different needs. Between the end-users and the maintainers, there's a huge potential for different vendors to create distros that address requirements of a variery of audiences (software engineeers, security engineers, software architects, automation engineers, finops teams, and more). Perhaps these "distros" should be upstreamed? Perhaps, but I'm not sure it's feasible.

ericmustin:
@nozik
> so this is really quite standard

do you mean there is an OTEP or portion of the specification (https://github.com/open-telemetry/opentelemetry-specification) related to this, or are you just speaking generally about industry observability sdks? It looks like you and your team have done a great job of doing a survey of the current landscape, it would be awesome to see some of this work formalized for basic standard things like, say, a config option name, or the arguments available to the lambda, etc.

The request/response hook config option is more or less just a continuation of user trends we've seen on the last 2 PRs to the Rack instrumentation, which both just added callables because the user needs were too broad and weren't well supported by current config. We've had similar discussions on the PRs for those features (although i think they may have been lost during the migration to a `-contrib` repo recently)

- https://github.com/open-telemetry/opentelemetry-ruby-contrib/commits/main/instrumentation/rack/lib/opentelemetry/instrumentation/rack/instrumentation.rb
- https://github.com/open-telemetry/opentelemetry-ruby-contrib/commit/11af088d2f7e0abc16228b0267b52a25541329ed

So, I generally agree with what you're saying. I think you touch one some good points and are operating in some areas that frankly are just not very clearly addressed by open-telemetry's specification or current design. I also want to enable you, as someone building on top of opentelemetry apis and sdks, to be able to. I'm just not sure if, when discussing the features you mentioned

>the potential use cases are virtually endless: collecting payloads (automated logging), extracting data from headers, detecting security signals, and more.

Whether a config option on the contrib rack tracing instrumentation is the appropriate way to offer many of these features. I certainly recognize that these features are important things that users want and vendors want to be able to offer, if they don't do so already.

Using the example of security signal detection you mention, one OSS prior art that's useful to reference is how datadog approaches this, which is to offer a separate package that injects a separate middleware that's inserted into the rack middleware chain in addition to the tracing middleware. That separate package does the song and dance of adding arbitrary stuff™️ to traces(if an active trace is present), an addition to lots of other security things. https://github.com/DataDog/dd-trace-rb/tree/master/lib/datadog/appsec/contrib/rack

Shoehorning all that functionality into a proc on the tracing instrumentation seems to be an antipattern, imo, and I am very hesitant to accidentally sign myself up to maintain support for that.

But otoh the burden for mimicking this approach in OpenTelemetry is pretty high. It would mean publishing a separate package within -contrib or via your company's gh/rubygems acct+registering it on the [OTel Registry](https://opentelemetry.io/registry/)). That seems pretty ridiculous for some potential basic usecases ("add some extra info about request if headers are certain values ") , and other languages have seemingly managed to offer this config option without footgunning themselves into a high support load.

Given all this (sorry for the long windedness), and that instrumentation is "experimental" technically, i think this sort of config option is a reasonable thing to support and i'm open to taking this PR. But I would hope that the more involved use cases you mention get formalized and moved forward as their own fully fleshed out specification or OTEPs, or at the very least separate packages, although I recognize that is a slow process.

nozik:
@ericmustin Thanks for the longer and detailed answer. By standard I wasn't referring to anything formal, just for the actual status of of instrumentations in other languages. Anyway, I do agree with your approach in general - there should be different SDK for different purposes - but they should all be built on top of the base SDK. Imagine that these SDKs would simply be thin config wrappers above the vanilla ruby sdk (this is actually what my team is currently building). Anyway, I agree there's definitely a bigger picture here. And finally - your PR would be much appreciated :) Thanks!

instrumentation/rack/test/opentelemetry/instrumentation/rack/middlewares/tracer_middleware_test.rb
```diff
@@ -84,6 +84,91 @@
...
+      let(:query_string_attribute) { 'query_string' }
+      let(:response_body_attribute) { 'response_body' }
+      let(:query_string) { 'abcd=1234' }
+
+      describe 'when hooks are valid' do
+        let(:config) do
+          {
+            request_hook: lambda do |span, request|
+              span.set_attribute(query_string_attribute, request['QUERY_STRING'])
+            end,
```
ericmustin, line 97:
```suggestion
            request_hook: lambda do |span, env|
              span.set_attribute(query_string_attribute, env['QUERY_STRING'])
            end,
```

instrumentation/rack/test/opentelemetry/instrumentation/rack/middlewares/tracer_middleware_test.rb
```diff
@@ -84,6 +84,91 @@
...
+            response_hook: lambda do |span, response|
+              span.set_attribute(response_body_attribute, response[0])
+            end
```
ericmustin, line 100: I'm not really sure the rack response body is something that's appropriate to expose as part of configuration, imo, especially given the lack of tests around how this behaves compared to the detail of the specification https://github.com/rack/rack/blob/main/SPEC.rdoc#the-body- . Reading the specification for it, there seems to be a number of untested code paths currently. Asking users to try to put in handling for all of these within their lambda seems unrealistic, and I tend to agree with @ahayworth that we may want to address specific use cases with configuration that we have confidence in.

instrumentation/rack/lib/opentelemetry/instrumentation/rack/middlewares/tracer_middleware.rb
```diff
@@ -150,7 +157,7 @@ def create_request_span_name(request_uri_or_path_info, env)
             end
           end
 
-          def set_attributes_after_request(span, status, headers, _response)
+          def set_attributes_after_request(span, status, headers, response)
```
ericmustin, line 160: I'm concerned that we're now exposing the rack `response` body to the user, but I may be off base, I am not an expert on rack/rails tbh, but I recall that it's unreliable and potentially non performant to attempt to access the rack response body. For example with streaming response the behavior differs significantly. 

- https://tenderlovemaking.com/2011/03/03/rack-api-is-awkward.html 
- https://api.rubyonrails.org/classes/ActionController/Streaming.html
- https://github.com/rack/rack/blob/main/SPEC.rdoc#the-body-

Have you tested this against any practical examples or applications, for example? I'm not sure the rack `response` body makes sense to pass into the hook.

instrumentation/rack/README.md
```diff
@@ -49,6 +49,23 @@ OpenTelemetry::SDK.configure do |c|
...
 ```
 
+### Request/response hooks
+
+You can pass hooks to extract additional attributes from the requests/responses.
+
+```ruby
+  c.use 'OpenTelemetry::Instrumentation::Rack', {
+    request_hook: lambda { |span, env|
```
ericmustin, line 59: can we link to the rack documentation on what the `env` argument is? Most users probably are not intimately familiar with rack internals.

ericmustin (changes requested):
I had a chance to review this and unfortunately I think I am touching the edges of my limits wrt rack internals. After reading up on the rack specification a bit, https://github.com/rack/rack/blob/main/SPEC.rdoc#the-body- I have some concerns/doubts about the response hook, specifically about exposing the response body to users. I am not a rack expert though so it's possible my concerns are mis-guided. I'd like to discuss this at the upcoming ruby otel SIG mtg (tuesday, 12PM EST on zoom, link here https://github.com/open-telemetry/community) to hopefully get the opinion of some folks with a bit more familiarity with rack than me. I've added it to the [agenda](https://docs.google.com/document/d/1D15bO8o340sQm2CVZiukEJuCO_XMMHKPuTznoEhyFqE/edit#), everyone is welcome at these meetings and they are recorded as well.

instrumentation/rack/test/opentelemetry/instrumentation/rack/middlewares/tracer_middleware_test.rb
```diff
@@ -84,6 +84,91 @@
...
+          }
+        end
+
+        it 'should collect data from hooks' do
```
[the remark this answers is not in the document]
nozik: This is not intended to run on untraced endpoints. Adding a test for that

nozik:
@fbogsany Thanks for your feedback.

1. I assure you it's not done because "everybody else does that" - but rather because it's an actual necessity. We are 
  building an OTEL distro that collects data automatically for our customers (we do that in several languages), and this is the most straight-forward approach to implement that.

2. Accessing `current_span` doesn't give me what I need, since it assumes I have control over the actual controller/ middleware code. Since this is an SDK I'm providing to my customers, and have no control over their code (nor do I expect them to make any code changes) this isn't possible.

fbogsany:
If you're building your own SDK and you need to add further (non-spec) information to Rack spans, I recommend you build your own Rack instrumentation (ideally, but not necessarily, based on the instrumentation from this project) and add the information you require. You can then configure your SDK to install your Rack instrumentation instead of the Rack instrumentation provided by this project.

Where I'd like us to go with our instrumentation is to provide a library of helpers that populate span attributes described by the OpenTelemetry semantic conventions. Hypothetically, the Rack instrumentation would then look something like:

```ruby
          def call(env)
            return OpenTelemetry::Common::Utilities.untraced { @app.call(env) } if untraced_request?(env)

            extracted_context = OpenTelemetry.propagation.extract(
              env,
              getter: OpenTelemetry::Common::Propagation.rack_env_getter
            )
            span_name = OpenTelemetry::Helpers.http_server_span_name(
              uri: env['REQUEST_URI'],
              path: env['PATH_INFO'],
              method: env['REQUEST_METHOD']
            )
            span_attributes = OpenTelemetry::Helpers.http_server_span_request_attributes(
              method: env['REQUEST_METHOD'],
              host: env['HTTP_HOST'] || 'unknown',
              scheme: => env['rack.url_scheme'],
              target: env['QUERY_STRING'].empty? ? env['PATH_INFO'] : "#{env['PATH_INFO']}?#{env['QUERY_STRING']}",
              user_agent: env['HTTP_USER_AGENT'],
              headers: allowed_request_headers(env)
            )
            OpenTelemetry::Context.with_current(extracted_context) do
              tracer.in_span(span_name, attributes: span_attributes, kind: :server) do |span|
                OpenTelemetry::Instrumentation::Rack.with_span(request_span) do
                  @app.call(env).tap do |status, headers, response|
                    OpenTelemetry::Helpers.apply_http_server_span_response_attributes(
                      span,
                      status_code: status,
                      headers: allowed_response_headers(headers)
                    )
                  end
                end
              end
            end
          end
```

This seems relatively straightforward to use and the instrumentation then follows a very simple pattern, and you can easily copy it and insert your own attributes.

One of the key requirements of the OpenTelemetry provided auto-instrumentation is that the provenance of the instrumentation is clear from the `instrumentation_library.name` and `instrumentation_library.version`. Adding hooks to include additional information incorrectly pushes the support burden from custom SDKs to the OpenTelemetry community because bugs in the hooks, leaked information, or non-spec attributes all appear to be problems with the instrumentation from this project instead of instrumentation/hooks your company has added.

nozik:
@fbogsany I see your point, but I think the community will never be able to keep up with all the potential use cases out there.  That's not scalable, and moreover - why would you even need to know about them, and spend time reviewing their code? If I create an OTEL distro with a buggy hook (which, frankly, I'm sure someone, if not me personally, will definitely do), why should it be a concern of the OTEL community? If I implement a bug in a Rails application, I don't think it should burden the Rails community.

fbogsany:
> the community will never be able to keep up with all the potential use cases out there

That's correct. It is not the responsibility of this repo to "keep up with all the potential use cases". It is the responsibility of this repo to implement auto-instrumentation that adheres to the OpenTelemetry semantic conventions for a reasonable subset of reasonably widely used Ruby gems.

As a vendor who wishes to create an OTel distro, you are free and very welcome to implement additional or replacement instrumentation, if you feel that the auto-instrumentation here is insufficient.

> If I create an OTEL distro with a buggy hook (which, frankly, I'm sure someone, if not me personally, will definitely do), why should it be a concern of the OTEL community?

Because the 🖕 points at us, not at your OTEL distro. As I pointed out, the `instrumentation_library.name` and `instrumentation_library.version` will reference the community's gem, even if the bug is in the hook. This delays problem resolution for your customers and increases the maintenance overhead for the folks here who have to investigate the bug report and redirect to you.

ericmustin:
Appreciate the thoughtful input from everyone here.  @nozik @fbogsany just adding here that we put this on the agenda and discussed on Tuesday's SIG Mtg, which everyone is welcome and encouraged to attend. The conclusion we reached is that we aren't going to be accepting a Rack hooks PR at this time in the `-contrib` repo, but we recognize that we should improve our documentation around how to add and include one's own instrumentation in your own distro. Mtg Notes can be found here https://docs.google.com/document/d/1D15bO8o340sQm2CVZiukEJuCO_XMMHKPuTznoEhyFqE/edit#

it's also worth noting that we're discussed that we're very open to supporting and adding more configuration options for rack if there is a specific use case that isn't addressed and can be implemented in a robust, performant way, but there hasn't been a specific use case provided yet that meets that criteria.

Also worth noting that we also discussed reverting the client hooks PR, although we didn't reach consensus on that.

I will try to push up some improved documentation on approaches to bringing your own instrumentation in the coming days.

plantfansam:
> I will try to push up some improved documentation on approaches to bringing your own instrumentation in the coming days

Tracking that [here](https://github.com/open-telemetry/opentelemetry-ruby-contrib/issues/104).

nozik:
@fbogsany Thanks for the detailed response. What I'm basically saying is that the hooks are essential enablers for such distros and that creating separate instrumentations for the same gems is something that should be avoided, as it contradicts one of the most important reasons (in my humble opinion) for the creation of OpenTelemetry - consolidation of the client-side instrumentations. I believe that the role of the community is, as you said, provide implementation for the semantic conventions, but also to enable downstream use cases. I don't agree that a bug in such a hook will be associated with OTEL rather than with the user who added it or the vendor who provided it, but this is something we can agree to disagree on.

Anyway, thanks everyone for the thoughtful comments. I urge you to reconsider your position; I will continue contributing regardless, of course :)

plantfansam:
I'm closing this, since the above discussion concludes that we are not going to merge rack hooks at this time. Follow-on work for improving docs on making new out-of-repo instrumentation gems is being tracked [here](https://github.com/open-telemetry/opentelemetry-ruby-contrib/issues/104).

PR #79, opened 2022-07-05
feat: Add GraphQL query context based tracing

toneymathews:
`opentelemetry-instrumentation-graphql` provides a `GraphQLTracer` ([source](https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/37696bd2c5c8c58c072937e69834b562bf2c42a5/instrumentation/graphql/lib/opentelemetry/instrumentation/graphql/tracers/graphql_tracer.rb)) based on `GraphQL::Tracing::PlatformTracing` ([source](https://github.com/rmosolgo/graphql-ruby/blob/master/lib/graphql/tracing/platform_tracing.rb)). This includes [several schema level configuration options](https://github.com/open-telemetry/opentelemetry-ruby/blob/opentelemetry-instrumentation-graphql/v0.19.2/instrumentation/graphql/lib/opentelemetry/instrumentation/graphql/instrumentation.rb#L38-L40) that can enable/disable specific parts of tracing.

The current options are global, but we could benefit from also having per request configuration. For example, a HTTP header may request verbose traces be recorded, ideally enabling `enable_platform_field` for only this request’s GraphQL execution. This way a client can be permitted to opt into tracing, without having to enable it for all requests.

To make this possible, the GraphQL execution context seemed like an ideal place to have this request specific configuration. Access to context is now possible as of graphql-ruby 1.13.13 and 2.0.9 (see [changelog](https://github.com/rmosolgo/graphql-ruby/blob/master/CHANGELOG.md), added in [#4077](https://github.com/rmosolgo/graphql-ruby/pull/4077)), meaning we have this option going forward, but potentially need to remain backwards compatible.

In `cached_platform_key`, the newer versions of graphql-ruby have both `context` and `trace_phase`, so we can determine which setting to check and which of `platform_field_key`, `platform_authorized_key`, `platform_resolve_type_key` would be called on a cache miss.

If we were to release a new version of `opentelemetry-instrumentation-graphql`, how should we handle checking the version of graphql-ruby, either being backwards compatible or having a minimum version?

The [CI failures](https://github.com/open-telemetry/opentelemetry-ruby-contrib/actions/runs/2619340318) have come up since `cached_patform_key` now expects 3 arguments (`trace_phase` being the new one) in newer versions of graphql-ruby.

cc @ravangen

### DIFF
--- a/instrumentation/graphql/Appraisals
+++ b/instrumentation/graphql/Appraisals
@@ -4,6 +4,10 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 
+appraise 'graphql-1.13' do
+  gem 'graphql', '~> 1.13.0'
+end
+
 appraise 'graphql-1.11' do
   gem 'graphql', '~> 1.11.0'
 end
--- a/instrumentation/graphql/README.md
+++ b/instrumentation/graphql/README.md
@@ -53,6 +53,27 @@ OpenTelemetry::SDK.configure do |c|
 end
 ```
 
+#### GraphQL context configuration
+
+Optionally, GraphQL execution context can include tracing option keys to control tracing behaviour independently per execution.
+
+An option's global configuration must be enabled for the associated context option to be respected. This is supported for `graphql-ruby` version `>= 1.13.13 < 2` or `>= 2.0.9`.
+
+| | Global Enabled | Global Disabled |
+|:---:|:---:|:---:|
+| **Context Unset** | ✅ | ❌ |
+| **Context Enabled** | ✅ | ❌ |
+| **Context Disabled** | ❌ | ❌ |
+
+```ruby
+query = GraphQL::Query.new(MyAppSchema, 'query { foo }')
+
+opentelemetry_context = query.context.namespace(:opentelemetry)
+opentelemetry_context[:enable_platform_field] = true
+opentelemetry_context[:enable_platform_authorized] = true
+opentelemetry_context[:enable_platform_resolve_type] = true
+```
+
 ## Examples
 
 An example of usage can be seen in [`example/graphql.rb`](https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/main/instrumentation/graphql/example/graphql.rb).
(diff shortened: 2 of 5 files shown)


plantfansam:
Thanks so much for the contribution @toneymathews! I need to familiarize myself with the changes in graphql-ruby and how `cached_platform_key` fits into the overall upstream tracing strategy before commenting on that.

On the topic of version compatibility: generally speaking, each instrumentation gem's `Instrumentation` class, which subclasses `Instrumentation::Base`, calls a [`compatible` block](https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/3b03ff7ea66b69c85ba205a369b85c2c33b712fe/instrumentation/base/lib/opentelemetry/instrumentation/base.rb#L136-L138) to determine if the version of the gem being instrumented is compatible with our instrumentation library.  I think that our current GraphQL `Instrumentation` class does not call this block, but it probably should. 

If we imagine a world where there is a `compatible` block in the our GraphQL `Instrumentation` class, then most likely we'd need to decide if the feature being proposed here is worth breaking compatibility with older versions.

A couple alternatives to that:

- We could dynamically select different code paths depending on which version of the gem is installed (that sounds like a maintenance headache)
- There might be some clever way to work around the problem you're facing in this specific instance, which means we might not have to solve the general problem.

The other @open-telemetry/ruby-contrib-maintainers may have some thoughts on a more holistic strategy for managing compatibility, here.

plantfansam:
Thanks for adding to Appraisals @toneymathews! Will take a look.

robertlaurin:
I've assigned this to myself, I'll work towards giving this a proper thorough review this week. It has a few implications, and unspecified behaviour so we need to be really detailed here.

ravangen:
Thank you for taking the time to review the suggestion and give it some thought.

GraphQL is not necessarily invoked through an HTTP request, although is the most common scenario. GraphQL being HTTP agnostic means that something needs to tell `GraphQLTracer` how to read the specific execution's configuration.

Practically speaking, I am unclear on an alternative to how the current request object or configuration is proposed to be passed down into the `GraphQLTracer`. Note that `context` was not usable prior to `graphql-ruby < 1.13.13` and `>= 2 and < 2.0.9`.

HTTP [`baggage`](https://www.w3.org/TR/baggage/) header is an option that can inform what is set in the GraphQL execution context. The current proposed approach is unopinionated, leaving it to the application code to know how to set these optional override settings.

> It would allow us as users to add a baggage header that says something like "baggage: ServiceC=enable_platform_field;" which could signal which application we want to enable high the verbosity request tracing.

A frontend client may not know the "service name" to put into a `baggage` header, and instead elect to just say "debug my request" through a query argument or request header for example. Some app logic could still transform such a flag into a baggage header for propagation internally.

Regardless, these changes can work with distributed tracing. Configuration can still be propagated across service boundaries using whatever tooling exists today I imagine. GraphQL wouldn't be concerned with this propagation as it is not directly the mechanism calling other services, instead it would be other clients (which are likely instrumented).

PR #83, opened 2022-07-08
Run rubocop against ruby 3.1

plantfansam:
We currently run Rubocop against Ruby 2.7 in CI. We do not run it against any other versions. I think we should run against Ruby 3.1 for the following reasons:

- I suspect most folks developing locally will be running the latest Ruby (3.1), and will run Rubocop against that. We should have CI match that workflow.
- Ruby is at version 3.1, so we should default to running our static analysis against that version.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -29,6 +29,7 @@ jobs:
         with:
           gem: "${{ matrix.gem }}"
           ruby: "3.1"
+          rubocop: true
       - name: "Test Ruby 3.0"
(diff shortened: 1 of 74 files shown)


plantfansam:
@robbkidd [made the excellent point](https://cloud-native.slack.com/archives/C01NWKKMKMY/p1657298502260789) that we should be running against lowest supported Ruby version. "That avoids Rubocop barking at us to use language features that appear in later Rubies that will break when run under older-but-still supported Rubies."

PR #84, opened 2022-07-08
Consolidate rubocop in root of repo

plantfansam:
Sorry for the barrage of Rubocop-related PRs (many of which were closed rapidly). I was trying to do this incrementally, which, it turned out, was a mistake. This PR:

- Adds a `rubocop.yml` file to the root of the repository and populates it with commonly-used configuration from the gems
- Points each gem's `.rubocop.yml` file to the `.rubocop.yml` file at the root of this repo
- Removes each gem's `test/rubocop.yml` file
- Makes some changes to the code that satisfies the root `rubocop.yml` (mostly automatic, generated via `rubocop . -a`)
- Adds some `# rubocop disable:` comments


instrumentation/active_job/test/test_helper.rb
```diff
@@ -42,12 +42,13 @@ def perform
...
 end
+
 class KeywordOnlyArgsJob < ::ActiveJob::Base
-  def perform(keyword1: 'default', keyword2:); end
+  def perform(keyword2:, keyword1: 'default'); end
 end
 
 class MixedArgsJob < ::ActiveJob::Base
-  def perform(arg1, arg2, keyword1: 'default', keyword2:); end
+  def perform(arg1, arg2, keyword2:, keyword1: 'default'); end
```
ahayworth, line 51: Why?

ahayworth (commented):
So mostly this just looks like an excellent cleanup! I had two questions. One is online, and relates to some keyword reordering that looked odd to me.

The other is: what happens to new instrumentation when the generator is run? We remove the rubocop template in this PR, but we don't remove all the "stub" rubocop.yml files in the already-existing instrumentation.

instrumentation/active_job/test/test_helper.rb
robbkidd: I'll posit that it was to appease [the Style/KeywordParametersOrder cop](https://docs.rubocop.org/rubocop/cops_style.html#stylekeywordparametersorder) where required keyword params come first and optional all appear at the end of the params list.

.rubocop.yml
```diff
@@ -0,0 +1,67 @@
...
+  Exclude:
+    - propagator/xray/lib/opentelemetry/propagator/xray/text_map_propagator.rb
+Lint/UnusedMethodArgument:
+  Enabled: false
+Naming/FileName:
+  Exclude:
+    - '**/opentelemetry-*'
+Naming/MethodParameterName:
+  Exclude:
+    - instrumentation/dalli/lib/opentelemetry/instrumentation/dalli/patches/server.rb
```
arielvalentin, line 32: We currently have a mix of disabling specific rules inline when we do not want them to apply to all files so I am not sure how I feel about overrides for individual files being moved into the parent Rubocop file. 

I know they probably already existed in the gem specific configs but IMO we should either disable these rules in individual source files or override them closer to a gem specific Rubocop config.

plantfansam: Yeah, great point. I'm for moving all exceptions to specific files. I can do a pass at that.

instrumentation/active_job/test/test_helper.rb
plantfansam: yep

ahayworth (approved):

arielvalentin (approved):

arielvalentin:
@plantfansam https://github.com/open-telemetry/opentelemetry-ruby-contrib/runs/7327567677?check_suite_focus=true#step:6:493

Issue #85, opened 2022-07-08
Instrumentation generator cannot load current OpenTelemetry API version


plantfansam:
When running `bin/generate_instrumentation` the code tries to load the current API version from `require_relative '../api/lib/opentelemetry/version'`, but it's not there anymore (since we split `-contrib` out of the main `opentelemetry-ruby` repo.

```
➜  opentelemetry-ruby-contrib git:(main) bin/instrumentation_generator 
Traceback (most recent call last):
        3: from bin/instrumentation_generator:3:in `<main>'
        2: from bin/instrumentation_generator:3:in `require_relative'
        1: from /Users/.../src/github.com/open-telemetry/opentelemetry-ruby-contrib/.instrumentation_generator/instrumentation_generator.rb:7:in `<top (required)>'
/Users/.../src/github.com/open-telemetry/opentelemetry-ruby-contrib/.instrumentation_generator/instrumentation_generator.rb:7:in `require_relative': cannot load such file -- /Users/.../src/github.com/open-telemetry/opentelemetry-ruby-contrib/api/lib/opentelemetry/version (LoadError)
```
**Share a simplified reproduction if possible**

`bin/instrumentation_generator`

We could look up the latest version from RubyGems, require the latest version of the otel API gem using [`gemfile do`](https://bundler.io/guides/bundler_in_a_single_file_ruby_script.html), or do something else!

PR #87, opened 2022-07-11
Update instrumentation generator to use inline bundler

plantfansam:
Addresses #85. Also bumps minimum ruby version to the current stable ruby in the 2.x series.


ahayworth (approved):
:shipit:

arielvalentin (approved):

Issue #107, opened 2022-07-23
Net::HTTP - Add ability to exclude domains


johnnyshields:
Rack instrumentation has an option `untraced_endpoints` which can be used to exclude tracing things like `/health`

I'd like to request an analogous feature for `Net::HTTP`. This can be used to filter out requests to Bugsnag/Sentry/etc. which are sent randomly in a non-webserver thread. Perhaps we could do `untraced_hosts`

ahayworth:
I believe we would welcome such a PR! This seems like it could be useful in a variety of circumstances.

johnnyshields:
PR raised: https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/108

PR #108, opened 2022-07-23
Add Net::HTTP :untraced_hosts option

johnnyshields:
Fixes https://github.com/open-telemetry/opentelemetry-ruby-contrib/issues/107 by adding `untraced_hosts` option.

Analogous to Rack instrumentation `untraced_endpoints`

This option is an Array which can be mixed String and Regexp. An example use case of Regexp is to exclude a domain and its subdomains:

```ruby
/\A(.*\.)?google\.com\z/
```

I've raised a related PR for Rack to support Regexp as well. https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/109


instrumentation/net_http/lib/opentelemetry/instrumentation/net/http/patches/instrumentation.rb
```diff
@@ -87,6 +93,14 @@ def annotate_span_with_response!(span, response)
...
               Net::HTTP::Instrumentation.instance.tracer
             end
+
+            def untraced?
+              return true if instrumentation_config[:untraced_hosts]&.any? do |host|
+                host === @address # rubocop:disable Style/CaseEquality
+              end
+
+              false
+            end
```
ahayworth, line 103:
```suggestion
            def untraced?
              instrumentation_config[:untraced_hosts].include?(@address)
            end
```

We should avoid `===`, since its uses are [far more interesting and varied](https://medium.com/rubyinside/triple-equals-black-magic-d934936a6379) than what we want here.

Given that the default for the option is `[]` and we are validating that it is an Array, I think we can simplify the entire method to be an `includes?` check.

johnnyshields: The intention here is to support both String and Regexp. Your proposed change doesn't work:

```ruby
[/foo/].include?('foo') #=> false
```

ahayworth: This PR didn't say that you were asking it to validate regular expressions as well. I don't know if we _should_ allow that. Triple-equals allows a lot of loopholes.

In the context of "this is an array of strings", it would work fine.

johnnyshields: We should allow Regexp. An example use case is to include a domain and its subdomains:

```
/\A(.*\.)?google\.com\z/
```

johnnyshields: Done.

ahayworth: I'm on the fence about it myself - I have the vaguest recollection of us deciding not to allow regular expressions before. Maybe @fbogsany or @robertlaurin can remember?

If they don't recall a specific thing, then perhaps it was a fever dream.

robertlaurin: I don't remember specifically not allowing regular expressions.

If I remember correctly for rack that case was made that we didn't want to add a config option for everyones preferred approach of filtering. [We ended up with two config options](https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/main/instrumentation/rack/lib/opentelemetry/instrumentation/rack/middlewares/tracer_middleware.rb#L95-L100), the first pass pass which simply accepted an array of strings representing endpoints to omit, and the ultra malleable lambda. 

I think this PR can go a couple of directions.

- We could say this should be solved via a sampler and drop this PR entirely, I've seen this approach more widely pushed in the instrumentation sigs whenever this subject has come up. 

That would look something like the (untested) snippet below, the benefit here is that not every http instrumentation gem needs to be updated. So it no longer becomes a concern of the OpenTelemetry maintainers as this approach is already supported.
```rb
class CustomSampler
  attr_reader :description

  RESULT = OpenTelemetry::SDK::Trace::Samplers::Result
  DECISION = OpenTelemetry::SDK::Trace::Samplers::Decision

  def initialize(untraced_hosts:, description:)
    @untraced_hosts = untraced_hosts
    @description = description
  end

  def ==(other)
    @untraced_hosts == other.untraced_hosts && @description == other.description
  end

  # @api private
  #
  # See {Samplers}.
  def should_sample?(trace_id:, parent_context:, links:, name:, kind:, attributes:)
    sample_decision = if kind == :client && untraced?(attributes['net.peer.name'])
      DECISION::DROP
    else
      DECISION::RECORD_AND_SAMPLE
    end

    RESULT.new(decision: sample_decision, tracestate: OpenTelemetry::Trace.current_span(parent_context).context.tracestate)
  end

  def untraced?(host)
    return true if @untraced_hosts&.any? do |untraced_host|
      host.is_a?(Regexp) ? host.match?(host) : untraced_host == host
    end

    false
  end
end
```

- If we decide that the sampler approach is not suitable for whatever reason, we could consider amending this PR to make it consistent with the [rack configuration options ](https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/main/instrumentation/rack/lib/opentelemetry/instrumentation/rack/middlewares/tracer_middleware.rb#L95-L100). Rack and http client gems have the common ground of the http semantic conventions, so it makes sense to me for the configuration options to have some parity.

- We could deviate to something a bit different, like what is proposed in this PR.

@johnnyshields you've been an active adopter and contributor to otel ruby for a while now, I'm interested to know what you think of the sampler approach for this instead of adding a config options.

johnnyshields: > We could consider amending this PR to make it consistent with the [rack configuration options](https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/main/instrumentation/rack/lib/opentelemetry/instrumentation/rack/middlewares/tracer_middleware.rb#L95-L100)

I actually originally a lambda option in this PR like Rack, but due to the gnarly specifics of `Net::HTTP`, it just doesn't feel as clean--you'd need 3 lambda args for `@address`, request, and request body respectively. (Rack *happens* to be designed for pluggability, which one `env` variable that is passed thru your middleware stack, and also into the lambda.)

I also couldn't think of a clear use case for the lambda that I couldn't address with Regexp. So I thought it would be better to implement in a separate PR if/when someone actually needs it.

> We could say this should be solved via a sampler and drop this PR entirely

Hmm... I think filtering out specific domains vs. "sampling" a subset of your overall requests are two different concerns/use cases, at least in my mind. It seems to be overkill to have to create a sampler class when I really just want to ignore one or two domains on this specific Net::HTTP instrumenter. Moreover, I'd then have to worry about chaining my "ignore domain" sampler with my overall "1-out-of-1000 requests" sampler, or else kludge the two concerns together. It also seems that OTEL Ruby currently [lacks basic programmatic configurability of samplers](https://github.com/open-telemetry/opentelemetry-ruby/issues/321#issuecomment-1189599314).

**TLDR;** I think we should merge this (minimal) PR as-is and defer decision for both Rack-style lambda arg and samplers.

fbogsany:
https://github.com/open-telemetry/opentelemetry-ruby/pull/594 is the PR that added the `untraced_endpoints` option to the Rack instrumentation. It contains a [discussion about Regex](https://github.com/open-telemetry/opentelemetry-ruby/pull/594/files#r568201520). The argument at the time was "let's keep it simple and address the use-cases we know about", which was primarily health-check pings. There wasn't a strong argument against Regexes other than possibly matching too aggressively.

We are currently exploring a "rules-based sampler", which is reasonably flexible and configurable for this use-case. Samplers are meant to handle this use-case, but the default set of samplers in the SDK is insufficient and the interface has grown a little complex. Our motivating example is very similar to the one for this PR - our logging framework is starting to use Kafka, which we're tracing, so we end up with a lot of noisy traces. The API we're experimenting with looks like:

```ruby
    # Returns a sampler configured to verbosely sample at the provided rate.
    # It ensures that spans forced with force_verbose_trace are always marked as verbose.
    # Sampling rules can be overridden by passing a different set of rules. For example:
    #   rules = [
    #     RulesSampler::Rule.new(kind: :producer, attributes: { 'messaging.destination' => 'logging.resque.v2.raw' }, delegate: ALWAYS_OFF),
    #   ]
    #   verbosity_sampler(probability: 0.01, rules: rules + DEFAULT_RULES)
    def verbosity_sampler(probability:, rules: DEFAULT_RULES)
```

We haven't considered regex matching for attribute values. That would add the flexibility you're looking for here.

Re: programmatic sampler configuration, while not part of the SDK `configure` block, it has been relatively easy since https://github.com/open-telemetry/opentelemetry-ruby/pull/822 to configure it like:

```ruby
OpenTelemetry.tracer_provider.sampler = OpenTelemetry::Shopify::VerbositySampler.new(
  probability: VERBOSITY_SAMPLER_PROBABILITY
)
```

Lastly, @robbkidd brought up in CNCF Slack:
> would allow/deny listing hosts from HTTP instrumentation also provide a mechanism to prevent trace/baggage headers from being sent to untraced hosts? Thinking that disabling trace/baggage prop could be a use case for this option that isn't addressed by sampling instead.

This is an excellent call-out. We had that capability in our Shopify-internal tracing library, and we lost it with the move to OpenTelemetry. We also only needed to consider `host`, although we matched a `Regexp.union` of regexes rather than an array. From that conversation:

> I think we'd need two options, though, since the `untraced` hosts may be different from the `unpropagated` hosts (e.g. you want to trace requests to 3rd party services, but you don't want to propagate traces to them).
>
> We would need to check unpropagated? here: https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/108/files#diff-f8e8b9417e4ef87bb82a060595c901af50759fa0d5a52f66ee60667553fca12bR40

This comment is a really long-winded way to say that I think this PR makes sense. Robb's question makes it clear that sampling alone won't deal with the `unpropagated` case. I'd love to see that added in this PR or a follow-on one.

johnnyshields:
> This comment is a really long-winded way to say that I think this PR makes sense.

Thanks for that 😅 Is this good to merge as-is then?

instrumentation/net_http/lib/opentelemetry/instrumentation/net/http/patches/instrumentation.rb
```diff
@@ -14,10 +14,16 @@ module Instrumentation
...
-            def request(req, body = nil, &block)
+            def request(req, body = nil, &block) # rubocop:disable Metrics/MethodLength
               # Do not trace recursive call for starting the connection
               return super(req, body, &block) unless started?
 
+              if untraced?
+                OpenTelemetry::Common::Utilities.untraced do
+                  return super(req, body, &block)
+                end
+              end
```
fbogsany, line 25: Do you need the `untraced` block here? Its only purpose is to avoid creating spans along a codepath where you may not have full control of the instrumentation. In this case, there should be no "deeper" instrumentation than the `kind=client` span we're creating below. I think it should be enough to:
```suggestion
              return super(req, body, &block) if untraced?
```

fbogsany:
Looking more closely at the code, `untraced_hosts` will also **NOT** propagate trace context (or any other OTel context) to those hosts. I'm not sure if that's the intent. As discussed above, it would be useful to trace-but-not-propagate in some cases. I'd like to see a clear statement of intent here, and documentation that clearly describes the behaviour of this option.

johnnyshields:
> I'd like to see a clear statement of intent here, and documentation that clearly describes the behaviour of this option.

@fbogsany my apologies I'm not quite getting your meaning here. I did merge your code suggestion, is it related to this?

instrumentation/net_http/lib/opentelemetry/instrumentation/net/http/instrumentation.rb
```diff
@@ -20,6 +20,8 @@ class Instrumentation < OpenTelemetry::Instrumentation::Base
             defined?(::Net::HTTP)
           end
 
+          option :untraced_hosts, default: [], validate: :array
```
fbogsany, line 23:
```suggestion
          ## Supported configuration keys for the install config hash:
          #
          # untraced_hosts: if a request's address matches any of the `String`
          #   or `Regexp` in this array, the instrumentation will not record a
          #   `kind = :client` representing the request and will not propagate
          #   context in the request.
          option :untraced_hosts, default: [], validate: :array
```

fbogsany:
> > I'd like to see a clear statement of intent here, and documentation that clearly describes the behaviour of this option.
> 
> @fbogsany my apologies I'm not quite getting your meaning here. I did merge your code suggestion, is it related to this?

I added a suggestion to document the configuration option. I'm 👍 with that change.

instrumentation/net_http/lib/opentelemetry/instrumentation/net/http/patches/instrumentation.rb
```diff
@@ -72,6 +74,14 @@ def annotate_span_with_response!(span, response)
             def tracer
               Net::HTTP::Instrumentation.instance.tracer
             end
+
+            def untraced?
+              return true if instrumentation_config[:untraced_hosts]&.any? do |host|
```
fbogsany, line 79:
```suggestion
              return true if Net::HTTP::Instrumentation.instance.config[:untraced_hosts]&.any? do |host|
```

johnnyshields:
OK works for me, done. Ready to merge?

ahayworth:
@johnnyshields the doc suggestion was [here](https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/108#discussion_r952702173) (if you want to do the honors) ... but the rubocop check is also failing 😢 

Afterwards: yes, I think so.

johnnyshields:
@ahayworth good to go.

ahayworth (approved):

ahayworth:
@fbogsany would you like to stamp this as well?

fbogsany (approved):

PR #109, opened 2022-07-23
Allow Rack :untraced_endpoints to include Regexp

johnnyshields:


instrumentation/rack/test/opentelemetry/instrumentation/rack/middlewares/tracer_middleware_test.rb
```diff
@@ -113,6 +113,20 @@
         end
       end
 
+      describe 'when a regexp is passed in' do
+        let(:config) { { untraced_endpoints: %r{\A/ping/?} } }
+
+        it 'traces everything' do
```
ahayworth, line 119: This seems like an odd test name...

instrumentation/rack/test/opentelemetry/instrumentation/rack/middlewares/tracer_middleware_test.rb
```diff
@@ -113,6 +113,20 @@
...
+          Rack::MockRequest.new(rack_builder).get('/ping', env)
+
+          ping_span = finished_spans.find { |s| s.attributes['http.target'] == '/ping' }
+          _(ping_span).wont_be_nil
```
ahayworth, line 123: If I pass `untraced_endpoints = %r{\A/ping/?}`, wouldn't I expect `must_be_nil` instead?

instrumentation/rack/lib/opentelemetry/instrumentation/rack/middlewares/tracer_middleware.rb
```diff
@@ -93,7 +93,9 @@ def call(env) # rubocop:disable Metrics/MethodLength
           private
 
           def untraced_request?(env)
-            return true if @untraced_endpoints.include?(env['PATH_INFO'])
+            return true if @untraced_endpoints.any? do |endpoint|
+              endpoint === env['PATH_INFO'] # rubocop:disable Style/CaseEquality
+            end
```
ahayworth, line 98: You've allowed a _lot_ more than just regular expressions this way. As noted [in this post](https://medium.com/rubyinside/triple-equals-black-magic-d934936a6379), triple-equals will test for a surprising range of truthiness.

For example, this would now allow us to pass:
- regular expressions (as intended)
- ranges
- classes
- procs
- ip addresses

...and we didn't want all of that. We should explicitly test for regular expressions if that's what we want to include, or the PR should state that it's allowing something much broader so we can figure out if we want that behavior or not.

ahayworth (commented):
Not sure about this in its present form - `===` is the kind of clever rubyisms that come back to bite you (in my opinion).

instrumentation/rack/test/opentelemetry/instrumentation/rack/middlewares/tracer_middleware_test.rb
johnnyshields: I've copy-pasted this from a related spec immediately above.

instrumentation/rack/test/opentelemetry/instrumentation/rack/middlewares/tracer_middleware_test.rb
johnnyshields: I've copy-pasted this from a related spec immediately above.

instrumentation/rack/lib/opentelemetry/instrumentation/rack/middlewares/tracer_middleware.rb
johnnyshields: `env['PATH_INFO']` will always be a String. Using `===` on any of those will return false, with the lone exception of the `String` class (`String === "foo" #=> true`)

Anyway, I can change it if you think it's an issue.

johnnyshields: Done.

instrumentation/rack/lib/opentelemetry/instrumentation/rack/middlewares/tracer_middleware.rb
```diff
@@ -93,7 +93,9 @@ def call(env) # rubocop:disable Metrics/MethodLength
           private
 
           def untraced_request?(env)
-            return true if @untraced_endpoints.include?(env['PATH_INFO'])
+            return true if @untraced_endpoints.any? do |endpoint|
+              endpoint.is_a?(Regexp) ? endpoint.match?(env['PATH_INFO']) : endpoint == env['PATH_INFO']
+            end
             return true if config[:untraced_requests]&.call(env)
```
robertlaurin, line 99: Curious why using the existing `untraced_requests` config option that allows for lambdas is insufficient here?

Added a test block to the existing rack instrumentation, and it appears it allows for the functionality you're looking to add.
```rb
        describe 'example using lambda' do
          let(:untraced_callable) do
            ->(env) { %r{\A/ping/?}.match?(env['PATH_INFO']) }
          end

          it 'does not trace things that match the regex' do
            Rack::MockRequest.new(rack_builder).get('/ping', env)

            ping_span = finished_spans.find { |s| s.attributes['http.target'] == '/ping' }
            _(ping_span).must_be_nil

            root_span = finished_spans.find { |s| s.attributes['http.target'] == '/' }
            _(root_span).wont_be_nil
          end
        end
      end
```

johnnyshields: Yep, I am doing same the currently, but its a lot simpler/less boilerplate in config to add a Regexp.

fbogsany: > why not remove `untraced_endpoints` entirely and have users do this?

It's a great idea. We didn't do that because we implemented `untraced_endpoints` first and we didn't want to break anyone who was already using it. It's also possible to configure `untraced_endpoints` via environment variable, but `callable` options can't be configured that way: https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/174e0fb80955522700f86671f1d156e797075e01/instrumentation/base/lib/opentelemetry/instrumentation/base.rb#L358-L362

instrumentation/rack/test/opentelemetry/instrumentation/rack/middlewares/tracer_middleware_test.rb
fbogsany: You've copy-pasted the wrong spec. This is only testing "the option expects an array", whereas we need a test for "when an array of regexes is passed in", based on the "when an array is passed in" spec.

instrumentation/rack/lib/opentelemetry/instrumentation/rack/middlewares/tracer_middleware.rb
johnnyshields: > It's a great idea

Hmmm... IMHO `untraced_endpoints` covers 80%+ of use cases here in a simpler fashion, I think its a good reason to keep it as its own option.

johnnyshields:
FYI the test failure here (rdkafka) looks unrelated to the changes in this PR.

kaylareopelle:
Hi @johnnyshields! Thanks for re-requesting a review on this PR. We're slowly trying to catch up on PRs that've been open for a long time.

@ahayworth, @fbogsany, @robertlaurin  - Have your concerns been resolved?

kaylareopelle:
cc [@open-telemetry/ruby-contrib-approvers](https://github.com/orgs/open-telemetry/teams/ruby-contrib-approvers) and [@open-telemetry/ruby-contrib-maintainers](https://github.com/orgs/open-telemetry/teams/ruby-contrib-maintainers) - please take a look if you're able!

arielvalentin:
Adding some additional work to this PR. 

This instrumentation now defaults to using `Rack::Events` as opposed to the classic middleware and will also need to be updated to support this option.  

https://github.com/open-telemetry/opentelemetry-ruby-contrib/tree/main/instrumentation/rack#rack-middleware-vs-rack-events

xuan-cao-swi (commented):
I think it's good idea for adding regex for untraced_endpoints, since I assume untraced_requests callable will have more complex Proc to process.

Issue #110, opened 2022-07-28
HTTP Header attribute name does not match semantic conventions


lilymara-onesignal:
According to the OpenTelemetry semantic conventions for HTTP traces, request headers should be recorded under the attribute prefix `http.request.header.#{x}`. This library reports `http.request.headers.#{x}`.This means there is a mismatch between this library and other implementations.

Code: https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/26a9b3b7cc6e37f71b29087ee06639aad6cd459a/instrumentation/rack/lib/opentelemetry/instrumentation/rack/middlewares/tracer_middleware.rb#L24

Spec: https://opentelemetry.io/docs/reference/specification/trace/semantic_conventions/http/#http-request-and-response-headers

ahayworth:
Thank you for pointing this out, it definitely looks like an oversight. #111 will fix it, as well! 😄

PR #111, opened 2022-07-29
fix: bring http.request.header and http.response.header in line with …

scbjans:
…semantic conventions.

Fixes: https://github.com/open-telemetry/opentelemetry-ruby-contrib/issues/110


ahayworth (approved):
Thank you! ❤️

arielvalentin (approved):

PR #113, opened 2022-08-03
Ensures the correct route is add to http.route span attribute

mcinbell:
The current implementation for the `enable_recognize_route` option will add the last of all matching routes to the `http.route` span attribute. In our experience, this has not been the correct route and we see all traces with a `http.route` of `/`. This fix sets the value based the longest match, rather than the last, which is working for us.

Update: To clarify the routing scenario this fix addresses, we have mounted rails engines on `/` and from further testing in our app it's _these_ routes that are being matched and turning up last in the order of matched routes and results in an incorrect value for `http.route`.

### DIFF
--- a/instrumentation/action_pack/lib/opentelemetry/instrumentation/action_pack/patches/action_controller/metal.rb
+++ b/instrumentation/action_pack/lib/opentelemetry/instrumentation/action_pack/patches/action_controller/metal.rb
@@ -27,11 +27,11 @@ module OpenTelemetry
             private
 
             def add_rails_route(rack_span, request)
-              matching_routes = []
               ::Rails.application.routes.router.recognize(request) do |route, _params|
-                matching_routes << route.path.spec.to_s
+                rack_span.set_attribute('http.route', route.path.spec.to_s)
+                # Rails will match on the first route - see https://guides.rubyonrails.org/routing.html#crud-verbs-and-actions
+                break
               end
-              rack_span.set_attribute('http.route', matching_routes.max_by(&:length))
             end
 
             def instrumentation_config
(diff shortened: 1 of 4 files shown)


ahayworth (commented):
This makes some intuitive sense, certainly. I'm wondering if we know where Rails does similar things internally? I would want to confirm that Rails is also taking the longest path, or something - if we're gonna fix it, we should _fix it_ you know? 😄

plantfansam:
Thanks for the contribution @mcinbell! I agree with @ahayworth, and maybe want to tack something on: we should check and see if this is what Rails does internally and document the results of that investigation with a code comment.

mcinbell:
Yup, fair comments. Not knowing the rails routing internals, I was assuming pattern length would be a reasonable proxy for specificity but rails could easily be just selecting the first match.

plantfansam (approved):
LGTM. [`.recognize`](https://github.com/rails/rails/blob/0c9e4069f71b139d1ea61ea1ba99e1982bd44bf2/actionpack/lib/action_dispatch/journey/router.rb#L65) runs a method called `find_routes` which looks to [order them by precedence](https://github.com/rails/rails/blob/0c9e4069f71b139d1ea61ea1ba99e1982bd44bf2/actionpack/lib/action_dispatch/journey/router.rb#L121). Returning the value yielded in the first iteration makes sense, to me.

mcinbell:
@plantfansam Thanks, are you able to approve running the workflows?

plantfansam:
@mcinbell looks like someone beat me to it — there was one (apparently) unrelated failure Rerunning failed specs now.

ahayworth:
If sidekiq instrumentation keeps failing on truffleruby, we can probably merge this anyways (I don't possibly see how it's related). Truffleruby support is intentionally best-effort and I don't want it to block this PR. 😄

ahayworth (approved):

PR #117, opened 2022-08-20
feat: Add :force_flush option to Resque instrumentation

nvolker:
## Summary

When a Resque worker picks up a job from the queue, it forks off a child process to perform the workload. The child then exits after it has finished performing the job and executed any defined **[job hooks](https://github.com/resque/resque/blob/master/docs/HOOKS.md#job-hooks)**.

The problem with this is that the child process is killed before the OTEL `span_processor` (in my case, the OTEL exporter's `BatchProcessor`) can flush its spans. Therefore the `send` span is exported from the parent process (the worker), but no `process` span is exported from the worker's forked child process.

The implementation has been adopted from the ActiveJob instrumentation. **[Code Link Here](https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/main/instrumentation/active_job/lib/opentelemetry/instrumentation/active_job/patches/active_job_callbacks.rb#L63)**

Associated PR: https://github.com/open-telemetry/opentelemetry-ruby/pull/726

## Related Conversation 

https://cloud-native.slack.com/archives/C01NWKKMKMY/p1660818697973079


instrumentation/resque/lib/opentelemetry/instrumentation/resque/instrumentation.rb
```diff
@@ -18,6 +18,28 @@ class Instrumentation < OpenTelemetry::Instrumentation::Base
...
+        # propagation_style: controls how the job's execution is traced and related
+        #   to the trace where the job was enqueued. Can be one of:
+        #
+        #   - :link (default) - the job will be executed in a separate trace. The
+        #     initial span of the execution trace will be linked to the span that
+        #     enqueued the job, via a Span Link.
+        #   - :child - the job will be executed in the same logical trace, as a direct
+        #     child of the span that enqueued the job.
+        #   - :none - the job's execution will not be explicitly linked to the span that
+        #     enqueued the job.
```
nvolker, line 40: Happy to split this, unrelated, commentary into a secondary, smaller PR.

instrumentation/resque/lib/opentelemetry/instrumentation/resque/instrumentation.rb
```diff
@@ -18,6 +18,28 @@ class Instrumentation < OpenTelemetry::Instrumentation::Base
           defined?(::Resque)
         end
 
+        ## Supported configuration keys for the install config hash:
+        #
+        # force_flush: when `true`, all completed spans will be synchronously flushed
+        #   at the end of a job's execution (default: `false`). You will likely wish to
+        #   enable this option for job systems that fork worker processes such as Resque.
```
fbogsany, line 25: It's a little odd to read `such as Resque` in instrumentation specifically for Resque. I *think* this option is useful when running Resque in a certain mode, but it's been a couple years since I was intimate with Resque. Can you clarify this a little?

instrumentation/resque/lib/opentelemetry/instrumentation/resque/instrumentation.rb
fbogsany: This doc is fantastic - very much appreciated. ❤️

instrumentation/resque/lib/opentelemetry/instrumentation/resque/instrumentation.rb
nvolker: Good catch, this is what I get for adopting and not adapting the ActiveJob implementation. 

Updated: https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/117/commits/88c18b457edb641e672cb30c948383684edb7794

instrumentation/resque/lib/opentelemetry/instrumentation/resque/instrumentation.rb
nvolker: Kudos goes to @ahayworth for doing the heavy lifting.

instrumentation/resque/lib/opentelemetry/instrumentation/resque/instrumentation.rb
nvolker: As for forking, as far as I know, this is the default behaviour for Resque **[Fork Documentation Link](https://github.com/resque/resque/#forking)**. While the documentation does begin with `"On certain platforms..."`, I am only familiar with using Resque when workers fork off child processes.

nvolker: I could append something like this to the `:force_flush` comment, what do you think?

```ruby
"... You will likely wish to enable this option if `Resque::Worker#fork_per_job?` is enabled."
```

ahayworth: I guess we're quibbling over semantics, but if it's `you will likely wish` then it makes me wonder "why isn't it the default?" (I know why, I'm just thinking as I would if this were my first time seeing it).

So: yes, I think adding that to the doc is wise, I would say "You may wish" instead.

instrumentation/resque/lib/opentelemetry/instrumentation/resque/instrumentation.rb
```diff
@@ -18,6 +18,27 @@ class Instrumentation < OpenTelemetry::Instrumentation::Base
           defined?(::Resque)
         end
 
+        ## Supported configuration keys for the install config hash:
+        #
+        # force_flush: when `true`, all completed spans will be synchronously flushed
+        #   at the end of a job's execution (default: `false`).
```
ahayworth, line 24:
```suggestion
        #   at the end of a job's execution (default: `false`). You may wish to
        #   enable this option if `Resque::Worker#fork_per_job?` is enabled.
```

ahayworth (approved):
LGTM! I added the minor docs thing as a suggestion if you want to just click the button and accept it.

instrumentation/resque/lib/opentelemetry/instrumentation/resque/instrumentation.rb
nvolker: Agreed!

instrumentation/resque/lib/opentelemetry/instrumentation/resque/instrumentation.rb
nvolker: Love suggestions, thanks. Committed! https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/117/commits/f7cd0c14458974f4a2c70f9ac481aee3cd2c1198

fbogsany (approved):

nvolker:
Just checking why Ruby 3.1 is sad with me:
```
  1) Error:
OpenTelemetry::Instrumentation::Resque::Patches::ResqueJob::#perform::force_flush::true#test_0001_does forcibly flush the tracer:
MockExpectationError: No more expects available for :force_flush: [] {}
    /home/runner/work/opentelemetry-ruby-contrib/opentelemetry-ruby-contrib/instrumentation/resque/lib/opentelemetry/instrumentation/resque/patches/resque_job.rb:60:in `perform'
    /home/runner/work/opentelemetry-ruby-contrib/opentelemetry-ruby-contrib/instrumentation/resque/test/opentelemetry/instrumentation/resque/patches/resque_job_test.rb:229:in `work_off_jobs'
    /home/runner/work/opentelemetry-ruby-contrib/opentelemetry-ruby-contrib/instrumentation/resque/test/opentelemetry/instrumentation/resque/patches/resque_job_test.rb:216:in `block (6 levels) in <top (required)>'
    /home/runner/work/opentelemetry-ruby-contrib/opentelemetry-ruby-contrib/instrumentation/resque/test/opentelemetry/instrumentation/resque/patches/resque_job_test.rb:215:in `block (5 levels) in <top (required)>'

23 runs, 57 assertions, 0 failures, 1 errors, 0 skips
```

https://github.com/open-telemetry/opentelemetry-ruby-contrib/runs/7973702409?check_suite_focus=true

nvolker:
@fbogsany, @ahayworth would either of you mind approving the running of the workflow? 🙏

ahayworth:
CI runs approved (I thought I did that already, weird).

nvolker:
> CI runs approved (I thought I did that already, weird).

My new commit is likely the culprit. Perhaps you have to approve CI each time?

instrumentation/resque/lib/opentelemetry/instrumentation/resque/instrumentation.rb
ahayworth: What if we asked the job instead?

The job class seems to have a [`worker` accessor](https://github.com/resque/resque/blob/8ea2e4b41bae7cf4c1242b81b25eaa39b9ccaf86/lib/resque/job.rb#L67), and while there is no explicit accessor, the [`fork_per_job?` method on the worker](https://github.com/resque/resque/blob/8ea2e4b41bae7cf4c1242b81b25eaa39b9ccaf86/lib/resque/worker.rb#L792-L795) is not private and memoizes the result ... so it should be low-impact to check on every job.

I could envision the code looking something like this:

```ruby
if worker.fork_per_job?
  OpenTelemetry.tracer_provider.force_flush unless !!!config[:force_flush]
end
```

... but that only works, of course, if `config[:force_flush]` can be left in an _unset_ state by default. If we could do that, then the "do we flush" truth table looks like:

|                                 | `fork_per_job? == true` | `fork_per_job? == false` |
|---------------------------------|-------------------------|--------------------------|
| `config[:force_flush] == true`  | ✅                      | ✅                       |
| `config[:force_flush] == false` | 🚫                      | 🚫                       |
| `config[:force_flush] == nil`   | ✅                      | 🚫                       |

... I don't know if allowing this option to be `nil` is something we want. But it would get around the brittleness of trying to check this at startup, and would also avoid force-flushing needlessly on platforms where that isn't happening.

ahayworth:
Maybe... 🤷

instrumentation/resque/lib/opentelemetry/instrumentation/resque/instrumentation.rb
fbogsany: If we go down that path, I'd recommend turning the option into an enum, e.g.:
```ruby
        option :force_flush,       default: :ask_the_job,  validate: %I[ask_the_job always never]
```
As noted above:
> Ideally, we'd be careful here not to break forks of Resque.

We should check `Hedwig`, for example, to make sure we don't break it. (We should also check whether we're using this instrumentation with `Hedwig` - if it has its own instrumentation, this is a moot point.)

fbogsany: I checked. We have services using the upstream `Resque` with this instrumentation. `Hedwig` does not use this instrumentation - it has its own first-party instrumentation using OpenTelemetry. So I think this is safe from Shopify's perspective.

nvolker: I like the direction this conversation has taken and I am in favour of the enum suggestion made by @fbogsany. If we can land on a consensus within the maintainer group I'd be happy to push a commit that reflects this suggestion.

nvolker: For the sake of conversation, I'd pushed [b07e9ed](https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/117/commits/b07e9ed0bb4d1a933bb3d7d4dc1857285934a07a).

nvolker: Not too much urgency here, but I would like to get this merged soonish as it will unblock enabling the resque instrumentation on our side (and avoid a monkey-patch).

ahayworth (approved):
Assuming CI passes, LGTM!

nvolker:
Oooof. The `Minitest::Mock` is still not working out for me on Ruby 3. I tried giving each scenario its own mock to try to combat this flakiness, but at this point I'm unsure. 

Some input on this would be appreciated!

instrumentation/resque/lib/opentelemetry/instrumentation/resque/instrumentation.rb
```diff
@@ -18,8 +18,38 @@ class Instrumentation < OpenTelemetry::Instrumentation::Base
...
+        #
+        # force_flush: controls if spans are forcibly flushed upon job completion
+        #   - :ask_the_job (default) - if `Resque::Worker#fork_per_job?` is set,
+        #     all completed spans will be synchronously flushed at the end of a
+        #     job's execution
+        #   - :always - all completed spans will be synchronously flushed at the
+        #     end of a job's execution
+        #   - :never - the job will not intervene with the processing of spans
+        #
+        # when `true`, all completed spans will be synchronously flushed
```
robbkidd, line 31: I ❤️  the documentation addition!

But I'm confused: when _what_ is `true`? `force_flush`? Is `true` an acceptable value for the `force_flush` config option? That seems to conflict with the list of symbols above and the `validate: %I[ask_the_job always never]` validations below.

instrumentation/resque/lib/opentelemetry/instrumentation/resque/instrumentation.rb
```diff
@@ -18,8 +18,38 @@ class Instrumentation < OpenTelemetry::Instrumentation::Base
...
+        #   at the end of a job's execution (default: `false`). You may wish to
+        #   enable this option if `Resque::Worker#fork_per_job?` is enabled.
```
robbkidd, line 33: What do you think of including the reason the reader may wish this?

I _think_ the recommendation is because the spans are queued in a processor+exporter instance that lives in a background thread of the forked worker process and so without force flushing, the spans there are at likely to remain unsent due to batch timing and then discarded when Sidekiq ends the forked process.

… And this pairs with my question about "when `true` …" at the beginning of this paragraph. Would I set `force_flush` to `:ask_the_job` or to `true` for the synchronous, blocking send behavior?

robbkidd (commented):
❤️ 

I have some non-blocking questions about the documentation added.

instrumentation/resque/lib/opentelemetry/instrumentation/resque/instrumentation.rb
nvolker: Good catch @robbkidd! This is a derp. This section should be deleted. Commit incoming.

nvolker: Peace has been restored [4855a4e](https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/117/commits/4855a4e3dcc2475fc4df7c761b5500f8286e7c67)

nvolker: Let me know what you think now @robbkidd, I'll refrain from responding to your other comment. Apologies for the confusion.

robbkidd (approved):
Sorry for the delay on re-review!

Looking good!

arielvalentin:
@nvolker would you mind taking a look at this again? It seems the test suite is non-deterministic and fails occasionally:

e.g. https://github.com/open-telemetry/opentelemetry-ruby-contrib/actions/runs/3790399834/jobs/6444991776#step:4:357

scbjans:
hey @nvolker, @arielvalentin 
I'm quite interested in this PR, so I took the liberty to check why the tests are flaky.

It seems that _sometimes_ the Resque job are not processed fast enough; so that calling `work_off_jobs` returns another job than expected.

I've altered the `before` to always delete all current jobs:
```
before do
    ::Resque::Job.destroy(:super_urgent, ActiveJob::QueueAdapters::ResqueAdapter::JobWrapper)
    ::Resque::Job.destroy(:super_urgent, DummyJob)
    ::Resque::Job.destroy(:super_urgent, BaggageTestingJob)
    ::Resque::Job.destroy(:super_urgent, ExceptionTestingJob)
    instrumentation.install(config)
    exporter.reset
  end
  ```
  
  That way, I cannot reproduce the flakyness anymore.
  It doesn't look really nice; so hopefully that can be cleaned up a bit; but I hope my $0.02 help.

scbjans:
@arielvalentin kind request to rerun CI for this PR. I believe https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/302 should have fixed the issues.

arielvalentin:
@scbjans My apologies. I missed the request and allowed this PR to become stale. I have merged main and am waiting on a rebuild.

arielvalentin:
@nvolker Looks like we have linter errors on this now with latest Rubocop https://github.com/open-telemetry/opentelemetry-ruby-contrib/actions/runs/4346678153/jobs/7593036414

PR #118, opened 2022-08-23
feat: support mongodb db statement without obfuscation

nozik:
Other Ruby instrumentation (e.g. Redis) support three db statement modes: omit, include and obfuscate. Currently, the include mode behaves like obfuscate. The change isn't backwards compatible (in case someone explicitly used "include"), but it's consistent with other instrumentations. Also, the current state is misleading as I'd expect "include" not to do anything with the statement

### DIFF
--- a/instrumentation/mongo/README.md
+++ b/instrumentation/mongo/README.md
@@ -35,10 +35,10 @@ end
 ```ruby
 OpenTelemetry::SDK.configure do |c|
   c.use 'OpenTelemetry::Instrumentation::Mongo', {
-    # The obfuscation of arguments in the db.statement attribute is enabled by default.
-    # To include the full query, set db_statement to :include.
-    # To obfuscate, set db_statement to :obfuscate.
-    # To omit the attribute, set db_statement to :omit.
+    # Sets how db_statement appears in your traces. The options are:
+    # :obfuscate - (default) query parameters are visible, but values are masked.
+    # :include - query parameters and values are both fully visible.
+    # :omit - db_statement is omitted entirely.
     db_statement: :include,
   }
 end
--- a/instrumentation/mongo/README.md
+++ b/instrumentation/mongo/README.md
@@ -30,6 +30,20 @@ OpenTelemetry::SDK.configure do |c|
 end
 ```
 
+### Configuration options
+
+```ruby
+OpenTelemetry::SDK.configure do |c|
+  c.use 'OpenTelemetry::Instrumentation::Mongo', {
+    # The obfuscation of arguments in the db.statement attribute is enabled by default.
+    # To include the full query, set db_statement to :include.
+    # To obfuscate, set db_statement to :obfuscate.
+    # To omit the attribute, set db_statement to :omit.
+    db_statement: :include,
+  }
+end
+```
+
 ## Example
 
 To run the example:
--- a/instrumentation/mongo/lib/opentelemetry/instrumentation/mongo/command_serializer.rb
+++ b/instrumentation/mongo/lib/opentelemetry/instrumentation/mongo/command_serializer.rb
@@ -14,10 +14,11 @@ module OpenTelemetry
 
         attr_reader :command, :command_name, :collection, :payload
 
-        def initialize(command)
+        def initialize(command, obfuscate)
           @command = command
           @command_name, @collection = command.first
-          @collection = MASK_VALUE unless @collection.is_a?(String) || @collection.is_a?(Integer)
+          @obfuscate = obfuscate
+          @collection = MASK_VALUE if @obfuscate && !(@collection.is_a?(String) || @collection.is_a?(Integer))
           @payload = {}
         end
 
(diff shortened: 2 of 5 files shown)


ahayworth (approved):
This seems alright to me, even if it's a breaking change. I'm curious what other @open-telemetry/ruby-contrib-approvers / @open-telemetry/ruby-contrib-maintainers think.

Can you do me a favor @nozik and reword the commit to have the title be `feat!: support mongodb db statement without obfuscation`? We use [conventional commits](https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/main/CONTRIBUTING.md#use-conventional-commit-messages) and our release tooling will then properly flag that as a breaking change in the changelogs.

arielvalentin:
I believe am in the minority here but I prefer obfuscation by default. I also do not consider this a breaking change but I think we should document the default behavior in the README.

nozik:
@arielvalentin Notice that the default behavior is still obfuscation

arielvalentin (approved):

ericmustin (changes requested):
can we have a test for the `obfuscate` option to prevent a regression please?

ericmustin:
small nit on test coverage, otherwise i think this is fine and won't argue against a majority.

my personal preference is that we should add this feature, keep the instrumentation defaulted to `include`, and label this a bug fix.

chesterton's fence here for this repo is that while this "obfuscate by default" behavior came during the [original port](https://github.com/open-telemetry/opentelemetry-ruby/pull/448/files) , the original `quantize by default` functionality arose from the [original author's](https://github.com/DataDog/dd-trace-rb/commit/9ae5b716af6d6b197d93786f64286fa3f91c2eb8#diff-49df72e8991cbd71985ce093f52dfa05f580c7fb2ae894a7c7b57f0115c5ff6fR30-R35) use of the obfuscated `db.statement` value as a Span `Resource`, which was a span field in their tracing system that has cardinality considerations similar to those of a OpenTelemetry Span Name.

from: https://github.com/DataDog/dd-trace-rb/commit/9ae5b716af6d6b197d93786f64286fa3f91c2eb8#diff-49df72e8991cbd71985ce093f52dfa05f580c7fb2ae894a7c7b57f0115c5ff6fR30-R35
```
      # removes the values from the given query; this quantization recursively
      # replace elements available in a given query, so that Arrays, Hashes and so
      # on are compacted. It ensures a low cardinality so that it can be used
      # as a Span resource.
      # TODO: the quantization may not be enough, we still need to check the
      # cardinality if it's reasonable
```

In our case, I don't think this "obfuscate by default" approach is relevant since `db.statement` is merely a span attribute in our case, and so cardinality considerations that original authors had wouldn't be relevant. They'd only apply if someone wanted to generate, say, RED Metrics on a per `db.statement` basis (in which case the attribute would have to be heavily normalized/quantized/etc). 

As far as defaulting to `include` vs `obfuscate` for security/pii posture reasons, i'd prefer that we keep the same posture as our other db instrumentations (default to `include`, though i think the spec _allows_ for either choice)

 https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/database.md#call-level-attributes

nozik:
@ericmustin Done, added a test that explicitly configures "obfuscate"

johnnyshields:
@ericmustin strongly believe that the default should be "obfuscate". There can potentially be personally identifiable information (PII) in queries. The principle of least surprise dictates that we should not "surprisingly" violate the GDPR / CCPA etc.

instrumentation/mongo/README.md
```diff
@@ -30,6 +30,20 @@ OpenTelemetry::SDK.configure do |c|
...
 
+### Configuration options
+
+```ruby
+OpenTelemetry::SDK.configure do |c|
+  c.use 'OpenTelemetry::Instrumentation::Mongo', {
+    # The obfuscation of arguments in the db.statement attribute is enabled by default.
+    # To include the full query, set db_statement to :include.
+    # To obfuscate, set db_statement to :obfuscate.
+    # To omit the attribute, set db_statement to :omit.
```
johnnyshields, line 41: Suggest the following:

```
    # Sets how db_statement appears in your traces. The options are:
    # :obfuscate - (default) query parameters are visible, but values are masked.
    # :include - query parameters and values are both fully visible.
    # :omit - db_statement is omitted entirely.
```

nozik: Done

instrumentation/mongo/lib/opentelemetry/instrumentation/mongo/command_serializer.rb
```diff
@@ -14,10 +14,11 @@ class CommandSerializer
 
         attr_reader :command, :command_name, :collection, :payload
 
-        def initialize(command)
+        def initialize(command, obfuscate)
           @command = command
           @command_name, @collection = command.first
-          @collection = MASK_VALUE unless @collection.is_a?(String) || @collection.is_a?(Integer)
+          @obfuscate = obfuscate
+          @collection = MASK_VALUE if @obfuscate && !(@collection.is_a?(String) || @collection.is_a?(Integer))
```
johnnyshields, line 21: If I recall correctly, the intention on this line was more about marking invalid data as "?" rather than obfuscating sensitive data; the collection name should always be a String AFAIK. Would suggest to revert this line or do something different without `@obfuscate` variable.

nozik: I personally don't see a reason to set a mask value unless we're obfusacting - happy to change the behavior if that's what's agreed upon by everyone.

nozik: Done

arielvalentin:
Thank you @johnnyshields for following up with your review on this PR. We will hold off on merging until you have had the opportunity to review this round of commits.

nozik:
@johnnyshields Are we ready to merge? Would be happy to address additional comments if necessary. Thanks!

johnnyshields:
@arielvalentin @nozik we're good, let's merge!

PR #121, opened 2022-09-02
redis-rb 5.0 and redis-client support

casperisfine:
redis-rb 5.0 ditched its internal client and now simply delegate to the `redis-client` gem.

@fbogsany @robertlaurin

### DIFF
--- a/instrumentation/redis/lib/opentelemetry/instrumentation/redis/instrumentation.rb
+++ b/instrumentation/redis/lib/opentelemetry/instrumentation/redis/instrumentation.rb
@@ -16,7 +16,7 @@ module OpenTelemetry
         end
 
         present do
-          defined?(::Redis)
+          defined?(::Redis) || defined?(::RedisClient)
         end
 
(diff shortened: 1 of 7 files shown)


instrumentation/redis/lib/opentelemetry/instrumentation/redis/instrumentation.rb
```diff
@@ -7,6 +7,73 @@
 module OpenTelemetry
   module Instrumentation
     module Redis
+      # Adapter for redis-client instrumentation interface
+      module ClientInstrumentation
```
robertlaurin, line 11: Lets move this module into its own file, either `patches/redis_client` or I guess `middlewares/redis_client` might be more appropriate here as well because its not a patch anymore.

robertlaurin (commented):
One comment on moving the location of the new module and then its good to go.

robertlaurin (approved):
Thanks for the PR

plantfansam (approved):
LGTM. Thanks for the PR!

PR #123, opened 2022-09-14
feat: name ActionPack spans with the HTTP method and route

dmathieu:
Many SDKs are naming HTTP spans with `#{HTTP_METHOD} #{ROUTE}`, for example `GET /user/{id}`.
While naming the span with the `ControllerName#action` works fine, it causes discrepancies in span names between requests that make it through rails, and others (such as the sidekiq admin interface for example).

With this change, setting `:span_naming` to `:controller_action` will keep the old behavior. The default behavior will be to name the span with the HTTP method and the route.


arielvalentin:
@dmathieu I am seeing this error in other projects as well. Looks like there may be some something wrong with gem metadata on rubygems.org

instrumentation/action_pack/lib/opentelemetry/instrumentation/action_pack/patches/action_controller/metal.rb
```diff
@@ -14,7 +14,12 @@ module Metal
             def dispatch(name, request, response)
               rack_span = OpenTelemetry::Instrumentation::Rack.current_span
               if rack_span.recording?
-                rack_span.name = "#{self.class.name}##{name}" unless request.env['action_dispatch.exception']
+                unless request.env['action_dispatch.exception']
+                  rack_span.name = case instrumentation_config[:span_naming]
+                                   when :route then "#{request.method} #{rails_route(request)}"
```
arielvalentin, line 19: If I am following along here, this will result in the route being computed multiple times when both `span_naming: :route` and `enable_recognize_route: true` are set. 

@robertlaurin Is there any reason why `enable_recognize_route` should _not_ rename the span instead of adding a separate `http.route` attribute?

Also, do any of the instrumentations add Rails specific code attributes? https://opentelemetry.io/docs/reference/specification/trace/semantic_conventions/span-general/#source-code-attributes

| semconv attribute | |
| --- | --- |
| `code.namespace` | `self.class.name` |
| `code.function` | `name` |

dmathieu: Yes, the route is computed twice. We could compute it once and set it in a variable. But we would always be computing it, even when we don't need it.

I think having both the span name and a separate `http.route` attribute has its usefulness, especially if the span name is method + route.

As for the code attributes, it does seem adding them with controller and action would be nice. I'm happy to do that, either in this PR or in another one.

arielvalentin: > I think having both the span name and a separate http.route attribute has its usefulness, especially if the span name is method + route.

The latter is a question I want to answer. Should we _always_ use the recognized route as the span name instead of using the class and method name?

dmathieu: That sounds good to me. I didn't want to introduce breaking change, but I agree it's a better default.
I've made the change, with an option to go back to the old format.

arielvalentin: @dmathieu Thank you again for your submission. I would like to give the original author(s), e.g. @robertlaurin, the opportunity to get a look at this PR and provide feedback, however he is not available for a few more days AFAIK.  

Thank you for your patience.

instrumentation/action_pack/lib/opentelemetry/instrumentation/action_pack/patches/action_controller/metal.rb
```diff
@@ -13,10 +13,16 @@ module ActionController
           module Metal
             def dispatch(name, request, response)
               rack_span = OpenTelemetry::Instrumentation::Rack.current_span
+              route = rails_route(request)
```
defeated, line 16: @dmathieu looks like `route` is only used within `if rack_span.recording?`, so should this be moved inside that condition?

instrumentation/action_pack/lib/opentelemetry/instrumentation/action_pack/patches/action_controller/metal.rb
arielvalentin: SemConv backs up using the templated route as the span name:

> Given an inbound request for a route (e.g. "/users/:userID?") the name attribute of the span SHOULD be set to this route. If the route does not include the application root, it SHOULD be prepended to the span name.

https://opentelemetry.io/docs/reference/specification/trace/semantic_conventions/http/#http-server-semantic-conventions

arielvalentin:
@robertlaurin A gentle reminder that I would like your feedback here before we make any decisions about next steps. Going to also ping you in slack...

instrumentation/action_pack/lib/opentelemetry/instrumentation/action_pack/patches/action_controller/metal.rb
```diff
@@ -14,9 +14,15 @@ module Metal
             def dispatch(name, request, response)
               rack_span = OpenTelemetry::Instrumentation::Rack.current_span
               if rack_span.recording?
-                rack_span.name = "#{self.class.name}##{name}" unless request.env['action_dispatch.exception']
+                route = rails_route(request)
```
robertlaurin, line 17: One blocker here, we should only calculate the rails route either the span naming is configured to use the normalized path, or if the http.route `enable_recognize_route` configuration is enabled.

The reason being as it is written right now if neither of those options are configured we are still paying the cost of calculating route even if it does not get used anywhere.

I'm aware that this is nitpicky, but the concern here is that we want to minimize the cost associated with instrumenting applications. Most code gets slow slowly through many changes of this nature. However I do think this is a valuable change as it introduces consistency and compliance with the semantic conventions. 

We just want to make sure we allow users to opt out if they choose.

I think we could memoize this here to get the desired effect, however I don't necessarily expect you to use my suggestion below so long as the code is changed to support avoiding calculating the route when it doesn't get used.
```rb
           def dispatch(name, request, response)
              if rack_span.recording?
                unless request.env['action_dispatch.exception']
                  rack_span.name = case instrumentation_config[:span_naming]
                                   when :controller_action then "#{self.class.name}##{name}"
                                   else "#{request.method} #{rails_route(request)}"
                                   end
                end

                rack_span.set_attribute('http.route', rails_route(request)) if instrumentation_config[:enable_recognize_route]

                rack_span.set_attribute('http.target', request.filtered_path) if request.filtered_path != request.fullpath
              end

              super(name, request, response)
            end

            private

            def rails_route(request)
              @rails_route ||= ::Rails.application.routes.router.recognize(request) do |route, _params|
                return route.path.spec.to_s
                # Rails will match on the first route - see https://guides.rubyonrails.org/routing.html#crud-verbs-and-actions
              end
            end
```

robertlaurin (commented):
One major comment on making sure we don't calculate the route if it doesn't get used. Otherwise I'm in favour of making this the new default as it supports the semantic conventions, which I think is the correct default.

instrumentation/action_pack/lib/opentelemetry/instrumentation/action_pack/patches/action_controller/metal.rb
arielvalentin: @robertlaurin I would like to confirm that `Metal` object instances are reinitialized per request. Introducing instance variables here may result in _incorrect routes being recorded_ if the same object instance is reused for different requests. 

Do you happen to know how `Metal` allocates object instances?

robertlaurin: I was under the impression that controller lifecycles start and end on a per request basis, as rack is the simplest form of a controller I expected that to hold true for this. However I'm not confident enough in my memory to assert this, we should either confirm this or choose a different approach. I have no preference here, so long as the route look up doesn't occur needlessly.

dmathieu: It seems the [Metal class itself](https://github.com/rails/rails/blob/8015c2c2cf5c8718449677570f372ceb01318a32/actionpack/lib/action_controller/metal.rb#L247) is using memoization (with the request and response). So I assume we can use it as well.

I have switched this PR to do so, with an additional test checking that memoization doesn't reuse previous values.

arielvalentin (approved):

robertlaurin (approved):

instrumentation/action_pack/lib/opentelemetry/instrumentation/action_pack/patches/action_controller/metal.rb
```diff
@@ -26,11 +31,10 @@ def dispatch(name, request, response)
 
             private
 
-            def add_rails_route(rack_span, request)
-              ::Rails.application.routes.router.recognize(request) do |route, _params|
-                rack_span.set_attribute('http.route', route.path.spec.to_s)
+            def rails_route(request)
+              @rails_route ||= ::Rails.application.routes.router.recognize(request) do |route, _params|
+                return route.path.spec.to_s
```
casperisfine, line 36: For context I'm investigating a bug where I have evidence that this PR cause `request.script_name = "/"`, resulting in duplicate `/` in `request.url` e.g. `request.url => "http://example.com//foo"`.

I haven't fully pinpointed what is going on, but in the meantime this change seem very suspect.

If you `return` here then this memoization is never gonna work. So might be totally orthogonal, but this is very fishy.

casperisfine: So what is happening in our app is that we reach this line: https://github.com/rails/rails/blob/a51e9676e611f578bdfd10fcf6891d9e2d762a23/actionpack/lib/action_dispatch/journey/router.rb#L68

We end up matching a catch all engine (`#<ActionDispatch::Journey::Route:0x000000012f5ae0d0 @name="security_reports_engine" ...`) which isn't anchored, and cause this bug.

Overall I don't think it's safe to call `recognize` where it's not expected.

arielvalentin: @casperisfine do you have a suggestion for what this should do instead or is it unsafe to use recognize_route at all?

casperisfine: I'm not that familiar with the router, but yeah I'd suggest to wait for the router to be called by rails and grab the route there instead. (just a guess, might not be possible).

We just turned off that option.

PR #129, opened 2022-09-25
fix: Rename HTTP CONNECT for low level connection spans

arielvalentin:
The current span name that representes `Net::HTTP#start` is a little misleading and should be renamed to `HTTP SESSION` to better represent the span.

According to `Net::HTTP#start` RDoc the `start` method is used to:

> Opens a TCP connection and HTTP session.
> When this method is called with a block, it passes the Net::HTTP object to the block, and closes the TCP connection and HTTP session after the block has been executed.
> When called with a block, it returns the return value of the block; otherwise, it returns self.
<https://ruby-doc.org/stdlib-3.1.2/libdoc/net/http/rdoc/Net/HTTP.html#method-i-start>

These operations are lower level operations that establish SSL connections and initiate an HTTP session, which is distinct from using the HTTP Protocol `CONNECT` method which is used to:

> starts two-way communications with the requested resource. It can be used to open a tunnel.
> For example, the CONNECT method can be used to access websites that use SSL (HTTPS).
> The client asks an HTTP Proxy server to tunnel the TCP connection to the desired destination.
> The server then proceeds to make the connection on behalf of the client.
> Once the connection has been established by the server, the Proxy server continues to proxy the TCP stream to and from the client.
> `CONNECT www.example.com:443 HTTP/1.1`

<https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/CONNECT>

`Net::HTTP` does use the `HTTP CONNECT` method but _only_ when the `Net::HTTP` object is configured to use a proxy:

https://github.com/ruby/net-http/blob/196f3d70f9ce154dc63a86a2cdbf1aec4df4982e/lib/net/http.rb#L1031

If we want to measure the time it takes to complete an `HTTP CONNECT` method request the code should create a separate span _or_ check the `proxy?` predicate.


instrumentation/net_http/lib/opentelemetry/instrumentation/net/http/patches/instrumentation.rb
```diff
@@ -57,7 +57,7 @@ def connect
                 OpenTelemetry::SemanticConventions::Trace::NET_PEER_PORT => conn_port
               }.merge!(OpenTelemetry::Common::HTTP::ClientContext.attributes)
 
-              tracer.in_span('HTTP CONNECT', attributes: attributes) do
+              tracer.in_span('HTTP SESSION', attributes: attributes) do
```
fbogsany, line 60: The recommendation in the semantic conventions is:
> HTTP client spans SHOULD be using conservative, low cardinality names formed from the available parameters of an HTTP request, such as `"HTTP {METHOD_NAME}"`.

While this is a recommendation, it's reasonable for users to assume HTTP client spans named `HTTP SESSION` represent HTTP requests with a method of `SESSION`, which ... isn't a thing.

I'd prefer it we checked for `use_ssl? && proxy?` and kept `HTTP CONNECT` as the span name in that case. If we also want to trace session establishment in the non-SSL-proxy case, we should probably do that with a different span. Ideally, that'd be an internal span (I think) that is always created on this path, wrapping the `HTTP CONNECT` client span in the SSL-proxy case, but that's challenging to implement without being incredibly brittle. In any case, it likely shouldn't be a HTTP client span, since it is lower-level and doesn't actually speak HTTP.

arielvalentin: > While this is a recommendation, it's reasonable for users to assume HTTP client spans named HTTP SESSION represent HTTP requests with a method of SESSION, which ... isn't a thing.

💯 I mentioned to the folks on the SIG call today that this was a terrible name and I was looking for ideas on a better name. I figured it was better to push out a PR with a bad name and get some feedback from folks and see if we could come up with a better name. 

> I'd prefer it we checked for use_ssl? && proxy? and kept HTTP CONNECT as the span name in that case. If we also want to trace session establishment in the non-SSL-proxy case, we should probably do that with a different span. Ideally, that'd be an internal span (I think) that is always created on this path, wrapping the HTTP CONNECT client span in the SSL-proxy case, but that's challenging to implement without being incredibly brittle. In any case, it likely shouldn't be a HTTP client span, since it is lower-level and doesn't actually speak HTTP.

Are you are asking me to include this implementation as part of this PR? i.e. conditionally add an `HTTP CONNECT` PR as well as add an internal spans to trace session establishment? Or is renaming the existing span OK?

arielvalentin: added a few more branches there.

fbogsany:
Screen Shot 2022-09-30 at 2 54 21 PM
💯

instrumentation/net_http/lib/opentelemetry/instrumentation/net/http/patches/instrumentation.rb
```diff
@@ -57,10 +58,19 @@ def connect
...
               }.merge!(OpenTelemetry::Common::HTTP::ClientContext.attributes)
 
-              tracer.in_span('HTTP CONNECT', attributes: attributes) do
+              if use_ssl? && proxy?
+                span_name = 'HTTP CONNECT'
+                span_kind = :client
+              else
+                span_name = 'connect'
+                span_kind = :internal
+              end
```
fbogsany, line 67: This is cool. Cleaner than I thought.

fbogsany (approved):

PR #130, opened 2022-09-28
feat: add public_endpoint option to rack middleware

dmathieu:
Public endpoints may receive requests that include an external `traceparent` header, making the SDK believe there is a parent span, when that span will never be found in the system, since it was emitted by whatever client made the request.

Example: A public API, where the client is using tracing as well, and therefore sends trace headers with every request.

This fix proposal is similar to what the [Go SDK is doing](https://github.com/open-telemetry/opentelemetry-go-contrib/blob/61ea3c1c30be3f63628aa72c02f9570c9ed11b85/instrumentation/net/http/otelhttp/handler.go#L132-L138).

It introduces a new `public_endpoint` option for the rack instrumentation, which is a lambda that takes two arguments: `env` from rack, and the context we just computed.

In our case for example, we inject some specific key/value to tracestate to identify that a request is coming from our internal systems.

### DIFF
--- a/instrumentation/rack/lib/opentelemetry/instrumentation/rack/instrumentation.rb
+++ b/instrumentation/rack/lib/opentelemetry/instrumentation/rack/instrumentation.rb
@@ -30,6 +30,7 @@ module OpenTelemetry
         option :untraced_endpoints,       default: [],    validate: :array
         option :url_quantization,         default: nil,   validate: :callable
         option :untraced_requests,        default: nil,   validate: :callable
+        option :public_endpoint,          default: nil,   validate: :callable
 
         private
 
--- a/instrumentation/rack/lib/opentelemetry/instrumentation/rack/middlewares/tracer_middleware.rb
+++ b/instrumentation/rack/lib/opentelemetry/instrumentation/rack/middlewares/tracer_middleware.rb
@@ -66,10 +66,7 @@ module OpenTelemetry
             end
 
             original_env = env.dup
-            extracted_context = OpenTelemetry.propagation.extract(
-              env,
-              getter: OpenTelemetry::Common::Propagation.rack_env_getter
-            )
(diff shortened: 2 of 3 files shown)


arielvalentin:
@dmathieu Thank you for your contribution. 

We briefly discussed this PR at the Ruby SIG 2022-10-11, if you have a moment please watch the discussion here: https://youtu.be/T20fKAUgeMM?t=2766

I think that @mwear @robertlaurin had the most insight related to the use of tracestate and how this should require some additional discussion with the specification working group/SIG. I will open the floor to them to add any additional thoughts or concerns about adding this functionality to the instrumentation.

mwear:
To summarize the discussion (briefly) I believe that the intended way to solve the problem you are running into is to use the [`tracestate` header](https://www.w3.org/TR/trace-context/#tracestate-header). Essentially, you would add an entry to `tracestate` when a trace enters your system. You can store whatever you want there (although size is a concern to be aware of), but for practical purposes, you'd want to store at least the current span id. When your system receives a request, you'd look at both the `traceparent` and your `tracestate` entry to make a decision on how you want to report the parent span id. If the parent id from tracestate matches the parent id in `traceparent`, you know this is a continuation of a trace that started in your system. If you don't have an entry in tracestate, or the parent ids do not match, this trace (likely) came from outside of your system and you can decide how to record the span for your purposes. I don't know that this scenario, or its interaction with auto-instrumentation has been discussed at the specification level. I'll bring it up at the spec meeting tomorrow and report back.

mwear:
I brought this up at the spec SIG earlier today. Everyone was in agreement that trust boundaries for trace context need to be established at the Spec level. There was an action item for the TC to come up with some proposals. The tracestate solution I mentioned did not seem very popular. I'll keep you updated on the progress as things evolve.

Issue #56, opened 2022-06-20
Sinatra instrumentation does not add caught exceptions to the span


ahayworth:
**Description of the bug**

The Sinatra auto-instrumentation does not record exceptions as error logs on the span, despite using our `in_span` helper. This is because Sinatra does not necessarily barf on exceptions unless configured to do so - much like Rails, it catches those exceptions and returns an appropriate HTTP status code but it does not _throw_ the exception. Thus, `in_span` never has anything to `rescue`. This feels like a bug to me, but I'm not sure.

**Share details about your runtime**

Operating system details: `Darwin janeway.local 21.5.0 Darwin Kernel Version 21.5.0: Tue Apr 26 21:08:29 PDT 2022; root:xnu-8020.121.3~4/RELEASE_ARM64_T8101 arm64`
RUBY_ENGINE: `ruby`
RUBY_VERSION: `3.1.2`
RUBY_DESCRIPTION: `ruby 3.1.2p20 (2022-04-12 revision 4491bb740a) [arm64-darwin21]`

**Share a simplified reproduction if possible**

You could trivially reproduce this by adding a `raise "boom"` into any of the examples we ship. The logged span from one such experiment looks like:

```
#<struct OpenTelemetry::SDK::Trace::SpanData
 name="GET /thing/:id",
 kind=:server,
 status=#<OpenTelemetry::Trace::Status:0x0000000104ae8eb8 @code=2, @description="">,
 parent_span_id="\x00\x00\x00\x00\x00\x00\x00\x00",
 total_recorded_attributes=4,
 total_recorded_events=0,
 total_recorded_links=0,
 start_timestamp=1655760015092533000,
 end_timestamp=1655760015092967000,
 attributes={"http.method"=>"GET", "http.url"=>"/thing/12345", "http.status_code"=>500, "http.route"=>"/thing/:id"},
 links=nil,
 events=nil,
 resource=
  #<OpenTelemetry::SDK::Resources::Resource:0x0000000104677050
   @attributes=
    {"service.name"=>"unknown_service",
     "process.pid"=>25942,
     "process.command"=>"./server.rb",
     "process.runtime.name"=>"ruby",
     "process.runtime.version"=>"3.1.2",
     "process.runtime.description"=>"ruby 3.1.2p20 (2022-04-12 revision 4491bb740a) [arm64-darwin21]",
     "telemetry.sdk.name"=>"opentelemetry",
[...]

PR #133, opened 2022-10-03
fix: add exceptions to sinatra spans.

scbjans:
This will add any caught Sinatra exceptions to the Span.

Fixes: https://github.com/open-telemetry/opentelemetry-ruby-contrib/issues/56

(edit: updated screenshot so that it correctly reflects outcome of this PR: only server side errors will mark a span as in error and attach an exception to the span.)
Screenshot 2022-10-05 at 08-41-43 Jaeger UI


scbjans:
Although the screenshot shows a span in error; I've removed that commit from this PR, since that should be handled in the Rack middleware.
If that is indeed the case; I'm happy to look into that in a separate PR.

instrumentation/sinatra/lib/opentelemetry/instrumentation/sinatra/middlewares/tracer_middleware.rb
```diff
@@ -27,6 +27,7 @@ def trace_response(env)
 
             span.set_attribute('http.route', env['sinatra.route'].split.last) if env['sinatra.route']
             span.name = env['sinatra.route'] if env['sinatra.route']
+            span.record_exception(env['sinatra.error']) if env['sinatra.error']
```
arielvalentin, line 30: Based on the screen shots from your example, it looks like Sinatra reports HTTP 404 `Sinatra::NotFound` in `sinatra.error`. 

I _think_ this code should only record an exception for 5xx responses or exceptions that would result in a 500 error. 

Similar to what `dd-trace` does: https://github.com/DataDog/dd-trace-rb/blob/master/lib/datadog/tracing/contrib/sinatra/tracer_middleware.rb#L78

arielvalentin: I really appreciate the time you are taking to look into addressing my concerns. 

> Long story short, this PR now only adds an Exception event, but does not mark the span as in Error; see below screenshot.

I do not think we should record exceptions for non-exceptional conditions that result in 4xx range _client_ errors (or any other ranges 1xx-3xx). 

I am a bit firm on _only_ using `span.record_exception` in cases where there is a _server side error_ i.e. 5xx range because they are issues related to the Sinatra application service that we are tracing. _Any objections or affirmations from the @open-telemetry/ruby-contrib-approvers?_

I do not have a lot of familiarity with Sinatra so I am afraid I do not have any recommendations other than digging into how ddtrace is able to disambiguate an exception that resulted in a server side error.

scbjans: thanks @arielvalentin and @ahayworth for your feedback.
Using a `404` for illustrating that we can add exceptions without marking the span as erroneous probably wasn't the best idea.

Anyway; https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/133/commits/edacd694e237d433446fba5f63dc88cc9205b664 fixes this, such that an exception and marking the span as erroneous is only done when there is a server error >500.

instrumentation/sinatra/test/opentelemetry/instrumentation/sinatra_test.rb
```diff
@@ -142,9 +142,10 @@
         'http.status_code' => 404,
         'http.target' => '/missing_example/not_present'
       )
+      _(exporter.finished_spans.first.events).must_be_nil
```
arielvalentin, line 145: Non-blocking suggesting to make this a little more functional.  

```suggestion
      _(exporter.finished_spans.flat_map(&:events).must_be_empty
```

instrumentation/sinatra/test/opentelemetry/instrumentation/sinatra_test.rb
```diff
@@ -156,6 +157,7 @@
         'http.scheme' => 'http',
         'http.target' => '/error'
       )
+      _(exporter.finished_spans.first.events.first.name).must_equal('exception')
```
arielvalentin, line 160: Non-blocking suggesting to make this a little more functional.  

```suggestion
      _(exporter.finished_spans.flat_map(&:events).map(&:name)).must_equal(['exception'])
```

instrumentation/sinatra/lib/opentelemetry/instrumentation/sinatra/middlewares/tracer_middleware.rb
```diff
@@ -16,17 +16,24 @@ def initialize(app)
...
-          def trace_response(env)
+          def trace_response(env, response)
             span = OpenTelemetry::Instrumentation::Rack.current_span
             return unless span.recording?
 
             span.set_attribute('http.route', env['sinatra.route'].split.last) if env['sinatra.route']
             span.name = env['sinatra.route'] if env['sinatra.route']
+
+            #response looks like [HTTP code, headers, [response body]]
+            sinatra_response = ::Sinatra::Response.new([], response.first)
```
arielvalentin, line 32: 🤔 `response.first`

What does it mean for there to be more than 1 response? Is response an array or enumerable structure?

arielvalentin: Oh silly me. That's the rack response array.

arielvalentin (approved):

ahayworth (approved):
LGTM!

PR #138, opened 2022-10-09
fix: Windows Ruby 3.1 Native Extensions Installs

arielvalentin:
Newer versions of gems were released for sqlite3 that have precompiled native dependenices included in the gem.

This change makes having to compile native extensions at install time unnecessary.

<https://github.com/ruby/setup-ruby#bundle-config> <https://github.com/sparklemotion/sqlite3-ruby#avoiding-the-precompiled-native-gem>

### DIFF
--- a/instrumentation/action_view/Gemfile
+++ b/instrumentation/action_view/Gemfile
@@ -11,5 +11,7 @@ gemspec
 group :test do
   gem 'opentelemetry-instrumentation-active_support', path: '../active_support'
   gem 'opentelemetry-instrumentation-base', path: '../base'
+  gem 'nokogiri', force_ruby_platform: true
   gem 'pry-byebug'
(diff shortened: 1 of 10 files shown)


instrumentation/action_view/Gemfile
```diff
@@ -11,5 +11,7 @@ gemspec
 group :test do
   gem 'opentelemetry-instrumentation-active_support', path: '../active_support'
   gem 'opentelemetry-instrumentation-base', path: '../base'
+  gem 'nokogiri', force_ruby_platform: true
```
johnnyshields, line 14: can you try:

```ruby
  gem 'nokogiri', platforms: [:ruby, :x64_mingw]
```

This should at least work on Windows Ruby 3.0 and earlier, and if you are installing a relatively recent Bundler, it will work on Windows Ruby 3.1 as well.

PR #142, opened 2022-10-11
fix: only take the first item in a comma-separated list for pg attrs

ahayworth:
In a recent upstream PR[^1], a change was made that introduced the
possibility of the `conninfo_hash` containing a comma-separated list if
the hostname we connect to resolves to multiple IP addresses.

In the tests, we say to connect to `localhost`: and sure enough, that
resolves to both `127.0.0.1` and `::1`. :facepalm:

I considered simply relaxing the validation here, but we'd break
semantic conventions: `net.peer.port` must be an integer, not a string -
so we can't just pass through the unmodified values. The values will
always be duplicate, comma-separated strings however: the actual
difference shows up in `net.peer.ip`.

Anyways, we add a helper function that tries to be smart about not doing
more work than is necessary.

[^1]: https://github.com/ged/ruby-pg/pull/485

### DIFF
--- a/instrumentation/pg/lib/opentelemetry/instrumentation/pg/patches/connection.rb
+++ b/instrumentation/pg/lib/opentelemetry/instrumentation/pg/patches/connection.rb
@@ -12,7 +12,7 @@ module OpenTelemetry
     module PG
       module Patches
         # Module to prepend to PG::Connection for instrumentation
-        module Connection
+        module Connection # rubocop:disable Metrics/ModuleLength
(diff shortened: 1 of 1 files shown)


arielvalentin (approved):

ahayworth:
To make it a little more clear: if you pass `localhost` as the connection host, then `conninfo_hash` looks like:
```
host: localhost,localhost
port: 5432,5432
address: 127.0.0.1, ::1
```

PR #147, opened 2022-10-13
Only set db statement if recording in mysql2 instrumentation

plantfansam:
Obfuscating db statement is potentially expensive, so only do it if current span is recording. Checking if the current span is recording also entails doing some computation, though, so we ought to benchmark this.

Code is pretty gnarly. Probably there's a simpler way to achieve this, but wanted to put a POC out there.

cc @aneda


instrumentation/mysql2/lib/opentelemetry/instrumentation/mysql2/patches/client.rb
```diff
@@ -52,23 +52,31 @@ module Client
...
-              attributes['db.statement'] = obfuscate_sql(sql)
-            end
+
             tracer.in_span(
               database_span_name(sql),
               attributes: attributes.merge!(OpenTelemetry::Instrumentation::Mysql2.attributes),
               kind: :client
-            ) do
+            ) do |span|
+              if OpenTelemetry::Trace.current_span.recording? && !OpenTelemetry::Instrumentation::Mysql2.attributes["db.statement"]
```
plantfansam, line 61: So we don't `merge!` attributes twice. Probably could do this more cleanly.

ahayworth: I kinda think it's not needed in general? In what circumstances would this situation actually happen? (Maybe it can, I'm just not sure when, where, or if it matters much)

instrumentation/mysql2/lib/opentelemetry/instrumentation/mysql2/patches/client.rb
```diff
@@ -52,23 +52,31 @@ module Client
...
 
+          def make_db_statement_attr(sql)
+            case config[:db_statement]
+            when :include
+              sql
+            when :obfuscate
+              obfuscate_sql(sql)
+            end
+          end
+
```
fbogsany, line 79:
```suggestion
```

instrumentation/mysql2/lib/opentelemetry/instrumentation/mysql2/patches/client.rb
```diff
@@ -52,23 +52,31 @@ module Client
...
-            when :include
-              attributes['db.statement'] = sql
-            when :obfuscate
-              attributes['db.statement'] = obfuscate_sql(sql)
-            end
+
             tracer.in_span(
               database_span_name(sql),
               attributes: attributes.merge!(OpenTelemetry::Instrumentation::Mysql2.attributes),
               kind: :client
```
fbogsany, line 59:
```suggestion
            overrides = OpenTelemetry::Instrumentation::Mysql2.attributes
            statement = config[:db_statement]
            statement = :override if overrides.key?('db.statement')

            attributes = client_attributes.merge!(overrides)
            attributes['db.statement'] = sql if statement == :include

            tracer.in_span(database_span_name(sql), attributes: attributes, kind: :client) do |span|
```

instrumentation/mysql2/lib/opentelemetry/instrumentation/mysql2/patches/client.rb
```diff
@@ -52,23 +52,31 @@ module Client
 
           def query(sql, options = {})
             attributes = client_attributes
```
fbogsany, line 54:
```suggestion
```

instrumentation/mysql2/lib/opentelemetry/instrumentation/mysql2/patches/client.rb
fbogsany: We need to be careful of a few things:
1. Overrides should always take precedence, including over the config.
2. As many attributes as possible should be supplied eagerly.
3. Hash merging and the implicit mutex access in `set_attribute` should be minimized.

instrumentation/mysql2/lib/opentelemetry/instrumentation/mysql2/patches/client.rb
```diff
@@ -52,23 +52,31 @@ module Client
...
+
             tracer.in_span(
               database_span_name(sql),
               attributes: attributes.merge!(OpenTelemetry::Instrumentation::Mysql2.attributes),
               kind: :client
-            ) do
+            ) do |span|
+              if OpenTelemetry::Trace.current_span.recording? && !OpenTelemetry::Instrumentation::Mysql2.attributes["db.statement"]
+                span.add_attributes({ 'db.statement' => make_db_statement_attr(sql) })
+              end
```
fbogsany, line 63:
```suggestion
              span['db.statement'] = obfuscate_sql(sql) if OpenTelemetry::Trace.current_span.recording? && statement == :obfuscate
```

fbogsany (changes requested):

instrumentation/mysql2/lib/opentelemetry/instrumentation/mysql2/patches/client.rb
plantfansam: Co-sign what @fbogsany said. I think there's a reasonable case for explicitly overriding `db.statement` for applications that default to obfuscating or redacting mysql `db.statement` attributes but then explicitly override `db.statement` using `.attributes` when the know that the query is safe to include.

instrumentation/mysql2/lib/opentelemetry/instrumentation/mysql2/patches/client.rb
plantfansam: maybe s/`statement =`/`statement_config =`

instrumentation/mysql2/lib/opentelemetry/instrumentation/mysql2/patches/client.rb
```diff
@@ -51,18 +51,16 @@ module Client
...
-              attributes['db.statement'] = sql
-            when :obfuscate
-              attributes['db.statement'] = obfuscate_sql(sql)
-            end
-            tracer.in_span(
-              database_span_name(sql),
-              attributes: attributes.merge!(OpenTelemetry::Instrumentation::Mysql2.attributes),
-              kind: :client
-            ) do
+            overrides = OpenTelemetry::Instrumentation::Mysql2.attributes
```
plantfansam, line 54: @fbogsany do we need to set this to a variable?

instrumentation/mysql2/lib/opentelemetry/instrumentation/mysql2/patches/client.rb
ahayworth: I generally agree with what is said, but I will `:homer-bushes:` here because I think I'm lacking some context to really know what I'm talking about. 😆 

Never mind me, I was never here! 😄

instrumentation/mysql2/lib/opentelemetry/instrumentation/mysql2/patches/client.rb
```diff
@@ -51,18 +51,16 @@ module Client
...
+            statement = config[:db_statement]
+            statement = :override if overrides.key?('db.statement')
+
+            attributes = client_attributes.merge!(overrides)
+            attributes['db.statement'] = sql if statement == :include
+
+            tracer.in_span(database_span_name(sql), attributes: attributes, kind: :client) do |span|
+              span['db.statement'] = obfuscate_sql(sql) if OpenTelemetry::Trace.current_span.recording? && statement == :obfuscate
```
robbkidd, line 62:
```suggestion
              span['db.statement'] = obfuscate_sql(sql) if statement == :obfuscate && OpenTelemetry::Trace.current_span.recording?
```

> Checking if the current span is recording also entails doing some computation, though, ...

If the `recording?` check is expensive, maybe switch the order of this guard to early return false when obfuscation is not enabled?

>  ... so we ought to benchmark this.

Was this ever benchmarked?

PR #149, opened 2022-10-14
refactor: perf tweak for metal attr append

robertlaurin:
Set attributes grabs a mutex, so lets only do it once, if we need to.

https://github.com/open-telemetry/opentelemetry-ruby/blob/47156c198d8a2214913b0932768500bc9c0fb9cc/sdk/lib/opentelemetry/sdk/trace/span.rb#L80


arielvalentin (approved):

robertlaurin:
@dmathieu did a quick follow up tweak I didn't think of right away on your PR here https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/123

PR #156, opened 2022-10-16
ci: Split Builds Into Multiple Workflows

arielvalentin:
Hoping that this makes tracking down broken builds a little easier


.github/workflows/ci-instrumentation.yml
```diff
@@ -0,0 +1,131 @@
...
+          - restclient
+          - rspec
+          - sinatra
+        os:
+          - ubuntu-latest
+          - macos-latest
+          - windows-latest
+        exclude:
+          - os: windows-latest
+            gem: ethon
```
robbkidd, line 50: Are there comments we could add (now or in followup PRs) about why these tests need / ought to be excluded for these OSes?

robbkidd (approved):
I like it.

PR #157, opened 2022-10-20
fix: declare span_naming option in action_pack instrumentation

boncri:
The `span_naming` option is not declared in `OpenTelemetry::Instrumentation::ActionPack::Instrumentation`, so this option is ignored if the class is configured using `OpenTelemetry::SDK.configure` method.

The `span_naming` is validated against 2 values: `:controller_action` and `:rails_route` and the default value is `:rails_route`.


arielvalentin (commented):
Good catch!  

May I ask you accompany this with a unit test so we may avoid regressions in the future?

arielvalentin (approved):

robertlaurin (approved):

PR #158, opened 2022-10-23
docs: Add information about ActionPack instrumentation span naming

richardmcmillen:
Since https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/123 has been merged I think the ActionPack README should be updated to reflect this new option and default behaviour.

This pull requests adds some documentation to satisfy this.

I have left the somewhat ominous, but not explained, line:

>This attribute is disabled by default because we use .recognize

I will create an issue for proposing some changes to the instrumentation and will address it there, as I would like to change the default setting for this attribute from `false` to `true`.

### DIFF
--- a/instrumentation/action_pack/README.md
+++ b/instrumentation/action_pack/README.md
@@ -32,7 +32,26 @@ end
 
 ### Configuration options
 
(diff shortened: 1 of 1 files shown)


richardmcmillen:
I am closing this pull request in favour of this proposal: https://github.com/open-telemetry/opentelemetry-ruby-contrib/issues/159

I will be more than happy to put together a pull request for that if it is a good direction.

PR #160, opened 2022-10-24
fix: handle encoding errors in mysql obfuscation

robertlaurin:
😭 
```
ArgumentError: invalid byte sequence in UTF-8
gems/opentelemetry-instrumentation-mysql2-0.21.0/lib/opentelemetry/instrumentation/mysql2/patches/client.rb:76 gsub	
gems/opentelemetry-instrumentation-mysql2-0.21.0/lib/opentelemetry/instrumentation/mysql2/patches/client.rb:76 obfuscate_sql	
gems/opentelemetry-instrumentation-mysql2-0.21.0/lib/opentelemetry/instrumentation/mysql2/patches/client.rb:59 query
```

### DIFF
--- a/instrumentation/mysql2/lib/opentelemetry/instrumentation/mysql2/patches/client.rb
+++ b/instrumentation/mysql2/lib/opentelemetry/instrumentation/mysql2/patches/client.rb
@@ -145,6 +145,7 @@ module OpenTelemetry
             QUERY_NAME_RE.match(sql) { |match| match[1].downcase } unless sql.nil?
           rescue StandardError => e
             OpenTelemetry.logger.debug("Error extracting sql statement type: #{e.message}")
+            nil
           end
         end
       end
--- a/instrumentation/mysql2/lib/opentelemetry/instrumentation/mysql2/patches/client.rb
+++ b/instrumentation/mysql2/lib/opentelemetry/instrumentation/mysql2/patches/client.rb
@@ -9,7 +9,7 @@ module OpenTelemetry
     module Mysql2
       module Patches
         # Module to prepend to Mysql2::Client for instrumentation
-        module Client
+        module Client # rubocop:disable Metrics/ModuleLength
           QUERY_NAMES = [
             'set names',
             'select',
@@ -73,10 +73,13 @@ module OpenTelemetry
             if sql.size > 2000
               'SQL query too large to remove sensitive data ...'
             else
-              obfuscated = sql.gsub(generated_mysql_regex, '?')
+              obfuscated = OpenTelemetry::Common::Utilities.utf8_encode(sql)
+              obfuscated = obfuscated.gsub(generated_mysql_regex, '?')
               obfuscated = 'Failed to obfuscate SQL query - quote characters remained after obfuscation' if detect_unmatched_pairs(obfuscated)
               obfuscated
             end
+          rescue StandardError
+            'OpenTelemetry error: failed to obfuscate sql'
           end
 
           def generated_mysql_regex
--- a/instrumentation/mysql2/lib/opentelemetry/instrumentation/mysql2/patches/client.rb
+++ b/instrumentation/mysql2/lib/opentelemetry/instrumentation/mysql2/patches/client.rb
@@ -73,7 +73,7 @@ module OpenTelemetry
             if sql.size > 2000
               'SQL query too large to remove sensitive data ...'
             else
-              obfuscated = OpenTelemetry::Common::Utilities.utf8_encode(sql)
+              obfuscated = OpenTelemetry::Common::Utilities.utf8_encode(sql, binary: true)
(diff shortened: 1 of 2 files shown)


ericmustin (approved):

plantfansam (approved):
👍

instrumentation/mysql2/lib/opentelemetry/instrumentation/mysql2/patches/client.rb
```diff
@@ -73,10 +73,13 @@ def obfuscate_sql(sql)
...
               'SQL query too large to remove sensitive data ...'
             else
-              obfuscated = sql.gsub(generated_mysql_regex, '?')
+              obfuscated = OpenTelemetry::Common::Utilities.utf8_encode(sql)
+              obfuscated = obfuscated.gsub(generated_mysql_regex, '?')
               obfuscated = 'Failed to obfuscate SQL query - quote characters remained after obfuscation' if detect_unmatched_pairs(obfuscated)
               obfuscated
             end
+          rescue StandardError
+            'OpenTelemetry error: failed to obfuscate sql'
```
ahayworth, line 82: Is it possible to make a more specific error message here?

robertlaurin: I wasn't sure how specific to make it, this is what will get plunked into the the db.statement attribute field. I'm being prudent about capturing any information about what was failed to be obfuscated.

What did you have in mind? Like the error class or something?

robertlaurin:
I mean technically true is not nil.
https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/cbcad7656e7552ad00d479ce76e37d801e170a7f/instrumentation/mysql2/lib/opentelemetry/instrumentation/mysql2/patches/client.rb#L98-L112

instrumentation/mysql2/lib/opentelemetry/instrumentation/mysql2/patches/client.rb
ahayworth: Ah, I didn't realize that this would get put into the `db.statement` field, which is my bad for not looking closely enough. I think this is fine. 👍

instrumentation/mysql2/lib/opentelemetry/instrumentation/mysql2/patches/client.rb
```diff
@@ -142,6 +145,7 @@ def extract_statement_type(sql)
             QUERY_NAME_RE.match(sql) { |match| match[1].downcase } unless sql.nil?
           rescue StandardError => e
             OpenTelemetry.logger.debug("Error extracting sql statement type: #{e.message}")
+            nil
```
robertlaurin, line 148: This error is covered by my new test.

instrumentation/mysql2/lib/opentelemetry/instrumentation/mysql2/patches/client.rb
```diff
@@ -73,10 +73,13 @@ def obfuscate_sql(sql)
             if sql.size > 2000
               'SQL query too large to remove sensitive data ...'
             else
-              obfuscated = sql.gsub(generated_mysql_regex, '?')
+              obfuscated = OpenTelemetry::Common::Utilities.utf8_encode(sql, binary: true)
```
robertlaurin, line 76: We're setting binary to true in redis, and dalli. 
https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/6df0122cbe829a410b7b00b1b732b578e27c5476/instrumentation/redis/lib/opentelemetry/instrumentation/redis/middlewares/redis_client.rb#L70
https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/6df0122cbe829a410b7b00b1b732b578e27c5476/instrumentation/dalli/lib/opentelemetry/instrumentation/dalli/utils.rb#L57

So I think it makes sense here. 
https://github.com/open-telemetry/opentelemetry-ruby/blob/18bfd391f2bda2c958d5d6935886c8cba61414dd/common/lib/opentelemetry/common/utilities.rb#L40-L63

ericmustin:
> Ya'll need to stop approving my failing PR lol

@robertlaurin this is merely a reflection of how high your trust battery charge is with me

ericmustin (approved):

instrumentation/mysql2/lib/opentelemetry/instrumentation/mysql2/patches/client.rb
fbogsany: Probably not something you need to address here, but it'd be useful to use a common prefix for all the "error" type substitutions (this and the two above) - it'd make it easier for Observability teams to monitor this behaviour. Metrics *might* be nice as well, but a competent engineer can build metrics from the `db.statement` in a span processor or in the collector.

fbogsany (approved):

PR #162, opened 2022-10-26
feat: Add ActionDispatch#request_id to rack span as 'action_dispatch.request_id'

### LINKED ISSUE #126: ActionDispatch Instrumentation - add request_id to the root span
I am currently working on instrumenting an established Rails application with OpenTelemetry. I was hoping to see the ActionDispatch request_id in the resulting trace. But looking at this repo I don't see ActionDispatch instrumented at all.

Here is the middleware that deals with the request_id:

* https://api.rubyonrails.org/classes/ActionDispatch/RequestId.html
* https://github.com/rails/rails/blob/main/actionpack/lib/action_dispatch/middleware/request_id.rb

I think adding instrumentation to set an attribute on the rack span would be useful for others too.

I would like to propose adding some instrumentation middleware for setting an attribute on the trace root. I'm unsure what the attribute should be named, not being super familiar with the standard, but the best one I have thought of would be `action_dispatch.request_id` to avoid conflict with other middleware or frameworks which may also wish to do the same.

I think all that would be needed here would be to ensure the instrumentation middleware is inserted after the ActionDispatch::RequestId middleware. The code could look something like this:

```ruby
module SomeMiddlewareName
  class TracingMiddlewareName
    def initialize(app)
      @app = app
    end

    def call(env)
      if env['action_dispatch.request_id']
        rack_span.set_attribute("action_dispatch.request_id", env['action_dispatch.request_id'])
      end

      status, headers, body = @app.call(env)

      [status, headers, body]
    end
  end
end
```

I am seeking guidance on this idea and approach before taking this any further.  Many thanks.

richardmcmillen:
ActionDispatch::RequestId is default middleware for Rails, it accepts a request ID from a header if it exists, otherwise it generates a new request ID with SecureRandom.uuid.

The default header used is X-Request-Id, though it is possible to configure which header to look in when initializing the middleware.

ActionDispatch::Request#uuid is an alias for ActionDispatch::Request#request_id

Ref: https://api.rubyonrails.org/classes/ActionDispatch/RequestId.html

As RequestId is the name of the middleware and the default header, this adds the new attribute 'action_dispatch.request_id' to the rack span.

Closes #126


ahayworth (approved):
This looks alright to me! We usually want two approvals on most PRs, so I'll wait to see if any other of the @open-telemetry/ruby-contrib-approvers want to weigh in - but this seems pretty straightforward.

Thank you! 😄 ❤️

instrumentation/action_pack/lib/opentelemetry/instrumentation/action_pack/patches/action_controller/metal.rb
```diff
@@ -24,6 +24,7 @@ def dispatch(name, request, response)
                 attributes_to_append = {}
                 attributes_to_append['http.route'] = rails_route(request) if instrumentation_config[:enable_recognize_route]
                 attributes_to_append['http.target'] = request.filtered_path if request.filtered_path != request.fullpath
+                attributes_to_append['action_dispatch.request_id'] = request.request_id if request&.request_id
```
fbogsany, line 27: According to [semantic conventions](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/http.md#http-request-and-response-headers) (which we should strive to adhere to in `-contrib` auto-instrumentation), this should be either `http.request.header.x_request_id` or `http.response.header.x_request_id`. The field is technically non-standard, but it is pretty common. Given that ActionDispatch bases RequestId on `X-Request-Id` if it is available in the request, I think we should pick one of those instead.

robertlaurin: As a follow on to the comment made by Francis, should we direct users to instead configure their rack middleware to do this work? 

https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/main/instrumentation/rack/lib/opentelemetry/instrumentation/rack/middlewares/tracer_middleware.rb#L19-L36

richardmcmillen: Thanks for the quick review and feedback everyone!

I will say that I didn't know about the `allowed_request_headers` option in Rack middleware. I might make a separate PR to document it and other features in the README.

That would definitely work for me, although, it does put the effort on the users to configure it, rather than getting it for free as part of standard rails instrumentation. If that is preferred then I am more than happy to close this PR and add something else around documentation. But I will address the other feedback above nevertheless.

Thanks for pointing out the semantic conventions for headers @fbogsany, given ActionDispatch::RequestId middleware is configurable, as in you can tell it to look in another header, for example 'X-Acme-Id', but still exposes the contents the header through the `ActionDispatch::Request.request_id` method do you still think we would always put the contents into the `http.request.header.x_request_id` even though it may not reflect where it was actually sourced. If we wanted to use the configured header in the attribute name I can't see a convenient way of finding what header was selected.

This is where I see the benefit of using something departed from headers as the attribute name - like `action_dispatch.request_id`. This is a rails framework feature, that happens to use headers, and using an attributes inline with the API could avoid the above confusion.

With all that being said, if you'd prefer to not incorporate this and instead document how to add headers to traces I can certainly do that. Thanks again.

richardmcmillen: In case I have swayed you accept the PR, perhaps a better attribute name would be `action_dispatch.request.request_id`? Allowing for other attributes to be nested?

I would also like to remove the safe operator here `request&.request_id` I think it is safe to assume we have the `request` object at this point and if anything we want to protect against it not responding to `#request_id`.

ahayworth: @fbogsany your point about the HTTP header conventions is well-taken, but I must respectfully disagree on that point. The request ID is typically carried in a header, yes, but I think there are a few reasons to treat this differently.

#### The request ID can be carried in any arbitrary header, or none at all

We can't _just_ stick the request ID in something like `http.request.header.x_request_id`, because if the user configures it to be carried instead in something like `X-Github-Request-Id` instead, then sticking it inside `http.request.header.x_request_id` would arguably be much more confusing.

We could vary the attribute we use based on where the request ID actually originated from, but as far as I can see in the Rails codebase we won't necessarily _know_ where that is coming from. We could inspect configuration, but that doesn't save us from the fact that Rails will _generate_ a request ID if one isn't found in the default header or the configured header for it. So we don't really know where it came from, and thus it's hard to say which http header semantic attribute should carry it.

#### Request / Response captured headers _should_ be allow-listed

If we're going to hew closely to the semantic conventions, they also say this:

> Instrumentations SHOULD require an explicit configuration of which headers are to be captured. Including all request/response headers can be a security risk - explicit configuration helps avoid leaking sensitive information.

Reasonable people can interpret that differently, and I'm inclined to say "this header is an exception!" of course. But if we are going to say that we must follow semantic conventions here, then we should _really_ follow them. 🙃 🙃 

But following it closely and completely means that this _should_ be opt-in; and I actually don't think we should do that. Capturing the Rails request ID seems like a very reasonable, automatic thing that our instrumentation should be doing because it increases the ability to correlate traces with signals from other areas (logging, downstream systems) that may not even have tracing. Our instrumentation is actually quite magical and usable in many ways, and I am loathe to take changes that move us away from that good user experience without a good reason.

(On the other hand, forcing people to specifically tell us which header is allowed for this feature also means it clarifies some of the "we don't know where it comes from" problem previously. `:lolsob:`)

#### A Request ID is really more than "just" a header

This reason is not something that I can really support with any kind of evidence, it's just an opinion. But: some things are carried in headers, but actually have a far stronger semantic meaning than "they are a header." It does more in this case than just signal something about the request... it gets passed into log lines, in many places it gets propagated downstream as a pseudo trace id, etc. It is important enough in Rails that it gets code in the framework to refer to it in an abstract way, and even to generate it if its not present. It's not merely a header.

We could also look to other examples here - trace IDs are typically carried in headers, but our abstractions around them do not require us to always deal with that fact whenever we use them (that would be horrifyingly inconvenient). We abstract that away and treat it as a higher-level concept than just something that showed up in a header.

---

So ... I disagree. I think this should be a top-level attribute that our instrumentation captures as something distinct from HTTP headers. It's not just a header, treating it as such would be a step back in user experience, and treating it as a header is arguably more complicated than this PR's approach.

arielvalentin: @ahayworth @fbogsany any resolution here?

robbkidd: I am in favor towards `action_dispatch.request_id` (or `action_dispatch.request.request_id`).

I am persuaded at the moment that recording the `request_id` is as "an aspect of the web framework component being instrumented" and not strictly part of the HTTP conventions, despite the ID potentially having been conveyed over an HTTP header vs generated by the middleware.

Also, as we shift more Rails instrumentation over to subscriptions on [ActiveSupport::Notifications events](https://edgeguides.rubyonrails.org/active_support_instrumentation.html), there will be other framework/component-specific fields on those events that will be useful to include on telemetry emitted by the OTel instrumentation. I was not able to find semconv recommendations for namespacing frameworky data. 😞

PR #167, opened 2022-10-28
Bump instrumentation-all 

ahayworth:
Bumps minimum gem versions to latest patch releases for everything in `-all`; and we also add the rake/rspec instrumentations to the gem.


ahayworth:
Ugh I can't github today

PR #174, opened 2022-11-02
test: Fix Rails tests for new ActionPack instrumentation default behaviour

richardmcmillen:
* ActionPack instrumentation now defaults enabled_recognize_route to true, and this means the http.route is added by default.

* This also corrects the tests description and adds the code.namspace and code.function attributes

--- 

https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/170 was merged while tests were failing and this addresses the failures on `main`

Sorry about that, it wasn't clear to me that the tests were failing in a different gem than I made the changes in. I missed that totally.

---
The rails tests in the rails gem are very minimal, is this expected? https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/main/instrumentation/rails/test/instrumentation/opentelemetry/instrumentation/rails/patches/action_controller/metal_test.rb

The ActionPack tests have much more to them https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/main/instrumentation/action_pack/test/opentelemetry/instrumentation/action_pack/patches/action_controller/metal_test.rb

Should all the tests that exist in the ActionPack gem, exist in the Rails tests as well? I also notice that other Rails instrumentation features, like ActiveSupport, aren't tested in the rails gem.

### DIFF
--- a/instrumentation/rails/test/instrumentation/opentelemetry/instrumentation/rails/patches/action_controller/metal_test.rb
+++ b/instrumentation/rails/test/instrumentation/opentelemetry/instrumentation/rails/patches/action_controller/metal_test.rb
@@ -17,7 +17,7 @@ describe OpenTelemetry::Instrumentation::Rails do
   # Clear captured spans
   before { exporter.reset }
 
-  it 'sets the span name to ControllerName#action' do
+  it 'sets the span name to the format: HTTP_METHOD /rails/route(.:format)' do
     get '/ok'
 
     _(last_response.body).must_equal 'actually ok'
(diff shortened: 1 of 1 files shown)


ahayworth (approved):
> Sorry about that, it wasn't clear to me that the tests were failing in a different gem than I made the changes in. I missed that totally.

No worries - stuff happens! We should also make sure that failing tests block PRs from being merged, and I'll go take a peek at our repo settings.

> The rails tests in the rails gem are very minimal, is this expected?

I don't know about _expected_ - more tests are always appreciated if you feel up to writing some of course. I'm not surprised that they are minimal though. The rails instrumentation has grown over the years.

> Should all the tests that exist in the ActionPack gem, exist in the Rails tests as well? I also notice that other Rails instrumentation features, like ActiveSupport, aren't tested in the rails gem.

I tend to think that only the things added by the `rails` instrumentation directly should be tested there, delegating the other tests to the supporting gems. But I don't feel super strongly about it (and can see a valid argument in the other direction as well).

PR #178, opened 2022-11-04
fix(pg): safeguard against host being nil

marsadle:
PR #142 extracted, but the `:host` param can be `nil` if a socket is used (e.g. Rails' default settings), which raises an exception in this case. 

This simple change provides a safeguard for it.


robertlaurin (commented):
Hey @marsadle, thanks for the contribution. Is this something we can add a test for to prevent future regressions?

ahayworth (approved):
Approved if a test case is added - thank you for the fix ❤️

instrumentation/pg/test/opentelemetry/instrumentation/pg/instrumentation_test.rb
```diff
@@ -310,5 +310,25 @@
         _(span.attributes['db.statement']).must_be_nil
       end
     end
+
+    describe '#first_in_list' do
```
arielvalentin, line 314: I see that this is testing a private method. 

Is there some way to reproduce this use case with the `pg` library and check the span attributes instead of invoking the private method directly?

marsadle: Please see my comment on the PR. It's complicated to set up with Docker Compose as it will require an actual connection via a socket to be able to  check the spans.

I gave it a shot for 30 minutes, but ran into different issues with it.

cbandy (approved):

marsadle:
So, I've looked into this more. I've not found a solution for the CI workflows as they are a little different to the docker-compose tests.

I'd be grateful if anybody had a suggestion how to make sockets work with GitHub Actions.

ahayworth:
@marsadle I managed to not push things correctly (git is hard), but I added two commits to the branch that should fix it up for Actions! It's basically docker, but the docs are confusing honestly.

I'm curious about all the volumes and things added to the `docker-compose.yml` file; were those all necessary? I was able to get this to work in actions just with one volume specification; maybe that's all the docker-compose tests need? (I can't actually confirm that locally right now - no docker on the work laptop - but if they aren't all necessary I would appreciate if we could strip out the unnecessary changes).

But ... otherwise, this will be good to go provided CI passes and whatnot.

marsadle:
@ahayworth Good to know, thanks!

I'm unsure if the same on all machines (it should, but GitHub Actions already showed us that it doesn't seem to be), but in my case (MacBook Pro M1) it needed the socket to be in `/tmp` instead of `/var/run/postgresql`, that's why both are in there.

The rest are just inherited from the base image.

ahayworth:
Ah okay. That's perfectly fine then !

arielvalentin (approved):

PR #182, opened 2022-11-09
feat: Add experimental traceresponse propagator to Rack instrumentation

wperron:
Adds an optional config for the rack instrumentation to use the [new traceresponse propagator][1].

[1]: https://github.com/open-telemetry/opentelemetry-ruby/pull/1397


instrumentation/rack/example/trace_demonstration.rb
```diff
@@ -11,7 +11,7 @@
 
 ENV['OTEL_TRACES_EXPORTER'] = 'console'
 OpenTelemetry::SDK.configure do |c|
-  c.use 'OpenTelemetry::Instrumentation::Rack'
+  c.use 'OpenTelemetry::Instrumentation::Rack', { response_propagator: [OpenTelemetry::Trace::Propagation::TraceContext::ResponseTextMapPropagator.new] }
```
plantfansam, line 14: Would rec. moving this to a separate example and noting that in `example/README`

instrumentation/rack/lib/opentelemetry/instrumentation/rack/instrumentation.rb
```diff
@@ -30,6 +30,7 @@ class Instrumentation < OpenTelemetry::Instrumentation::Base
         option :untraced_endpoints,       default: [],    validate: :array
         option :url_quantization,         default: nil,   validate: :callable
         option :untraced_requests,        default: nil,   validate: :callable
+        option :response_propagator,      default: [],    validate: :array
```
plantfansam, line 33: Pluralize, since the user could specify multiple propagators to run

instrumentation/rack/lib/opentelemetry/instrumentation/rack/middlewares/tracer_middleware.rb
```diff
@@ -5,6 +5,7 @@
 # SPDX-License-Identifier: Apache-2.0
 
 require 'opentelemetry/trace/status'
+require 'pry'
```
plantfansam, line 8: Remove

instrumentation/rack/opentelemetry-instrumentation-rack.gemspec
```diff
@@ -28,6 +28,7 @@ Gem::Specification.new do |spec|
   spec.add_dependency 'opentelemetry-api', '~> 1.0'
   spec.add_dependency 'opentelemetry-common', '~> 0.19.3'
   spec.add_dependency 'opentelemetry-instrumentation-base', '~> 0.21.0'
+  spec.add_dependency 'opentelemetry-sdk-experimental'
```
plantfansam, line 31: I wonder if we should include some kind of note during upgrade to next release about how `sdk-experimental` won't replace the SDK. Just don't want to freak users out.

plantfansam (changes requested):
Seems like a great direction! Next up, tests 😄

instrumentation/rack/example/trace_demonstration.rb
wperron: Since there's no README in the `example/` directory, I simply moved this to another file and added a comment explaining what's going on here.

instrumentation/rack/opentelemetry-instrumentation-rack.gemspec
wperron: Actually, since making this optional, it occurs to me that we don't need to add the experimental SDK to the dependencies since it's up to the user to inject the `ResponseTextMapPropagator` in the middleware config.

plantfansam: You're right 🧠! Hot dog!

instrumentation/rack/Gemfile
```diff
@@ -7,8 +7,8 @@
 source 'https://rubygems.org'
 
 gemspec
-
 group :test do
   gem 'opentelemetry-instrumentation-base', path: '../base'
+  gem 'opentelemetry-sdk-experimental'
```
plantfansam, line 12: Do we need this here _and_ in the gemspec as a dev dependency?

instrumentation/rack/example/Gemfile
```diff
@@ -4,7 +4,9 @@ source 'https://rubygems.org'
 
 gem 'opentelemetry-api'
 gem 'opentelemetry-common'
-gem 'opentelemetry-instrumentation-rack'
+gem 'opentelemetry-instrumentation-rack', path: '../'
 gem 'opentelemetry-sdk'
+gem 'opentelemetry-sdk-experimental'
 gem 'rack'
 gem 'rack-test'
+gem 'pry'
```
plantfansam, line 12: Remove?

instrumentation/rack/example/trace_demonstration.rb
```diff
@@ -25,4 +25,6 @@
 end
 
 # demonstrate tracing (span output to console):
-puts Rack::MockRequest.new(builder).get('/')
+res = Rack::MockRequest.new(builder).get('/')
```
plantfansam, line 28: Do we still need this patch?

instrumentation/rack/example/trace_demonstration3.rb
```diff
@@ -0,0 +1,31 @@
...
+# SPDX-License-Identifier: Apache-2.0
+
+require 'rubygems'
+require 'bundler/setup'
+
+Bundler.require
+
+ENV['OTEL_TRACES_EXPORTER'] = 'console'
+OpenTelemetry::SDK.configure do |c|
+  # Uses the experimental ResponseTextMapPropagator option
```
plantfansam, line 14: Would personally move this comment to line 7 and rephrase as "This uses the response_propagators option. `ResponseTextMapPropagator` is from the `opentelemetry-sdk-experimental` gem," but it's not a blocker.

instrumentation/rack/test/opentelemetry/instrumentation/rack/middlewares/tracer_middleware_test.rb
```diff
@@ -311,6 +311,17 @@
...
+    describe 'with ResponseTextMapPropagator' do
+      let(:config) { default_config.merge(response_propagators: [OpenTelemetry::Trace::Propagation::TraceContext::ResponseTextMapPropagator.new]) }
+
+      it 'injects the traceresponse header' do
+        res = Rack::MockRequest.new(rack_builder).get('/ping', env)
+        _(res.headers).must_include('traceresponse')
+      end
+    end
+  end
+
```
plantfansam, line 324: Can we add tests for:

- when this option is not configured? 
- when the propagator throws an exception?

plantfansam (changes requested):

instrumentation/rack/Gemfile
wperron: I thought so, but re-testing just now it doesn't seem necessary 🤔

instrumentation/rack/example/trace_demonstration.rb
wperron: nope

plantfansam (changes requested):
Looking good! I think we just need a small change to the language in one test ...

instrumentation/rack/lib/opentelemetry/instrumentation/rack/middlewares/tracer_middleware.rb
```diff
@@ -82,6 +82,7 @@ def call(env) # rubocop:disable Metrics/MethodLength
                 OpenTelemetry::Instrumentation::Rack.with_span(request_span) do
                   @app.call(env).tap do |status, headers, response|
                     set_attributes_after_request(request_span, status, headers, response)
+                    config[:response_propagators].each { |propagator| propagator.inject(headers) }
```
plantfansam, line 85: To state the unstated: each propagator should be prepared to grab `Context.current` since we're not explicitly passing context, here. This is indeed the case for the [`ResponseTextMapPropagator`](https://github.com/open-telemetry/opentelemetry-ruby/blob/main/sdk_experimental/lib/opentelemetry/sdk/trace/propagation/trace_context/response_text_map_propagator.rb#L27).

instrumentation/rack/lib/opentelemetry/instrumentation/rack/instrumentation.rb
```diff
@@ -30,6 +30,7 @@ class Instrumentation < OpenTelemetry::Instrumentation::Base
         option :untraced_endpoints,       default: [],    validate: :array
         option :url_quantization,         default: nil,   validate: :callable
         option :untraced_requests,        default: nil,   validate: :callable
+        option :response_propagators,     default: [],    validate: :array
```
plantfansam, line 33: 👌

plantfansam (approved):
LGTM ✨

robertlaurin (approved):

PR #185, opened 2022-11-10
fix: Use attributes from the active PG connection

cbandy:
As discovered and addressed in #142, the `pg` gem and its underlying `libpq` library allow multiple hosts and network addresses in the connection string. The solution applied there, however, does not always return accurate information about the connected peer.

When there are multiple hosts or addresses, the library [tries each in order until one succeeds](https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-MULTIPLE-HOSTS). The [`target_session_attrs`](https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-CONNECT-TARGET-SESSION-ATTRS) setting allows both primary and replica servers to be specified in the connection string, and the library will figure which is which. Always using the first value in the connection string ignores this stuff.

I don't see an explanation for why `conninfo_hash` was used in 463c14c3581641ce57270c1c7cfb91715cfd2c3e, so perhaps I'm missing some benefit to it.

The separate methods (`host`, `hostaddr`, `port`, `user`, and `db`) describe the singular active connection from the resolved connection string (IPv4, IPv6, primary, replica, etc). I'm pretty sure we should use these when they're available.

There are a couple of older versions of `libpq` that allow multiple hosts/addresses but lack a `hostaddr` method. For those, `conninfo_hash` is accurate when it returns a single address. When it returns multiple addresses, we can use the first like #142 or omit `net.peer.ip`. The latter seems better to me, so that's what I've done in this PR.

~TODO:~
- [x] I would like to rebase this after #178 lands for the UNIX socket test there.

### DIFF
--- a/instrumentation/pg/lib/opentelemetry/instrumentation/pg/patches/connection.rb
+++ b/instrumentation/pg/lib/opentelemetry/instrumentation/pg/patches/connection.rb
@@ -166,12 +166,12 @@ module OpenTelemetry
           def transport_port
             # The port method can fail in older versions of the gem. It is
             # accurate and safe to use when the DEF_PGPORT constant is defined.
-            return port.to_s if defined?(::PG::DEF_PGPORT)
+            return port if defined?(::PG::DEF_PGPORT)
 
             # As a fallback, we can use the port of the parsed connection
             # string when there is exactly one.
             p = conninfo_hash[:port]
-            return p unless p.nil? || p.empty? || p.include?(',')
+            return p.to_i unless p.nil? || p.empty? || p.include?(',')
           end
         end
       end
--- a/instrumentation/pg/lib/opentelemetry/instrumentation/pg/patches/connection.rb
+++ b/instrumentation/pg/lib/opentelemetry/instrumentation/pg/patches/connection.rb
@@ -97,7 +97,7 @@ module OpenTelemetry
           end
 
           def span_name(operation)
-            [validated_operation(operation), database_name].compact.join(' ')
+            [validated_operation(operation), db].compact.join(' ')
           end
 
           def validated_operation(operation)
@@ -123,42 +123,55 @@ module OpenTelemetry
             @generated_postgres_regex ||= Regexp.union(PG::Constants::POSTGRES_COMPONENTS.map { |component| PG::Constants::COMPONENTS_REGEX_MAP[component] })
           end
 
-          def database_name
-            conninfo_hash[:dbname]&.to_s
-          end
-
-          def first_in_list(item)
-            return unless item
-
-            if (idx = item.index(','))
-              item[0...idx]
-            else
-              item
-            end
-          end
-
           def client_attributes
             attributes = {
               'db.system' => 'postgresql',
-              'db.user' => conninfo_hash[:user]&.to_s,
-              'db.name' => database_name,
-              'net.peer.name' => first_in_list(conninfo_hash[:host]&.to_s)
+              'db.user' => user,
+              'db.name' => db
             }
             attributes['peer.service'] = config[:peer_service] if config[:peer_service]
 
             attributes.merge(transport_attrs).reject { |_, v| v.nil? }
           end
 
+          def transport_addr
+            # The hostaddr method is available when the gem is built against
+            # a recent version of libpq.
+            return hostaddr if defined?(hostaddr)
+
+            # As a fallback, we can use the hostaddr of the parsed connection
+            # string when there is only one. Some older versions of libpq allow
+            # multiple without any way to discern which is presently connected.
+            addr = conninfo_hash[:hostaddr]
+            return addr unless addr&.include?(',')
+          end
+
           def transport_attrs
-            if conninfo_hash[:host]&.start_with?('/')
-              { 'net.transport' => 'Unix' }
+            h = host
+            if h&.start_with?('/')
+              {
+                'net.sock.family' => 'unix',
+                'net.peer.name' => h
+              }
             else
               {
-                'net.transport' => 'IP.TCP',
-                'net.peer.ip' => conninfo_hash[:hostaddr]&.to_s,
-                'net.peer.port' => first_in_list(conninfo_hash[:port]&.to_s)
+                'net.transport' => 'ip_tcp',
+                'net.peer.name' => h,
+                'net.peer.ip' => transport_addr,
+                'net.peer.port' => transport_port
               }
             end
+          end
+
+          def transport_port
+            # The port method can fail in older versions of the gem. It is
+            # accurate and safe to use when the DEF_PGPORT constant is defined.
+            return port.to_s if defined?(::PG::DEF_PGPORT)
+
+            # As a fallback, we can use the port of the parsed connection
+            # string when there is exactly one.
+            p = conninfo_hash[:port]
+            return p unless p.nil? || p.empty? || p.include?(',')
           end
         end
       end
--- a/instrumentation/pg/Appraisals
+++ b/instrumentation/pg/Appraisals
@@ -1,5 +1,15 @@
 # frozen_string_literal: true
 
-appraise 'pg-1.2.3' do
+# 1.2.3 appears to be the most popular version at RubyGems.org
+appraise 'pg-1.2' do
+  gem 'pg', '~> 1.2.3'
+end
+
+# 1.3.0 significantly changed connecting and the handling of connection strings
+appraise 'pg-1.3' do
+  gem 'pg', '~> 1.3.5'
+end
+
+appraise 'pg-latest' do
   gem 'pg'
 end
--- a/instrumentation/pg/test/opentelemetry/instrumentation/pg/instrumentation_test.rb
+++ b/instrumentation/pg/test/opentelemetry/instrumentation/pg/instrumentation_test.rb
@@ -315,11 +315,33 @@ describe OpenTelemetry::Instrumentation::PG::Instrumentation do
       let(:host) { nil }
       let(:port) { nil }
 
-      it 'sets empty span attributes for host and port' do
+      it 'sets attributes for the socket directory and family' do
         client.query('SELECT 1')
 
-        _(span.attributes['net.peer.name']).must_equal ''
-        _(span.attributes['net.peer.port']).must_equal ''
+        _(span.attributes['net.peer.name']).must_match %r{^/}
+        _(span.attributes['net.peer.port']).must_be_nil
+        _(span.attributes['net.sock.family']).must_equal 'unix'
+      end
+    end
+
+    describe 'when connection has multiple hosts' do
+      before { skip 'requires libpq >= 10.0' if ::PG.library_version < 10_00_00 } # rubocop:disable Style/NumericLiterals
+
+      let(:client) do
+        ::PG::Connection.open(
+          host: ['nowhere.', host].join(','),
+          port: ['20823', port].join(','),
+          user: user,
+          dbname: dbname,
+          password: password
+        )
+      end
+
+      it 'sets attributes of the active connection' do
+        client.query('SELECT 1')
+
+        _(span.attributes['net.peer.name']).must_equal host
+        _(span.attributes['net.peer.port']).must_equal port if ::PG.const_defined?('DEF_PORT')
       end
     end
   end unless ENV['OMIT_SERVICES']
--- a/instrumentation/pg/test/opentelemetry/instrumentation/pg/instrumentation_test.rb
+++ b/instrumentation/pg/test/opentelemetry/instrumentation/pg/instrumentation_test.rb
@@ -114,7 +114,7 @@ describe OpenTelemetry::Instrumentation::PG::Instrumentation do
         _(span.attributes['db.statement']).must_equal 'SELECT 1'
         _(span.attributes['db.operation']).must_equal 'SELECT'
         _(span.attributes['net.peer.name']).must_equal host.to_s
-        _(span.attributes['net.peer.port']).must_equal port.to_s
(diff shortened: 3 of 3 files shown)


instrumentation/pg/lib/opentelemetry/instrumentation/pg/patches/connection.rb
```diff
@@ -123,38 +123,38 @@ def generated_postgres_regex
...
-            if conninfo_hash[:host]&.start_with?('/')
+            if host&.start_with?('/')
               { 'net.transport' => 'Unix' }
             else
               {
                 'net.transport' => 'IP.TCP',
-                'net.peer.ip' => conninfo_hash[:hostaddr]&.to_s,
-                'net.peer.port' => first_in_list(conninfo_hash[:port]&.to_s)
+                'net.peer.ip' => transport_addr&.to_s,
+                'net.peer.port' => port&.to_s
```
cbandy, line 157: 🤔 I'm curious about this line. Semantic Conventions say that `peer.port` should be an integer. Is `to_s` appropriate here?

- https://opentelemetry.io/docs/reference/specification/trace/semantic_conventions/span-general/

ahayworth: I don't actually recall what the reasons were for stringification here; which suggests that perhaps there was no valid reason and it was just a mistake. If you want to include a fix for this in the PR, that would be a welcome change. 😄

ahayworth:
This looks like a good change, I will happily review it once you're ready. I agree with everything outlined in the PR description, including the `TODO` you added. Thank you so much for this!

---

> I don't see an explanation for why conninfo_hash was used in https://github.com/open-telemetry/opentelemetry-ruby-contrib/commit/463c14c3581641ce57270c1c7cfb91715cfd2c3e, so perhaps I'm missing some benefit to it.

For context, this was contributed before we split off the `-contrib` repo, so the original PR that added this instrumentation is here: https://github.com/open-telemetry/opentelemetry-ruby/pull/659

I don't recall having any specific reason for using `conninfo_hash`; if there are better ways to get that information than we should use it. I think it was just my best attempt at getting the right information since I am not an expert on the `pg` gem itself. We are most grateful for contributions that improve the quality of instrumentation in this way - we are certainly not experts and we're covering a very wide surface area in this repo. 😍

cbandy:
I haven't forgotten about this! `let(:port) { nil }` exposed a bug in the `pg` gem I'm waiting on now: https://github.com/ged/ruby-pg/issues/491

The implementation here will consider versions of that gem with and without whatever fix comes.

ahayworth:
@cbandy looks like they just merged a fix, too!

cbandy:
Rebased and ready to go. I've added a test verifying that active connection information is returned when using "multiple hosts."

IP address and port depend on the version of the gem and the underlying driver. The original implementation was correct to use `conninfo_hash` in the past. Recent versions have the desired behavior, so this detects and uses those.

I made three other changes to the transport attributes after reading the [semantic conventions](https://opentelemetry.io/docs/reference/specification/trace/semantic_conventions/span-general/):

1. `net.peer.port` is now an integer.
2. `net.transport` is now `ip_tcp` when not using a Unix socket.
3. `net.transport` is _not set_ when using a Unix socket. Instead, `net.sock.family` is set to `unix`.

I wanted to fill out `net.sock.peer.addr` for the Unix socket case, but there isn't a good way to get that information from the gem or driver. I left `net.peer.name` set to the _directory_ containing the socket, as we've done before.

cbandy:
Amended for Rubocop.

instrumentation/pg/lib/opentelemetry/instrumentation/pg/patches/connection.rb
```diff
@@ -123,43 +123,56 @@ def generated_postgres_regex
...
             }
             attributes['peer.service'] = config[:peer_service] if config[:peer_service]
 
             attributes.merge(transport_attrs).reject { |_, v| v.nil? }
           end
 
+          def transport_addr
+            # The hostaddr method is available when the gem is built against
+            # a recent version of libpq.
+            return hostaddr if defined?(hostaddr)
```
ahayworth, line 140: Huh. I actually didn't know you could use `defined?` in that way (I thought this was maybe calling `defined?(result-of-unintentional-hostaddr-call)` which wouldn't be a good thing to rely on). But wouldn't you know it, it's right there in the docs 🤦 

https://ruby-doc.org/docs/keywords/1.9/Object.html#method-i-defined-3F

Thank you, I learned something new today. 😄

instrumentation/pg/lib/opentelemetry/instrumentation/pg/patches/connection.rb
```diff
@@ -123,43 +123,56 @@ def generated_postgres_regex
...
-          end
-
           def client_attributes
             attributes = {
               'db.system' => 'postgresql',
-              'db.user' => conninfo_hash[:user]&.to_s,
-              'db.name' => database_name,
-              'net.peer.name' => first_in_list(conninfo_hash[:host]&.to_s)
+              'db.user' => user,
+              'db.name' => db
```
ahayworth, line 130: I'm guessing these get defined in the `pg` connection?

instrumentation/pg/lib/opentelemetry/instrumentation/pg/patches/connection.rb
```diff
@@ -123,43 +123,56 @@ def generated_postgres_regex
...
+
           def transport_attrs
-            if conninfo_hash[:host]&.start_with?('/')
-              { 'net.transport' => 'Unix' }
+            host = self.host
+            if host&.start_with?('/')
+              {
+                'net.sock.family' => 'unix',
+                'net.peer.name' => host
+              }
```
ahayworth, line 155: This kind of scope-shenanigans makes my eye twitch, as I try to parse exactly which `host` we're actually looking at (the variable? the method call?). Even if it is nicer to read!

Would you mind doing something silly like `the_host` (or whatever feels most readable to you) so that we aren't shadowing the host method in this scope here? I know it's a minor thing, but...

ahayworth (commented):
No major issues, and I'm grateful that this PR is so thorough. And because you've been so thorough and helpful I have rewarded you with a little finicky comment to address before we merge it. 😆 ❤️  Great job catching the semantic convention changes, as well!

As a meta question, should we test multiple versions of the PG gem via the `Appraisals` file? I don't think I have a strong opinion either way.

instrumentation/pg/lib/opentelemetry/instrumentation/pg/patches/connection.rb
cbandy: Yes. These are methods of ::PG:: Connection. I don't mind using `self.db` to make that more clear. Would that be better?

instrumentation/pg/lib/opentelemetry/instrumentation/pg/patches/connection.rb
cbandy: Agreed. I wasn't sure what to name it either. Is `h` too short?

instrumentation/pg/lib/opentelemetry/instrumentation/pg/patches/connection.rb
cbandy: I learned it during this PR as well. I saw it as part of module prepending: `super if defined? super`

In this case, we can use `respond_to?`. I wonder if one is definitely faster? 🤷🏻

instrumentation/pg/lib/opentelemetry/instrumentation/pg/patches/connection.rb
ahayworth: No need to change, this one was actually just a question. Thank you!

instrumentation/pg/lib/opentelemetry/instrumentation/pg/patches/connection.rb
ahayworth: `h` is fine with me 😄

instrumentation/pg/lib/opentelemetry/instrumentation/pg/patches/connection.rb
ahayworth: I mean; either is probably fine. But my first meeting of the morning cancelled so that means its time for [MicroBenchmark, Thanksgiving Edition 🦃](https://gist.github.com/ahayworth/bdd5f12c5ce08a075289bfa4530a468c)

Looks like `defined?` may be a tiny bit faster? It probably doesn't have to go through things like `method_missing` and whatever. Here are the results:
```
/-------------------------------------------------------------\
|                  user     system      total        real     |
|    defined?:   0.094393   0.000956   0.095349 (  0.095374)  |
| respond_to?:   0.099900   0.001003   0.100903 (  0.100944)  |
\-------------------------------------------------------------/
   \
    \                                  ,+*^^*+___+++_
     \                           ,*^^^^              )
      \                       _+*                     ^**+_
       \                    +^       _ _++*+_+++_,         )
                _+^^*+_    (     ,+*^ ^          \+_        )
               {       )  (    ,(    ,_+--+--,      ^)      ^\
              { (@)    } f   ,(  ,+-^ __*_*_  ^^\_   ^\       )
             {:;-/    (_+*-+^^^^^+*+*<_ _++_)_    )    )      /
            ( /  (    (        ,___    ^*+_+* )   <    <      \
             U _/     )    *--<  ) ^\-----++__)   )    )       )
              (      )  _(^)^^))  )  )\^^^^^))^*+/    /       /
            (      /  (_))_^)) )  )  ))^^^^^))^^^)__/     +^^
           (     ,/    (^))^))  )  ) ))^^^^^^^))^^)       _)
            *+__+*       (_))^)  ) ) ))^^^^^^))^^^^^)____*^
            \             \_)^)_)) ))^^^^^^^^^^))^^^^)
             (_             ^\__^^^^^^^^^^^^))^^^^^^^)
               ^\___            ^\__^^^^^^))^^^^^^^^)\\
                    ^^^^^\uuu/^^\uuu/^^^^\^\^\^\^\^\^\^\
                       ___) >____) >___   ^\_\_\_\_\_\_\)
                      ^^^//\\_^^//\\_^       ^(\_\_\_\)
                        ^^^ ^^ ^^^ ^
```

instrumentation/pg/lib/opentelemetry/instrumentation/pg/patches/connection.rb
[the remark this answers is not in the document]
ahayworth: @ericmustin I think what would happen is that we would end up checking the attribute's validity [here](https://github.com/open-telemetry/opentelemetry-ruby/blob/a4efaf3a44b0efc6ca805a6d0a5bdc4f13a01b34/sdk/lib/opentelemetry/sdk/internal.rb#L27), and that doesn't like nil-things. The net effect when you follow things around the codebase is that we would log a warning, but ultimately just [keep only the valid values](https://github.com/open-telemetry/opentelemetry-ruby/blob/a4efaf3a44b0efc6ca805a6d0a5bdc4f13a01b34/sdk/lib/opentelemetry/sdk/trace/span.rb#L341-L345).

You're correct that `nil` would be correctly represented as the empty string when the destination is supposed to be an OTLP `AnyValue`, but I think the actual controlling part of the spec for attribute values is [here](https://opentelemetry.io/docs/reference/specification/common/#attribute) ... and that explicitly mentions that `null` (`nil`, in our  case) is actually not allowed as an attribute value. `AnyValue` is used [more restrictively](https://opentelemetry.io/docs/reference/specification/common/attribute-type-mapping/) from what I can gather - at the collector, and in the logging SDK?

But regardless, we shouldn't let `nil` sneak through and I'm really glad you brought it up. @cbandy can we assert with any amount of confidence that the `user` and `host` methods never return `nil`? If they can (even if only in corner cases), we should make sure to `reject` away `nil` values.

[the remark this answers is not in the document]
ericmustin: i truly do not know C but i think looking at the c extension, nil is a possible return value https://github.com/ged/ruby-pg/blob/cf1443dd4473ffc0b1ebbe401c6a700b1cd50476/ext/pg_connection.c#L615-L634 so i think we should coerce to empty string to be safe

cbandy:
> As a meta question, should we test multiple versions of the PG gem via the `Appraisals` file? I don't think I have a strong opinion either way.

I'll try that out. I can use the [RubyGems dataset](https://play.honeycomb.io/ruby-together/datasets/rubygems.org/) at Honeycomb to pick some useful versions.

cbandy:
Amended `host` variable and added appraisals for some interesting `pg` versions:

```
>> BUNDLE_GEMFILE=/app/instrumentation/pg/gemfiles/pg_1.2.gemfile bundle list pg
/bundle/gems/pg-1.2.3
>> BUNDLE_GEMFILE=/app/instrumentation/pg/gemfiles/pg_1.3.gemfile bundle list pg
/bundle/gems/pg-1.3.5
>> BUNDLE_GEMFILE=/app/instrumentation/pg/gemfiles/pg_latest.gemfile bundle list pg
/bundle/gems/pg-1.4.5
```

ahayworth (approved):

ahayworth:
@cbandy Thanks so much for the back-and-forth on this. One more ✅ and I think we're good to merge. (cc @ericmustin since you've been looking)

ericmustin (approved):

ericmustin:
kicked the ci, i think you have to pull in https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/211 for things to pass. :shakes-fist-at-jruby:

PR #196, opened 2022-11-15
Return and scan latest result for each check

plantfansam:
This patch updates release utils' `github_check_errors` function to:

1. collect all GitHub checks (currently, we just get the first page of results)
2. aggregate them by check name
3. check the result of the _most recent run_ when evaluating a SHA's readiness for release

This should let us retry releases (using the retry release [workflow](https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/main/.github/workflows/release-retry.yml)) when that release's SHA had failed commit checks on `main` that were successfully retried.

## Does it work?

This is a hard patch to check (I don't think there's a test harness for Toys stuff), but I have tested out the logic locally (see [gist](https://gist.github.com/plantfansam/f027699cfa6842a2af00016537e7cea3); had to use `open3` since there was some weirdness with `exec` on my machine):

####  Run against ref c0e2e26bbbdce8254bdcf9d1315d9de21384e884, which has failed checks that were successfully retried:
```
# https://github.com/open-telemetry/opentelemetry-ruby-contrib/commit/c0e2e26bbbdce8254bdcf9d1315d9de21384e884
➜  opentelemetry-ruby-contrib git:(check-status) ✗ bundle exec ruby foo.rb
foo.rb: warning: argument of top-level return is ignored
[]
```
#### Run against ref b53f9925c043fbf319bbc322b4755d8d7e94b32f, which has failing checks
```
# https://github.com/open-telemetry/opentelemetry-ruby-contrib/commit/b53f9925c043fbf319bbc322b4755d8d7e94b32f
➜  opentelemetry-ruby-contrib git:(check-status) ✗ bundle exec ruby foo.rb
foo.rb: warning: argument of top-level return is ignored
["Latest gitHub check \"rails / macos-latest\" was not successful.", "Latest gitHub check \"rails / ubuntu-latest\" was not successful."]
```


.toys/release/.lib/release_utils.rb
```diff
@@ -311,18 +312,43 @@ def github_check_errors(ref)
...
-        results << "GitHub check #{name.inspect} was not successful"
+
+    (1..pages).each do |idx|
+      result = exec(["gh", "api", "repos/#{repo_path}/commits/#{ref}/check-runs?page=#{idx}",
+        "-H", "Accept: application/vnd.github.antiope-preview+json"],
+       out: :capture, e: false)
+      return ["Failed to obtain GitHub check results for #{ref}"] unless result.success?
+      checks = ::JSON.parse(result.captured_out)["check_runs"]
+      results << "No GitHub checks found for #{ref}" if checks.empty?
+      checks.each_with_index do |check|
```
ahayworth, line 329: Why `each_with_index` here? You aren't using the index.

```suggestion
      checks.each do |check|
```

ahayworth (approved):
Seems okay enough to me, if it works.

robertlaurin (approved):

PR #202, opened 2022-11-16
OpenTelemetry Hanami Instrumentation

Ozkee:


ahayworth:
@Ozkee Hey there - just noticed this while combing through my notifications. Is there a plan to put this instrumentation elsewhere, or something like that? ❤️

PR #205, opened 2022-11-23
fix: remove `job_options` when using `bulk_enqueue`

msalei:
`bulk_enqueue` was added with `Que` version `2.1.0` and it requires `job_options` to be passed to `bulk_enqueue` itself rather than `enqueue`.

Here, however, `job_options` is forcibly passed to `enqueue` containing injected tags, which means that the following error is raised:

```
When using .bulk_enqueue, job_options must be passed to that method rather than .enqueue
```

This commit adds a workaround to be able to use `bulk_enqueue` without forcing the library users to upgrade `que` to `2.1.0`.

[Que changelog](https://github.com/que-rb/que/blob/master/CHANGELOG.md#210-2022-08-25)
[bulk_enqueue docs](https://github.com/que-rb/que/tree/master/docs#enqueueing-jobs-in-bulk)


instrumentation/que/lib/opentelemetry/instrumentation/que/patches/que_job.rb
```diff
@@ -43,27 +43,35 @@ def enqueue(*args, job_options: {}, **arg_opts)
                   OpenTelemetry.propagation.inject(tags, setter: TagSetter)
                 end
 
-                job_options = job_options.merge(tags: tags)
-                job = super(*args, job_options: job_options, **arg_opts)
+                # In Que version 2.1.0 `bulk_enqueue` was introduced and in order
+                # for it to work, we must pass `job_options` to `bulk_enqueue` instead of enqueue.
+                if Gem.loaded_specs['que'].version >= Gem::Version.new('2.1.0') && Thread.current[:que_jobs_to_bulk_insert]
```
ahayworth, line 48: Is checking a thread-local var the only way to do it? I would feel better about it if we could use something the `Que` provides to manipulate this rather than manipulating these ourselves (feels like we'll have compatibility problems with this down the line if we aren't careful).

(Also: is there a way to not check `Gem.loaded_specs` every time? I feel like we could / should memoize that at least, feels like a lot more work than we really want to do if we can avoid it. Not like it's going to change on us while it's running! At least ... I hope not 😅 )

ahayworth (commented):
I'm not opposed to this - but see comments for some clarification.

Also, thank you for adding an Appraisal to this so that we can test both versions! 🎉 ❤️

instrumentation/que/lib/opentelemetry/instrumentation/que/patches/que_job.rb
msalei: The thread-local variable is modified without using any function within `enqueue` and `bulk_enqueue` methods and they don't provide any way of doing it through `Que` as well so I don't think there is a _good_ way to do it here as well unless they update it in the future.

Also, I added memoization to the gem version to not call `Gem.loaded_specs` every time.

instrumentation/que/Rakefile
```diff
@@ -26,3 +26,15 @@ if RUBY_ENGINE == 'truffleruby'
...
+  sh 'bundle exec appraisal install'
+end
+
+task test_que1: :install_test_deps do
+  sh 'bundle exec appraisal que-1.2 rake test'
+end
+
+task test_que2: :install_test_deps do
+  sh 'bundle exec appraisal que-2.1 rake test'
+end
```
ahayworth, line 40: I think we should remove these:
- running `bundle exec appraisal` is enough to install deps
- running tests can be done with `bundle exec appraisal rake test`, which tests all versions in the `Appraisal` file in series.

Maybe we can just document this in the README for the gem instead?

ahayworth (approved):
Sad to see that there's no better way to modify this than with the thread local variable, but - you're right, it really doesn't look like there's another supported way to do it upstream.

I left a comment about the `Rakefile`, but I don't feel strongly enough to block on it or anything. ❤️ 

Thank you so much for the contribution! We'll wait to get at least one more approving review before we merge and release. 😄

instrumentation/que/Rakefile
msalei: Thanks! I removed the rake tasks and added a section to README but I'd like feedback on that :) https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/205/commits/31d541b7fe781c079fee83c0b805ef241bcace09

arielvalentin:
@indrekj Since you are a major contributor to this instrumentation and user of this gem I would appreciate a review from you on this.

indrekj (approved):
lgtm

msalei:
I fixed Ruby requirement in Que 2.1 appraisal, it should pass in CI now.

Que dropped support for Ruby < 2.7 but they also introduced support for Ruby 3 so `~> 2.7` was not necessary.

https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/205/commits/5d80527d9e261767061898f73e3f9b79704a571e

msalei:
@ahayworth can you confirm that tests in CI are passing now, please?

ahayworth:
@msalei apologies, I was on an extended leave. However, the PR was merged! Yay! 🎉

PR #206, opened 2022-11-23
feat: Add option 'trace_root_spans' to instrumentation PG

AS-AlStar:
Add option 'trace_root_spans' to instrumentation PG

### DIFF
--- a/instrumentation/pg/lib/opentelemetry/instrumentation/pg/patches/connection.rb
+++ b/instrumentation/pg/lib/opentelemetry/instrumentation/pg/patches/connection.rb
@@ -133,13 +133,6 @@ module OpenTelemetry
             conninfo_hash[:dbname]&.to_s
           end
 
-
-          # NOTE: caught a error: undefined local variable or method `conninfo_hash' for PG::Connection
-          # HACK for fix it
-          def conninfo_hash
-            super || {}
-          end
-
           def first_in_list(item)
             if (idx = item.index(','))
               item[0...idx]
--- a/instrumentation/pg/lib/opentelemetry/instrumentation/pg/patches/connection.rb
+++ b/instrumentation/pg/lib/opentelemetry/instrumentation/pg/patches/connection.rb
@@ -133,6 +133,13 @@ module OpenTelemetry
             conninfo_hash[:dbname]&.to_s
           end
 
+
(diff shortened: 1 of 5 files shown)


instrumentation/pg/lib/opentelemetry/instrumentation/pg/patches/connection.rb
```diff
@@ -15,6 +15,8 @@ module Patches
         module Connection # rubocop:disable Metrics/ModuleLength
           PG::Constants::EXEC_ISH_METHODS.each do |method|
             define_method method do |*args|
+              return super(*args) unless config[:trace_root_spans] || OpenTelemetry::Trace.current_span.context.valid?
```
fbogsany, line 18:
```suggestion
              return super(*args) unless config[:trace_root_spans] || OpenTelemetry::Trace.current_span.recording?
```

fbogsany:
@AS-AlStar can you please provide more context on the issue this PR is trying to solve?

AS-AlStar:
> 

I don't want spans to be created on their own. For example, when I make a request to the database in the console "User.last". I spied this option in the redis.
Check in 'rails c', when **trace_root_spans: true**:
`User.last` -  will created span 
when **trace_root_spans: false**:
`User.last` -  don't will created span
but this code generate span:
```
tracer.in_span('operation-name') do |span|
  User.first
end
```

instrumentation/pg/lib/opentelemetry/instrumentation/pg/patches/connection.rb
AS-AlStar: does this change do the same? I think it's a little different

fbogsany: It does the same thing. This is the officially supported way to decide whether to do "expensive" things, based on whether the data will actually be recorded.

ahayworth:
@AS-AlStar @fbogsany I think we should hold off on merging this right now - we discussed this in the SIG today and we feel that this could be accomplished in other ways without needing another config option. We also discussed how our documentation does not really make that obvious or clear, and we will provide examples shortly. 😄

arielvalentin:
@ahayworth do we have docs for this?

ericmustin:
this should stay open i think

ericmustin:
ergh, wrong pr, disregard

PR #207, opened 2022-11-25
Fix `format_command`'s terrible performance

casperisfine:
I recently got pulled on a memcached performance issue, we were seeing horribly slow memcached operations.

After profiling it turns out that 99% of the time was spent in opentelemetry.

This was particularly bad when the payload we were writing in memcache was a large (1MB) binary blob.

benchmark:

```ruby
require 'bundler/inline'

gemfile do
  source 'https://rubygems.org'
  gem 'opentelemetry-instrumentation-dalli'
  gem 'benchmark-ips'
end

require 'benchmark/ips'
require 'opentelemetry-instrumentation-dalli'
require 'opentelemetry/instrumentation/dalli/utils'

operation = :set
args = ["foo", Random.bytes(1_000_000), 0, 0, nil].freeze

ORIG = OpenTelemetry::Instrumentation::Utils

module Opt
  extend self

  CMD_MAX_LEN = 500

  def format_command(operation, args)
    placeholder = "#{operation} BLOB (OMITTED)"

    command = operation.to_s
    args.each do |arg|
      str = arg.to_s
      if str.bytesize >= CMD_MAX_LEN
        break
      else
        command << " " << str
      end
    end

    command = OpenTelemetry::Common::Utilities.utf8_encode(command, binary: true, placeholder: placeholder)
    OpenTelemetry::Common::Utilities.truncate(command, CMD_MAX_LEN)
  rescue StandardError => e
    OpenTelemetry.logger.debug("Error sanitizing Dalli operation: #{e}")
    placeholder
  end
end

Benchmark.ips do |x|
  x.report("original") { ORIG.format_command(operation, args) }
  x.report("opt") { Opt.format_command(operation, args) }
  x.compare!(order: :baseline)
end
```

Results:

```
Warming up --------------------------------------
            original     4.000  i/100ms
                 opt    99.088k i/100ms
            original     42.831  (± 2.3%) i/s -    216.000  in   5.044141s
                 opt    990.037k (± 0.4%) i/s -      4.954M in   5.004339s

Comparison:
            original:       42.8 i/s
                 opt:   990037.0 i/s - 23114.82x  (± 0.00) faster
```

cc @robertlaurin

### DIFF
--- a/instrumentation/dalli/lib/opentelemetry/instrumentation/dalli/utils.rb
+++ b/instrumentation/dalli/lib/opentelemetry/instrumentation/dalli/utils.rb
@@ -53,7 +53,17 @@ module OpenTelemetry
 
       def format_command(operation, args)
         placeholder = "#{operation} BLOB (OMITTED)"
-        command = [operation, *args].join(' ').strip
+
+        command = +operation.to_s
+        args.flatten.each do |arg|
+          str = arg.to_s
+          if str.bytesize >= CMD_MAX_LEN
+            command << ' BLOB (OMITTED)'
+          elsif !str.empty?
+            command << ' ' << str
+          end
+        end
+
         command = OpenTelemetry::Common::Utilities.utf8_encode(command, binary: true, placeholder: placeholder)
         OpenTelemetry::Common::Utilities.truncate(command, CMD_MAX_LEN)
       rescue StandardError => e


casperisfine:
I updated the implementation to generate the same output as before based on the test suite, so it's not longer as fast but still way better:

```
Warming up --------------------------------------
            original     4.000  i/100ms
                 opt    63.244k i/100ms
            original     43.899  (± 2.3%) i/s -    220.000  in   5.012839s
                 opt    627.939k (± 0.8%) i/s -      3.162M in   5.036167s

Comparison:
            original:       43.9 i/s
                 opt:   627939.2 i/s - 14304.28x  (± 0.00) faster
```

benoittgt:
Shouldn't we have a non regression performance test ? I know it is a pain to maintain, but I am wondering if could help to avoid futur regression.

instrumentation/dalli/lib/opentelemetry/instrumentation/dalli/utils.rb
```diff
@@ -53,7 +53,17 @@ def opname(operation, multi)
 
       def format_command(operation, args)
         placeholder = "#{operation} BLOB (OMITTED)"
-        command = [operation, *args].join(' ').strip
+
+        command = +operation.to_s
```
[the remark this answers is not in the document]
casperisfine: Ensure that `command` is mutable: https://ruby-doc.org/core-3.1.2/String.html#method-i-2B-40

Some object sometimes return a frozen string though `to_s`.

robertlaurin (approved):

ahayworth (approved):

adlroux:
Great find!

instrumentation/dalli/lib/opentelemetry/instrumentation/dalli/utils.rb
```diff
@@ -53,7 +53,17 @@ def opname(operation, multi)
...
+        args.flatten.each do |arg|
+          str = arg.to_s
+          if str.bytesize >= CMD_MAX_LEN
+            command << ' BLOB (OMITTED)'
```
drinkbeer, line 61: How about making `'  BLOB (OMITTED)'` a variable `delimiter` which can be shared at line 55 and line 61?

instrumentation/dalli/lib/opentelemetry/instrumentation/dalli/utils.rb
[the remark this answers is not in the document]
drinkbeer: Thanks for the explanation. This will make the string mutable, and not always create a new string object in memory, right?

instrumentation/dalli/lib/opentelemetry/instrumentation/dalli/utils.rb
```diff
@@ -53,7 +53,17 @@ def opname(operation, multi)
 
       def format_command(operation, args)
         placeholder = "#{operation} BLOB (OMITTED)"
-        command = [operation, *args].join(' ').strip
+
+        command = +operation.to_s
+        args.flatten.each do |arg|
+          str = arg.to_s
+          if str.bytesize >= CMD_MAX_LEN
```
drinkbeer, line 60: We have the `if` which checks if the `arg` bytesize exceeds limitation. In line 68, we are truncating the total concentrated command lines. I am thinking could we do something like:

```
if command.bytesize >= CMD_MAX_LEN
  break
elsif !str.empty?
  command << ' ' << str
end

command = OpenTelemetry::Common::Utilities.utf8_encode(command, binary: true, placeholder: placeholder)
```

We can remote line 68?

drinkbeer (approved):
Great finding!

fbogsany (approved):
Thanks for the find, and the fix.

I don't think a performance regression test is necessary. There isn't high churn in this instrumentation code and the maintainers are pretty scrupulous about checking "why did we do this again?"

instrumentation/dalli/lib/opentelemetry/instrumentation/dalli/utils.rb
casperisfine: It's already shared because we have `# frozen_string_literals: true`, so identical literal strings point to the same object.

instrumentation/dalli/lib/opentelemetry/instrumentation/dalli/utils.rb
casperisfine: Yeah, I should have made it clear that I made no attempt at optimizing the most out of it here, it's just a quick fix for a pathological case.

PR #213, opened 2022-11-30
fix: add untraced check to the Net::HTTP connect instrumentation

yoheyk:
We are observing that the Net::HTTP connect instrumentation does not respect untraced hosts.

I think this check should have been included in #108 but let me know if not having the check is intentional

### DIFF
--- a/instrumentation/net_http/test/opentelemetry/instrumentation/net/http/instrumentation_test.rb
+++ b/instrumentation/net_http/test/opentelemetry/instrumentation/net/http/instrumentation_test.rb
@@ -150,6 +150,18 @@ describe OpenTelemetry::Instrumentation::Net::HTTP::Instrumentation do
         _(exporter.finished_spans.size).must_equal 0
       end
 
+      it 'does not create a span on connect when request ignored using a regexp' do
+        WebMock.allow_net_connect!
+
+        uri = URI.parse('http://bazqux.com')
+        Net::HTTP.start(uri.host, uri.port) do |http|
+          http.get('/')
+        end
+        _(exporter.finished_spans.size).must_equal 0
+      ensure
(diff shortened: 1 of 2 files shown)


instrumentation/net_http/test/opentelemetry/instrumentation/net/http/instrumentation_test.rb
```diff
@@ -150,6 +150,18 @@
         _(exporter.finished_spans.size).must_equal 0
       end
 
+      it 'does not create a span on connect when request ignored using a regexp' do
+        WebMock.allow_net_connect!
```
arielvalentin, line 154: I think disabling Webmock will potentially result in non-deterministic tests. 

Is there a way to test this without disabling webmock?

yoheyk: If it's okay to call private methods, we can do something like this: 074748501160bb2f7a029fb2616a1c092151d155

arielvalentin: So Webmock is not exercising TLS/SSL connection negotiation and therefore never calling connect?

yoheyk: Webmock is [patching the `start` method](https://github.com/bblimke/webmock/blob/833291d4ce2d5927a738f929db26b594bf4fa7f5/lib/webmock/http_lib_adapters/net_http.rb#L135-L143) and it will skip connect unless you configure `net_http_connect_on_start` to true.

fbogsany (approved):

arielvalentin (approved):

PR #220, opened 2022-12-16
String-ify the code.function Span attribute

stevenharman:
Some times these come through as symbols, for example when using Devise's `FailureApp`, the name will be `:respond`. So we'll String-ify the value before sending it on.


arielvalentin (commented):
Thanks for your contribution @stevenharman ! 

Would you kindly add an automated test so that we are able to avoid regressions in the future?

stevenharman:
@arielvalentin Sure thing. I just pushed up a rebased commit with a test.

arielvalentin (approved):

robbkidd (approved):

stevenharman:
I know this has already been merged (and recently released), but I was auditing the changes across all of the Otel instrumentation libs we depend on to see about upgrading and realized that I'd not added this fix to the CHANGELOG. Would you be opposed to me opening another PR to add that retroactively?

ahayworth:
Sure, go ahead! @arielvalentin that probably won't make the release tooling unhappy, right? Probably not? 😆

arielvalentin:
We'll find out!

PR #222, opened 2022-12-19
feat: add option to configure span name

fbogsany:
This adds a configuration option for MySQL instrumentation to control the span name. At present, we default to expensive client-side parsing of the expression to determine the statement type, however this is contrary to recommendations in the Semantic Conventions.

This adds 3 options, defaulting to the current behaviour: `statement_type`, `db_name` and `db_operation_and_name`. The `db_name` option extracts the database name from the `query_options`, if available. The `db_operation_and_name` option additionally looks for `db.operation` in the `OpenTelemetry::Instrumentation::Mysql2.attributes` (from the current context). This allows callers to add `db.operation` to the context, if they know it.

In all cases, if no useful span name can be determined, we default to `mysql`.

### TODO
- [x] Add tests for new config option

### DIFF
--- a/instrumentation/mysql2/lib/opentelemetry/instrumentation/mysql2/instrumentation.rb
+++ b/instrumentation/mysql2/lib/opentelemetry/instrumentation/mysql2/instrumentation.rb
@@ -30,6 +30,7 @@ module OpenTelemetry
         option :peer_service, default: nil, validate: :string
         option :enable_sql_obfuscation, default: false, validate: :boolean
         option :db_statement, default: :include, validate: %I[omit include obfuscate]
+        option :span_name, default: :statement_type, validate: %I[statement_type db_name db_operation_and_name]
 
         private
 
--- a/instrumentation/mysql2/lib/opentelemetry/instrumentation/mysql2/patches/client.rb
+++ b/instrumentation/mysql2/lib/opentelemetry/instrumentation/mysql2/patches/client.rb
@@ -95,7 +95,7 @@ module OpenTelemetry
             %r{'|"|\/\*|\*\/}.match(obfuscated)
           end
 
-          def database_span_name(sql)
+          def database_span_name(sql) # rubocop:disable Metrics/CyclomaticComplexity
             case config[:span_name]
             when :statement_type
               extract_statement_type(sql)
--- a/instrumentation/mysql2/lib/opentelemetry/instrumentation/mysql2/patches/client.rb
+++ b/instrumentation/mysql2/lib/opentelemetry/instrumentation/mysql2/patches/client.rb
@@ -96,19 +96,22 @@ module OpenTelemetry
           end
 
           def database_span_name(sql)
-            # Setting span name to the SQL query without obfuscation would
-            # result in PII + cardinality issues.
-            # First attempt to infer the statement type then fallback to
-            # current Otel approach {database.component_name}.{database_instance_name}
-            # https://github.com/open-telemetry/opentelemetry-python/blob/39fa078312e6f41c403aa8cad1868264011f7546/ext/opentelemetry-ext-dbapi/tests/test_dbapi_integration.py#L53
-            # This creates span names like mysql.default, mysql.replica, postgresql.staging etc.
-
-            statement_type = extract_statement_type(sql)
-
-            return statement_type unless statement_type.nil?
-
-            # fallback
-            database_name ? "mysql.#{database_name}" : 'mysql'
+            case config[:span_name]
+            when :statement_type
+              extract_statement_type(sql)
+            when :db_name
+              database_name
+            when :db_operation_and_name
+              op = OpenTelemetry::Instrumentation::Mysql2.attributes['db.operation']
+              name = database_name
+              if op && name
+                "#{op} #{name}"
+              elsif op
+                op
+              elsif name
+                name
+              end
+            end || 'mysql'
           end
 
           def database_name
--- a/instrumentation/mysql2/test/opentelemetry/instrumentation/mysql2/instrumentation_test.rb
+++ b/instrumentation/mysql2/test/opentelemetry/instrumentation/mysql2/instrumentation_test.rb
@@ -309,6 +309,136 @@ describe OpenTelemetry::Instrumentation::Mysql2::Instrumentation do
           end
         end
       end
+
+      describe 'when span_name is set as statement_type' do
+        it 'sets span name to statement type' do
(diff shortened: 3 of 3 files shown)


robertlaurin (approved):

instrumentation/mysql2/lib/opentelemetry/instrumentation/mysql2/instrumentation.rb
```diff
@@ -30,6 +30,7 @@ class Instrumentation < OpenTelemetry::Instrumentation::Base
         option :peer_service, default: nil, validate: :string
         option :enable_sql_obfuscation, default: false, validate: :boolean
         option :db_statement, default: :include, validate: %I[omit include obfuscate]
+        option :span_name, default: :statement_type, validate: %I[statement_type db_name db_operation_and_name]
```
ericmustin, line 33: not exactly the most descriptive configuration option but i don't really have any better ideas

instrumentation/mysql2/lib/opentelemetry/instrumentation/mysql2/patches/client.rb
```diff
@@ -95,20 +95,23 @@ def detect_unmatched_pairs(obfuscated)
...
-            return statement_type unless statement_type.nil?
-
-            # fallback
-            database_name ? "mysql.#{database_name}" : 'mysql'
+          def database_span_name(sql) # rubocop:disable Metrics/CyclomaticComplexity
+            case config[:span_name]
+            when :statement_type
+              extract_statement_type(sql)
+            when :db_name
+              database_name
```
plantfansam, line 103: Not a new problem and hence not a blocker, but worth flagging that `database_name` could be an empty string (see code [here](https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/222/files#diff-09613a195baead8b773b7041065d5878ed88c7742a035db6e9bee436cbc05c0eR119)).

ericmustin (approved):
lgtm, if i recall correctly there wasn't a lot of intention when i wrote this originally, was mostly just c/p from dd-trace-rb, so makes sense to have some non performant code paths

instrumentation/mysql2/test/opentelemetry/instrumentation/mysql2/instrumentation_test.rb
```diff
@@ -309,6 +309,136 @@
...
+          OpenTelemetry::TestHelpers.with_env('OTEL_RUBY_INSTRUMENTATION_MYSQL2_CONFIG_OPTS' => 'span_name=db_name') do
+            instrumentation.instance_variable_set(:@installed, false)
+            instrumentation.install
+
+            sql = "SELECT * from users where users.id = 1 and users.email = 'test@test.com'"
+            expect do
+              client.query(sql)
+            end.must_raise Mysql2::Error
+
+            _(span.name).must_equal 'mysql' # TODO: change the db name so we can distinguish it from the default
```
plantfansam, line 354: Could also stub out the `database_name` method in the test, but that is less good than what's proposed in this TODO 🤷

instrumentation/mysql2/test/opentelemetry/instrumentation/mysql2/instrumentation_test.rb
```diff
@@ -309,6 +309,136 @@
...
+          end
+        end
+
+        describe 'when db name is nil' do
```
plantfansam, line 358: Is this meant to nest under line 343? Could also just update test language to read `when span_name=db_name and db name is nil` or something.

instrumentation/mysql2/lib/opentelemetry/instrumentation/mysql2/instrumentation.rb
plantfansam, line 33: Something feels weird to me about the `db_operation_and_name` option. I wonder if the option should communicate that it's reading off the `Mysql2.attributes` hash, somehow?  And writing that, maybe the option should _just_ defer to a `Mysql2.attributes` attribute and not include the `db_name`? And if users want to also inject db name, there could be another option for that? `%I[statement db_name mysql_attr enriched_mysql_attr]` or something. To be clear, I'm just musing, and don't think this is a blocker.

plantfansam (approved):
This seems totally reasonable to me. I have a couple thoughts inline, but nothing I feel strongly enough about to block on.

instrumentation/mysql2/lib/opentelemetry/instrumentation/mysql2/instrumentation.rb
fbogsany: The argument for `"#{db_operation} #{db_name}"` is [that that's what the spec suggests](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/database.md):
> If `db.sql.table` is not available due to its semantics, the span SHOULD be named `<db.operation> <db.name>`.

Of course, that raises the question: can we get the table somehow without parsing?
> SQL spans SHOULD be named the following way, unless the statement is known to be of low cardinality: `<db.operation> <db.name>.<db.sql.table>`, provided that `db.operation` and `db.sql.table` are available.

instrumentation/mysql2/lib/opentelemetry/instrumentation/mysql2/patches/client.rb
fbogsany: Empty string or `nil`? The `nil` case is handled below with `|| 'mysql'`. I'm not sure which code you linked to.

instrumentation/mysql2/test/opentelemetry/instrumentation/mysql2/instrumentation_test.rb
fbogsany: It is meant to nest, yes - this lets me ... `let(:database) { nil }`, overriding the default in the `describe 'tracing'` block on line 36.

instrumentation/mysql2/lib/opentelemetry/instrumentation/mysql2/instrumentation.rb
plantfansam: Ah, RTFS 😄 - makes sense.

instrumentation/mysql2/lib/opentelemetry/instrumentation/mysql2/patches/client.rb
plantfansam: I was reading the code wrong, but was looking at the [database_name definition](https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/main/instrumentation/mysql2/lib/opentelemetry/instrumentation/mysql2/patches/client.rb#L117-L120C14).  It can result in a `nil` (`(nil)&.to_s => nil`), which is handled, here.

Issue #32, opened 2021-12-18
SQL Instrumentation Improvements


arielvalentin:
> not blocking bc it's a huge change and out of scope for this PR but, along with some broad http helpers that oughta be abstracted into a gem, to be DRY we should also probably move the mysql obfuscation regex into a db helper gem of some sort eventually/ideally as an early todo in 2022.
> 
> _Originally posted by @ericmustin in https://github.com/open-telemetry/opentelemetry-ruby/pull/1063#r771863967_

There is structural duplication in SQL instrumentation gems, e.g. obfuscation and semconv, as well as knowledge duplication between the `mysql2` and `triology` instrumentations. To minimize maintenance burden of these libraries it will likely be best to extract reusable code and provides more focused test coverage. 

In addition to that there is not any test coverage for UDS connections and should update semantic conventions to reflect these values.

arielvalentin:
@ahayworth @ericmustin I started to look at this and realized we have a little bit of re-inventing the wheel going on here. 

It looks like the original code was based off the NewRelic agent, which has an all-in-one set of Database utilities that supports multiple dialects. It's unclear to me why we didn't migrate this code into this repo, which I imagine has quite a bit of production usage. 

Instead of refactoring out common functionality from our classes I would like to propose we revisit this and import the code from the NewRelic agent and credit them with authorship.

https://github.com/newrelic/newrelic-ruby-agent/blob/dev/lib/new_relic/agent/database/obfuscation_helpers.rb

https://github.com/newrelic/newrelic-ruby-agent/tree/dev/test/fixtures/cross_agent_tests/sql_obfuscation

Apache 2 License https://github.com/newrelic/newrelic-ruby-agent/blob/dev/LICENSE

ericmustin:
Happy to see this refactored into something reusable and more dry, i forget what the resolution was previously but my understanding is if we have concerns about OSS crediting we can just include a reference to the NR library in a license-3rdparty.csv file rather than adding a dependency, which id prefer to avoid. there were some New Relic folks who were involved in the SIG at the time this was contributed and didn’t raise any objections to our crediting approach (just a code comment iirc) 

Example: https://github.com/DataDog/dd-trace-rb/blob/master/LICENSE-3rdparty.csv

arielvalentin:
[...]

PR #224, opened 2022-12-23
feat: add config[:obfuscation_limit] to pg and mysql2

reid-rigo:
Enable the following:

```ruby
c.use 'OpenTelemetry::Instrumentation::PG', db_statement: :obfuscate, obfuscation_limit: 1000
```
to allow users to control obfuscation performance via a configurable limit.

Additionally, when SQL is over the limit, truncate it to the first regex match to make output more useful for analysis:
```
SELECT * from users where users.id =... 
SQL truncated (> 1000 characters)
```

### DIFF
--- a/instrumentation/mysql2/test/opentelemetry/instrumentation/mysql2/instrumentation_test.rb
+++ b/instrumentation/mysql2/test/opentelemetry/instrumentation/mysql2/instrumentation_test.rb
@@ -172,6 +172,20 @@ describe OpenTelemetry::Instrumentation::Mysql2::Instrumentation do
         _(span.attributes['net.peer.name']).must_equal host.to_s
         _(span.attributes['net.peer.port']).must_equal port.to_s
       end
+
+      describe 'with obfuscation_limit' do
+        let(:config) { { db_statement: :obfuscate, obfuscation_limit: 10 } }
+
+        it 'truncates SQL using config limit' do
+          sql = "SELECT * from users where users.id = 1 and users.email = 'test@test.com'"
+          obfuscated_sql = "SELECT * from users where users.id = \nSQL truncated (> #{config[:obfuscation_limit]} characters)"
+          expect do
+            client.query(sql)
+          end.must_raise Mysql2::Error
+
+          _(span.attributes['db.statement']).must_equal obfuscated_sql
+        end
+      end
     end
 
     describe 'when db_statement set as omit' do
--- a/instrumentation/mysql2/test/opentelemetry/instrumentation/mysql2/instrumentation_test.rb
+++ b/instrumentation/mysql2/test/opentelemetry/instrumentation/mysql2/instrumentation_test.rb
@@ -178,7 +178,7 @@ describe OpenTelemetry::Instrumentation::Mysql2::Instrumentation do
 
         it 'truncates SQL using config limit' do
           sql = "SELECT * from users where users.id = 1 and users.email = 'test@test.com'"
-          obfuscated_sql = "SELECT * from users where users.id = \nSQL truncated (> #{config[:obfuscation_limit]} characters)"
+          obfuscated_sql = "SELECT * from users where users.id = ...\nSQL truncated (> 10 characters)"
           expect do
             client.query(sql)
           end.must_raise Mysql2::Error
--- a/instrumentation/mysql2/lib/opentelemetry/instrumentation/mysql2/instrumentation.rb
+++ b/instrumentation/mysql2/lib/opentelemetry/instrumentation/mysql2/instrumentation.rb
@@ -31,6 +31,7 @@ module OpenTelemetry
         option :enable_sql_obfuscation, default: false, validate: :boolean
         option :db_statement, default: :include, validate: %I[omit include obfuscate]
         option :span_name, default: :statement_type, validate: %I[statement_type db_name db_operation_and_name]
+        option :obfuscation_limit, default: 2000, validate: :integer
 
         private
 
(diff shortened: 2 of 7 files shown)


rigoleto:
I'm working with my employer on the CLA and will leave this as a draft until I'm authorized.

ahayworth:
@reid-rigo Thanks for the PR, and let us know once the CLA is signed.

---

That said, I'm a little conflicted about this PR:
- If you (personally) are looking for a _higher_ limit, doesn't that imply that we are perhaps being too conservative here and should instead just raise the limit for everyone?
- If you (personally) are looking for a _lower_ limit, doesn't that imply that we are not being conservative enough and should instead lower the limit for everyone?

It may help to know that while span size _is_ a concern in general, the main motivation behind the truncation limit is to avoid running a regular expression that might a long time in a hot code path. Truncating it after the first match implies running that regular expression, and I think that defeats one of the the purposes of the check to begin with.

With all that in mind - can you elaborate on your goals with this PR? That might help us understand what the right course of action should be.

---

Also worth mentioning - we might consider conditionally enabling or disabling this at all based on Ruby version. The latest release has greatly improved the speed of regular expressions and also introduced a "timeout" that could also help keep this code path speedy.

ref: https://bugs.ruby-lang.org/issues/19104 https://bugs.ruby-lang.org/issues/17837

reid-rigo:
@ahayworth, in my opinion the current limit is far too low. I really want to understand the performance of big queries, and the milliseconds spent obfuscating the SQL is likely to be dwarfed by the query itself. As a reference, New Relic's [limit is 16384](https://github.com/newrelic/newrelic-ruby-agent/blob/edd2cc413dce0e90f5f9a9e030b3dc97af2af99b/lib/new_relic/agent/database.rb#L26).

As for the use or `index(regex)`, in my testing it's dramatically faster than running the full gsub. And of course the truncated query still gives me a lot to look at.

The Ruby regex development is really interesting, and I saw it shortly after working on this. I haven't tested it yet, but increasing the default limit based on Ruby version could make a lot of sense.

arielvalentin:
cc: https://github.com/open-telemetry/opentelemetry-ruby-contrib/issues/32

ahayworth:
> As for the use or index(regex), in my testing it's dramatically faster than running the full gsub. And of course the truncated query still gives me a lot to look at.

Interesting, that's not what I would have expected. Ruby always has interesting surprises for me, even after many years with it. 😄 It's a good surprise, though!

> in my opinion the current limit is far too low. I really want to understand the performance of big queries, and the milliseconds spent obfuscating the SQL is likely to be dwarfed by the query itself. As a reference, New Relic's [limit is 16384](https://github.com/newrelic/newrelic-ruby-agent/blob/edd2cc413dce0e90f5f9a9e030b3dc97af2af99b/lib/new_relic/agent/database.rb#L26).
> ...
> but increasing the default limit based on Ruby version could make a lot of sense

I would support increasing it, regardless of Ruby version based on New Relic's defaults actually. I am curious if @robertlaurin has any thoughts here; I recall he had done some benchmarking on sanitization performance last year.

arielvalentin:
Anecdotes: Sanitizing the SQL queries is _very slow_ in our prod environments. I do not have any profiles to share at the moment but I will do my best to share something with y'all. 

I am still interested in finding a good way to optimize or offload the SQL sanitization from the users request path but did not invest enough time in that recently. Technically we should not be doing any scrubbing in the SDK but there isn't anything in the collector that does this for us AFAIK. 

I think that is why https://github.com/open-telemetry/opentelemetry-ruby-contrib/issues/32 will be critical for our users so we may optimize as much as we can in a single module/package.

reid-rigo:
@ahayworth, the CLA is signed. I'd be happy to separate out the truncation part of this to keep this PR purely about the obfuscation limit.

@arielvalentin, would you able to benchmark truncation with one of your big SQL queries? Here's what I used to test:

```ruby
require "benchmark"

COMPONENTS_REGEX_MAP = {
  single_quotes: /'(?:[^']|'')*?(?:\\'.*|'(?!'))/,
  dollar_quotes: /(\$(?!\d)[^$]*?\$).*?(?:\1|$)/,
  uuids: /\{?(?:[0-9a-fA-F]\-*){32}\}?/,
  numeric_literals: /-?\b(?:[0-9]+\.)?[0-9]+([eE][+-]?[0-9]+)?\b/,
  boolean_literals: /\b(?:true|false|null)\b/i,
  comments: /(?:#|--).*?(?=\r|\n|$)/i,
  multi_line_comments: /\/\*(?:[^\/]|\/[^*])*?(?:\*\/|\/\*.*)/
}.freeze

POSTGRES_COMPONENTS = %i[
  single_quotes
  dollar_quotes
  uuids
  numeric_literals
  boolean_literals
  comments
  multi_line_comments
].freeze

generated_postgres_regex = Regexp.union(POSTGRES_COMPONENTS.map { |component| COMPONENTS_REGEX_MAP[component] })
sql = <<SQL
  ...
SQL

puts "SQL length: #{sql.length}"

times = 1000
Benchmark.bm(16) do |x|
  x.report("full gsub") do
    times.times do
      sql.gsub(generated_postgres_regex, '?')
    end
  end
  x.report("first index") do
    times.times do
      sql[..sql.index(generated_postgres_regex) - 1]
    end
  end
end
```

arielvalentin:
Very sorry about this falling off of my radar. I did not try the proposed changes out against some of our more gnarly production queries but I will schedule time to do this later in the week of 2023-04-03.

I will also add that we use `trilogy` and not `mysql2` since that gem has not had active maintenance for quite some time.

@ahayworth would you mind re-engaging here? I would like to see about resolving your concerns.

ericmustin:
this should stay open i think

arielvalentin:
I agree! 

I think we should review this again and see about getting it merged.

arielvalentin (approved):

instrumentation/mysql2/test/opentelemetry/instrumentation/mysql2/instrumentation_test.rb
```diff
@@ -172,7 +172,7 @@
         _(span.attributes['net.peer.name']).must_equal host.to_s
         _(span.attributes['net.peer.port']).must_equal port.to_s
       end
-
+
```
arielvalentin, line 175:
```suggestion

```

PR #227, opened 2022-12-26
chore: Enable Style/RedundantConstantBase

arielvalentin:
https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Style/RedundantConstantBase


ericmustin (approved):
as far as i can tell this is all `_test.rb` and `.gemspec` files, so, yolo. but ya a bit of a doozy wrt files changed 😄

PR #232, opened 2022-12-26
fix: Reduce Hash Allocations in PG Instrumentation

arielvalentin:
https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Style/CollectionCompact

### DIFF
--- a/instrumentation/pg/lib/opentelemetry/instrumentation/pg/patches/connection.rb
+++ b/instrumentation/pg/lib/opentelemetry/instrumentation/pg/patches/connection.rb
@@ -86,7 +86,7 @@ module OpenTelemetry
             attrs = { 'db.operation' => validated_operation(operation), 'db.postgresql.prepared_statement_name' => statement_name }
             attrs['db.statement'] = sql unless config[:db_statement] == :omit
             attrs.merge!(OpenTelemetry::Instrumentation::PG.attributes)
-            attrs.reject! { |_, v| v.nil? }
+            attrs.compact!
 
             [span_name(operation), client_attributes.merge(attrs)]
           end
@@ -131,7 +131,7 @@ module OpenTelemetry
             }
             attributes['peer.service'] = config[:peer_service] if config[:peer_service]
 
-            attributes.merge(transport_attrs).reject { |_, v| v.nil? }
+            attributes.merge(transport_attrs).compact
           end
 
           def transport_addr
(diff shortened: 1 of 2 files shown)


instrumentation/pg/lib/opentelemetry/instrumentation/pg/patches/connection.rb
```diff
@@ -86,7 +86,7 @@ def span_attrs(kind, *args)
             attrs = { 'db.operation' => validated_operation(operation), 'db.postgresql.prepared_statement_name' => statement_name }
             attrs['db.statement'] = sql unless config[:db_statement] == :omit
             attrs.merge!(OpenTelemetry::Instrumentation::PG.attributes)
-            attrs.reject! { |_, v| v.nil? }
+            attrs.compact!
```
ericmustin, line 89: is there any perf benefit to inlining this and not using a destructive `!` method?, ie in the line below

```
            [span_name(operation), client_attributes.merge(attrs.compact)]
```

ericmustin (approved):
minor comment but not blocking, lgtm

instrumentation/pg/lib/opentelemetry/instrumentation/pg/patches/connection.rb
arielvalentin: I think instrumentations should limit the amount of Objects they create since it will contribute to triggering GC and require more memory.

instrumentation/pg/lib/opentelemetry/instrumentation/pg/patches/connection.rb
```diff
@@ -131,7 +131,7 @@ def client_attributes
             }
             attributes['peer.service'] = config[:peer_service] if config[:peer_service]
 
-            attributes.merge(transport_attrs).reject { |_, v| v.nil? }
+            attributes.merge(transport_attrs).compact
```
arielvalentin, line 134: Now that I am reading this... it should use `merge!` and `compact!`

instrumentation/pg/lib/opentelemetry/instrumentation/pg/patches/connection.rb
ericmustin: ok so then uh...perf regression. yea let's not do that lol

arielvalentin: Checkout cd98a8aad8bf044a0d106ab2d2ecdb7a7ffc38a9

ericmustin (approved):
🚀

PR #235, opened 2022-12-26
chore: Enable Lint/RedundantRequireStatement

arielvalentin:
https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Lint/RedundantRequireStatement


instrumentation/active_record/lib/opentelemetry/instrumentation/active_record/instrumentation.rb
```diff
@@ -60,7 +60,6 @@ def patch
         end
 
         def require_dependencies
-          require 'ruby2_keywords'
```
ericmustin, line 63: mmmm are we sure we can rip out this dependency i feel like there's some context here i'm missing (or have forgotten as my brain cells continue to dwindle thanks to my gas stove).

ericmustin (approved):
it seems suspicious to me that we just have this dependency in here unnecessarily, especially as ruby2_keywords is sorta pseudo special ruby gem. approving as i trust your judgement but wouldnt mind someone else taking a look on this one

instrumentation/active_record/lib/opentelemetry/instrumentation/active_record/instrumentation.rb
arielvalentin: There may be some corner cases for Ruby 2.7 so I will revert this.

arielvalentin:
```console
🤘 bundle _2.4.3_ exec rubocop
Inspecting 24 files
......W.................

Offenses:

lib/opentelemetry/instrumentation/active_record/instrumentation.rb:63:11: W: [Correctable] Lint/RedundantRequireStatement: Remove unnecessary require statement.
          require 'ruby2_keywords'
          ^^^^^^^^^^^^^^^^^^^^^^^^
```

When I enable the rule and revert this line I get this warning.

arielvalentin: I need to look at the docs to see _when_ this is required. I still feel like this should be required by the _application_ using the instrumentation gems and _not_ by our gems.

arielvalentin: Looking at the linter rule again it is looking for a file that was _already_ required. Bundler is requiring this file for us so I am going to disable this check for that line since our AR monkey patches are using ruby2_keywords.

arielvalentin: Fixed a421b4500dd15ab566c464ba7a6971c2cf1b788a

ahayworth (approved):
I think this is okay to ship as-is: we're not changing the behavior of the instrumentation, just making rubocop shut up about re-requiring `ruby2_keywords`.

Whether or not `ruby2_keywords` is necessary at all is another matter entirely, but not one that must be solved here. So ... 🚢 🚢 🚢

PR #238, opened 2022-12-26
chore: Enable Lint/AmbiguousOperatorPrecedence

arielvalentin:
https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Lint/AmbiguousOperatorPrecedence


instrumentation/redis/lib/opentelemetry/instrumentation/redis/middlewares/redis_client.rb
```diff
@@ -59,7 +59,7 @@ def serialize_commands(commands)
             serialized_commands = commands.map do |command|
               # If we receive an authentication request command we want to obfuscate it
               if obfuscate || command[0].match?(/\A(AUTH|HELLO)\z/)
-                command[0].to_s.upcase + ' ?' * (command.size - 1)
+                command[0].to_s.upcase + (' ?' * (command.size - 1))
```
ericmustin, line 62: Maybe there's an easy performance win here if we just reference a frozen string literal `?` as a constant rather than use object allocations to create this dynamic length `?????` string. idk wdyt? cc @plantfansam who rubys better than me.

ericmustin (approved):
i had one minor point of feedback but not blocking

instrumentation/redis/lib/opentelemetry/instrumentation/redis/middlewares/redis_client.rb
arielvalentin: i would say it is very likely and we should be able to use `concat` instead of allocating new string objects. 

```
irb(main):001:0> t = "hello".to_s.upcase
=> "HELLO"
irb(main):002:0> t.object_id
=> 33160
irb(main):003:0> u = ( ' ?' * 10)
=> " ? ? ? ? ? ? ? ? ? ?"
irb(main):004:0> u.object_id
=> 93080
irb(main):005:0> v = t + u
=> "HELLO ? ? ? ? ? ? ? ? ? ?"
irb(main):006:0> v.object_id
=> 111380
irb(main):009:0> t.concat(u)
=> "HELLO ? ? ? ? ? ? ? ? ? ?"
irb(main):010:0> t.object_id
=> 33160
```

arielvalentin: @ericmustin I think I am going to defer that optimization and ship it in a different PR

Issue #218, opened 2022-12-06
Migrate Rails instrumentation to use only ActiveSupport::Notifications


ahayworth:
We recently uncovered an issue wherein our [action_pack instrumentation caused a production problem](https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/123#discussion_r1035711863). There are many ways to consider such an issue, but one way is to acknowledge that monkey-patching is fraught with peril, and that we should consider different methods of instrumentation.

For Rails, we should be able to use ActiveSupport::Notifications. We already use this to great effect in our [action_view instrumentation](https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/main/instrumentation/action_view/lib/opentelemetry/instrumentation/action_view/railtie.rb), and we think we can expand this further to the rest of our Rails instrumentation.

To do this, we need to:

- [ ] Verify that Rails 7 implements safer subscription start/end behavior (a crashing subscriber should not prevent us from finishing our spans correctly).
- [ ] Audit the instrumentation we have for Rails, and figure out what ActiveSupport notifications could provide similar levels of instrumentation:
  - [ ] action_pack
  - [ ] action_view
  - [ ] active_job
  - [ ] active_record
  - [ ] active_support
- [ ] Augment our [custom span subscriber](https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/main/instrumentation/active_support/lib/opentelemetry/instrumentation/active_support/span_subscriber.rb) in any way necessary to support a more broad use-case than what it does now
  - We need to figure out if there is a better way to do this in Rails 7 and if we should adjust this span subscriber.
  - We also should look and see if we could backport the Rails 7 fixes here ... or get that backported upstream.
- [ ] Work through instrumentation and convert them.
  - [ ] Depending on the outcome of the safety/backport investigations, we can decide whether or not we wish to keep the monkey-patched versions around for older Rails installations or not.

ahayworth:
This is the PR that improved safety in Rails 7: https://github.com/rails/rails/pull/43282

plantfansam:
I think this is a great idea. Even if we discover issues with ASN, that sounds like a great reason to contribute fixes upstream 😄

robbkidd:
Talked about this during the SIG today. Attendees are fans of the idea to move away from monkeypatching and towards AS::N subscriptions.

tjefferson08:
Can AS::N support `current_span` ?

[...]

PR #242, opened 2022-12-27
fix: Drop Rails dependency for ActiveSupport Instrumentation

arielvalentin:
This makes it possible to use this instrumentation in non-Rails applications that leverage AS::Notifications.

In addition to that it mitigates a race condition where a subscription may be registered by a separate thread before obtaining a lock on the notifications object.

See https://github.com/open-telemetry/opentelemetry-ruby-contrib/issues/218

Note to the reviewers:

This code evaluates the ActiveSupport version for each subscription in order to determine how to look up the existing subscriber list and should likely be computed during installation instead of subscription time and I chose not to address this as part of this PR.

It looks like this code exists to accommodate support for Rails 5.2, which is EOL, and I will drop support for Rails 5 in a separate PR.


instrumentation/active_support/lib/opentelemetry/instrumentation/active_support/span_subscriber.rb
```diff
@@ -29,18 +32,36 @@ def self.subscribe(
...
+            )
           end
         end
         subscriber_object
       end
 
+      def self.find_subscribers(pattern)
+        active_support_major_version = ::ActiveSupport.version.canonical_segments.first
+
+        # TODO: Drop support for Rails 5 since it is EOL since 2022-06-01
```
mwear, line 57: Do we have an official policy for Rails support? It would be nice to give a grace period beyond the EOL, but it is probably a larger discussion. I would imagine it being something reasonable, like EOL + 1 yr. Whatever we decide, we should have a clear policy.

mwear (approved):

instrumentation/active_support/lib/opentelemetry/instrumentation/active_support/span_subscriber.rb
arielvalentin: Nothing official. These are all pre 1.0 so they are subject to change at any time. 

Please see this related PR to drop support for Rails 5

https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/259

mwear: I 👀 @fbogsany approves. I don't have any strong opinions other than, whatever works for the community works for us. If we drop support and nobody complains, then we're probably fine.

instrumentation/active_support/lib/opentelemetry/instrumentation/active_support/span_subscriber.rb
```diff
@@ -29,18 +32,36 @@ def self.subscribe(
...
         subscriber_object = ::ActiveSupport::Notifications.subscribe(pattern, subscriber)
+
         ::ActiveSupport::Notifications.notifier.synchronize do
-          if ::Rails::VERSION::MAJOR >= 6
-            s = ::ActiveSupport::Notifications.notifier.instance_variable_get(:@string_subscribers)[pattern].pop
-            ::ActiveSupport::Notifications.notifier.instance_variable_get(:@string_subscribers)[pattern].unshift(s)
+          subscribers = find_subscribers(pattern)
+
+          if subscribers.nil?
+            OpenTelemetry.handle_error(
```
ericmustin, line 40: should this be a debug log rather than an error? a bit of semantics (since the default error handler merely logs the exception.message and message), defer to your judgement.

ericmustin (approved):

instrumentation/active_support/lib/opentelemetry/instrumentation/active_support/span_subscriber.rb
arielvalentin: I chose handle_error since it is an unexpected and unrecoverable. 

The intent is to bring it to our attention early on or have users file a bug report.

PR #244, opened 2022-12-28
chore: Use Bundler bootstrap dependencies

arielvalentin:
This change leverages bundler to require all of the appropriate dependencies uses in tests instead of requireing them manually.


mwear (approved):
LGTM

Issue #249, opened 2022-12-28
Drop support for Rails 5.2


arielvalentin:
EOL was on 2022-06-01

pb-dod:
😢 This makes it very difficult to instrument legacy services.

arielvalentin:
@pb-dod You will still be able to instrument services with EoL versions of rails as long as you pin to a specific version of the instrumentation:

https://github.com/open-telemetry/opentelemetry-ruby-contrib/tree/main/instrumentation/rails#version-compatability

PR #259, opened 2022-12-29
feat!: Drop Rails 5 Support

arielvalentin:
Closes https://github.com/open-telemetry/opentelemetry-ruby-contrib/issues/249

### DIFF
--- a/instrumentation/active_job/opentelemetry-instrumentation-active_job.gemspec
+++ b/instrumentation/active_job/opentelemetry-instrumentation-active_job.gemspec
@@ -28,7 +28,7 @@ Gem::Specification.new do |spec|
   spec.add_dependency 'opentelemetry-api', '~> 1.0'
   spec.add_dependency 'opentelemetry-instrumentation-base', '~> 0.21.0'
 
-  spec.add_development_dependency 'activejob', '>= 5.2.0'
+  spec.add_development_dependency 'activejob', '>= 6.0.0', '< 8.0.0'
   spec.add_development_dependency 'appraisal', '~> 2.2.0'
   spec.add_development_dependency 'bundler', '~> 2.4'
   spec.add_development_dependency 'minitest', '~> 5.0'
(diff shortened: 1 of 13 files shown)


ericmustin (approved):

instrumentation/active_job/opentelemetry-instrumentation-active_job.gemspec
```diff
@@ -28,7 +28,7 @@ Gem::Specification.new do |spec|
   spec.add_dependency 'opentelemetry-api', '~> 1.0'
   spec.add_dependency 'opentelemetry-instrumentation-base', '~> 0.21.0'
 
-  spec.add_development_dependency 'activejob', '>= 5.2.0'
+  spec.add_development_dependency 'activejob', '>= 6.0.0', '< 8.0.0'
```
ahayworth, line 31: Any reason not to optimistically support future activejobs?

ahayworth (approved):
Overall LGTM but I did have one question about the dependency version changes (but no real worries about it either way)

instrumentation/active_job/opentelemetry-instrumentation-active_job.gemspec
arielvalentin: Personally speaking I think gem requirements send a more explicit signal about what we support and what we don't but I do not have to add that constraint here if you prefer I remove it.

ahayworth: 🤷 Honestly I don't have any strong feelings either way; I wasn't sure if it was an implicit signal that there may be problems with future versions or something. 😄

arielvalentin:
Summarizing our discussion from SIG on 2023-01-07

Before Merging This PR...

1. Open a discussion and post in CNCF slack about this PR along with release timelines
2. Open a PR that prints deprecation warnings and release a patch version of the gem
3. Provide instructions on what version to pin OTel instrumentation for Rails 5 apps

arielvalentin:
@ahayworth et.al. I have had yet _another_ change of heart. 

I would like to merge this before migrating active support instrumentation. I would like to minimize any challenges we will run into due to workarounds with limitations in Rails 5 when switching to AS. 

I will also remove the restrictions I added to development dependencies before merging

ahayworth (approved):
🚢

Issue #250, opened 2022-12-28
Docker issues


arielvalentin:
I am unable to runs some of the examples on an M1 Mac due to comparability issues with older image architectures. 

It also looks like the image we try to build is running in Ruby 2.6 by default.

I think we should upgrade it to run something that would allow us to reproduce the CI environment to improve our ability to debug issues with specific platforms and versions.

PR #263, opened 2022-12-30
chore: Upgrade some Docker dependencies to latest images

arielvalentin:
This commit introduces a few changes to our development docker image:

1. Sets the default Ruby version to 3.1.3 as well as removes the ability to set a custom bundler version and uses the latest gem version
2. Upgrades to mysql 8 since there are no 5.7 images available for Apple Silicon
3. Makes it possible to start up the main service, named `app`, with all of the dependencies to make it possible to run tests in a container
4. Overrides GitHub Action MySQL service to use a different entrypoint script, which is not currently supported via the YAML configurations

See https://github.com/open-telemetry/opentelemetry-ruby-contrib/issues/250
See https://stackoverflow.com/questions/60902904/how-to-pass-mysql-native-password-to-mysql-service-in-github-actions
See https://github.com/github/trilogy/issues/26


arielvalentin:
Blocked by https://github.com/github/trilogy/issues/26

robertlaurin (approved):

mwear (approved):

PR #267, opened 2023-01-09
Check if env['sinatra.error'] exists before recording it

rdoherty-fastly:
Fixes https://github.com/open-telemetry/opentelemetry-ruby/issues/1420

Needs tests, I would like tips on if that would be required since afaict there aren't tests for this type of guard code. Thanks!

### DIFF
--- a/instrumentation/sinatra/lib/opentelemetry/instrumentation/sinatra/middlewares/tracer_middleware.rb
+++ b/instrumentation/sinatra/lib/opentelemetry/instrumentation/sinatra/middlewares/tracer_middleware.rb
@@ -31,7 +31,7 @@ module OpenTelemetry
             sinatra_response = ::Sinatra::Response.new([], response.first)
             return unless sinatra_response.server_error?
 
-            span.record_exception(env['sinatra.error'])
+            span.record_exception(env['sinatra.error']) if env['sinatra.error']
             span.status = OpenTelemetry::Trace::Status.error
           end
         end


arielvalentin:
@rdoherty-fastly any luck getting sign off on CLA?

rdoherty-fastly:
> @rdoherty-fastly any luck getting sign off on CLA?

@arielvalentin I am kickstarting this now so I could get a copy of the CLA and make sure everything on my end was setup right. I'm starting the conversation now with my org to figure out who should be CLA manager and having our legal team review the CLA. Thanks!

rdoherty-fastly:
@arielvalentin CLA signed off and I am approved. LMK if there are any other steps I need to do for this, thank you!

instrumentation/sinatra/lib/opentelemetry/instrumentation/sinatra/middlewares/tracer_middleware.rb
```diff
@@ -31,7 +31,7 @@ def trace_response(env, response)
             sinatra_response = ::Sinatra::Response.new([], response.first)
             return unless sinatra_response.server_error?
 
-            span.record_exception(env['sinatra.error'])
+            span.record_exception(env['sinatra.error']) if env['sinatra.error']
```
arielvalentin, line 34: Would you kindly add an automated test that we may use to avoid a future regression?

rdoherty-fastly: I've been digging for how to remove/make-nil `env['sinatra.error']` for a while and can't seem to find a good example of how to intercept calls to that variable that is passed into `call`. If you have any tips or places to look lmk, thank you!

arielvalentin: Hmmm... how does it manifest itself in production? Does the Sinatra code base have a test case that covers this that you may use for inspiration?

rdoherty-fastly: We have a custom error handler class that that calls [Sinatra's Base error](https://www.rubydoc.info/gems/sinatra/1.4.6/Sinatra%2FBase.error), and then calls:
```
ensure
    env.delete("sinatra.error")
end
```
I can't put all of it here b/c it's our closed source code unfortunately. 

I have a few options:

* Add some new mocking libraries that allow me to intercept all calls to Hashes and when `sinatra.error` is accessed, return nil. 
* Figure out if some middleware can delete that key (tried this already, can't seem to make it work)
* Call `trace_response` with mock/stubbed `env` and `response` input and verify it works when `env` is missing `sinatra.error`
* Intercept calls to `TracerMiddleware#trace_response`, remove `sinatra.error` and then call the original function. I like this option, but can't figure out how to make it work.
* ?

Thoughts?

ahayworth (approved):
It's definitely an odd circumstance that `sinatra_response.server_error?` would be true, while `env['sinatra.error']` would not be an exception.

But: if you've got an application where that is the case, than this is a bug we can and should fix. 👍

instrumentation/sinatra/lib/opentelemetry/instrumentation/sinatra/middlewares/tracer_middleware.rb
ahayworth: > Call trace_response with mock/stubbed env and response input and verify it works when env is missing sinatra.error

This might be the easiest way.

arielvalentin: @ahayworth should we merge to get past the bug and ask @rdoherty-fastly to follow up with test coverage?

ahayworth: I would be okay with releasing without the test case. A follow-on PR for test coverage would be fine with me.

arielvalentin (approved):
Please follow up with a test case at your earliest convenience.

rdoherty-fastly:
Thank you @arielvalentin and @ahayworth ! I agree it is a weird situation. 

Regarding tests, I think one reason I am having problems getting a test case is because of bug https://github.com/open-telemetry/opentelemetry-ruby-contrib/issues/292 - the exception handling code is throwing an exception. If we figure out that problem I believe I can proceed. Thanks!

PR #277, opened 2023-01-15
fix: Micro-optimizations for ActiveSupport Instrumentation

arielvalentin:
- Reduces the number of hashes required to instrument notifications
- Extracts constants to avoid allocating small arrays
- Use Set operations instead of linear search whenever possible

### DIFF
--- a/instrumentation/active_support/lib/opentelemetry/instrumentation/active_support/span_subscriber.rb
+++ b/instrumentation/active_support/lib/opentelemetry/instrumentation/active_support/span_subscriber.rb
@@ -4,6 +4,8 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 
+require 'set'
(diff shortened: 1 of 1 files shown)


arielvalentin:
> Do we have any before/after measurements of how much this helps? I'm suspicious about the changes in `valid_payload_value?` making much of a difference, but I'm excited to learn that I am probably very ignorant of the performance implications there.

I will extract this code into a Ruby benchmark but I do not think I can easily provide you with GC pressure relief or memory allocation metrics.

arielvalentin:
Good thing you promoted me to do the benchmarks. I ran some rudimentary benchmarks and using sets is "slower" than arrays for items under ~10 entries.

Now I want to profile it to see what happens with allocations when I switch the arrays to constants.

PR #278, opened 2023-01-15
ci: Break out datastores tests into separate builds

arielvalentin:
Builds have been unstable lately due to failing services starting up and SEGFAULT in kafka gems.

This change is an attempt at splitting up builds by datastore so they fail earlier and make it possible to rerun if necessary.

The disadvantage of this approach is that it will require additional runners to parallelize the builds ~84 checks.

image

image

_I tried to build dynamic json arrays from file names to set up a matrix but could not figure out the incantation to make it work so everything is static in the workflow files_

```bash
matrix=$((
    find . -iname "*.gemspec" -type f -exec basename {} .gemspec ';' | jq -R -s -c 'split("\n")[:-1]'
  ) | jq -c .)
  echo $matrix
  echo $matrix | jq .
```

Would always return an empty array with syntax errors: https://github.com/open-telemetry/opentelemetry-ruby-contrib/actions/runs/3924404255/jobs/6708640959#step:2:3


arielvalentin:
Ehhh... I don't know. The matrix build is unpleasant for the linters. 

Is there a better way to do this that gives us the same feedback with less resources?

arielvalentin:
@ahayworth thanks! I appreciate the support here. I feel like the matrix build adds mental-overhead and complexity for the minor benefit but I do feel strongly that removing the linter/docs/build steps into its own job helps so I want to try and do that. 

I am going to revert the changes that extract the linters into their own build for now until we can find a better option.

ahayworth:
> the matrix build adds mental-overhead and complexity for the minor benefit

I agree, actually. Matrix builds _almost_ worked really well but then I ran into issues ... and we ended up with a gross in-between world. I say this mainly to emphasize that I have no attachment to matrix builds, and if we want/need to ditch them we should!

arielvalentin:
@ahayworth et.al. I reverted separating out misc tasks and keeping them running in the 2.7 matrix builds.

.github/workflows/ci-instrumentation-with-services.yml
```diff
@@ -58,30 +46,99 @@ jobs:
...
+        gem:
+          - mysql2
+          - trilogy
+        os:
+          - ubuntu-latest
+    name: mysql / ${{ matrix.gem }} / ${{ matrix.os }}
+    runs-on: ${{ matrix.os }}
+    steps:
+      - uses: actions/checkout@v3
+      - name: "Test Ruby 3.2"
```
robbkidd, line 77: Would declaring these steps once with a YAML anchor and then referencing that anchor by alias in the multiple jobs be too DRY?

robbkidd (approved):
This all looks good. 👍🏻 

Non-blocking questions in comments below.

.github/workflows/ci-instrumentation-with-services.yml
arielvalentin: Please add a +1 to this issue: https://github.com/actions/runner/issues/1182

.github/workflows/ci-instrumentation-with-services.yml
```diff
@@ -5,34 +5,22 @@ on:
...
     branches:
       - main
-      - 'release/**'
   pull_request:
     branches:
       - main
 
 jobs:
-  # These builds require sidecar services (postgres, redis, etc) in order to run their test suites.
   instrumentation_with_services:
```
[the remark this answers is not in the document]
robbkidd: A-ha. Makes sense!

.github/workflows/ci-instrumentation-with-services.yml
robbkidd: Oof.

- Philosophically: not too DRY.
- Realistically: not possible using _only_ GitHub Actions workflow YAML.

OK. 😞

arielvalentin: Yeah the best we could do was extract the test scripts into a reusable file. That script requires parameters which is what these jobs are supposed to do.

PR #279, opened 2023-01-20
redis: Handle redis with unix socket configuration

tonobo:
The instrumentation package expected to always have host and port configured, which isn't the case in path was set. Only setting the peer name in case it has been configured.


tonobo:
Sorry guys, but i don't want to sign the cla stuff.

PR #281, opened 2023-01-21
chore: Enable dependabot

arielvalentin:
Dependabot does not support subdirectories and I did not want to bother with scripting the output:

- https://github.com/dependabot/dependabot-core/issues/2178

Dependabot does not support YAML aliases:

- https://github.com/dependabot/dependabot-core/issues/1582

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,179 @@
+---
+version: 2
+updates:
+- package-ecosystem: github-actions
+  directory: "/"
+  schedule:
+    interval: daily
+- package-ecosystem: bundler
+  directory: "/"
+  schedule:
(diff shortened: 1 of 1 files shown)


ahayworth (approved):
Approved, but one meta-question: will this significantly add to the workload? If so, maybe we do monthly instead. Just a thought.

arielvalentin:
@ahayworth I think the way it will manifest itself will be... if we can't get to it we can let it linger. if it sees an update it will close the existing PR and open a new one.

ahayworth:
Totally fine by me! Just wanted to make sure we didn't make ourselves feel worse about workloads and all. 😃 

🚢

PR #291, opened 2023-01-24
Normalize GraphQL span names for easier aggregation analysis

ravangen:
For the GraphQL integration, this:
- introduces new attributes to existing platform spans
- seeks to normalize platform span names for easier identification and future aggregation

### New Attributes

Builds on https://github.com/open-telemetry/opentelemetry-specification/pull/2456 and work done in [opentelemetry-js-contrib](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/79b2d3ff08904ce84c6bc48427cd98906c2f0d79/plugins/node/opentelemetry-instrumentation-graphql/src/utils.ts#L125-L129)

There are some differences between the GraphQL implementations in different languages that may make it harder to have globally consistent names without working with the library authors. For example, `graphql-ruby` separates out reporting of lazy (delayed, possibly async) work from its sync work, resulting in multiple spans for one [`graphql-js` "resolve" concept](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/79b2d3ff08904ce84c6bc48427cd98906c2f0d79/plugins/node/opentelemetry-instrumentation-graphql/src/enum.ts#L51).

#### Field resolve/execute

- `graphql.field.parent`: Field's parent type. For interfaces, this is the concrete type's name, not interface name.
- `graphql.field.name`: Field's name, matches [`opentelemetry-js`](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/79b2d3ff08904ce84c6bc48427cd98906c2f0d79/plugins/node/opentelemetry-instrumentation-graphql/src/utils.ts#L126)
- `graphql.lazy`: Is this the lazy part of the resolver. There will always at least an inline/sync portion (e.g. `false`), and will sometimes have lazy work to be subsequently done at a later time. No `opentelemetry-js` equivalent as they use promises.
- `graphql.field.path`: Did **not** add at this time, but the [information is available via `graphql-ruby`](https://graphql-ruby.org/api-doc/2.0.16/GraphQL/Tracing) if we want to report it. There is a [`opentelemetry-js`](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/79b2d3ff08904ce84c6bc48427cd98906c2f0d79/plugins/node/opentelemetry-instrumentation-graphql/src/utils.ts#L127) equivalent. Thoughts?
 
#### Type authorization

- `graphql.type.name`: Type name having its authorization check run
- `graphql.lazy`: Is this the lazy part of the authorization.

#### Type resolution

When a field’s return type is an interface, GraphQL has to figure out what specific object type to use for the return value.
- `graphql.type.name`: Interface name having its object type determined
- `graphql.lazy`: Is this the lazy part of the type resolution.

### Span Name Normalization

In talking with @robertlaurin about how we can use this data for analysis across many requests, we found we wanted to have a consistent span name to easily identify specify types of spans. So I moved the type/field names to attributes.

This change was made in a backwards compatible manner via `legacy_platform_span_names` config. To keep the original behaviour, set `true`. I figure if/when this gem moves to v1, we can remove this config option.

### DIFF
--- a/instrumentation/graphql/README.md
+++ b/instrumentation/graphql/README.md
@@ -41,6 +41,11 @@ OpenTelemetry::SDK.configure do |c|
     enable_platform_authorized: false,
     # enable_platform_resolve_type maps to the resolve_type and resolve_type_lazy keys
     enable_platform_resolve_type: false
+
+    # Controls if platform tracing (field/authorized/resolve_type)
+    # should use the legacy span names (e.g. "MyType.myField") or the
+    # new normalized span names (e.g. "graphql.execute_field").
+    legacy_platform_span_names: false
   }
 end
 ```
--- a/instrumentation/graphql/README.md
+++ b/instrumentation/graphql/README.md
@@ -40,7 +40,7 @@ OpenTelemetry::SDK.configure do |c|
     # enable_platform_authorized maps to the authorized and authorized_lazy keys
     enable_platform_authorized: false,
     # enable_platform_resolve_type maps to the resolve_type and resolve_type_lazy keys
-    enable_platform_resolve_type: false
+    enable_platform_resolve_type: false,
 
     # Controls if platform tracing (field/authorized/resolve_type)
     # should use the legacy span names (e.g. "MyType.myField") or the
--- a/instrumentation/graphql/lib/opentelemetry/instrumentation/graphql/tracers/graphql_tracer.rb
+++ b/instrumentation/graphql/lib/opentelemetry/instrumentation/graphql/tracers/graphql_tracer.rb
@@ -86,14 +86,14 @@ module OpenTelemetry
             attributes = {}
             case key
             when 'execute_field', 'execute_field_lazy'
-              attributes['graphql.field.parent'] = data[:owner].graphql_name # owner is the concrete type, not interface
-              attributes['graphql.field.name'] = data[:field].graphql_name
+              attributes['graphql.field.parent'] = data[:owner]&.graphql_name # owner is the concrete type, not interface
+              attributes['graphql.field.name'] = data[:field]&.graphql_name
               attributes['graphql.lazy'] = key == 'execute_field_lazy'
             when 'authorized', 'authorized_lazy'
-              attributes['graphql.type.name'] = data.fetch(:type).graphql_name
+              attributes['graphql.type.name'] = data[:type]&.graphql_name
               attributes['graphql.lazy'] = key == 'authorized_lazy'
             when 'resolve_type', 'resolve_type_lazy'
-              attributes['graphql.type.name'] = data.fetch(:type).graphql_name
+              attributes['graphql.type.name'] = data[:type]&.graphql_name
               attributes['graphql.lazy'] = key == 'resolve_type_lazy'
             when 'execute_query'
               attributes['graphql.operation.name'] = data[:query].selected_operation_name if data[:query].selected_operation_name
--- a/instrumentation/graphql/lib/opentelemetry/instrumentation/graphql/tracers/graphql_tracer.rb
+++ b/instrumentation/graphql/lib/opentelemetry/instrumentation/graphql/tracers/graphql_tracer.rb
@@ -82,7 +82,7 @@ module OpenTelemetry
             GraphQL::Instrumentation.instance.config
           end
 
-          def attributes_for(key, data)
+          def attributes_for(key, data) # rubocop:disable Metrics/CyclomaticComplexity
             attributes = {}
             case key
             when 'execute_field', 'execute_field_lazy'
--- a/instrumentation/graphql/lib/opentelemetry/instrumentation/graphql/tracers/graphql_tracer.rb
+++ b/instrumentation/graphql/lib/opentelemetry/instrumentation/graphql/tracers/graphql_tracer.rb
@@ -45,19 +45,31 @@ module OpenTelemetry
           def platform_field_key(type, field)
             return unless config[:enable_platform_field]
 
-            "#{type.graphql_name}.#{field.graphql_name}"
+            if config[:legacy_platform_span_names]
+              "#{type.graphql_name}.#{field.graphql_name}"
+            else
+              'graphql.execute_field'
+            end
           end
 
           def platform_authorized_key(type)
             return unless config[:enable_platform_authorized]
 
-            "#{type.graphql_name}.authorized"
+            if config[:legacy_platform_span_names]
+              "#{type.graphql_name}.authorized"
+            else
+              'graphql.authorized'
+            end
           end
 
           def platform_resolve_type_key(type)
             return unless config[:enable_platform_resolve_type]
 
-            "#{type.graphql_name}.resolve_type"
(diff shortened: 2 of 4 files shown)


robertlaurin:
>graphql.field.path: Did not add at this time, but the [information is available via graphql-ruby](https://graphql-ruby.org/api-doc/2.0.16/GraphQL/Tracing) if we want to report it. There is a [opentelemetry-js](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/79b2d3ff08904ce84c6bc48427cd98906c2f0d79/plugins/node/opentelemetry-instrumentation-graphql/src/utils.ts#L127) equivalent. Thoughts?

Just to clarify I understand what his value would represent.

Using one of the test examples of `{ vehicle { model } }`, would the path for `model` field be `vehicle.model`?

instrumentation/graphql/lib/opentelemetry/instrumentation/graphql/tracers/graphql_tracer.rb
```diff
@@ -73,6 +85,16 @@ def config
           def attributes_for(key, data)
             attributes = {}
             case key
+            when 'execute_field', 'execute_field_lazy'
+              attributes['graphql.field.parent'] = data[:owner].graphql_name # owner is the concrete type, not interface
+              attributes['graphql.field.name'] = data[:field].graphql_name
+              attributes['graphql.lazy'] = key == 'execute_field_lazy'
+            when 'authorized', 'authorized_lazy'
+              attributes['graphql.type.name'] = data.fetch(:type).graphql_name
```
robertlaurin, line 93: Is it possible for the `data` hash here to not contain the key `:type`? Fetch will raise if it is absent, which I believe in this code path would introduce a runtime exception.

ravangen:
> > graphql.field.path: Did not add at this time, but the [information is available via graphql-ruby](https://graphql-ruby.org/api-doc/2.0.16/GraphQL/Tracing) if we want to report it. There is a [opentelemetry-js](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/79b2d3ff08904ce84c6bc48427cd98906c2f0d79/plugins/node/opentelemetry-instrumentation-graphql/src/utils.ts#L127) equivalent. Thoughts?
> 
> Just to clarify I understand what his value would represent.
> 
> Using one of the test examples of `{ vehicle { model } }`, would the path for `model` field be `vehicle.model`?

Yes. It would be a [path](https://spec.graphql.org/draft/#example-072c4) with array elements joined by `.`. A field with a list of elements would use the index of the element, `vehicles.1.model`

instrumentation/graphql/lib/opentelemetry/instrumentation/graphql/tracers/graphql_tracer.rb
ravangen: I believe it is fine as it would have raised in `PlatformTracing#trace` before now. [Ever since that `key` code path has been added for authorized and resolve_type](https://github.com/rmosolgo/graphql-ruby/pull/2660/files#diff-957133c12f5c1bb43a25cf51171630b8b7baff115db50158543d40f2b938e615R55-R72), it uses `fetch` so I feel safe with this.

ahayworth: I find myself wondering why we're using `fetch` here at all, in fact. I would change it to either:
- Just grab it from the hash and use safe nav: `attributes['graphql.type.name'] = data[:type]&.graphql_name`
- Only add it if we know it's present: `attributes['graphql.type.name'] = data[:type].graphql_name if data[:type]`

instrumentation/graphql/README.md
```diff
@@ -41,6 +41,11 @@ OpenTelemetry::SDK.configure do |c|
     enable_platform_authorized: false,
     # enable_platform_resolve_type maps to the resolve_type and resolve_type_lazy keys
     enable_platform_resolve_type: false
+
+    # Controls if platform tracing (field/authorized/resolve_type)
+    # should use the legacy span names (e.g. "MyType.myField") or the
+    # new normalized span names (e.g. "graphql.execute_field").
+    legacy_platform_span_names: false
```
ahayworth, line 48: I saw in the PR message that we would consider flipping this from `false` -> `true` if instrumentation ever hits 1.0 (we're not in a rush).

Why wait? It's not stable now; and if we think this is a useful enough change to flip at `1.0` why not now?

This is not a blocking comment, I'm just curious what you think here.

instrumentation/graphql/lib/opentelemetry/instrumentation/graphql/tracers/graphql_tracer.rb
ahayworth: Unless `data` isn't a hash, of course.

ahayworth (approved):
I have no major concerns here - I should note that I do not really know much about graphql in general so the utility of this is not something I can comment on. However, I think I see the value of normalizing these names for analysis.

instrumentation/graphql/README.md
ravangen: Wasn't sure on this gem's policy on breaking changes. We can change the default to still give folks the option of using the old behaviour for now, no strong opinions.

instrumentation/graphql/lib/opentelemetry/instrumentation/graphql/tracers/graphql_tracer.rb
ravangen: [Docs](https://graphql-ruby.org/api-doc/2.0.16/GraphQL/Tracing) covering `data` contents. I believe it is a hash.

instrumentation/graphql/README.md
ahayworth: I don't think we have a super strong policy for things that are pre-1.0 honestly.

instrumentation/graphql/lib/opentelemetry/instrumentation/graphql/tracers/graphql_tracer.rb
robertlaurin: If it is programmatically impossible for some of these values to not be present we probably don't need to introduce defensive programming, but we do need to follow the error handling guidelines https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/error-handling.md#basic-error-handling-principles

My personal take here is that I'm inclined to err to the side of caution.

robertlaurin (approved):

robertlaurin:
@ravangen feel free to just add rubocop disable comment for this.
```
lib/opentelemetry/instrumentation/graphql/tracers/graphql_tracer.rb:85:11: C: Metrics/CyclomaticComplexity: Cyclomatic complexity for attributes_for is too high. [10/7]
          def attributes_for(key, data) ...
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

arielvalentin:
👋🏼 @rmosolgo if you have any insights you would be interested in sharing.

Issue #292, opened 2023-01-26
Incorrect exception handling when routes in Sinatra raise exceptions


rdoherty-fastly:
### Description of the bug

When an exception is raised in a Sinatra route, the call to `@app.call(env)` in https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/main/instrumentation/sinatra/lib/opentelemetry/instrumentation/sinatra/middlewares/tracer_middleware.rb#L19 raises, which means `response` is nil. Therefore the calls to `response` in `trace_response` will throw NoMethodError exceptions, which is what ends up being put into the traces, which is incorrect.

This was not caught before because the test cases were only checking if the span is marked as having an exception event, but not the contents of it. The higher level Rack tracing code *seems* to be the one marking the span with an exception, but I am not 100% sure.

Commit with failing test case, there are other things that need to be verified but this will give some direction:
https://github.com/rdoherty-fastly/opentelemetry-ruby-contrib/commit/194f8ec6f15d759fa70c68a44d213056a4817dda

I'm not sure the exact next steps for this, probably catching any errors raised and a special code path for that? 🤷 

### Test output:
```
  1) Failure:
OpenTelemetry::Instrumentation::Sinatra::tracing#test_0009_adds exception type to events when the app raises errors [/Users/rdoherty/src/opentelemetry-ruby-contrib/instrumentation/sinatra/test/opentelemetry/instrumentation/sinatra_test.rb:170]:
--- expected
+++ actual
@@ -1 +1,3 @@
-"CustomError"
+# encoding: US-ASCII
+#    valid: true
+"NoMethodError"
```

### Expected behavior
It sets the exception information to the exception raised in the Sinatra route.

**Share details about your runtime**

Operating system details: OSX 13.1
RUBY_ENGINE: "ruby"
RUBY_VERSION: "2.7.4"
RUBY_DESCRIPTION: "ruby 2.7.4p191 (2021-07-07 revision a21a3b7d23) [arm64-darwin21]"

PR #317, opened 2023-02-03
Fix incorrect error type being recorded when Sinatra route raises exception

rdoherty-fastly:
Fixes https://github.com/open-telemetry/opentelemetry-ruby-contrib/issues/292

Wait, we're *deleting* code? YUP!

tl;dr: The Rack instrumentation already catches exceptions, adds the data to the current span and marks the error status on it. 

The primary problem is if an exception is raised in `@app.call(env)`, `response` would be nil, so the error checking code in `trace_response` will throw a new exception for trying to use `response`. That *new* exception is then caught by the Rack instrumentation (I believe, or higher level instrumentation that I found when debugging). 

So, if we were to make this work and catch exceptions, record them and then raise it again (as we should since this is middleware and we don't want to interfere with other exception handling), the higher level OTEL exception handler will also catch it and record it, giving us 2 exception events. That is what I originally did and saw 2 exception events, which is not correct.

This is a draft, please feel free to do a full review and make suggestions or ask questions. I'm new to this codebase so I could have done this completely wrong!

### DIFF
--- a/instrumentation/sinatra/lib/opentelemetry/instrumentation/sinatra/middlewares/tracer_middleware.rb
+++ b/instrumentation/sinatra/lib/opentelemetry/instrumentation/sinatra/middlewares/tracer_middleware.rb
@@ -16,23 +16,17 @@ module OpenTelemetry
           end
 
           def call(env)
-            response = @app.call(env)
+            @app.call(env)
           ensure
-            trace_response(env, response)
+            trace_response(env)
           end
 
-          def trace_response(env, response)
+          def trace_response(env)
             span = OpenTelemetry::Instrumentation::Rack.current_span
             return unless span.recording?
 
             span.set_attribute('http.route', env['sinatra.route'].split.last) if env['sinatra.route']
             span.name = env['sinatra.route'] if env['sinatra.route']
-
-            sinatra_response = ::Sinatra::Response.new([], response.first)
-            return unless sinatra_response.server_error?
-
-            span.record_exception(env['sinatra.error']) if env['sinatra.error']
-            span.status = OpenTelemetry::Trace::Status.error
           end
         end
       end
--- a/instrumentation/sinatra/test/opentelemetry/instrumentation/sinatra_test.rb
+++ b/instrumentation/sinatra/test/opentelemetry/instrumentation/sinatra_test.rb
@@ -168,6 +168,7 @@ describe OpenTelemetry::Instrumentation::Sinatra do
       get '/one/error'
 
       _(exporter.finished_spans.first.events[0].attributes['exception.type']).must_equal('CustomError')
+      _(exporter.finished_spans.first.events[0].attributes['exception.message']).must_equal('custom message')
     end
   end
 end
--- a/instrumentation/sinatra/test/opentelemetry/instrumentation/sinatra_test.rb
+++ b/instrumentation/sinatra/test/opentelemetry/instrumentation/sinatra_test.rb
@@ -12,6 +12,10 @@ describe OpenTelemetry::Instrumentation::Sinatra do
   let(:instrumentation) { OpenTelemetry::Instrumentation::Sinatra::Instrumentation.instance }
   let(:exporter) { EXPORTER }
 
+  class CustomError < StandardError
+
+  end
+
   let(:app_one) do
     Class.new(Sinatra::Application) do
       get '/endpoint' do
@@ -19,7 +23,7 @@ describe OpenTelemetry::Instrumentation::Sinatra do
       end
 
       get '/error' do
-        raise
+        raise CustomError, 'custom message'
       end
 
       template :foo_template do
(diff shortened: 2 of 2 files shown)


arielvalentin (approved):
@rdoherty-fastly Thanks again for your help with this instrumentation, and I agree with you that since this instrumentation requires Rack to be enabled then we should probably only do the error handling in one place.

rdoherty-fastly:
Thanks @arielvalentin! I'm super excited to get this working! It's one of the blockers for doing a proof of concept at my job.

scbjans:
This unfortunately does not add the sinatra exception to the span anymore. It would be great if we could do the error handling in the Rack instrumentation and record the exception there; but unfortunately I don't know how.

arielvalentin:
> This unfortunately does not add the sinatra exception to the span anymore. It would be great if we could do the error handling in the Rack instrumentation and record the exception there; but unfortunately I don't know how.

@scbjans are you saying that when sinatra handles and exception it does not bubble it up to Rack so this PR isn't working as expected?

rdoherty-fastly:
> This unfortunately does not add the sinatra exception to the span anymore. It would be great if we could do the error handling in the Rack instrumentation and record the exception there; but unfortunately I don't know how.

AFAIK env['sinatra.error'] is the stacktrace from the exception, and that is added to the event via the Rack instrumentation. We could check `env['sinatra.error']` in this sinatra instrumentation and add it to the span if we want, I can test that. It would probably duplicate what Rack is doing though.

Before this PR, sinatra.error wasn't being recorded at all due to the error described in the PR description.

scbjans:
@arielvalentin @rdoherty-fastly I did a quick test and it seems that since `span.record_exception(env['sinatra.error'])` is removed; the exception is not added to the span anymore (but the span is correctly marked as erroneous).

Before:
afbeelding

With this PR:
afbeelding

I'll look further into this next week.

arielvalentin:
Thanks again. I will keep on the lookout for an update.  

@scbjans thanks for double checking this! I will scrutinize the changes a little more next time when there are interdependent instrumentations like this.

rdoherty-fastly:
@scbjans That is weird because when I inspect the finished spans in the test case, I can see a stacktrace:

` exporter.finished_spans.first.events[0]
=> #<struct OpenTelemetry::SDK::Trace::Event name="exception", attributes={"exception.type"=>"CustomError", "exception.message"=>"custom message", "exception.stacktrace"=>"/Users/rdoherty/src/opentelemetry-ruby-contrib/instrumentation/sinatra/test/opentelemetry/instrumentation/sinatra_test.rb:26.....`

I probably don't understand the correct method of handling exceptions, I'll step through the test cases from `main` and this branch to see what's different.

arielvalentin:
@rdoherty-fastly what is the instrumentation scope for the finished span? Is this possibly an artifact of rack-test vs when running plan old rack on an app server?

scbjans:
@rdoherty-fastly @arielvalentin 
I'm so sorry, apparantly I was testing on an outdated clone.. :(

If I do test it on a freshly cloned repo; it all works as expected.
Sorry for the confusion!

arielvalentin:
@scbjans No worries! I do appreciate you doing exploratory testing since I am not a sinatra user. 

I am glad we got that sorted out.

rdoherty-fastly:
@scbjans no worries, thanks for testing it! \o/

arielvalentin (approved):
Approved

scbjans:
I didn't fully understand why my simple tests did only work on a freshly cloned repo.
Although my branch was fully up-to-date before testing; I had an outdated named docker volume `bundle`. Resulting in testing against an old set of gems 🤦 

It seems the original PR (https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/133) that added this wasn't even necessary in the first place, but since it was released simultaneously with the PR that based sinatra middleware on top of rack middleware (https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/54) I guess I just simply never noticed.

Anyway, always good to see code cleaned up where it can. 
I learned (and re-learned 😄  ) stuff along the way.
Thank you both.

rdoherty-fastly:
You're welcome @scbjans ! Thanks for reviewing!

rdoherty-fastly:
So what is the next step on this PR? I see it was approved but I don't have privileges to merge. Someone else will have to do that for me :)

arielvalentin:
@rdoherty-fastly we have a policy of having at least 2 approvals from @open-telemetry/ruby-contrib-approvers and I am the only one who has had a chance to review this PR so far. Once we get another approval I will merge this PR and release the gem. 

I have added this PR to the agenda[^1] for the Ruby SIG so hopefully that will be enough of a signal boost. Please join us if you are available!   

[^1]: https://docs.google.com/document/d/1D15bO8o340sQm2CVZiukEJuCO_XMMHKPuTznoEhyFqE/edit#

rdoherty-fastly:
Oops ok thanks for the context @arielvalentin!

mwear (approved):

robertlaurin (approved):

PR #325, opened 2023-02-09
ci: disable a vexatious rubocop rule, and correct a sorting

ahayworth:
Rubocop has been failing our canary builds every night for a longer
period of time than I'm willing to admit to in a public forum.

One of these failures was because we weren't strictly ensuring that
development dependencies were listed in the Gemfile as opposed to the
gemspec. I really think that's a little over-pedantic, so I've turned
it off with the support of my esteemed colleage, @arielvalentin.

The other failure we fix is related to the sorting of dependencies
within the gemspec. Mind you, I think this failure is also
overly-pedantic and I would be inclined to disable it. However, rubocop
was able to autofix it with no effort on my part, and so I've kept the fix.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -9,8 +9,6 @@ AllCops:
     - "**/**/vendor/bundle/**/*"
 Bundler/OrderedGems:
   Enabled: false
-Gemspec/DevelopmentDependencies:
-  Enabled: false
 Gemspec/RequiredRubyVersion:
   Enabled: false
 Lint/ConstantDefinitionInBlock:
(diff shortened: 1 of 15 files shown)


arielvalentin (approved):

ahayworth:
Oh wow, this made it so much worse. Looks like this is only in rubocop 1.44 and up: https://docs.rubocop.org/rubocop/cops_gemspec.html#gemspecdevelopmentdependencies

I'll fix it up later today or something.

arielvalentin (approved):

Issue #327, opened 2023-02-10
Sinatra exceptions are missing for erroneous spans


scbjans:
https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/317 defers error handling from sinatra to the rack middleware. This only applies to marking a span as erroneous (see: https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/main/instrumentation/rack/lib/opentelemetry/instrumentation/rack/middlewares/tracer_middleware.rb#L155), not for adding the exception to it (see: https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/317#issuecomment-1416297686)

Although I did review that PR, my conclusion was eventually that I messed up my own configuration; but apparently I messed that up as well :)

scbjans:
I'm currently a bit low on bandwidth; so for our own installation I've pinned `opentelemetry-instrumentation-sinatra` to `0.21.3` which still works for us.
I'm happy to look into this myself in a couple of weeks.

arielvalentin:
@rdoherty-fastly mind taking a look at this?

PR #328, opened 2023-02-10
fix: add exceptions to sinatra spans, ruboproof test.

scbjans:
This will essentially revert https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/317 , but adds a guard clause since unhandled exceptions in Sinatra will return `response = nil`.

I've tried to check if I could add the exception event in the Rack middleware; but the Sinatra exception is not available there.

@rdoherty-fastly since your tests pass, I believe it's OK. Nevertheless I would love your :eyes: on this as well.

Fixes: https://github.com/open-telemetry/opentelemetry-ruby-contrib/issues/327

### DIFF
--- a/instrumentation/sinatra/lib/opentelemetry/instrumentation/sinatra/middlewares/tracer_middleware.rb
+++ b/instrumentation/sinatra/lib/opentelemetry/instrumentation/sinatra/middlewares/tracer_middleware.rb
@@ -16,17 +16,25 @@ module OpenTelemetry
           end
 
           def call(env)
-            @app.call(env)
+            response = @app.call(env)
           ensure
-            trace_response(env)
+            trace_response(env, response)
           end
 
-          def trace_response(env)
+          def trace_response(env, response)
             span = OpenTelemetry::Instrumentation::Rack.current_span
             return unless span.recording?
 
             span.set_attribute('http.route', env['sinatra.route'].split.last) if env['sinatra.route']
             span.name = env['sinatra.route'] if env['sinatra.route']
+
+            return if response.nil?
+
+            sinatra_response = ::Sinatra::Response.new([], response.first)
+            return unless sinatra_response.server_error?
+
+            span.record_exception(env['sinatra.error']) if env['sinatra.error']
+            span.status = OpenTelemetry::Trace::Status.error
           end
         end
       end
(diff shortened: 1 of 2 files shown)


instrumentation/sinatra/lib/opentelemetry/instrumentation/sinatra/middlewares/tracer_middleware.rb
```diff
@@ -16,17 +16,25 @@ def initialize(app)
...
             span.set_attribute('http.route', env['sinatra.route'].split.last) if env['sinatra.route']
             span.name = env['sinatra.route'] if env['sinatra.route']
+
+            return if response.nil?
+
+            sinatra_response = ::Sinatra::Response.new([], response.first)
+            return unless sinatra_response.server_error?
+
+            span.record_exception(env['sinatra.error']) if env['sinatra.error']
+            span.status = OpenTelemetry::Trace::Status.error
```
scbjans, line 37: I'm not sure if we should tag the span as erroneous here since that's also done in the Rack middleware. However I've left it as-is(/was) to keep exception+status setting together

scbjans:
BTW, I've tried to check if I could add the exception event in the Rack middleware; but the Sinatra exception is not available there.

instrumentation/sinatra/lib/opentelemetry/instrumentation/sinatra/middlewares/tracer_middleware.rb
arielvalentin: We should _only_ tag it as an error if the response is in 5xx range

scbjans: It will only set it for a `>500`. That is done by checking `sinatra_response.server_error?`

arielvalentin: > It will only set it for a >500. That is done by checking sinatra_response.server_error?

Fair. 

> should we keep it here? (to keep exception and status setting together) or should we only set exception here, and set status in rack middleware?

I do not object to setting it here since it is more explicit.

I am not a sinatra user so I admit that I have not done the appropriate due diligence reviews here. 

May I ask you to share some screen shots of Jaeger with the traces before and after so we know what users should see?

scbjans: Nice, I believe setting error status and adding the exception go hand-in-hand and I like it to be explicit.

Current behavior in 0.21.4 (sets status correctly (in Rack middleware), but does not add exception):
216688451-975a6ff5-a5b8-44a8-b770-1f34409e67ff

Behavior after this PR (sets status correctly, and adds exception):
Screenshot 2023-02-11 at 19-55-31 Jaeger UI

Essentially, this is the same behavior as in 0.21.3, but additionally it should support the use-case of @rdoherty-fastly as well

arielvalentin (approved):

rdoherty-fastly:
Thanks for digging into this @scbjans!

Issue #179, opened 2022-11-07
Feature Request - New Resource Detector - dockerCGroupV1Detector #2353


tomerfriedman:
- Same as the @opentelemetry/resource-detector-docker for NodeJS
- Add Container Tag attributes

tomerfriedman:
https://github.com/DataDog/dd-trace-rb/blob/f1a8956baa5c1f552c89f8a6c72cc04e075e5d5a/lib/datadog/core/environment/container.rb

ahayworth:
I think we'd be happy to accept a PR that added this resource detector. At the moment, we are a little constrained by resources and unlikely to add this ourselves right away - but if anyone in the community is interested, we'd be happy to review such a PR and work with you to get it merged. 😄

PR #340, opened 2023-02-16
feat: add container resource detector.

scbjans:
This PR adds a resource detector for Containers.

It will add the container ID if possible.

~~For containers running on a Docker runtime, it will also try to add~~
~~* container.image.name~~
~~* container.image.tag~~
~~* container.name~~
~~* container.runtime~~

~~but that depends on the availability of a readable Docker Socket.~~

I've decided to remove setting additional container resource attributes and only support setting the `container.id`.

Reasoning behind that:
* Setting other container resource attributes requires a readable `docker.sock`/`containerd.sock`, which requires correct - probably most of the time equals to elevated - permissions. Most easy way to do that is of course running the container as `root`. I do not want to promote that.
* I could try and add support for configuring access to the `docker.sock` using TLS certs; but that would make auto-instrumentation a lot less auto
* Other languages also only seem to be setting `container.id`

Screenshot 2023-02-16 at 15-42-04 Jaeger UI

Fixes: #179


resource_detectors/lib/opentelemetry/resource/detectors/container.rb
```diff
@@ -0,0 +1,51 @@
...
+        end
+
+        private
+
+        def container_id
+          [CGROUP_V2_PATH, CGROUP_V1_PATH].each do |cgroup|
+            next unless File.readable?(cgroup)
+
+            File.readlines(cgroup, chomp: true).each do |line|
+              next unless line.include?(Socket.gethostname)
```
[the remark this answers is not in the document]
scbjans: Thanks @mwear you're absolutely right. I've planned on using regexes to determine the container id, but somewhere I've lost track of that and didn't implement a testcase for cgroupv1 so I didn't notice 🤦
I will rework this PR a bit to fix both these problems

mwear (commented):
This mostly LGTM. I just had a question about cgroup v2 this time. I'm not requesting changes, but I'd like to clarify that the strategy we're using is at least as reliable as others I've seen.

resource_detectors/lib/opentelemetry/resource/detectors/container.rb
```diff
@@ -0,0 +1,70 @@
...
+          nil
+        end
+
+        def container_id_v2
+          File.readlines(CGROUP_V2_PATH, chomp: true).each do |line|
+            parts = line.split('/')
+            parts.shift
+
+            parts.each do |part|
+              return part if part.match?(CONTAINER_REGEX)
```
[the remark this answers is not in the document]
scbjans: Thanks for your re-review @mwear.

My guess would be that filtering on `"hostname"` is done to limit the amount of regexmatching?
I initially included the filter on `"hostname"` as well; but decided to remove that; since it cannot be used reliably. e.g. when a user set the `hostname` themselves in [kubernetes](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.19/#pod-v1-core) or [swarm](https://docs.docker.com/compose/compose-file/compose-file-v3/#domainname-hostname-ipc-mac_address-privileged-read_only-shm_size-stdin_open-tty-user-working_dir); we would not be able to determine the `container.id` (although I think setting the `"hostname"` will probably be an edge case).

If we decide that we don't want to filter on `"hostname"` before regex matching; we might also remove `container_id_v1`; since `def container_id_v2` will also work for `cgroupv1` - although I still want to benchmark that before doing so - to keep the code a bit simpler. WDYT?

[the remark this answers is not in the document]
mwear: I don't have a strong preference. I am most interested in the detector reliably and consistently detecting the container id. As far as performance goes, detecting the container id will be something that happens once at startup. While we don't want to give performance away unnecessarily, this is not a hot code path, so favoring simplicity is reasonable thing to do here. Let me know when you're happy with everything and I'll review it again.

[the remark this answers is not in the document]
scbjans: I've simplified the code a bit; according to me this should do the trick.

mwear (approved):
Looks great! Thanks @scbjans.

mwear:
@robertlaurin could we get an additional set of eyes on this when you get a chance?

scbjans:
@mwear @robertlaurin from the SIG I understood that the plan was to split the resource detectors gem into separate gems. That totally makes sense to me; and I'll rework this PR a bit to reflect that. Once that's done and approved, I'm happy to also split of the other two into separate gems.

One point on which I would like advice or some directions is the addition of other container resource attributes (see PR description, and - currently outdated - image). It'll be helpful to add the other container resource attributes as well; but I do not want to promote running the container as root. Do you perhaps have another idea on how I could add them?

resource_detectors/lib/opentelemetry/resource/detectors/container.rb
```diff
@@ -0,0 +1,59 @@
...
+
+        private
+
+        # Returns the container.id if it can be determined from cgroup
+        # or nil if container.id could not be determined
+        #
+        # @return [String] container.id
+        #   May be nil.
+        def container_id
+          [CGROUP_V2_PATH, CGROUP_V1_PATH].each do |cgroup|
```
arielvalentin, line 41: Would emitting some kind of warning or error message be helpful to folks trying to debug why this resource detector is unable to read the `cgroup` files?

resource_detectors/lib/opentelemetry/resource/detectors/container.rb
```diff
@@ -0,0 +1,59 @@
...
+        CONTAINER_PATTERN = '[0-9a-f]{64}'
+        CONTAINER_REGEX = /(?<container>#{UUID_PATTERN}|#{CONTAINER_PATTERN})(?:.scope)?$/
+        CGROUP_V1_PATH = '/proc/self/cgroup'
+        CGROUP_V2_PATH = '/proc/self/mountinfo'
+
+        # Detects container attributes and creates a {OpenTelemetry::SDK::Resources::Resource} with the specified attributes.
+        # If no container attributes could be determined an empty resource is returned
+        #
+        # @return [OpenTelemetry::SDK::Resources::Resource]
+        def detect
```
arielvalentin, line 24: I am not entirely certain what the spec says about error handling in Resource detectors but I recommend you consider adding error handling with diagnostic information that will clue folks into why this resource detector may be failing and avoid raising errors that would cause the application to crash at boot time.

resource_detectors/lib/opentelemetry/resource/detectors/container.rb
```diff
@@ -0,0 +1,59 @@
...
+        # or nil if container.id could not be determined
+        #
+        # @return [String] container.id
+        #   May be nil.
+        def container_id
+          [CGROUP_V2_PATH, CGROUP_V1_PATH].each do |cgroup|
+            next unless File.readable?(cgroup)
+
+            File.readlines(cgroup, chomp: true).each do |line|
+              parts = line.split('/')
```
arielvalentin, line 45: Is there some nuance I am missing here. Does scanning the full string result in false positives?

scbjans: scanning the full string works for cgroupv1; but I couldn't get it to work for cgroupv2. This way it's compatible with both

resource_detectors/lib/opentelemetry/resource/detectors/container.rb
scbjans: Originally my idea was that we do not want to flood the logs if someone uses this gems `Autodetector.detect`, but since we're going to split the resource detectors up into separate gems, that reasoning is no longer valid.

I'll add some logs once I get to splitting this up into it's own gem.

resource_detectors/lib/opentelemetry/resource/detectors/container.rb
scbjans: Will do so once I split this of into it's own separate gem (see comment above). However, I do believe that currently this resource detector will not crash if it cannot determine the `container.id`?

scbjans:
The `container.id` cannot be determined reliably when the container has additional copy-on-write layers written to it's filesystem. I'll have to sort that out first -> returning to draft mode.

scbjans:
Regarding adding the other container resource attributes: I've learned that I can use the [otelcols k8sattributesprocessor](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/processor/k8sattributesprocessor) to fix that.

ericmustin:
@arielvalentin lgtm

arielvalentin:
@ericmustin would you mind merging and releasing?

scbjans:
@mwear @ericmustin @arielvalentin anything I can do to get this merged?

scbjans:
CI is broken; but there's already a PR which fixes that: https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/581

Issue #341, opened 2023-02-17
Rack Span not available in `Rack::BodyProxy` 


arielvalentin:
`Rack::BodyProxy` defers executing a user specified block until after the request is complete, however the Rack instrumentation middleware closes the span before the block is evaluated and then detaches the current context.

When users try to access the current span in the block, Rack returns an invalid span. In the snippet below I was able to reproduce the issue using RSpec as a test harness:

```ruby
require "spec_helper"

RSpec.describe "Rack::BodyProxy Integration" do
  include Rack::Test::Methods

  let(:app) do
    Rack::Builder.new {
      use OpenTelemetry::Instrumentation::Rack::Middlewares::TracerMiddleware
      run lambda { |_arg|
        puts "app.call #{OpenTelemetry::Instrumentation::Rack.current_span.context.hex_trace_id}"
        body = Rack::BodyProxy.new([]) { puts "BodyProxy.close #{OpenTelemetry::Instrumentation::Rack.current_span.context.hex_trace_id}" }
        [
          200,
          { "Content-Type" => "application/json" },
          body,
        ]
      }
    }
  end

  it "does something" do
    get "/"
    expect(last_response).to be_ok
  end
end

```

In the mix of the spec output you can see that the trace id is that of an `INVALID_SPAN`: 

```console
Rack::BodyProxy Integration
app.call 65a385eada410a721a79431171960814
BodyProxy.close 00000000000000000000000000000000
  does something

Finished in 0.08231 seconds (files took 2.31 seconds to load)
1 example, 0 failures
```

I am unsure what the correct behavior should be here but this is related since it is basically my problem where we are unable to correlate Rack Requests to Traces when using `Rack::BodyProxy` or any of its descendants. https://github.com/DataDog/dd-trace-rb/pull/1746

arielvalentin:
Is using `Rack::Events` the best way to handle this going forward? https://www.rubydoc.info/gems/rack/Rack/Events

> ## Order
> 
> `on_start` is called on the handlers in the order that they were passed to the constructor. `on_commit`, on_send`, `on_finish`, and `on_error` are called in the reverse order. `on_finish` handlers are called inside an `ensure` block, so they are guaranteed to be called even if something raises an exception. If something raises an exception in a `on_finish` method, then nothing is guaranteed.

ahayworth:
[...]

PR #342, opened 2023-02-18
feat: Use Rack::Events for instrumentation

arielvalentin:
This is an alternate implementation of the Rack instrumentation that leverages the [`Rack::Events` API](https://www.rubydoc.info/gems/rack/Rack/Events) instead of a custom`Rack::Middleware`.

Why am I suggesting we change this instrumentation? At GitHub we leverage `Rack::BodyProxy` to write Rack Request logs after the request is complete, however Rack span is already finished and its related `Context` has already been detached. This means we are not able to correlate request logs to our traces.

The advantage of using `Rack::Events` is that handlers are triggered during difference stages of a request, including for deferred operations like [`Rack::BodyProxy`](https://www.rubydoc.info/gems/rack/Rack/BodyProxy), as opposed to middlewares who _only_ are invoked inline.

The disadvantage of this API is that it makes managing the request more difficult and we have to track the Context positions to detach in the `Rack::Env`.

This implementation will be released along side the existing instrumtation to give users the option to use the middleware instead of the `Rack::Events` handler until we are able to run this in some of our heavy production workloads.

Other noteworthy things:

1. The `http.proxy` span has been converted to a Span Event and ignores the `record_frontend_span` option
2. There are no _new_ events added that we brainstormed in this discussion: https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/342#discussion_r1113462298
3. All configuration options for this middleware that required special processing/registration were moved into the `Rack::Instrumentation.install` 
4. The Rack event now **includes** the time it takes to write the response to the socket.
5. Sinatra 1.4 depends on Rack 1 so we must support both Rack Middleware and Events unless we choose to drop support for it.

Fixes https://github.com/open-telemetry/opentelemetry-ruby-contrib/issues/341 Related https://github.com/DataDog/dd-trace-rb/pull/1746


instrumentation/rack/lib/opentelemetry/instrumentation/rack/middlewares/event_handler.rb
```diff
@@ -0,0 +1,255 @@
...
+            span.record_exception(error)
+            span.status = OpenTelemetry::Trace::Status.error
+          end
+
+          # Finishes the span making it eligible to be exported and cleans up existing contexts
+          #
+          # @note does nothing if the span is a non-recording span
+          # @param [Rack::Request] The current HTTP request
+          # @param [Rack::Response] The current HTTP response
+          def on_finish(request, response)
```
fbogsany, line 152: It might also be nice to add `on_send` and record a span event for "sent first byte" or something similar. It is really nice to know, from a perf analysis perspective, when the application has finalized the response and the remaining work is "out of our hands".

An interesting philosophical question: when should we consider the server span "finished"? When application processing is complete or when the response has been completely written to the socket? I _think_ the current middleware implementation is effectively equivalent to "when application processing is complete", before anything has been written to the socket.

arielvalentin: > It might also be nice to add `on_send` and record a span event for "sent first byte" or something similar. It is really nice to know, from a perf analysis perspective, when the application has finalized the response and the remaining work is "out of our hands".
> 

I like this idea a lot! I will add it to my todo list.

> 
> An interesting philosophical question: when should we consider the server span "finished"? When application processing is complete or when the response has been completely written to the socket? I _think_ the current middleware implementation is effectively equivalent to "when application processing is complete", before anything has been written to the socket.

- 💭 I'm not sure. I lean towards after the middleware has completed writing the response but does that make it asymmetric from when it's processing the incoming request? I'm not entirely certain.

instrumentation/rack/lib/opentelemetry/instrumentation/rack/middlewares/event_handler.rb
```diff
@@ -0,0 +1,277 @@
...
+          # @param [Rack::Response] This is nil in practice
+          # @return [void]
+          def on_start(request, _)
+            return if untraced_request?(request.env)
+
+            extracted_context = extract_remote_context(request)
+            if record_frontend_span?
+              request_start_time = OpenTelemetry::Instrumentation::Rack::Util::QueueTime.get_request_start(request.env)
+              unless request_start_time.nil?
+                frontend_span = tracer.start_span(
```
arielvalentin, line 61: @fbogsany @mwear Am I correct in that the frontend span is here for parity with `dd-trace-rb`? Do we have any insight into how popular this feature is? Would it be ok to consider dropping it?

https://github.com/DataDog/dd-trace-rb/blob/2b163dd91c8d349ceb92370c16f4a7be4a691c35/docs/GettingStarted.md#http-request-queuing

I see you all discussed how this should work early: https://github.com/open-telemetry/opentelemetry-specification/issues/51

AFAICT representing proxy spans remains unresolved: https://github.com/open-telemetry/opentelemetry-specification/issues/51

But it looks like we took this approach:
https://github.com/open-telemetry/opentelemetry-specification/pull/337#issuecomment-549567281

One thing I noticed about porting this functionality to use the events API is that the `http.proxy` span does not include semantic attributes although it is the `server` span. Is that intentional? 

https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/main/instrumentation/rack/lib/opentelemetry/instrumentation/rack/middlewares/tracer_middleware.rb#L109

instrumentation/rack/lib/opentelemetry/instrumentation/rack/middlewares/event_handler.rb
```diff
@@ -0,0 +1,277 @@
...
+
+          def url_quantization
+            config[:url_quantization]
+          end
+
+          def response_propagators
+            config[:response_propagators]
+          end
+
+          def allowed_request_headers
```
arielvalentin, line 245: @open-telemetry/ruby-contrib-approvers When I started this implementation I used constructor injection, but then soon realized there are some limitations with that approach that supports _injecting_ this middleware _after_ the Rack app is initialized like in this example: 

https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/main/instrumentation/rack/example/trace_demonstration2.rb#L27

- Manual instantiation required:  The existing implementation requires users to explicitly add Rack tracer middleware to the builder via `use`. Following the same pattern for the `EventHandler` object instance _may_ be initialized _prior_ to the SDK initialization.  
- Stale config variables: The SDK is lazily initialized so the configuration options may not yet be set on the instrumentation instance. That means the `EventHandler` has to defer looking up the configuration options until _after_ the SDK is initialized, however it does not know when that happens because there aren't any lifecycle hooks that would notify this object instance that the config options are available. Using instance variables made it so the EventHandler had a set of default options and did not work correctly. 
- Tracer not upgraded: The object instance had a stale instance of a tracer, which did not make any sense to me as I expected the tracer to be a upgraded once the SDK was initialized so none of the test passed.

The middleware [instrumentation](https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/main/instrumentation/rack/lib/opentelemetry/instrumentation/rack/middlewares/tracer_middleware.rb) _memoizes_ these options as class variables. That caught me by surprise since I thought it made more sense for these options to be configured/memoized as part of the instrumentation `instance`. 

Is there a case where Rack requests will start to be processed before the SDK will be initialized?

Does the existing rack middleware account for the instrumentation being disabled via environment variables or the config attributes?

instrumentation/rack/lib/opentelemetry/instrumentation/rack/middlewares/event_handler.rb
arielvalentin: We discussed this at the SIG today. 

@robbkidd and I are going to swarm on resolving proxy spans with the HTTP WG but we will be dropping this feature.

Endusers may add `X-Request-Start` or `X-Queue-Time` via `allowed_request_headers` and process the span attributes in their vendor backends.

instrumentation/rack/lib/opentelemetry/instrumentation/rack/middlewares/event_handler.rb
```diff
@@ -0,0 +1,277 @@
...
+          #
+          # @param [Rack::Request] The current HTTP request
+          # @param [Rack::Response] This current HTTP response
+          # @return [void]
+          def on_commit(request, response)
+            span = OpenTelemetry::Instrumentation::Rack.current_span
+            return unless span.recording?
+
+            begin
+              response_propagators&.each { |propagator| propagator.inject(response.headers) }
```
arielvalentin, line 100: hmm... I should not stop here on the _first_ failure but attempt to process all of the propagators until it has gone through the entire list.

instrumentation/rack/lib/opentelemetry/instrumentation/rack/middlewares/event_handler.rb
```diff
@@ -0,0 +1,277 @@
...
+            span = OpenTelemetry::Instrumentation::Rack.current_span
+            return unless span.recording?
+
+            if response
+              span.status = OpenTelemetry::Trace::Status.error unless GOOD_HTTP_STATUSES.include?(response.status.to_i)
+              attributes = extract_response_attributes(response)
+              span.add_attributes(attributes)
+            end
+
+            request.env[TOKENS_KEY]&.reverse&.each do |token|
```
arielvalentin, line 135: hmm... should I add error handling here and attempt to detach each token anyway?

instrumentation/rack/lib/opentelemetry/instrumentation/rack/middlewares/event_handler.rb
fbogsany: I don't recall why we added this, but parity with DD seems likely. I would actually prefer to add an Event with the `request_start_time` - that's more meaningful than a "fake span" and is similar to what we do with Sidekiq job instrumentation with `created_at` and `enqueued_at` https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/57cabe3fac44a0921af542b62dc34ef285104a5c/instrumentation/sidekiq/lib/opentelemetry/instrumentation/sidekiq/middlewares/server/tracer_middleware.rb#L35-L36

In general, I am not fond of encoding timestamps in attributes (as integers or strings) and then parsing them on the backend. This sort of thing is exactly what Events are intended for.

arielvalentin: We did discuss recording it as an event but we didn't know what it would mean for a span event to have a timestamp that was _earlier_ than start timestamp of the event, so I thought it was best not to add it at all.

If users choose to add it to the allow list still that gives them access to the header without us having to add special logic.

fbogsany: There is ample precedent for an event timestamp earlier than the span start time - in particular, see the Sidekiq examples I shared. I strongly believe this should be an event rather than an attribute or a span. Requiring backends to parse and interpret stringly-typed attributes is bad form IMO, especially when we have a well structured and typed alternative.

ericmustin: Fine to drop and fine to restructure as an event. It was a thing that came from dd and wasn’t even standard across dd sdks, it predated the open tracing Nginx instrumentation iirc and was added as a one off work around for a specific user, although over time it did grow to become a somewhat popular feature with ddtracerb users. I believe the correct approach is to “simply” instrument your proxy.

arielvalentin: I have moved it into an event.

instrumentation/rack/lib/opentelemetry/instrumentation/rack/middlewares/event_handler.rb
arielvalentin: I moved all of the configuration setup into the `install` block.

instrumentation/rack/lib/opentelemetry/instrumentation/rack/middlewares/event_handler.rb
arielvalentin: Going to hold off adding other events until I get this initial implementation merged. 

I am also going to stick with `on_finish`, i.e. after the response has been written to the socket and the `Rack::BufferedResponse` is closed.

arielvalentin:
moving back to draft until I am able to determine why tests are failing

arielvalentin:
Tests are failing for combination of rack >= 2.2 and rack test 2.1

arielvalentin:
I have opened an issue with `rack-test` https://github.com/rack/rack-test/issues/335

Hopefully I will be able to dig in something soon and see what is going on.

arielvalentin:
`jeremyevans` set me straight. The test case setup was wrong.

instrumentation/rack/lib/opentelemetry/instrumentation/rack/instrumentation.rb
```diff
@@ -29,12 +28,56 @@ class Instrumentation < OpenTelemetry::Instrumentation::Base
         option :url_quantization,         default: nil,   validate: :callable
         option :untraced_requests,        default: nil,   validate: :callable
         option :response_propagators,     default: [],    validate: :array
+        # This option is only valid for applicaitons using Rack 2.0 or greater
+        option :use_rack_events,          default: false, validate: :boolean
+
+        # Temporary Helper for Sinatra and ActionPack middleware to use during installation
```
fbogsany, line 34: What's the long term plan (you may have answered this elsewhere)?

instrumentation/rack/lib/opentelemetry/instrumentation/rack/middlewares/tracer_middleware.rb
```diff
@@ -6,8 +6,6 @@
 
 require 'opentelemetry/trace/status'
 
-require_relative '../util/queue_time'
-
 module OpenTelemetry
   module Instrumentation
     module Rack
```
fbogsany, line 11: Any reason not to update this to use the `config[:allowed_rack_request_headers]` and `config[:allowed_rack_response_headers]` precomputed in the `Instrumentation` class?

instrumentation/rack/lib/opentelemetry/instrumentation/rack/middlewares/event_handler.rb
```diff
@@ -0,0 +1,271 @@
...
+            config[:response_propagators]
+          end
+
+          def allowed_request_headers
+            config[:allowed_rack_request_headers]
+          end
+
+          def allowed_response_headers
+            config[:allowed_rack_response_headers]
+          end
```
fbogsany, line 237: Is there any reason not to cache all these config options in instance variables?

instrumentation/rack/lib/opentelemetry/instrumentation/rack/instrumentation.rb
arielvalentin: This configuration setting and helper will exist until...

1. We drop support for Sinatra versions that require Rack 1.x
2. A future release that makes the Event Handler the default way for users to instrument their applications

instrumentation/rack/lib/opentelemetry/instrumentation/rack/middlewares/event_handler.rb
arielvalentin: The existing instrumentation supports being able to inject the middleware and then initializing the SDK afterwards. 

Using instance variables made it so the instance in the middleware had stale values and did not reflect the values after the instrumentation was installed so I didn't have a choice but to look them up.

robbkidd: My understanding is that caching config option values in instance variables is not a blocker to adding the event-based implementation. 👍🏻 

I think caching (I'm assuming "via memoization" here) is something we could consider doing in a future release if/when the middleware-based implementation is _removed_.

instrumentation/rack/lib/opentelemetry/instrumentation/rack/instrumentation.rb
```diff
@@ -29,12 +28,56 @@ class Instrumentation < OpenTelemetry::Instrumentation::Base
...
+        end
 
         private
 
         def require_dependencies
+          require_relative 'middlewares/event_handler' if defined?(Rack::Events)
           require_relative 'middlewares/tracer_middleware'
         end
+
+        def config_options(user_config)
```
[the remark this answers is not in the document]
robbkidd: @fbogsany @robertlaurin Is this a blocker for merging and releasing the event-based implementation?

robbkidd (approved):
It hangs together, new behavior is behind a feature flag config option, old behavior works in the handful of Racky test apps I have.

```ruby
# in a Gemfile
gem 'opentelemetry-instrumentation-rack', git: 'https://github.com/open-telemetry/opentelemetry-ruby-contrib', glob: 'instrumentation/*/*.gemspec'
```

I'd like to see the instrumentation's options appear documented in the gem's README, but (1) of the _existing_ options, only `url_quantization` is mentioned so this is not really a hard criteria for this new option and (2) I can be the change I want to see in the world (I'll open a PR updating the README).

arielvalentin:
OOOOOOOOPPPPPPPPPPPPSSSSSS I had set this to auto-merge by accident. I hope that is not a problem.

PR #343, opened 2023-02-21
ActiveJob instrumentation:  Adds ActiveJob::Core `provider_job_id` attribute

rscnt:
Some instrumentations like the one for Sidekiq or DelayedJob, use their own id as the value for `messaging.message_id` while the spans produced by active job instrumentation use the `job.job_id` UUID.

Examples:

- Que: https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/6e89c92f189bc6e187da06ea2af4e38531b93601/instrumentation/que/lib/opentelemetry/instrumentation/que/patches/que_job.rb#L78
- DelayedJob: https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/8f40eda20e0257bc740508b4b50758285d93e895/instrumentation/delayed_job/lib/opentelemetry/instrumentation/delayed_job/plugins/tracer_plugin.rb#L37
- Sidekiq: https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/a11d8b135d9ac4c28521619dc3b4744692ae2e6e/instrumentation/sidekiq/lib/opentelemetry/instrumentation/sidekiq/middlewares/client/tracer_middleware.rb#L19

We could use the `provider_job_id` since it's [guaranteed to be there](https://github.com/rails/rails/blob/812bae8167488ee5984043ef71125e055352b873/activejob/lib/active_job/core.rb#L26), at least with a nil value, and most adapter seem to define a correct value:

- Sidekiq adapter: https://github.com/rails/rails/blob/927b3b5a8449638a81ef934b0cd56def8bed4a74/activejob/lib/active_job/queue_adapters/sidekiq_adapter.rb#L29
- DelayedJob adapter: https://github.com/rails/rails/blob/c50721b88b624c4014f97fd21145b6c641c1e540/activejob/lib/active_job/queue_adapters/delayed_job_adapter.rb#L21

Having this attribute in each active job span could make it possible to eventually link both traces.

### DIFF
--- a/instrumentation/active_job/lib/opentelemetry/instrumentation/active_job/patches/active_job_callbacks.rb
+++ b/instrumentation/active_job/lib/opentelemetry/instrumentation/active_job/patches/active_job_callbacks.rb
@@ -75,7 +75,7 @@ module OpenTelemetry
               'messaging.active_job.scheduled_at' => job.scheduled_at,
               'messaging.active_job.priority' => job.priority
             }
-            
+
             otel_attributes['net.transport'] = 'inproc' if %w[async inline].include?(job.class.queue_adapter_name)
 
             otel_attributes.compact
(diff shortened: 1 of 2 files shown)


arielvalentin:
@rscnt Welcome and thank you for your contribution! 

Please provide us with some test coverage for this change and possibly screenshots of these changes in action.

ahayworth (approved):
Agreed on the test changes (and maybe a screenshot), but otherwise this LGTM! I verified that these changes were available in Rails 6.0, so we should be able to safely merge this. 👍

PR #345, opened 2023-02-24
fix: Ensure encoding errors handled during SQL obfuscation for Trilogy

adrianna-chang-shopify:
Shopify is planning to start migrating apps from Mysql2 to Trilogy as a SQL database client. Ensuring parity between OpenTelemetry instrumentation for Mysql2 and Trilogy is important before we move to Trilogy.

This PR ensures that invalid byte sequences are handled appropriately when obfuscating SQL, taking inspiration from https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/160. I also added some additional tests to the Trilogy spec around configuring `db_statement` via an environment variable.

I noticed the Mysql2 test around invalid byte encoding was removed in https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/265 when the `enable_sql_obfuscation` config went away, and I'm not sure if that was intentional. I've added a test to the Mysql2 spec to ensure there's coverage for an invalid byte sequence appearing in SQL.

### DIFF
--- a/instrumentation/mysql2/lib/opentelemetry/instrumentation/mysql2/patches/client.rb
+++ b/instrumentation/mysql2/lib/opentelemetry/instrumentation/mysql2/patches/client.rb
@@ -78,7 +78,8 @@ module OpenTelemetry
               obfuscated = 'Failed to obfuscate SQL query - quote characters remained after obfuscation' if detect_unmatched_pairs(obfuscated)
               obfuscated
             end
-          rescue StandardError
+          rescue StandardError => e
+            OpenTelemetry.handle_error(message: 'Failed to obfuscate SQL', exception: e)
             'OpenTelemetry error: failed to obfuscate sql'
           end
 
--- a/instrumentation/mysql2/test/opentelemetry/instrumentation/mysql2/instrumentation_test.rb
+++ b/instrumentation/mysql2/test/opentelemetry/instrumentation/mysql2/instrumentation_test.rb
@@ -172,6 +172,19 @@ describe OpenTelemetry::Instrumentation::Mysql2::Instrumentation do
         _(span.attributes['net.peer.name']).must_equal host.to_s
         _(span.attributes['net.peer.port']).must_equal port.to_s
       end
+
+      it 'encodes invalid byte sequences for db.statement' do
+        # \255 is off-limits https://en.wikipedia.org/wiki/UTF-8#Codepage_layout
+        sql = "SELECT * from users where users.id = 1 and users.email = 'test@test.com\255'"
+        obfuscated_sql = 'SELECT * from users where users.id = ? and users.email = ?'
+
+        expect do
+          client.query(sql)
+        end.must_raise Mysql2::Error
+
+        _(span.name).must_equal 'mysql'
+        _(span.attributes['db.statement']).must_equal obfuscated_sql
+      end
     end
 
     describe 'when db_statement set as omit' do
--- a/instrumentation/trilogy/lib/opentelemetry/instrumentation/trilogy/patches/client.rb
+++ b/instrumentation/trilogy/lib/opentelemetry/instrumentation/trilogy/patches/client.rb
@@ -87,10 +87,13 @@ module OpenTelemetry
             if sql.size > 2000
               'SQL query too large to remove sensitive data ...'
             else
-              obfuscated = sql.gsub(FULL_SQL_REGEXP, '?')
+              obfuscated = OpenTelemetry::Common::Utilities.utf8_encode(sql, binary: true)
+              obfuscated = obfuscated.gsub(FULL_SQL_REGEXP, '?')
               obfuscated = 'Failed to obfuscate SQL query - quote characters remained after obfuscation' if detect_unmatched_pairs(obfuscated)
               obfuscated
             end
+          rescue StandardError
+            'OpenTelemetry error: failed to obfuscate sql'
           end
 
           def detect_unmatched_pairs(obfuscated)
@@ -140,6 +143,7 @@ module OpenTelemetry
             QUERY_NAME_RE.match(sql) { |match| match[1].downcase } unless sql.nil?
           rescue StandardError => e
             OpenTelemetry.logger.error("Error extracting sql statement type: #{e.message}")
+            nil
           end
         end
       end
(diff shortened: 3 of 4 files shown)


ericmustin:
Hey @adrianna-chang-shopify just ack'ing this. This looks good to me but i'll admit i'm very out of touch with Trilogy at the moment, and would appreciate input from @arielvalentin on this. I will try to give this a closer look this evening but seems straightforward imo and appreciate you keeping this instrumentation in line with mysql2

instrumentation/mysql2/test/opentelemetry/instrumentation/mysql2/instrumentation_test.rb
```diff
@@ -172,6 +172,19 @@
         _(span.attributes['net.peer.name']).must_equal host.to_s
         _(span.attributes['net.peer.port']).must_equal port.to_s
       end
+
+      it 'encodes invalid byte sequences for db.statement' do
```
arielvalentin, line 176: Thank you for adding this test back. That was clearly an oversight on my part.

instrumentation/trilogy/lib/opentelemetry/instrumentation/trilogy/patches/client.rb
```diff
@@ -87,10 +87,13 @@ def obfuscate_sql(sql)
...
               'SQL query too large to remove sensitive data ...'
             else
-              obfuscated = sql.gsub(FULL_SQL_REGEXP, '?')
+              obfuscated = OpenTelemetry::Common::Utilities.utf8_encode(sql, binary: true)
+              obfuscated = obfuscated.gsub(FULL_SQL_REGEXP, '?')
               obfuscated = 'Failed to obfuscate SQL query - quote characters remained after obfuscation' if detect_unmatched_pairs(obfuscated)
               obfuscated
             end
+          rescue StandardError
+            'OpenTelemetry error: failed to obfuscate sql'
```
arielvalentin, line 96: This should probably report the error using the error handler to make it easier for end users to track down.

arielvalentin (approved):

instrumentation/trilogy/lib/opentelemetry/instrumentation/trilogy/patches/client.rb
adrianna-chang-shopify: Done!

instrumentation/trilogy/lib/opentelemetry/instrumentation/trilogy/patches/client.rb
```diff
@@ -87,10 +87,14 @@ def obfuscate_sql(sql)
...
               'SQL query too large to remove sensitive data ...'
             else
-              obfuscated = sql.gsub(FULL_SQL_REGEXP, '?')
+              obfuscated = OpenTelemetry::Common::Utilities.utf8_encode(sql, binary: true)
+              obfuscated = obfuscated.gsub(FULL_SQL_REGEXP, '?')
               obfuscated = 'Failed to obfuscate SQL query - quote characters remained after obfuscation' if detect_unmatched_pairs(obfuscated)
               obfuscated
             end
+          rescue StandardError => e
+            OpenTelemetry.logger.error("Failed to obfuscate sql: #{e.message}")
```
arielvalentin, line 96:
```suggestion
            OpenTelemetry.handle_error(message: "Failed to obfuscate SQL", exception: e)
```

ericmustin (approved):
lgtm pending @arielvalentin suggested change wrt error handling

instrumentation/trilogy/lib/opentelemetry/instrumentation/trilogy/patches/client.rb
```diff
@@ -87,10 +87,14 @@ def obfuscate_sql(sql)
...
+            nil
```
robertlaurin, line 97: Lets keep the default message in addition to the error handling Ariel suggested above. The reasoning here is that if we default to nil, that attribute will get dropped when the span goes through its finalization steps. I think dropping the opentelemetry error as an attribute value on the span is a little bit on the crude side, but if a user is looking at a trace and sees a DB span with this message on the `db.statement` at least they have some sense that something went wrong, with nil it will just be absent which is a bit more confusing IMO.
```suggestion
            'OpenTelemetry error: failed to obfuscate sql'
```

adrianna-chang-shopify: Got it! Thanks for clarifying

instrumentation/mysql2/lib/opentelemetry/instrumentation/mysql2/patches/client.rb
```diff
@@ -78,8 +78,9 @@ def obfuscate_sql(sql)
               obfuscated = 'Failed to obfuscate SQL query - quote characters remained after obfuscation' if detect_unmatched_pairs(obfuscated)
               obfuscated
             end
-          rescue StandardError
-            'OpenTelemetry error: failed to obfuscate sql'
+          rescue StandardError => e
+            OpenTelemetry.logger.error("Failed to obfuscate sql: #{e.message}")
```
arielvalentin, line 82:
```suggestion
            OpenTelemetry.handle_error(message: "Failed to obfuscate SQL", exception: e)
```

robertlaurin (approved):
LGTM

adrianna-chang-shopify:
Oops let me just update the CHANGELOG quickly

robertlaurin:
> Oops let me just update the CHANGELOG quickly

The deploy robot will handle that for us, but if you've already updated it that's ok!

instrumentation/trilogy/CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Release History: opentelemetry-instrumentation-trilogy
 
+### v0.51.2 / 2023-xx-xx
+
+* Ensure SQL obfuscation can handle invalid byte encoding. [#345](https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/345)
+
```
arielvalentin, line 6:
```suggestion
```

The release scripts dynamically generate release notes based on PR titles that follow conventional commits format. https://www.conventionalcommits.org/en/v1.0.0/

Please rename this PR's title to use conventional commits.

arielvalentin (changes requested):

instrumentation/trilogy/CHANGELOG.md
adrianna-chang-shopify: Neat, TIL about conventional commits! 😮 I've dropped the CHANGELOG commit and updated the PR description.

arielvalentin (approved):

PR #347, opened 2023-02-27
feat: add with_attributes context propagation to Trilogy instrumentation

adrianna-chang-shopify:
Attempting to reach feature parity between Mysql2 instrumentation and Trilogy instrumentation. I believe we'll want to make additional attributes available to Trilogy spans as with Mysql2 spans.

PR #348, opened 2023-02-27
feat: add option to configure span name for trilogy

adrianna-chang-shopify:
Dependent on: https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/347

Shopify is planning to start migrating apps from Mysql2 to Trilogy as a SQL database client. Ensuring parity between OpenTelemetry instrumentation for Mysql2 and Trilogy is important before we move to Trilogy.

This PR adds configurable span names to the Trilogy instrumentation.

### DIFF
--- a/instrumentation/trilogy/lib/opentelemetry/instrumentation/trilogy/patches/client.rb
+++ b/instrumentation/trilogy/lib/opentelemetry/instrumentation/trilogy/patches/client.rb
@@ -106,20 +106,27 @@ module OpenTelemetry
             %r{'|"|\/\*|\*\/}.match(obfuscated)
           end
 
-          def database_span_name(sql)
-            # Setting span name to the SQL query without obfuscation would
-            # result in PII + cardinality issues.
-            # First attempt to infer the statement type then fallback to
-            # current Otel approach {database.component_name}.{database_instance_name}
-            # https://github.com/open-telemetry/opentelemetry-python/blob/39fa078312e6f41c403aa8cad1868264011f7546/ext/opentelemetry-ext-dbapi/tests/test_dbapi_integration.py#L53
-            # This creates span names like mysql.default, mysql.replica, postgresql.staging etc.
+          def database_span_name(sql) # rubocop:disable Metrics/CyclomaticComplexity
+            case config[:span_name]
+            when :statement_type
+              extract_statement_type(sql)
+            when :db_name
+              database_name
+            when :db_operation_and_name
+              op = OpenTelemetry::Instrumentation::Trilogy.attributes['db.operation']
+              name = database_name
+              if op && name
+                "#{op} #{name}"
+              elsif op
+                op
+              elsif name
+                name
+              end
+            end || 'mysql'
+          end
 
-            statement_type = extract_statement_type(sql)
-
-            return statement_type unless statement_type.nil?
-
-            # fallback
-            'mysql'
+          def database_name
+            connection_options[:database]
           end
 
           def net_peer_name
--- a/instrumentation/trilogy/lib/opentelemetry/instrumentation/trilogy/instrumentation.rb
+++ b/instrumentation/trilogy/lib/opentelemetry/instrumentation/trilogy/instrumentation.rb
@@ -24,6 +24,7 @@ module OpenTelemetry
 
         option :peer_service, default: nil, validate: :string
         option :db_statement, default: :obfuscate, validate: %I[omit include obfuscate]
+        option :span_name, default: :statement_type, validate: %I[statement_type db_name db_operation_and_name]
 
         private
 
--- a/instrumentation/trilogy/test/opentelemetry/instrumentation/trilogy/instrumentation_test.rb
+++ b/instrumentation/trilogy/test/opentelemetry/instrumentation/trilogy/instrumentation_test.rb
@@ -340,5 +340,135 @@ describe OpenTelemetry::Instrumentation::Trilogy do
         end
       end
     end
+
+    describe 'when span_name is set as statement_type' do
+      it 'sets span name to statement type' do
+        OpenTelemetry::TestHelpers.with_env('OTEL_RUBY_INSTRUMENTATION_TRILOGY_CONFIG_OPTS' => 'span_name=statement_type') do
+          instrumentation.instance_variable_set(:@installed, false)
+          instrumentation.install
+
+          sql = "SELECT * from users where users.id = 1 and users.email = 'test@test.com'"
+          expect do
+            client.query(sql)
+          end.must_raise Trilogy::Error
+
+          _(span.name).must_equal 'select'
+        end
+      end
+
+      it 'sets span name to mysql when statement type is not recognized' do
+        OpenTelemetry::TestHelpers.with_env('OTEL_RUBY_INSTRUMENTATION_TRILOGY_CONFIG_OPTS' => 'span_name=statement_type') do
+          instrumentation.instance_variable_set(:@installed, false)
+          instrumentation.install
+
+          sql = 'DESELECT 1'
+          expect do
+            client.query(sql)
+          end.must_raise Trilogy::Error
+
+          _(span.name).must_equal 'mysql'
+        end
+      end
+    end
+
+    describe 'when span_name is set as db_name' do
+      it 'sets span name to db name' do
+        OpenTelemetry::TestHelpers.with_env('OTEL_RUBY_INSTRUMENTATION_TRILOGY_CONFIG_OPTS' => 'span_name=db_name') do
+          instrumentation.instance_variable_set(:@installed, false)
+          instrumentation.install
+
+          sql = "SELECT * from users where users.id = 1 and users.email = 'test@test.com'"
+          expect do
+            client.query(sql)
+          end.must_raise Trilogy::Error
+
+          _(span.name).must_equal 'mysql'
+        end
+      end
+
+      describe 'when db name is nil' do
+        let(:database) { nil }
+
+        it 'sets span name to mysql' do
+          OpenTelemetry::TestHelpers.with_env('OTEL_RUBY_INSTRUMENTATION_TRILOGY_CONFIG_OPTS' => 'span_name=db_name') do
+            instrumentation.instance_variable_set(:@installed, false)
+            instrumentation.install
+
+            sql = "SELECT * from users where users.id = 1 and users.email = 'test@test.com'"
+            expect do
+              client.query(sql)
+            end.must_raise Trilogy::Error
+
+            _(span.name).must_equal 'mysql'
+          end
+        end
+      end
+    end
+
+    describe 'when span_name is set as db_operation_and_name' do
+      it 'sets span name to db operation and name' do
+        OpenTelemetry::TestHelpers.with_env('OTEL_RUBY_INSTRUMENTATION_TRILOGY_CONFIG_OPTS' => 'span_name=db_operation_and_name') do
+          instrumentation.instance_variable_set(:@installed, false)
+          instrumentation.install
+
+          sql = "SELECT * from users where users.id = 1 and users.email = 'test@test.com'"
+          OpenTelemetry::Instrumentation::Trilogy.with_attributes('db.operation' => 'foo') do
+            expect do
+              client.query(sql)
+            end.must_raise Trilogy::Error
+          end
+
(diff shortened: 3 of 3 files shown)


instrumentation/trilogy/README.md
```diff
@@ -40,6 +40,17 @@ OpenTelemetry::SDK.configure do |c|
 end
 ```
 
+The `trilogy` instrumentation allows the user to supply additional attributes via the `with_attributes` method. This makes it possible to supply additional attributes on trilogy spans. Attributes supplied in `with_attributes` supersede those automatically generated within `trilogy`'s automatic instrumentation. If you supply a `db.statement` attribute in `with_attributes`, this library's `:db_statement` configuration will not be applied.
+
+```ruby
+require 'opentelemetry/instrumentation/trilogy'
```
arielvalentin, line 46:
```suggestion
require 'opentelemetry-instrumentation-trilogy'
```

instrumentation/trilogy/lib/opentelemetry/instrumentation/trilogy.rb
```diff
@@ -11,6 +11,45 @@ module OpenTelemetry
   module Instrumentation
     # Contains the OpenTelemetry instrumentation for the Trilogy gem
     module Trilogy
+      extend self
+
+      CURRENT_ATTRIBUTES_KEY = Context.create_key('mysql-attributes-hash')
```
arielvalentin, line 16: Although these two context keys will have a different has entry I think it will be confusing to me if I inspected in the has in a debugger and saw that they key name was `mysql-attributes-hash`. 

I also wonder if you all will be running the mysql driver and trilogy drivers simultaneously in your apps. 

Would having a common `MySQL::Context` helper make sense?

instrumentation/trilogy/lib/opentelemetry/instrumentation/trilogy/patches/client.rb
```diff
@@ -103,19 +103,14 @@ def detect_unmatched_pairs(obfuscated)
...
-
-            statement_type = extract_statement_type(sql)
-
-            return statement_type unless statement_type.nil?
-
-            # fallback
-            'mysql'
+            # TODO: When https://github.com/github/trilogy/pull/48/commits/6e2377ae2d737962a7a1a46b8f63438868034c7b is released,
+            # we will have access to the database name, so we can support :span_name configs of :db_name and :db_operation_and_name.
+            case config[:span_name]
```
arielvalentin, line 108: Is there a way we can evaluate this once? Instead of switching everytime?

instrumentation/trilogy/lib/opentelemetry/instrumentation/trilogy/patches/client.rb
```diff
@@ -103,19 +103,14 @@ def detect_unmatched_pairs(obfuscated)
...
+            when :statement_type
+              extract_statement_type(sql)
+            when :db_operation
+              OpenTelemetry::Instrumentation::Trilogy.attributes['db.operation']
```
arielvalentin, line 112: Does this mean it is always user supplied?

instrumentation/trilogy/lib/opentelemetry/instrumentation/trilogy/patches/client.rb
ericmustin: wdyt about just keeping `:db_operation_and_name` and the behavior of the `mysql2` instrumentation being consistent

```ruby
            when :db_operation_and_name
              op = OpenTelemetry::Instrumentation::Mysql2.attributes['db.operation']
              name = database_name
              if op && name
                "#{op} #{name}"
              elsif op
                op
              elsif name
                name
              end
```

In this case we can let `database_name` be a no-op (or just add a code comment with the TODO), document the configuration limitation in instrumentation readme. I'd like to avoid a place where we add, and then immediately deprecate, the `:db_operation` configuration option.

instrumentation/trilogy/lib/opentelemetry/instrumentation/trilogy.rb
adrianna-chang-shopify: Applications should be using either Mysql2 or Trilogy as the database client, but not both simultaneously (am I understanding your question correctly?). Either way, having a common MySQL::Context SGTM 👍 Ultimately they're both propagating context for a SQL database. I can go ahead and extract a helper.

Can you clarify whether you think the context key should change? Maybe we should name them `mysql2-attributes-hash` and `trilogy-attributes-hash` respectively?

instrumentation/trilogy/lib/opentelemetry/instrumentation/trilogy/patches/client.rb
adrianna-chang-shopify: Makes sense! The `database_name` should be available in the next release, yes, just not sure when that will be 😄

instrumentation/trilogy/lib/opentelemetry/instrumentation/trilogy/patches/client.rb
arielvalentin: IDK. I am not familiar with the motivation to add it to the mysql2 instrumentation so I was hoping to learn more about why we want to add a config option for this.

instrumentation/trilogy/lib/opentelemetry/instrumentation/trilogy/patches/client.rb
arielvalentin: > but that seems like a fair amount of added complexity to avoid a fairly simple case statement. What do you think?

My instinct is to replace conditional logic with polymorphism but not required. 

In our case at GH we will always use `statement_type` so the case statement will always hit the first conditional so I suspect it won't have much of an impact for us to add a `===` call for each SQL span. If I see anything come up in our profiles I will report back to you all after we test this release.

arielvalentin: @ericmustin I have not kept a close eye on Trilogy releases but if `database_name` becomes available that would be pretty great.

instrumentation/trilogy/lib/opentelemetry/instrumentation/trilogy/patches/client.rb
adrianna-chang-shopify: From the [PR](https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/222) where this option was added to Mysql2:
> The db_operation_and_name option additionally looks for db.operation in the OpenTelemetry::Instrumentation::Mysql2.attributes (from the current context). This allows callers to add db.operation to the context, if they know it.

[This comment](https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/222#discussion_r1053552076) also points to the OTel spec, which suggests that
>  If db.sql.table is not available due to its semantics, the span SHOULD be named <db.operation> <db.name>

So my understanding was that OTel is in favour of including the db operation in the span name, if it's made available.

instrumentation/trilogy/lib/opentelemetry/instrumentation/trilogy.rb
arielvalentin: > Applications should be using either Mysql2 or Trilogy as the database client, but not both simultaneously (am I understanding your question correctly?). Either way, having a common MySQL::Context SGTM 👍 Ultimately they're both propagating context for a SQL database. I can go ahead and extract a helper.

My question was specifically about your migration process to Trilogy. I imagine your more complex applications are the like the GH monolith where we have may have multiple databases so I do imagine a world where you may have both drivers and instrumentations running in parallel. 

> Can you clarify whether you think the context key should change? Maybe we should name them mysql2-attributes-hash and trilogy-attributes-hash respectively? Or would we pull everything out to the shared SQL helper, including the key?

If it is a shared helper then the key should be the same. If you stick with the Trilogy specific Context key then I think naming it `trilogy-attributes-hash` would be best.

adrianna-chang-shopify: > My question was specifically about your migration process to Trilogy.

Gotcha. At Shopify, we intend to only be using a single db client per application, even in our multi-db applications. Migrating to Trilogy will involve migrating all of our dbs at the same time, so on our end I don't think there's a concern with sharing a key between the Mysql2 / Trilogy drivers.

I'll stick with the shared helper and a single key for now.

arielvalentin: Thank you!

instrumentation/trilogy/lib/opentelemetry/instrumentation/trilogy/patches/client.rb
adrianna-chang-shopify: Daniel Colson from Github's Ruby Archeology team has kindly offered to cut a new release of Trilogy tomorrow, so I'll get the proposed changes in tmrw.

instrumentation/trilogy/lib/opentelemetry/instrumentation/trilogy.rb
robertlaurin: Hey @arielvalentin I think we should probably pull the mysql context into common similarly to how we did it with the HTTP context. We could duplicate the behaviour of mysql into trilogy for now, and then extract it after. WDYT?

arielvalentin: > Hey @arielvalentin I think we should probably pull the mysql context into common similarly to how we did it with the HTTP context. We could duplicate the behaviour of mysql into trilogy for now, and then extract it after. WDYT?

Works for me. Thank you friends!

instrumentation/trilogy/lib/opentelemetry/instrumentation/trilogy/patches/client.rb
arielvalentin: @composerinteralia 🙇🏼

adrianna-chang-shopify:
cc @arielvalentin @robertlaurin @ericmustin This should be ready for another review, folks! 🙇‍♀️ 

I've updated https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/347 with the changes that were suggested in this PR -- as discussed, we'll stick with duplicating the attributes behaviour between Mysql2 and Trilogy for now, and I'll leave our OTel team to handle a future refactor :) 

I'll go ahead and rebase this after #347 ships, let me know if there's anything else missing!

instrumentation/trilogy/lib/opentelemetry/instrumentation/trilogy.rb
arielvalentin, line 16:
```suggestion
      CURRENT_ATTRIBUTES_KEY = Context.create_key('trilogy-attributes-hash')
```

arielvalentin (approved):

ericmustin (approved):

instrumentation/trilogy/lib/opentelemetry/instrumentation/trilogy.rb
adrianna-chang-shopify: @arielvalentin I thought the plan was to eventually extract this into a shared MySQL context helper 🤔 Shouldn't the key be kept the same in that case? (e.g. https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/348#discussion_r1120727262)

adrianna-chang-shopify: I can change this to be `trilogy-attributes-hash`, but will this change to `mysql-attributes-hash` when the helper is eventually extracted?

arielvalentin: That will be OK then. I think since they are different Key Object Instances in this commit I would prefer them to have different names.

adrianna-chang-shopify: Sounds good! Will update 👍

adrianna-chang-shopify: Okay, https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/347 is updated with the change. Will rebase this PR once the aforementioned one ships!

PR #356, opened 2023-02-28
feat!: Remove retain_middleware_names Rack Option

arielvalentin:
This configuration option sets a rack environment variable that isn't used in the instrumentation or its dependencies.

dd-trace-rb used this option to set the span name to the Rack application Name:

https://github.com/DataDog/dd-trace-rb/blob/d76de947d66fe50f5ac6c7a6016ebbfa5c0de986/lib/datadog/tracing/contrib/rack/middlewares.rb#L158

It looks like some of the functionality was ported over but applying it to the span name was never completed:

https://github.com/open-telemetry/opentelemetry-ruby/pull/166


ahayworth:
Is this just dead code?

arielvalentin:
@ahayworth AFAICT it was copy/paste incomplete...

PR #357, opened 2023-03-01
feat: Elasticsearch client auto instrumentation

estolfo:
These changes add instrumentation of the [Elasticsearch client](https://github.com/elastic/elasticsearch-ruby). More technically, it instruments the [`elastic-transport` gem](https://github.com/elastic/elastic-transport-ruby), which is a component/dependency of the elasticsearch client.

The instrumentation is more or less a copy of what's in the Elastic APM agent, found [here](https://github.com/elastic/apm-agent-ruby/blob/main/lib/elastic_apm/spies/elasticsearch.rb).

There are some open questions with this instrumentation. Namely, there are Elasticsearch client auto instrumentations for the Java, Python, .NET OpenTelemetry projects but they are inconsistent with each other. So I have [an open issue](https://github.com/open-telemetry/opentelemetry-specification/issues/3239) with the OTel specifications group to add a spec for Elasticsearch client instrumentation. Until that spec is written and approved, this PR will be in flux.

At this point, I think the code can be reviewed for structure and functionality. I will move it from draft status into "ready for review" when the above-mentioned spec is approved.

Note that this instrumentation is only for Elasticsearch client >= v8.0

TODO:

- [ ] Verify the versioning scheme in the auto instrumentation gems. Have I correctly set the initial version to `0.0.1`?
- [ ] Spec written and approved by the OTel specs group
- [ ] Complete `TODO`s in `client.rb`
- [ ] Find out where documentation of `sanitize_field_names` should go


instrumentation/elasticsearch/lib/opentelemetry/instrumentation/elasticsearch/patches/client.rb
```diff
@@ -0,0 +1,62 @@
...
+          # rubocop:disable Metrics/CyclomaticComplexity, Metrics/PerceivedComplexity
+          def perform_request(method, path, *args, &block)
+            config = Elasticsearch::Instrumentation.instance.config
+            attributes = {
+              'db.system' => 'elasticsearch',
+              # TODO: should we set db.name?
+              # 'db.name' => database_name,
+              'db.operation' => method,
+              # TODO: is this true for all elasticsearch client requests?
+              'net.transport' => IP_TCP,
```
kares, line 25: that field isn't that interesting here given the http client itself might be instrumented?
and if that isn't the case than shouldn't the http.* attributes be of more interest here?

instrumentation/elasticsearch/lib/opentelemetry/instrumentation/elasticsearch/version.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+# Copyright The OpenTelemetry Authors
+#
+# SPDX-License-Identifier: Apache-2.0
+
+module OpenTelemetry
+  module Instrumentation
+    module Elasticsearch
+      VERSION = '0.0.1'
```
ahayworth, line 10: I think we'd want `0.1.0` here but I don't feel strongly.

instrumentation/elasticsearch/lib/opentelemetry/instrumentation/elasticsearch/wildcard_pattern.rb
```diff
@@ -0,0 +1,47 @@
...
+
+# Copyright The OpenTelemetry Authors
+#
+# SPDX-License-Identifier: Apache-2.0
+
+module OpenTelemetry
+  module Instrumentation
+    module Elasticsearch
+      # Object representing a user-supplied key pattern that behaves as a regex
+      class WildcardPattern
```
ahayworth, line 11: Why not just allow a regex in this case? Is this mimicking an upstream Elasticsearch wildcard pattern? (And, if so, could we just borrow their implementation?)

ahayworth (commented):
Thank you for this! I left a few initial thoughts, but I'll wait until you mark it ready for review before doing much deeper.

The most recent attempt at this was by @plantfansam in https://github.com/open-telemetry/opentelemetry-ruby/pull/1189 - that version is slightly different; perhaps we could harmonize the approaches?

instrumentation/elasticsearch/lib/opentelemetry/instrumentation/elasticsearch/patches/client.rb
estolfo: I didn't see that there was an PR already for this instrumentation, thanks for pointing that out. I'll take a look at it and see how these can be merged.
When you say that the approach is one you would recommend, you mean that there's something about @plantfansam's approach that is preferable to the one in this PR? If so, I might have some follow-up questions to clarify what exactly you'd prefer to have in the instrumentation as I take a look.

instrumentation/elasticsearch/lib/opentelemetry/instrumentation/elasticsearch/wildcard_pattern.rb
estolfo: The reason I used a special object for this and didn't require a regex was to allow the config option `sanitize_field_names` to be set as a list of strings or a string representing a list. So the config option could be set as:
`{ sanitize_field_names: 'foo.*,*.bar' } }`
or
`{ sanitize_field_names: ['Auth*tion', 'abc*', '*xyz'] }`

I decided to do that instead of requiring a regex or list of regexes so that the config could be set via an environment variable, if needed. Maybe that's not a valid use case in OTel instrumentations? I'm happy to change this, just let me know what makes more sense for your users.

instrumentation/elasticsearch/lib/opentelemetry/instrumentation/elasticsearch/instrumentation.rb
```diff
@@ -0,0 +1,60 @@
...
+          present? && gem_version >= MINIMUM_VERSION
+        end
+
+        def patch
+          ::Elastic::Transport::Client.prepend(Patches::Client)
+        end
+
+        option :peer_service, default: nil, validate: :string
+        option :db_statement, default: :obfuscate, validate: %I[omit obfuscate include]
+        option :sanitize_field_names, default: nil, validate: ->(v) { v.is_a?(String) || v.is_a?(Array) }
```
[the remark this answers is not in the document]
estolfo: See my comment [here](https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/357#discussion_r1146599050), I'm accepting a string representing a list as well but happy to change this if it's overkill.

instrumentation/elasticsearch/lib/opentelemetry/instrumentation/elasticsearch/patches/client.rb
estolfo: Update: I've taken a look at the other PR and I think it makes a lot of sense to use `OpenTelemetry::Common::HTTP::ClientContext.with_attributes`. We also have a way of avoiding redundant spans in the Elastic APM agent's Elasticsearch client instrumentation. So I'll add that in, including the config option, so the user has a choice.

Is there anything else about the approach in the other PR that you'd like to include in this one?

estolfo: Update: I've added a config option `capture_es_spans` to the instrumentation so that the user can decide whether they want the http span to be enriched with elasticsearch instrumentation attributes or if they want separate spans for es and http.

estolfo:
Just to recap where we are at this point: I have [an open PR with the semantic conventions group](https://github.com/open-telemetry/opentelemetry-specification/pull/3358) with a proposal for Elasticsearch client instrumentation.
There are a few instrumentations that already exist (.NET, Python, Java) and we are checking with those groups to see if they have any concerns.
I should have more info next week, as we are bringing it up in the language SIGs this week. Once that above PR is merged, I can return to this PR and update it and mark is as ready for review.
I believe I've addressed all the comments and questions specific to this PR at this point.

ericmustin:
added the `keep` label here, as far as i can tell this is mostly blocked by semconv consensus, which (reviewing the associated semconv pr) seems to be progressing/has some approvals/is close to merging. that being said, if this stays stuck in semconv purgatory for too long, i'd err toward merging this pr with the intention to update when semconv gets finalized, don't want to let perfect be the enemy of good.

estolfo:
Hey @ericmustin thanks for adding the `keep` label.
Based on discussions with the semantic conventions group, it seems like native otel instrumentation is a better path forward. So I've been working with the Elasticsearch clients team to write native instrumentation into the client libraries so that we don't have separate instrumentations that have to be maintained.
With that said, the proposed semantic conventions haven't been merged yet and the native instrumentation work is still in progress. Concerning this PR, we can close it if you'd like as it's unlikely this will be the path forward for OTel ES client instrumentation. Maybe I could have the Ruby group review the native instrumentation in the ES client when it's complete though?

cheempz:
Late to the party but I have been watching this one with interest because we would like to provide elasticsearch instrumentation in our OTel custom distro for Ruby.  I agree that client native instrumentation is the future, but wondering if/how native client instrumentation would support those using older versions of elastic clients?

estolfo:
@arielvalentin Yes, I mean instrumentation (use of the OTel API, creation of spans, etc) in the Elasticsearch client code itself. We refer to it as "native instrumentation" but maybe this is confusing in the context of Ruby because "native" usually means a c extension? Is there another term you can recommend using in the context of Ruby?
[Here](https://github.com/elastic/elastic-transport-ruby/compare/main...estolfo:elastic-transport-ruby:otel-instrumentation?expand=1) is a rough outline of what it would look like in Ruby. There are also changes in the `elasticsearch-api` gem.
[Here](https://github.com/elastic/elasticsearch-java/pull/588) is the native instrumentation in the Elasticsearch Java client my colleague as been working on.
I would gladly come to the Ruby SIG today to chat more about it but I not able to make the time. I can come next week though :)

estolfo:
@cheempz yes, sure

PR #359, opened 2023-03-05
chore: Add Rack 3.0 compatibility tests

arielvalentin:


instrumentation/rack/Appraisals
```diff
@@ -4,10 +4,14 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 
+appraise 'rack-3.0' do
+  gem 'rack', '~> 3.0.0'
+end
+
```
arielvalentin, line 10:
```suggestion

appraise 'rack-2.2.x' do
  gem 'rack', '~> 2.2.0'
end

```

ericmustin (approved):

robbkidd (approved):
We're tracking minors.

PR #360, opened 2023-03-05
ci: Add gem test details

arielvalentin:
It is sometimes difficult to debug the CI test suite because it lacks enough details to reproduce the error.

This change adds a few more details to help track down issues so we are able to reproduce them locally.

### DIFF
--- a/.github/actions/test_gem/action.yml
+++ b/.github/actions/test_gem/action.yml
@@ -86,7 +86,9 @@ runs:
       run: |
         if [[ -f "Appraisals" ]]; then
           for i in `bundle exec appraisal list | sed 's/-/_/g' `; do
+            echo "Appraisal install and test #{i}"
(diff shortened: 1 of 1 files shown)


ericmustin (approved):

.github/actions/test_gem/action.yml
```diff
@@ -86,7 +86,9 @@ runs:
       run: |
         if [[ -f "Appraisals" ]]; then
           for i in `bundle exec appraisal list | sed 's/-/_/g' `; do
+            echo "Appraisal install and test #{i}"
```
arielvalentin, line 89:
```suggestion
            echo "Appraisal install and test ${i}"
```

ericmustin, line 89: syntax issue

Issue #374, opened 2023-03-13
GraphQL 2.0.18 Tracer API Incompatible with Instrumentation <= 0.23.0


arielvalentin:
https://github.com/open-telemetry/opentelemetry-ruby-contrib/actions/runs/4404390868/jobs/7713940824

```console
1) Failure:
OpenTelemetry::Instrumentation::GraphQL::Tracers::GraphQLTracer::#platform_trace::when platform_resolve_type is enabled with legacy naming#test_0001_traces .resolve_type [/home/runner/work/opentelemetry-ruby-contrib/opentelemetry-ruby-contrib/instrumentation/graphql/test/instrumentation/graphql/tracers/graphql_tracer_test.rb:218]:
Expected nil to not be nil.

  2) Failure:
OpenTelemetry::Instrumentation::GraphQL::Tracers::GraphQLTracer::#platform_trace::when platform_authorized is enabled#test_0002_includes attributes [/home/runner/work/opentelemetry-ruby-contrib/opentelemetry-ruby-contrib/instrumentation/graphql/test/instrumentation/graphql/tracers/graphql_tracer_test.rb:205]:
Expected nil to not be nil.

  3) Failure:
OpenTelemetry::Instrumentation::GraphQL::Tracers::GraphQLTracer::#platform_trace::when platform_authorized is enabled#test_0001_traces .authorized [/home/runner/work/opentelemetry-ruby-contrib/opentelemetry-ruby-contrib/instrumentation/graphql/test/instrumentation/graphql/tracers/graphql_tracer_test.rb:186]:
Expected nil to not be nil.

  4) Failure:
OpenTelemetry::Instrumentation::GraphQL::Tracers::GraphQLTracer::#platform_trace::when platform_authorized is enabled with legacy naming#test_0001_traces .authorized [/home/runner/work/opentelemetry-ruby-contrib/opentelemetry-ruby-contrib/instrumentation/graphql/test/instrumentation/graphql/tracers/graphql_tracer_test.rb:168]:
Expected nil to not be nil.

  5) Failure:
OpenTelemetry::Instrumentation::GraphQL::Tracers::GraphQLTracer::#platform_trace::when platform_field is enabled#test_0002_includes attributes [/home/runner/work/opentelemetry-ruby-contrib/opentelemetry-ruby-contrib/instrumentation/graphql/test/instrumentation/graphql/tracers/graphql_tracer_test.rb:155]:
Expected nil to not be nil.

[...]

PR #375, opened 2023-03-13
fix!: Lock graphql max version to 2.0.17

arielvalentin:
2.0.18 introduces a refactoring of the GraphQL Trace API that makes it incompatible with the OTel Instrumentation.

This change locks the existing implementation to lower than 2.0.18

See https://github.com/open-telemetry/opentelemetry-ruby-contrib/issues/374 See https://github.com/rmosolgo/graphql-ruby/issues/4380

### DIFF
--- a/instrumentation/graphql/lib/opentelemetry/instrumentation/graphql/instrumentation.rb
+++ b/instrumentation/graphql/lib/opentelemetry/instrumentation/graphql/instrumentation.rb
@@ -12,7 +12,7 @@ module OpenTelemetry
       # The Instrumentation class contains logic to detect and install the GraphQL instrumentation
       class Instrumentation < OpenTelemetry::Instrumentation::Base
         compatible do
-          gem_version <= Gem::Version.new('2.0.16')
+          gem_version <= Gem::Version.new('2.0.17')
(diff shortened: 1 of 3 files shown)


instrumentation/graphql/lib/opentelemetry/instrumentation/graphql/instrumentation.rb
```diff
@@ -12,7 +12,7 @@ module GraphQL
       # The Instrumentation class contains logic to detect and install the GraphQL instrumentation
       class Instrumentation < OpenTelemetry::Instrumentation::Base
         compatible do
-          gem_version < Gem::Version.new('3.0.0')
+          gem_version <= Gem::Version.new('2.0.16')
```
ericmustin, line 15: is the PR Title incorrect? if not then shouldn't it be: 
```suggestion
          gem_version <= Gem::Version.new('2.0.17')
```

arielvalentin:
@ericmustin sloppy git work by me. I did not stage my commits properly and I just pushed up the 2.0.17.

robertlaurin (approved):

PR #379, opened 2023-03-14
fix!: Add support for GraphQL 2.0.19

arielvalentin:
This change adds support for GraphQL 2.0.19, which fixes an incompatibility with tracing changes introduced in 2.0.18.

This _does not_ transition to the new lower overhead API https://github.com/rmosolgo/graphql-ruby/blob/master/CHANGELOG.md#2018-9-march-2023

Details here: https://github.com/rmosolgo/graphql-ruby/issues/4378

Fixes https://github.com/open-telemetry/opentelemetry-ruby-contrib/issues/374


ericmustin (approved):
seems fine to me. maybe worth socializing the state of things in the SIG today though? As far as I understand it

- 2.0.18 added a new tracing api with improved performance `GraphQL::Tracing::Trace`
- 2.0.18 also unintentionally broke existing tracers that subclass off the `platform` tracer
- 2.0.19 fixes this bug
- 2.0.19 does not auto-magically enable the lower overhead tracing api for otel instrumentation. that would be on us to deliver, as far as i can tell.

Do i have that right?

arielvalentin:
@ericmustin Yessir. Its already on the agenda. 

Also https://github.com/rmosolgo/graphql-ruby/issues/4380

arielvalentin:
Much to my dismay. Tests are non-deterministic:

`bundle exec rake test TESTOPTS="--seed=8625"`

arielvalentin:
@ericmustin looks like I got the non-deterministic test failure under control but it required that I break encapsulation of the library to degrees that make me uncomfortable.

PR #385, opened 2023-03-16
refactor: use semantic calls in the mysql2 instrumentation

lloydpick:
Unsure if you want the references to include the full module name or not. eg. `SemanticConventions::Trace::DB_STATEMENT` vs `OpenTelemetry::SemanticConventions::Trace::DB_STATEMENT`, as I tried to see if there was an convention in the repository, but it seems to be mismatched [[1](https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/95707f9ce9fb1a485fd7437c313615bb66dc1ffc/instrumentation/action_pack/lib/opentelemetry/instrumentation/action_pack/patches/action_controller/metal.rb#L20-L21)] [[2](https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/95707f9ce9fb1a485fd7437c313615bb66dc1ffc/instrumentation/aws_sdk/lib/opentelemetry/instrumentation/aws_sdk/messaging_helper.rb#L32-L35)].

### DIFF
--- a/instrumentation/mysql2/lib/opentelemetry/instrumentation/mysql2/patches/client.rb
+++ b/instrumentation/mysql2/lib/opentelemetry/instrumentation/mysql2/patches/client.rb
@@ -54,9 +54,9 @@ module OpenTelemetry
             attributes = client_attributes
             case config[:db_statement]
             when :include
-              attributes['db.statement'] = sql
+              attributes[SemanticConventions::Trace::DB_STATEMENT] = sql
             when :obfuscate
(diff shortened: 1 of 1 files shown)


ahayworth (approved):
I don't think we have a strong preference yet on fully-qualified names or not. 🤷

Either way though - thank you very much for this! Every little bit of cleanup like this helps keep the project sustainable, and so it's very much appreciated. ❤️ 😄 ❤️

arielvalentin (approved):

Issue #372, opened 2023-03-07
Non Deterministic Test Rack


arielvalentin:
https://github.com/open-telemetry/opentelemetry-ruby-contrib/actions/runs/4356739174/jobs/7615156371

```console
/opt/hostedtoolcache/Ruby/3.2.1/x64/lib/ruby/gems/3.2.0/gems/yard-doctest-0.1.17/lib/yard-doctest.rb:57: warning: `&' interpreted as argument prefix
Run options: --seed 52002

# Running:

F................................................

Finished in 0.030398s, 1611.[96](https://github.com/open-telemetry/opentelemetry-ruby-contrib/actions/runs/4356739174/jobs/7615156371#step:3:106)62 runs/s, 2796.2679 assertions/s.

  1) Failure:
OpenTelemetry::Instrumentation::Rack::Instrumentation::when rack gem does not exist#test_0001_skips installation [/home/runner/work/opentelemetry-ruby-contrib/opentelemetry-ruby-contrib/instrumentation/rack/test/opentelemetry/instrumentation/rack/instrumentation_test.rb:45]:
Expected nil (NilClass) to respond to #empty?.

49 runs, 85 assertions, 1 failures, 0 errors, 0 skips
rake aborted!
Command failed with status (1)
/opt/hostedtoolcache/Ruby/3.2.1/x64/bin/bundle:25:in `load'
/opt/hostedtoolcache/Ruby/3.2.1/x64/bin/bundle:25:in `<main>'
Tasks: TOP => test
(See full trace by running task with --trace)
```

ahayworth:
This will be fixed by #388

PR #388, opened 2023-03-18
ci: do not test a config that will never succeed

### LINKED ISSUE #387: Rack tests are non-deterministic
**Description of the bug**

Rack tests are non-deterministic. It is trivial to reproduce -

Good run:
```
opentelemetry-ruby-contrib/instrumentation/rack on  ahayworth/rack-test-ordering-failure [?] via 💎 v3.0.4 took 2s
zsh ❯ bundle exec rake test SEED=1
/Users/andrew/.rbenv/versions/3.0.4/lib/ruby/gems/3.0.0/gems/yard-doctest-0.1.17/lib/yard-doctest.rb:57: warning: `&' interpreted as argument prefix
Run options: --seed 1

# Running:

.................................................

Finished in 0.014089s, 3477.8905 runs/s, 6813.8262 assertions/s.

49 runs, 96 assertions, 0 failures, 0 errors, 0 skips
```

Bad run:

```

opentelemetry-ruby-contrib/instrumentation/rack on  main via 💎 v3.0.4
zsh ❯ bundle exec rake test SEED=6372
/Users/andrew/.rbenv/versions/3.0.4/lib/ruby/gems/3.0.0/gems/yard-doctest-0.1.17/lib/yard-doctest.rb:57: warning: `&' interpreted as argument prefix
Run options: --seed 6372


# Running:

F................................................

Finished in 0.015968s, 3068.6373 runs/s, 5323.1463 assertions/s.

  1) Failure:
OpenTelemetry::Instrumentation::Rack::Instrumentation::when rack gem does not exist#test_0001_skips installation [/Users/andrew/projects/opentelemetry-ruby-contrib/instrumentation/rack/test/opentelemetry/instrumentation/rack/instrumentation_test.rb:45]:
Expected nil (NilClass) to respond to #empty?.

49 runs, 85 assertions, 1 failures, 0 errors, 0 skips
rake aborted!
Command failed with status (1)
/Users/andrew/.rbenv/versions/3.0.4/bin/bundle:23:in `load'
/Users/andrew/.rbenv/versions/3.0.4/bin/bundle:23:in `<main>'
Tasks: TOP => test
(See full trace by running task with --trace)
```

It seems as if the tests in `test/opentelemetry/instrumentation/rack/instrumentation_test.rb` are run _first_, then the instrumentation's `config` hash is always empty (and thus further assertions on the hash keys fail). It seems to be related to how we are resetting the instrumentation in that test with an explicitly empty config hash - the defaults aren't getting loaded as we expect.

cc @arielvalentin - I think this change started failing after #356

ahayworth:
In this test, we are asserting that the instrumentation is _not_
installed when the `Rack` constant is not present (see the `before`
block for this test case). We then go on to test that the configuration
is the "default" configuration that you'd get with a fresh installation.

The problem is that if the `Rack` constant is not present, then the
`instrumentation-base` code that sets all the defaults for our options
is never run (and logically, why would it?). So these test lines can
never actually succeed.

Unless, of course, the `instrumentation` object we're referring to is
_not_ a pristine, new object. And in fact, depending on what order the
tests run in, our object is _not_ a pristine, new object. If you run
basically any _other_ test in this suite before, then we actually end up
recycling an object that is partially initialzied from a previous test,
and has an internal `@config` object that has some default options.

And so, the test is actually passing some times because of this
non-deterministic behavior. For example, if you run with `SEED=1`, then
the test suite passes (because other tests run first that initialize the
object completely). If you run with `SEED=6372`, it fails (because it is
the very first test run).

The _real_ bug here is that we do not have proper test isolation. I
think that's actually a problem all over the code base, if I'm being
honest about it.

However, I elect not to fix that problem today. We'd need some way to
"reset" instrumentation and the registry in between tests (maybe not
_that_ hard, honestly). That's more than I want to do on a Saturday
afternoon. So to fix the current issue, we just don't bother testing
things that logically could never pass anyways. What we actually care
about is that the instrumentation reports it was not installed, which
does work correctly as it is.

Fixes #387

### DIFF
--- a/instrumentation/rack/test/opentelemetry/instrumentation/rack/instrumentation_test.rb
+++ b/instrumentation/rack/test/opentelemetry/instrumentation/rack/instrumentation_test.rb
@@ -42,14 +42,6 @@ describe OpenTelemetry::Instrumentation::Rack::Instrumentation do
 
     it 'skips installation' do
       _(instrumentation).wont_be :installed?
-      _(instrumentation.config[:allowed_request_headers]).must_be_empty
-      _(instrumentation.config[:allowed_response_headers]).must_be_empty
-      _(instrumentation.config[:application]).must_be_nil
-      _(instrumentation.config[:record_frontend_span]).must_equal false
-      _(instrumentation.config[:untraced_endpoints]).must_be_empty
-      _(instrumentation.config[:url_quantization]).must_be_nil
-      _(instrumentation.config[:untraced_requests]).must_be_nil
-      _(instrumentation.config[:response_propagators]).must_be_empty
     end
   end
 end


ahayworth:
Fixes #372

arielvalentin (approved):

PR #390, opened 2023-03-20
refactor: use semantic calls in the sidekiq instrumentation

lloydpick:
simple change to use semantic names over strings, will slowly make my way through the different instrumentations when I have time


arielvalentin (approved):

ahayworth (commented):
@lloydpick I ... think we might actually need to declare a dependency on the `semantic_conventions` gem? It's a dependency of the _SDK_, which is a development dependency for the gem - so tests pass. But I think if someone were to install the API and instrumentation (but _not_ the SDK, which is technically valid) it might break?

(If so, we should probably fix this elsewhere too 👀 )

mwear (approved):

Issue #9, opened 2020-05-12
Grape instrumentation adapter


mwear:
Convert Datadog instrumentation for Grape to OpenTelemetry. 

See
  * [Datadog porting guide](https://github.com/open-telemetry/opentelemetry-ruby/blob/master/adapters/datadog-porting-guide.md)
  * [Semantic Conventions](https://github.com/open-telemetry/opentelemetry-specification/tree/master/specification/trace/semantic_conventions)
  * [Datadog implementation](https://github.com/DataDog/dd-trace-rb/tree/master/lib/ddtrace/contrib/grape)

muripic:
Hi there!

We've picked up the Grape instrumentation at Catawiki and we're already working on it.
We are using the ddog implementation as a reference, using ActiveSupport notifications. It's still a WIP, but just for reference here's the fork we're working on: https://github.com/catawiki/opentelemetry-ruby-contrib 😄

PR #394, opened 2023-03-23
feat: add Grape instrumentation

muripic:
This PR addresses https://github.com/open-telemetry/opentelemetry-ruby-contrib/issues/9 by adding the basic instrumentation for [Ruby Grape](https://github.com/ruby-grape/grape).
It uses `ActiveSupport::Notifications` and it is heavily based on the [DataDog instrumentation](https://github.com/DataDog/dd-trace-rb/tree/3c4c4baaf43f31849e1f7e51ebf6839281f58df9/lib/datadog/tracing/contrib/grape), although it differs from it in some aspects (which will be explained in comments in the relevant sections of the changes).


instrumentation/grape/lib/opentelemetry/instrumentation/grape/instrumentation.rb
```diff
@@ -0,0 +1,46 @@
...
+#
+# SPDX-License-Identifier: Apache-2.0
+
+module OpenTelemetry
+  module Instrumentation
+    module Grape
+      # The Instrumentation class contains logic to detect and install the Grape instrumentation
+      class Instrumentation < OpenTelemetry::Instrumentation::Base
+        # Minimum Grape version needed for compatibility with this instrumentation
+        MINIMUM_VERSION = Gem::Version.new('1.2.0')
```
muripic, line 13: This is the minimum version required so we don't have to support [this workaround](https://github.com/DataDog/dd-trace-rb/blob/a02ccfc38123c087451f204ac6e822a6240216d5/lib/datadog/tracing/contrib/grape/endpoint.rb#L201).

instrumentation/grape/lib/opentelemetry/instrumentation/grape/event_handler.rb
```diff
@@ -0,0 +1,133 @@
...
+          end
+
+          def request_method(endpoint)
+            endpoint.options[:method]&.first
+          end
+
+          def path(endpoint)
+            namespace = endpoint.routes.first.namespace
+            version = endpoint.routes.first.options[:version] || ''
+            prefix = endpoint.routes.first.options[:prefix]&.to_s || ''
```
muripic, line 119: This aspect of our implementation differs from DDog: I decided to include the prefix and the version in the full path, because they are part of the path so I don't see why we shouldn't. If you can think of any objections, please let me know 🙏 😄

instrumentation/grape/test/test_helper.rb
```diff
@@ -0,0 +1,43 @@
...
+end
+
+# Helper functions
+def uninstall_and_cleanup
+  instrumentation.instance_variable_set('@installed', false)
+  unsubscribe
+  EXPORTER.reset
+end
+
+def unsubscribe
```
muripic, line 31: For a complete cleanup, it was necessary to fully unsubscribe from all built in `ActiveSupport::Notifications`. If I didn't do this, install would run multiple times because of the way the tests need to be set up and are randomised, and this created multiple subscriptions. A span was generated for each of these subscriptions.
I felt that this was the correct approach since in real scenarios, the installation will only happen once.

instrumentation/grape/lib/opentelemetry/instrumentation/grape/event_handler.rb
```diff
@@ -0,0 +1,133 @@
...
+            end
+          end
+
+          private
+
+          def tracer
+            Grape::Instrumentation.instance.tracer
+          end
+
+          def span_name(endpoint)
```
muripic, line 86: According to the [DDog porting guidelines](https://github.com/catawiki/opentelemetry-ruby-contrib/blob/55b6be133bd038fc597d8a08f7da661734eb0d38/instrumentation/datadog-porting-guide.md#L49), `resource` in DDog is equivalent to `span.name` here.

`resource` in DDog has the following structure: `BasicAPI GET /hello`, which includes the API class name, the method and the route. We decided to remove the API class name because we felt it didn't add much to the name (it's still in the attributes) and without this part it follows the recommended practice for http spans, as described [here](https://opentelemetry.io/docs/reference/specification/trace/semantic_conventions/http/#name).

Also, I understand that we should keep cardinality low, and this is bound to have a relatively high cardinality on services with multiple endpoints, but if we went for something like 'GET' then it would be harder to group traces, so I'm wondering how low it should be in this case 🤔 And also, maybe a lower cardinality is needed for other instrumentations, but not for Grape? These are all open questions to us, so we'd be really grateful if you could provide more input and I'll be happy to make the necessary adjustments to the span name 🙏 😄

instrumentation/grape/README.md
```diff
@@ -0,0 +1,90 @@
...
+The Grape instrumentation is a community-maintained instrumentation for [Grape][grape], a REST-like API framework for Ruby.
+
+It relies on the Grape built-in support for `ActiveSupport::Notifications` (more info [here](https://github.com/ruby-grape/grape#active-support-instrumentation)).
+
+It currently supports the following events:
+
+- `endpoint_run.grape`
+- `endpoint_render.grape`
+- `endpoint_run_filters.grape`
+- `format_response.grape`
```
muripic, line 12: `format_response` is not part of the DDog instrumentation but we decided to add it because we felt it could add value with formatting errors (we've had some experiences where this would have been useful).

However, it is important to point out that this event is not a child of `endpoint_run`, so we'll be relying on another parent span (e.g. from Rack or Sinatra) to connect it to the trace. In other words, while `endpoint_render` and `endpoint_run_filters` have `endpoint_run` as their parent span, `format_response` does not.

Because of this, we're open to removing it if you feel it would cause trouble 😄

instrumentation/grape/README.md
```diff
@@ -0,0 +1,90 @@
...
+
+Example:
+
+```ruby
+OpenTelemetry::SDK.configure do |c|
+  c.use 'OpenTelemetry::Instrumentation::Grape', { ignored_events: [:endpoint_run_filters] }
+end
+```
+
+Note that the `endpoint_run` event cannot be ignored. If you need to disable the instrumentation, set `:enabled` to `false`:
```
muripic, line 61: We decided to exclude `endpoint_run` from the "ignorable" events because, since it's the "main" span, IMO it wouldn't make a lot of sense to ignore it. It's the also span that has all the more informative attributes, and also if it's excluded all the other spans would be orphaned (at least as far as Grape is concerned, because they would still grouped together into the trace by the middleware span).

instrumentation/grape/lib/opentelemetry/instrumentation/grape/event_handler.rb
```diff
@@ -0,0 +1,133 @@
...
+            endpoint = payload[:endpoint]
+            path = path(endpoint)
+            {
+              'component' => 'web',
+              'operation' => 'endpoint_run',
+              'grape.route.endpoint' => endpoint.options[:for]&.base.to_s,
+              'grape.route.path' => path,
+              'grape.route.method' => request_method(endpoint),
+              OpenTelemetry::SemanticConventions::Trace::HTTP_METHOD => request_method(endpoint),
+              OpenTelemetry::SemanticConventions::Trace::HTTP_ROUTE => path
```
muripic, line 100: We are aware that `grape.route.method` and `grape.route.path` here repeat the http attributes, but we've kept them this way to follow the [DDog instrumentation](https://github.com/DataDog/dd-trace-rb/blob/a02ccfc38123c087451f204ac6e822a6240216d5/lib/datadog/tracing/contrib/grape/endpoint.rb#L99-L105).

Please let us know what would be your preference regarding keeping or removing the duplicates 🙏 😄 

Another important point here is that we experimented a bit with adding `HTTP_STATUS_CODE`, not present in the DDog instrumentation. In case you think it could add value we could add it back 😄 [Here's the commit](https://github.com/catawiki/opentelemetry-ruby-contrib/pull/3/commits/f1abe62ab1ab9ed17008e12d1436b53d68415a86) where we removed it, in case you want to see what it would look like.

.github/workflows/ci-instrumentation-canary.yml
```diff
@@ -25,6 +25,7 @@ jobs:
           - ethon
           - excon
           - faraday
+          - grape
```
muripic, line 28: Sorry, I just noticed after the CI runs that this was missing.

instrumentation/grape/lib/opentelemetry/instrumentation/grape/event_handler.rb
```diff
@@ -0,0 +1,133 @@
...
+          # Handles the format_response.grape event
+          def format_response(_name, start, _finish, _id, payload)
+            endpoint = payload[:env]['api.endpoint']
+            name = span_name(endpoint)
+            attributes = {
+              'component' => 'template',
+              'operation' => 'format_response',
+              'grape.formatter.type' => formatter_type(payload[:formatter])
+            }
+            tracer.in_span(name, attributes: attributes, start_timestamp: start, kind: :server) do |span|
```
arielvalentin, line 75: ♻️ Is this a span or event? finish timestamp?

instrumentation/grape/lib/opentelemetry/instrumentation/grape/event_handler.rb
```diff
@@ -0,0 +1,133 @@
...
+
+          def span_name(endpoint)
+            "#{request_method(endpoint)} #{path(endpoint)}"
+          end
+
+          def run_attributes(payload)
+            endpoint = payload[:endpoint]
+            path = path(endpoint)
+            {
+              'component' => 'web',
```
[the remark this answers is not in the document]
arielvalentin: I am not keen on the `component` or `operation` attributes. They should either use `code.*` semconv or be nested under `grape` if they do not map to any semconv attributes.

instrumentation/grape/lib/opentelemetry/instrumentation/grape/event_handler.rb
arielvalentin: Please remove duplicates. We want to avoid having vendor specific requirements in this repository.

instrumentation/grape/opentelemetry-instrumentation-grape.gemspec
```diff
@@ -0,0 +1,53 @@
...
+  spec.homepage    = 'https://github.com/open-telemetry/opentelemetry-ruby-contrib'
+  spec.license     = 'Apache-2.0'
+
+  spec.files = Dir.glob('lib/**/*.rb') +
+               Dir.glob('*.md') +
+               ['LICENSE', '.yardopts']
+  spec.require_paths = ['lib']
+  spec.required_ruby_version = '>= 2.7.6'
+
+  spec.add_dependency 'activesupport'
```
arielvalentin, line 28: We must not introduce transitive runtime dependencies to 3rd party libraries. If the `ActiveSupport` is not defined, then they should not be able to use this instrumentation. 

```suggestion
  spec.add_development_dependency 'activesupport'
```

arielvalentin (changes requested):
Thank you for your contribution and I appreciate your patience while you have waited for a review! 

My required changes here are the following:

1. Updating gem dependencies to move activesupport to a dev dependency
2. Remove Datadog vendor specific attribute usage and use semconv instead

I have a question about:

1. The choice of using Span vs Span Events. I would like to understand why you want to record some of the events as spans.
2. Why you are ignoring the finish time instead of using it for the span end time?
3. Does Grape have an operating mode where it may run without Rack? If not, should this instrumentation _enrich_ the current Rack span instead of creating additional child spans?

mwear (commented):
Thanks for the contribution. This is off to a really great start @muripic! I made a first pass through the core part of the instrumentation. We will need to have some discussions on how we want to model the telemetry, but the pieces are definitely here. I mention this in a comment, but it's worth reiterating here. If this is meant to be used in conjunction with the Rack instrumentation, then Rack will create the HTTP server span for us. Typically, we will enrich that span as we get further information from the framework. If we look to the Sinatra instrumentation as an example:

https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/3efcbdf322b68f9dbb200020f40d18e028d67b38/instrumentation/sinatra/lib/opentelemetry/instrumentation/sinatra/middlewares/tracer_middleware.rb#L24-L38

then, we probably do not need to create a separate span on `endpoint_run`, but can decorate the Rack span with additional data.

The other point (also discussed in the comments) is that there are well defined semantic conventions for [http spans](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/http.md). We  should adhere to them when we can. Things are more fuzzy for framework specific spans and attributes. We will figure those out through discussions on the PR.

Lastly, I have not been very active on this project lately. I wanted to provide some feedback to get the ball rolling, but I wouldn't run with my suggestions until others chime in and confirm they are the direction we want to take this.

muripic:
Thank you @arielvalentin and @mwear for your review and your detailed comments :bowing_woman: 
I'll start reworking my PR in line with what you pointed out and follow up in the corresponding threads for the pending discussion points :smile:

instrumentation/grape/lib/opentelemetry/instrumentation/grape/event_handler.rb
muripic: Yes, the URLs are templated, so I'm leaving the span name as it is then 😄

mwear:
@muripic I just wanted to check in to see if you have what you need, or if there are any questions we can help answer.

muripic:
> @muripic I just wanted to check in to see if you have what you need, or if there are any questions we can help answer.

Thank you @mwear 🙇‍♀️ I apologize for the delay, I started working on the changes already but I've been OOO for the past week.
I'll try to wrap it up this week and reach out if there's anything blocking me 😄

instrumentation/grape/lib/opentelemetry/instrumentation/grape/event_handler.rb
muripic: To give some extra context, I decided to add the prefix and version to be able to differentiate cases such as the following:

```ruby
class V1API < Grape::API
  prefix :api
  version 'v1'
  get :hello do
    { message: 'Hello, world!' }
  end
end

class V2API < Grape::API
  prefix :api
  version 'v2'
  get :hello do
    { message: 'Hello, new world!' }
  end
end
```

If these are not added, the path will be `/hello` for both, and I thought it would be helpful to be able to differentiate them. At Catawiki we have several cases like this one where it would be useful for debugging to know the full path.

instrumentation/grape/lib/opentelemetry/instrumentation/grape/event_handler.rb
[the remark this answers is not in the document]
muripic: I removed those attributes 👍

instrumentation/grape/lib/opentelemetry/instrumentation/grape/event_handler.rb
muripic: Done ✅

instrumentation/grape/lib/opentelemetry/instrumentation/grape/event_handler.rb
```diff
@@ -0,0 +1,133 @@
...
+          end
+
+          # Handles the endpoint_render.grape event
+          def endpoint_render(_name, start, _finish, _id, payload)
+            name = span_name(payload[:endpoint])
+            attributes = {
+              'component' => 'template',
+              'operation' => 'endpoint_render'
+            }
+            tracer.in_span(name, attributes: attributes, start_timestamp: start, kind: :server) do |span|
```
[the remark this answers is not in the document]
muripic: Great point, I agree that these should be events. I turned them all into events of the parent Rack span.

instrumentation/grape/lib/opentelemetry/instrumentation/grape/event_handler.rb
```diff
@@ -0,0 +1,133 @@
...
+            # Prevent submitting empty filters
+            return if (!filters || filters.empty?) || !type || (finish - start).zero?
+
+            name = span_name(payload[:endpoint])
+            attributes = {
+              'component' => 'web',
+              'operation' => 'endpoint_run_filters',
+              'grape.filter.type' => type.to_s
+            }
+            tracer.in_span(name, attributes: attributes, start_timestamp: start, kind: :server) do |span|
```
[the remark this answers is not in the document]
muripic: Same as above, now they're all span events added to the parent Rack span 😄

instrumentation/grape/lib/opentelemetry/instrumentation/grape/event_handler.rb
```diff
@@ -0,0 +1,133 @@
...
+            end
+          end
+
+          # Handles the format_response.grape event
+          def format_response(_name, start, _finish, _id, payload)
+            endpoint = payload[:env]['api.endpoint']
+            name = span_name(endpoint)
+            attributes = {
+              'component' => 'template',
+              'operation' => 'format_response',
```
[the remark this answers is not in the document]
muripic: I decided to remove these attributes completely, since they were just leftovers from the DDog instrumentation, and I don't think they add much value. The operation is now the name of the event.

instrumentation/grape/opentelemetry-instrumentation-grape.gemspec
muripic: Done ✅

instrumentation/grape/lib/opentelemetry/instrumentation/grape/event_handler.rb
```diff
@@ -0,0 +1,133 @@
...
+module OpenTelemetry
+  module Instrumentation
+    module Grape
+      # Handles the events instrumented with ActiveSupport notifications.
+      # These handlers contain all the logic needed to create and connect spans.
+      class EventHandler
+        class << self
+          # Handles the start of the endpoint_run.grape event (the parent event), where the context is attached
+          def endpoint_run_start(_name, _id, payload)
+            name = span_name(payload[:endpoint])
```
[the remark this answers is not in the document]
muripic: Yes, great point. I turned all of these into events.

instrumentation/grape/lib/opentelemetry/instrumentation/grape/event_handler.rb
```diff
@@ -0,0 +1,133 @@
...
+            span = tracer.start_span(name, attributes: run_attributes(payload), kind: :server)
```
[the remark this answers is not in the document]
muripic: I decided to go for this, thank you for the suggestion 🙇‍♀️

instrumentation/grape/lib/opentelemetry/instrumentation/grape/event_handler.rb
```diff
@@ -0,0 +1,133 @@
...
+            span.finish
+            OpenTelemetry::Context.detach(token)
+          end
+
+          # Handles the endpoint_render.grape event
+          def endpoint_render(_name, start, _finish, _id, payload)
+            name = span_name(payload[:endpoint])
+            attributes = {
+              'component' => 'template',
+              'operation' => 'endpoint_render'
```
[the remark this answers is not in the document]
muripic: I removed the `component` attribute and turned these spans into events, so the `operation` is now the name of those events 😃

instrumentation/grape/lib/opentelemetry/instrumentation/grape/event_handler.rb
```diff
@@ -0,0 +1,133 @@
...
+          # Handles the end of the endpoint_run.grape event (the parent event), where the context is detached
+          def endpoint_run_finish(_name, _id, payload)
+            span = payload.delete(:__opentelemetry_span)
+            token = payload.delete(:__opentelemetry_ctx_token)
+            return unless span && token
+
+            handle_payload_exception(span, payload[:exception_object]) if payload[:exception_object]
+
+            span.finish
+            OpenTelemetry::Context.detach(token)
```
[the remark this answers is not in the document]
muripic: I ended up removing this code, so I didn't implement this. Since now the code reuses the Rack span, I don't know if we need this somewhere, but please let me know if we do 🙏

instrumentation/grape/README.md
muripic: I decided to keep the exclusion of `endpoint_run` from the "ignorable" events with the new changes too because it's the parent event, where the Rack span name is modified and the most important attributes are added.

arielvalentin (approved):

instrumentation/grape/example/trace_demonstration.rb
```diff
@@ -0,0 +1,52 @@
...
+
+require 'bundler/setup'
+
+Bundler.require
+
+require 'opentelemetry-api'
+require 'opentelemetry-sdk'
+require 'opentelemetry-instrumentation-grape'
+require 'opentelemetry-instrumentation-rack'
+require 'grape'
```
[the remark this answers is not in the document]
muripic: Fixed in https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/394/commits/383c1b54d2c718b3ad63d99b71269d211f7d7344 😄

instrumentation/grape/lib/opentelemetry/instrumentation/grape/event_handler.rb
```diff
@@ -0,0 +1,107 @@
...
+          # Handles the start of the endpoint_run event, modifying the parent Rack span
+          # and recording it as a span event
+          def endpoint_run(_name, start, _finish, _id, payload)
+            span = OpenTelemetry::Instrumentation::Rack.current_span
+            return unless span.recording?
+
+            endpoint = payload[:endpoint]
+            span.name = span_name(endpoint)
+            span.set_attribute('code.namespace', endpoint.options[:for]&.base.to_s)
+            span.set_attribute(OpenTelemetry::SemanticConventions::Trace::HTTP_ROUTE, path(endpoint))
```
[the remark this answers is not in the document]
muripic: My bad, totally overlooked this one 😅 
Fixed in https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/394/commits/5decb4598a2759e1334c7bfe531ef675f340088f

instrumentation/grape/lib/opentelemetry/instrumentation/grape/event_handler.rb
```diff
@@ -0,0 +1,107 @@
...
+          end
+
+          def request_method(endpoint)
+            endpoint.options[:method]&.first
+          end
+
+          def path(endpoint)
+            namespace = endpoint.routes.first.namespace
+            version = endpoint.routes.first.options[:version] || ''
+            prefix = endpoint.routes.first.options[:prefix]&.to_s || ''
```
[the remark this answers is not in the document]
muripic: I don't think so, because if the endpoint does not match `endpoint_run` does not run at all. But just to be 100% safe (in the event of a bug in Grape or something like that) I added [this](https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/394/commits/5decb4598a2759e1334c7bfe531ef675f340088f#diff-413cd403b56aef0b2f6ecbbc3b77991071e8e77b62c0be98f6bc653e0b6a9356R93), wdyt?

instrumentation/grape/example/trace_demonstration.rb
```diff
@@ -0,0 +1,54 @@
...
+  get 'users/:id' do
+    { id: params[:id], name: 'John Doe', email: 'johndoe@example.com' }
+  end
+end
+
+# Set up fake Rack application
+builder = Rack::Builder.app do
+  # Integration is automatic in web frameworks but plain Rack applications require this line.
+  # Enable it in your config.ru.
+  use OpenTelemetry::Instrumentation::Rack::Middlewares::TracerMiddleware
```
arielvalentin, line 48: Does Grape have access to the Rack middleware in anyway? It would be really nice to have it inject the middleware automatically into the rack builder without having to have users require it themselves like we do in Sinatra and ActionPack

https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/main/instrumentation/sinatra/lib/opentelemetry/instrumentation/sinatra/extensions/tracer_extension.rb#L31

https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/main/instrumentation/action_pack/lib/opentelemetry/instrumentation/action_pack/railtie.rb#L15

arielvalentin (approved):

arielvalentin:
@muripic looks like you have a failing linter: https://github.com/open-telemetry/opentelemetry-ruby-contrib/actions/runs/4690521811/jobs/8314700821?pr=394#step:5:329

muripic:
> @muripic looks like you have a failing linter: https://github.com/open-telemetry/opentelemetry-ruby-contrib/actions/runs/4690521811/jobs/8314700821?pr=394#step:5:329

Thanks for pointing it out, fixed in https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/394/commits/10f809ef04cedf1b232f4dbd940828501d7170cb

mwear (approved):
LGTM. Thanks @muripic!

PR #406, opened 2023-04-07
feat: add semantic_conventions gem dependency

lloydpick:
This came up in #390...

> I ... think we might actually need to declare a dependency on the `semantic_conventions` gem? It's a dependency of the _SDK_, which is a development dependency for the gem - so tests pass. But I think if someone were to install the API and instrumentation (but _not_ the SDK, which is technically valid) it might break?
> 
> (If so, we should probably fix this elsewhere too 👀 )

I imagine the idea is that each individual instrumentation gem should be using semantic conventions where possible to keep things aligned with the spec.

How you want to depend on that gem (and keep it up to date) I've no idea. I'm suggesting here that the instrumentation_base gem gets a direct dependency on it rather than adding it to each individual gem. This should also help keep the gems sticking to the spec, and not lagging behind because they set a different version.

I also don't know how often this situation comes up, where someone isn't using the SDK, I suspect potentially never? But if the instrumentation gems use the semantic conventions, they need to have a dependency somewhere (either in base or individually) and not it coming through because you happen to always use another gem that does (SDK).

> Just with a quick look, I think the only instrumentation gem that will work without the SDK that is also using semantic names is the trilogy one. All the others would fail.


mwear (commented):
Thanks for submitting this @lloydpick. We discussed this at a recent SIG meeting, and the tl;dr is that we're really not sure how to best handle semantic conventions dependencies; it's kind of a can of worms, especially as we look into the future, and expected churn there. I wouldn't expect us to move forward with the PR in the short term, but we can keep you in the loop as plans evolve.

PR #412, opened 2023-04-11
chore: [sinatra] include a web server in example so it boots without changes

pollosp:
Hi, 
I always like examples working by default , so sending you easy "patch" for booting the example without modifying the example into the repo. 
 
 Solves:
 
 ```
 ./start_server.sh 
I, [2023-04-11T17:13:34.785064 #15]  INFO -- : Instrumentation: OpenTelemetry::Instrumentation::Sinatra was successfully installed with the following options {}
/usr/local/bundle/gems/rack-2.2.6.4/lib/rack/handler.rb:45:in `pick': Couldn't find handler for: thin, falcon, puma, HTTP, webrick. (LoadError)
        from /usr/local/bundle/gems/sinatra-3.0.6/lib/sinatra/base.rb:1526:in `run!'
        from ./server.rb:45:in `<class:App>'
        from ./server.rb:20:in `<main>'
root@2419e60e122c:/data# ls
 ```


robbkidd:
@pollosp Yep! This does solve that problem. Would you be willing to run through the CLA signature process so the project can accept your changes?

pollosp:
Thanks @robbkidd , looks I have issues with EasyCLA as I already signed the document. 
I have opened a ticket and I'm waiting for the response as will be nice to do more contributions in the future.

instrumentation/sinatra/example/Gemfile
```diff
@@ -5,7 +5,7 @@
 # SPDX-License-Identifier: Apache-2.0
 
 source 'https://rubygems.org'
-
+gem 'puma'
```
arielvalentin, line 8: Does webrick not work by default?

arielvalentin: Yeah I guess I am surprised that it does not work and it should probably use whatever is bundled in the Ruby stdlib, which is what I thought sinatra used by default? but I guess not. 

Puma should be fine.

arielvalentin (approved):

robbkidd (approved):
Nancy Sinatra approves of the boot.

robbkidd:
@pollosp thank you for making life a little easier for folks trying things out! 💞

PR #414, opened 2023-04-13
feat: Adding queue latency metrics for ActiveJob

jdlubrano:


instrumentation/active_job/lib/opentelemetry/instrumentation/active_job/patches/active_job_callbacks.rb
```diff
@@ -81,6 +86,10 @@ def job_attributes(job)
             otel_attributes.compact
           end
 
+          def milliseconds_since(timestamp)
```
jdlubrano, line 89: I'm a little concerned about "polluting" the ActiveJob method namespace with this method, so I'm open to alternatives.  It would also seem okay to me to inline this logic within the `job_attributes` method 🤷‍♂️

instrumentation/active_job/test/test_helper.rb
```diff
@@ -25,6 +25,15 @@ def perform
   end
 end
 
+class RetryWithWaitJob < ActiveJob::Base
+  retry_on StandardError, wait: 0, attempts: 2
+
+  def perform
+    sleep(0.01)
```
jdlubrano, line 32: I initially tried setting `wait` on this job, but then the retry attempt was not running in the tests.  I think that makes sense as the retry would go back into the queue.  I don't love throwing a `sleep` in here, and I am open to suggestions.  This seemed like the simplest way to simulate the passage of time, for the `first_enqueued_at` tests, though.

instrumentation/active_job/lib/opentelemetry/instrumentation/active_job/patches/active_job_callbacks.rb
```diff
@@ -72,6 +73,10 @@ def job_attributes(job)
               'messaging.destination' => job.queue_name,
               'messaging.message_id' => job.job_id,
               'messaging.active_job.provider_job_id' => job.provider_job_id,
+              'messaging.active_job.enqueued_at' => job.enqueued_at,
+              'messaging.active_job.enqueued_duration_ms' => milliseconds_since(job.enqueued_at),
+              'messaging.active_job.first_enqueued_at' => job.first_enqueued_at,
+              'messaging.active_job.enqueued_and_retry_duration_ms' => milliseconds_since(job.first_enqueued_at),
```
jdlubrano, line 79: I have **no** strong opinion on the naming of `enqueued_duration_ms` or `enqueued_and_retry_duration_ms`.  My other thought was `enqueued_wait_ms` and `enqueued_and_retry_wait_ms`.  I chose `duration` because OTel already captures a `duration_ms` metric for spans, so I opted for consistency with the naming here.

instrumentation/active_job/lib/opentelemetry/instrumentation/active_job/patches/active_job_callbacks.rb
```diff
@@ -72,6 +73,10 @@ def job_attributes(job)
               'messaging.destination' => job.queue_name,
               'messaging.message_id' => job.job_id,
               'messaging.active_job.provider_job_id' => job.provider_job_id,
+              'messaging.active_job.enqueued_at' => job.enqueued_at,
+              'messaging.active_job.enqueued_duration_ms' => milliseconds_since(job.enqueued_at),
```
jdlubrano, line 77: The time that the job spent enqueued before _this_ execution.

instrumentation/active_job/lib/opentelemetry/instrumentation/active_job/patches/active_job_callbacks.rb
jdlubrano, line 79: The time since the job was first enqueued.  This captures:
* Time spent enqueued
* Time spent waiting to be retried
* Duration of prior executions

arielvalentin (changes requested):
@jdlubrano Thank you for taking the time to contribute to OTel Ruby instrumentation.

We are not currently accepting instrumentation changes that adds metric data as span attributes. Normally I would recommend that we capture these as OTel Metrics but our implementation is incomplete at this time. 

Some options that I am able to recommend: 

- Represent latency metrics as spans or span events. It will be symmetric with how we represent this data with other instrumentations, e.g. Sidekiq[^1]. The challenge here I think is that there are little known semantics around representing intermediaries, e.g. retries or time spent in the queue https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/messaging.md
- Enrich or create custom spans: Trickier to do but in your `ApplicationJob` code perhaps you can grab the current span and add these attributes. The challenge I see here is that your custom implementation may not be compatible with new features your vendor may build around Intermediary message semconv and instrumentation. 

[^1]: https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/main/instrumentation/sidekiq/lib/opentelemetry/instrumentation/sidekiq/middlewares/server/tracer_middleware.rb#L44

jdlubrano:
Thanks for the suggestions @arielvalentin !  Closing this to pursue the options that you recommended within my own application 👍

arielvalentin:
Thank you for understanding. I really hope this does not do discourage you from submitting future contributions!

stevenharman:
Firstly, thank you, both, for this discussion. I'm once again on my semi-regular journey to figure out a semantic and consistent way to extract latency data from traces for background jobs/messaging. 😆

Regarding the first alternative proposed above: latency by way of Span Events. That is currently how Sidekiq works, but [it still might not be enough to be usable for SLOs, triggers, etc… depending on your OTel vendor](https://github.com/open-telemetry/opentelemetry-ruby/discussions/1030). I haven't checked in a while, but to my knowledge, Honeycomb, for example, cannot extract those event timestamps into something you can use to calculate latency (e.g., a derived column).

Given that, I'm now considering the second option you proposed, @arielvalentin, wherein we enrich the current span (in Sidekiq-land, via a Server Middleware) to add the attributes ourselves. I agree that a downside of this is drift between our custom attributes/code and potential features our vendor (i.e., Honeycomb) might build around the `messaging` Semantic Convention. But that seems a worthy tradeoff given that those features don't exist right now. If they ever do show up, we could adapt our custom instrumentation to support them, or remove our custom instrumentation in favor of whatever SemConv things happens.

The other downside, of course, is that we're all left to build this bespoke custom attributes machinery ourselves. Which might be preferable to an `opentelemetry-ruby-contrib-contrib` Gem. 😄

robbkidd:
I just fiddled with using OTel Collector Contrib (v0.83.0) to hoist times from these span events onto the parent span with the [transform processor](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/processor/transformprocessor). Maybe more experimentation can happen there to see what fits and to inform suggestions for semantic conventions around jobs and/or messaging.

It went something like this:

```yaml
# collector config snippet
processors:
  … 
  transform:
    error_mode: ignore
    trace_statements:
      - context: spanevent
        statements:
          - set(span.attributes["job.created_at"], time_unix_nano) where IsMatch(instrumentation_scope.name, "Sidekiq") and name == "created_at"
          - set(span.attributes["job.enqueued_at"], time_unix_nano) where IsMatch(instrumentation_scope.name, "Sidekiq") and name == "enqueued_at"
          - set(span.attributes["job.latency_ms"], ((span.start_time_unix_nano - time_unix_nano) / 1000000)) where IsMatch(instrumentation_scope.name, "Sidekiq") and name == "enqueued_at"
```

A Rails app with these bits inside:

```ruby
class HomeController < ApplicationController
  def home
    HardJob.perform_in(2.seconds) if params[:run_job] == 'true'
    …
  end
  … 
end

class HardJob
  include Sidekiq::Job

  def perform(*args); sleep(0.03) ; end
end
```

Run with ApacheBench to abuse it:

```shell
❯ ab -n 10000 -c 20 "http://localhost:3000/?run_job=true"
```

Screenshot 2023-08-24 at 6 40 05 PM

PR #415, opened 2023-04-16
chore: Fix broken graphql test

arielvalentin:
`BUNDLE_GEMFILE=gemfiles/graphql_2.x.gemfile  bundle exec rake test TESTOPTS="--seed=14209"`

e.g. https://github.com/open-telemetry/opentelemetry-ruby-contrib/actions/runs/4703630683/jobs/8342356744


robbkidd (approved):
It works for me with the error-producing seed and several dozen other random seeds.

For what it's worth, I removed the `after` block entirely just to see if resetting before _and_ after is necessary, and the same several dozen other random seeds also passed. Not sure the `after` is needed any more.

arielvalentin:
Thanks @robbkidd 

Are you saying the entire block can be removed?

robbkidd:
> Are you saying the entire block can be removed?

Yep. The test suite seemed to work with the `after` block removed entirely on the 50 or so runs I did without a fixed seed.

PR #428, opened 2023-04-17
feat: Use Rack Middleware Helper

arielvalentin:
The Rack instrumentation supports the ability to use the Rack::Events API for instrumentation.

This change makes updates instrumentations that depend on the Rack instrumentation to use the Events based instrumentation.

### DIFF
--- a/instrumentation/action_pack/lib/opentelemetry/instrumentation/action_pack/railtie.rb
+++ b/instrumentation/action_pack/lib/opentelemetry/instrumentation/action_pack/railtie.rb
@@ -14,7 +14,7 @@ module OpenTelemetry
 
           app.middleware.insert_before(
             0,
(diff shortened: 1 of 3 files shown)


mwear (approved):
LGTM

robbkidd (approved):
It works!

dazuma (changes requested):
As discussed in the SIG meeting, consider `feat:` as the conventional commit tag.

robbkidd:
Updated the convcommit preamble. With this change, a user can twiddle the Rack instrumentation config to use the Rack Events-based instrumentation, so new capabilities!

dazuma (approved):
👍🏼

arielvalentin:
@dazuma once this is merged would you mind releasing a new version of the gems that were impacted by this change? I believe that would be:

- `actionpack`
- `sinatra`
- `rails`
- `all`

PR #448, opened 2023-05-01
fix: Grape Instrumentation handle status code symbol

dhruvCW:
#### Context
ever since grape 1.7.0 grape endpoints can return status codes as rack status code symbols, 
these symbols are then converted into integers by `Grape::Middleware::Error`. 

In `event_handler` we make the assumption that exception status would be an integer, which since grape 1.7.0 is incorrect.

#### Changes:
we are converting symbols to integers using `Rack::Utils.status_code` just like grape [does](https://github.com/ruby-grape/grape/blob/master/lib/grape/middleware/error.rb#L75).


instrumentation/grape/lib/opentelemetry/instrumentation/grape/event_handler.rb
```diff
@@ -79,7 +81,7 @@ def handle_payload_exception(span, exception)
             # Only record exceptions if they were not raised (i.e. do not have a status code in Grape)
             # or do not have a 5xx status code. These exceptions are recorded by Rack.
             # See instrumentation/rack/lib/opentelemetry/instrumentation/rack/middlewares/tracer_middleware.rb#L155
-            return unless exception.respond_to?('status') && exception.status.to_i < 500
+            return unless exception.respond_to?('status') && Utils.status_code(exception.status) < 500
```
arielvalentin, line 84: Am I missing something? What is preventing you from using the `Rack::Utils.status_code` method here?

instrumentation/grape/lib/opentelemetry/instrumentation/grape/version.rb
```diff
@@ -8,7 +8,7 @@ module OpenTelemetry
   module Instrumentation
     module Grape
       # Current gem version
-      VERSION = '0.1.1'
+      VERSION = '0.1.2'
```
arielvalentin, line 11:
```suggestion
      VERSION = '0.1.1'
```

Our release process automatically bumps versions. Please revert this.

arielvalentin (changes requested):
Thanks for helping with Grape Instrumentation maintenance! 

I have some questions regarding this implementation and a request to revert the version bump. Once we have those resolved I will prepare a release ASAP.

instrumentation/grape/lib/opentelemetry/instrumentation/grape/event_handler.rb
dhruvCW: nothing, and I am happy to use `Rack::Utils` if it's okay to add `rack` as a dependency.

instrumentation/grape/lib/opentelemetry/instrumentation/grape/version.rb
dhruvCW: oh my bad, reverted the change ✅

instrumentation/grape/lib/opentelemetry/instrumentation/grape/event_handler.rb
arielvalentin: Yes please go ahead. 

Rack is a dependency of this instrumentation already, however in your gemspec please make sure to keep it as a `development` dependency.

arielvalentin:
Looks like there are failing tests: https://github.com/open-telemetry/opentelemetry-ruby-contrib/actions/runs/4852574508/jobs/8647749969?pr=448

robbkidd (approved):
grape heart

instrumentation/grape/lib/opentelemetry/instrumentation/grape/event_handler.rb
```diff
@@ -4,6 +4,8 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 
+require 'rack'
```
arielvalentin, line 7: hmm interesting. I think this needs to be a test dependency and not a production dependency if that makes sense.

If the app does not use Rack, then it should not load this file. 

That _should_ be guarded by this:

https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/main/instrumentation/grape/lib/opentelemetry/instrumentation/grape/instrumentation.rb#L22

Grape requires rack as a Runtime dependency so this should not be necessary: https://github.com/ruby-grape/grape/blob/master/grape.gemspec#L27

I am confused why the test suite is not requiring `rack` since it is loading `grape`.

dhruvCW: You're correct, I believe this is a redundant require statement. Let me create a follow-up PR to remove this. We can also safely not add rack as a development dependency

dhruvCW: as promised the followup PR https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/450

PR #450, opened 2023-05-01
fix: remove redundant require statement for 'rack' from grape instrumentation

dhruvCW:
and remove rack as a development dependency since we already have it as a transitive dependency from grape.


arielvalentin (approved):

robbkidd (approved):
Clean up.

Issue #475, opened 2023-05-18
Action View instrumentation fails if Rails is not defined


jmorrell:
**Description of the bug**

When using the `OpenTelemetry::Instrumentation::ActionView` instrumentation outside of a Rails app the instrumentation will fail with an error:

```
ERROR -- : OpenTelemetry error: Instrumentation: OpenTelemetry::Instrumentation::ActionView unhandled exception during install: 
[{stack trace ...}] - uninitialized constant Rails
```

The instrumentation does not check for `::Rails` when [testing for compatibility](https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/main/instrumentation/action_view/lib/opentelemetry/instrumentation/action_view/instrumentation.rb#L18)

```ruby
        present do
          defined?(::ActionView)
        end
```

but does require `::Rails::Railtie` when [installing the instrumentation](https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/main/instrumentation/action_view/lib/opentelemetry/instrumentation/action_view/railtie.rb#L17)

```ruby
      # This Railtie sets up subscriptions to relevant ActionView notifications
      class Railtie < ::Rails::Railtie
```

If the instrumentation cannot be made to work without depending on `Rails::Railtie` then it would be good to update the detection to fail to install in the case that `Rails` is not defined:

```
        compatible do
          gem_version >= MINIMUM_VERSION && defined?(::Rails)
        end
```

**Share details about your runtime**

Operating system details: Linux, Ubuntu 20.04 LTS
RUBY_ENGINE: "ruby"
RUBY_VERSION: "3.2.1"
RUBY_DESCRIPTION: "ruby 3.2.1 (2023-02-08 revision 31819e82c8) [arm64-darwin22]"

**Share a simplified reproduction if possible**

```rb
require "bundler/inline"

gemfile(true) do
  source "https://rubygems.org"

  gem "opentelemetry-api"
  gem "opentelemetry-sdk"
  gem 'opentelemetry-instrumentation-all'
  gem "actionview"
end

require "action_view"
require "opentelemetry-api"
require "opentelemetry-sdk"
require 'opentelemetry/instrumentation/all'

OpenTelemetry::SDK.configure do |c|
  c.service_name = 'Demo'
  c.use("OpenTelemetry::Instrumentation::ActionView", {})
end
```

arielvalentin:
Thanks for reporting this. Are you using actionview without Rails?

[...]

PR #476, opened 2023-05-23
Check for rails when determining whether ActionView instrumentation is compatible

jmorrell:
Fixes https://github.com/open-telemetry/opentelemetry-ruby-contrib/issues/475

When using the `OpenTelemetry::Instrumentation::ActionView` instrumentation outside of a Rails app the instrumentation will fail with an error:

```
ERROR -- : OpenTelemetry error: Instrumentation: OpenTelemetry::Instrumentation::ActionView unhandled exception during install: 
[{stack trace ...}] - uninitialized constant Rails
```

The instrumentation does not check for `::Rails` when [testing for compatibility](https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/main/instrumentation/action_view/lib/opentelemetry/instrumentation/action_view/instrumentation.rb#L18)

```ruby
        present do
          defined?(::ActionView)
        end
```

but does require `::Rails::Railtie` when [installing the instrumentation](https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/main/instrumentation/action_view/lib/opentelemetry/instrumentation/action_view/railtie.rb#L17)

```ruby
      # This Railtie sets up subscriptions to relevant ActionView notifications
      class Railtie < ::Rails::Railtie
```

I'm not sure how to instrument `ActionView` outside of `RailTie`, but at the least the instrumentation should fail it's compatibility check if `::Rails` is not defined.


jmorrell:
I'm working to fix the CLA issue ☝️

This change does not include tests because it was not clear how I should approach testing this. If the maintainers have a suggested approach I would be happy to implement it.

arielvalentin:
@jmorrell check for use of `hide_const` for examples of how we test it. 

https://github.com/search?q=repo%3Aopen-telemetry%2Fopentelemetry-ruby-contrib%20hide_const&type=code

instrumentation/action_view/lib/opentelemetry/instrumentation/action_view/instrumentation.rb
```diff
@@ -19,7 +19,7 @@ class Instrumentation < OpenTelemetry::Instrumentation::Base
         end
 
         compatible do
-          gem_version >= MINIMUM_VERSION
+          gem_version >= MINIMUM_VERSION && defined?(::Rails)
```
arielvalentin, line 22: Const checks should be done in the `present` block. 

The compatible block is an additional check to check for version compatibility.

PR #480, opened 2023-05-30
Provide support for rdkafka fork for Karafka

mensfeld:
[Karafka](karafka.io/) is a Ruby and Rails multi-threaded efficient Kafka processing framework built on top of rdkafka with its own rdkafka fork that is API compatible.

This PR aims to provide support for rdkafka telemetry when rdkafka comes from karafka-rdkafka.

Why fork? karafka-rdkafka receives patches and stability improvements much faster than rdkafka itself. The same applies to updates to the underlying librdkafka C code.

P.S. Karafka integration with opentelemetry is also something I am looking into but it is a separate issue.

This PR requires only change in the detection of source (rdkafka vs karafka-rdkafka), rest within the appropriate API scope works and will work exactly the same.

### DIFF
--- a/instrumentation/rdkafka/lib/opentelemetry/instrumentation/rdkafka/instrumentation.rb
+++ b/instrumentation/rdkafka/lib/opentelemetry/instrumentation/rdkafka/instrumentation.rb
@@ -11,8 +11,11 @@ module OpenTelemetry
       class Instrumentation < OpenTelemetry::Instrumentation::Base
         MINIMUM_VERSION = Gem::Version.new('0.10.0')
 
+        GEM_VARIANTS = %w[rdkafka karafka-rdkafka]
+        
         compatible do
-          Gem.loaded_specs['rdkafka'].version >= MINIMUM_VERSION
+          variant = GEM_VARIANTS.find { |variant| !Gem.loaded_specs[variant].nil? }
+          variant.version >= MINIMUM_VERSION
         end
 
         install do |_config|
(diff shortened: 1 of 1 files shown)


instrumentation/rdkafka/lib/opentelemetry/instrumentation/rdkafka/instrumentation.rb
```diff
@@ -11,8 +11,11 @@ module Rdkafka
       class Instrumentation < OpenTelemetry::Instrumentation::Base
         MINIMUM_VERSION = Gem::Version.new('0.10.0')
 
+        GEM_VARIANTS = %w[rdkafka karafka-rdkafka]
+        
         compatible do
-          Gem.loaded_specs['rdkafka'].version >= MINIMUM_VERSION
+          variant = GEM_VARIANTS.find { |variant| !Gem.loaded_specs[variant].nil? }
```
arielvalentin, line 17: Is there a `VERSION` constant you can use here instead?

mensfeld: You mean a direct reference via `Rdkafka::VERSION`?

arielvalentin: Yes please.

mensfeld:
@arielvalentin should I create a new separate namespace for karafka-rdkafka or just update the `Appraisals` and hope for the best? ;)

arielvalentin:
@mensfeld Since it is a fork of the same gem I think its OK to start with using the same appraisal file and adding a separate section for now. 

Once it grows to include support for multiple versions then I think we can do something a bit more sophisticated.

mensfeld:
Hey @arielvalentin I am going to close this PR. I think we're reaching a consensus with other forks and original gem in regards to future development and we will join forces. This means, that we would have to reverse this in the upcoming months (as would be come obsolete).

I am really sorry for time you spent on the review.

arielvalentin:
No apologies necessary! Thank you for taking the time to wanting to make this instrumentation better!

PR #482, opened 2023-05-30
fix: GraphQL tracing

### LINKED ISSUE #481: ArgumentError `GraphqlController#execute` wrong number of arguments (given 1, expected 0; required keyword: field) (ArgumentError)
```
ArgumentError: wrong number of arguments (given 1, expected 0; required keyword: field) (ArgumentError)
  from opentelemetry-instrumentation-graphql (0.26.0) lib/opentelemetry/instrumentation/graphql/tracers/graphql_trace.rb:141:in `platform_execute_field_key'
  from activesupport (7.0.4.3) lib/active_support/notifications.rb:206:in `instrument'
  from actionpack (7.0.4.3) lib/action_dispatch/middleware/stack.rb:60:in `call'
  from rack (2.2.7) lib/rack/deflater.rb:44:in `call'
  from actionpack (7.0.4.3) lib/action_dispatch/middleware/stack.rb:61:in `block in call'
```

Upgrading to 0.26 I get this error

swalkinshaw:
Fixes #481

The `legacy_tracing_requirement_satisfied?` logic was returning `true` for `2.0.22` which meant that the tests were _never_ being in non-legacy mode which was hiding many bugs.

This updates the logic to be strict on matching `2.0.19` which will properly detect `2.0.22` as a non-legacy version now.

### DIFF
--- a/instrumentation/graphql/lib/opentelemetry/instrumentation/graphql/instrumentation.rb
+++ b/instrumentation/graphql/lib/opentelemetry/instrumentation/graphql/instrumentation.rb
@@ -34,7 +34,7 @@ module OpenTelemetry
         end
 
         def legacy_tracing_requirement_satisfied?
-          Gem::Requirement.new('<= 2.0.17').satisfied_by?(gem_version) || Gem::Requirement.new('~> 2.0.19').satisfied_by?(gem_version)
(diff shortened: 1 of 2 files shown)


ericmustin (approved):
my bad should've caught this on review

arielvalentin:
Any way to add a test for this to ensure we catch a regression?

PR #487, opened 2023-05-31
feat!: Separate logical MySQL host from connected host

arielvalentin:
MySQL server instances have a logical host name and an instance node names.

This is particularly useful when running in clustered mode where a logical hostname is used to connect to a cluster but you also want to record what specific node is executing your query.

Trilogy exposes the instance node name via `@connected_host` as well as the configured `host` name in its `connection_options`.

This breaking change ensures that the `net.peer.name` is always set to the logical host name, while the individual instance name is being recorded in a separate attribute.

cc: @cschleiden

### DIFF
--- a/instrumentation/trilogy/lib/opentelemetry/instrumentation/trilogy/patches/client.rb
+++ b/instrumentation/trilogy/lib/opentelemetry/instrumentation/trilogy/patches/client.rb
@@ -48,11 +48,6 @@ module OpenTelemetry
 
           FULL_SQL_REGEXP = Regexp.union(MYSQL_COMPONENTS.map { |component| COMPONENTS_REGEX_MAP[component] })
 
-          def initialize(args)
-            @_otel_net_peer_name = args[:host]
-            super
-          end
-
           def query(sql)
             tracer.in_span(
               database_span_name(sql),
@@ -68,11 +63,12 @@ module OpenTelemetry
           def client_attributes(sql)
             attributes = {
               ::OpenTelemetry::SemanticConventions::Trace::DB_SYSTEM => 'mysql',
-              ::OpenTelemetry::SemanticConventions::Trace::NET_PEER_NAME => net_peer_name
+              ::OpenTelemetry::SemanticConventions::Trace::NET_PEER_NAME => connection_options.fetch(:host, 'unknown sock'),
             }
 
             attributes[::OpenTelemetry::SemanticConventions::Trace::DB_NAME] = database_name if database_name
             attributes[::OpenTelemetry::SemanticConventions::Trace::PEER_SERVICE] = config[:peer_service] unless config[:peer_service].nil?
+            attributes['db.mysql.instance.host.name'] = @connected_host if defined?(@connected_host)
 
             case config[:db_statement]
             when :obfuscate
@@ -129,16 +125,6 @@ module OpenTelemetry
 
           def database_name
             connection_options[:database]
-          end
-
-          def net_peer_name
-            if defined?(@connected_host)
-              @connected_host
-            elsif @_otel_net_peer_name
-              @_otel_net_peer_name
-            else
-              'unknown sock'
-            end
           end
 
           def tracer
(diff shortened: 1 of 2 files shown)


instrumentation/trilogy/lib/opentelemetry/instrumentation/trilogy/patches/client.rb
```diff
@@ -68,11 +63,12 @@ def query(sql)
           def client_attributes(sql)
             attributes = {
               ::OpenTelemetry::SemanticConventions::Trace::DB_SYSTEM => 'mysql',
-              ::OpenTelemetry::SemanticConventions::Trace::NET_PEER_NAME => net_peer_name
+              ::OpenTelemetry::SemanticConventions::Trace::NET_PEER_NAME => connection_options.fetch(:host, 'unknown sock'),
             }
 
             attributes[::OpenTelemetry::SemanticConventions::Trace::DB_NAME] = database_name if database_name
             attributes[::OpenTelemetry::SemanticConventions::Trace::PEER_SERVICE] = config[:peer_service] unless config[:peer_service].nil?
+            attributes['db.mysql.instance.host.name'] = @connected_host if defined?(@connected_host)
```
arielvalentin, line 71: I have no idea what this be named and I am struggling here. 

What do you think @adrianna-chang-shopify @cschleiden ?

arielvalentin: cc: @open-telemetry/specs-semconv-approvers I could use some of your input/feedback here as well.

cschleiden: I see https://github.com/open-telemetry/opentelemetry-ruby/blob/main/semantic_conventions/lib/opentelemetry/semantic_conventions/trace.rb#L71 

looking at https://github.com/open-telemetry/opentelemetry-specification/pull/3402, maybe
`db.mysql.instance.address` or `db.mysql.address`? But `db.mysql.instance.host.name` would also work for me

cschleiden (approved):
Thanks for this change!

instrumentation/trilogy/lib/opentelemetry/instrumentation/trilogy/patches/client.rb
arielvalentin: Nice find!

instrumentation/trilogy/lib/opentelemetry/instrumentation/trilogy/patches/client.rb
```diff
@@ -68,11 +63,12 @@ def query(sql)
           def client_attributes(sql)
             attributes = {
               ::OpenTelemetry::SemanticConventions::Trace::DB_SYSTEM => 'mysql',
-              ::OpenTelemetry::SemanticConventions::Trace::NET_PEER_NAME => net_peer_name
+              ::OpenTelemetry::SemanticConventions::Trace::NET_PEER_NAME => connection_options.fetch(:host, 'unknown sock'),
```
cschleiden, line 66:
```suggestion
              ::OpenTelemetry::SemanticConventions::Trace::NET_PEER_NAME => connection_options.fetch(:host, 'unknown sock')
```

to make the linter happy

instrumentation/trilogy/lib/opentelemetry/instrumentation/trilogy/patches/client.rb
arielvalentin: cc: @andyedison

ericmustin (approved):
lgtm, may wanna confirm with known users of the instrumentation (cc @adrianna-chang-shopify ) that this works for them.

ericmustin:
ah @arielvalentin looks like a couple typos, my apols on the early approval, that aside this lgtm

cschleiden (approved):

plantfansam (approved):
Love this! As a trilogy instrumentation user, we can handle the churn in attr name 😄.

PR #529, opened 2023-06-22
refactor!: Move shared sql behavior to helper gems

### LINKED ISSUE #32: SQL Instrumentation Improvements

kaylareopelle:
Mysql2, trilogy, and pg instrumentation contained duplicated constants and methods. This refactor creates two new gems, `opentelemetry-helpers-mysql` and `opentelemetry-helpers-sql-obfuscation`, to hold the shared code.

In addition, it also adds the ability to obfuscate SQLite, Cassandra, and Oracle databases, though there isn't any instrumentation available to automatically incorporate obfuscation for those database types.

It also improves MySQL statement extraction and SQL obfuscation behavior to better support comments prepended to SQL statements.

Resolves #32 

_Outdated content below this line_

----

> Questions for maintainers:
> * _Would you prefer the obfuscation and MySql refactors to be in their own commits/PRs?_ (I have a branch on standby with separate commits, so it's easy to pivot) => doesn't matter, going with one commit 
> * _What do you think about this pattern? These names? Is there a different design you'd like me to try?_ => new gems, named like `opentelemetry-helpers-<type>` in a directory at a peer level with `instrumentation`
> * _The public methods for the refactored instrumentation remained the same. Only constants and private methods were moved into helper modules. The instrumentation calling the modules continues to exercise the code. Would you like me to write tests for the new modules regardless?_ => This changed when we moved from modules to gems.  The gems have full test coverage for their methods.
> 
> 
> Remaining Work: 
> - [x] Write the readmes & rdocs
>     - [x] mysql
>     - [x] obfuscation
> - [X] Update CI actions to include the new gems 
> - [x] Write tests
>     - [x] mysql
>     - [x] obfuscation
> - [x] Resolve bundler error

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -10,6 +10,14 @@ updates:
   schedule:
     interval: weekly
 - package-ecosystem: bundler
+  directory: "/helpers/mysql"
+  schedule:
+    interval: weekly
+- package-ecosystem: bundler
+  directory: "/helpers/sql-obfuscation"
+  schedule:
+    interval: weekly
+- package-ecosystem: bundler
   directory: "/propagator/ottrace"
   schedule:
     interval: weekly
--- a/helpers/mysql/CHANGELOG.md
+++ b/helpers/mysql/CHANGELOG.md
@@ -1,0 +1,1 @@
+# Release History: opentelemetry-helpers-mysql
--- a/helpers/mysql/lib/opentelemetry/helpers/mysql.rb
+++ b/helpers/mysql/lib/opentelemetry/helpers/mysql.rb
@@ -1,0 +1,91 @@
+# frozen_string_literal: true
+
+# Copyright The OpenTelemetry Authors
+#
+# SPDX-License-Identifier: Apache-2.0module OpenTelemetry
+require 'opentelemetry-common'
+
+module OpenTelemetry
+  module Helpers
+    # This module contains helpers for MySQL instrumentation libraries, like mysql2 and
+    # trilogy. It is intended for use by instrumentation developers and not to
+    # be called directly by an application.
+    #
+    # To use this in your instrumentation, the `Instrumentation` class for
+    # your gem must contain configuration options for:
+    #  * `:span_name`
+    #  Example:
+    #  `option :span_name, default: :statement_type, validate: %I[statement_type db_name db_operation_and_name]`
+    #
+    # @api public
+    module MySQL
+      module_function
+
+      QUERY_NAMES = [
+        'set names',
+        'select',
+        'insert',
+        'update',
+        'delete',
+        'begin',
+        'commit',
+        'rollback',
+        'savepoint',
+        'release savepoint',
+        'explain',
+        'drop database',
+        'drop table',
+        'create database',
+        'create table'
+      ].freeze
+
+      # Ignore query names that might appear in comments prepended to the
+      # statement.
+      PREPENDED_COMMENTS_REGEX = %r{(?:\/\*.*?\*\/)}m
+      QUERY_NAME_REGEX = Regexp.new("^\s*(?:#{PREPENDED_COMMENTS_REGEX})?\s*\\b(#{QUERY_NAMES.join('|')})\\b.*", Regexp::IGNORECASE)
+
+      # This is a span naming utility intended for use in MySQL database
+      # adapter instrumentation.
+      #
+      # @param sql [String] The SQL statement for the span.
+      # @param operation [String] The database operation.
+      # @param database_name [String] The name of the database.
+      # @param config [Hash] The user's configuration for the database adapter.
+      #  Desired keys:
+      #    * `:span_name` => A symbol describing the type of name desired. Expected options are `:statement_type`, `:db_name`, and `:db_operation_and_name`. A nil or unknown `:span_name` will return 'mysql' as the span name
+      # @return [String] The span name.
+      # @api public
+      def database_span_name(sql, operation, database_name, config)
+        case config[:span_name]
+        when :statement_type
+          extract_statement_type(sql)
+        when :db_name
+          database_name
+        when :db_operation_and_name
+          db_operation_and_name(operation, database_name)
+        end || 'mysql'
+      end
+
+      # @api private
+      def extract_statement_type(sql)
+        sql = OpenTelemetry::Common::Utilities.utf8_encode(sql, binary: true)
+
+        QUERY_NAME_REGEX.match(sql) { |match| match[1].downcase } unless sql.nil?
+      rescue StandardError => e
+        OpenTelemetry.handle_error(message: 'Error extracting SQL statement type', exception: e)
+        nil
+      end
+
+      # @api private
+      def db_operation_and_name(operation, database_name)
+        if operation && database_name
+          "#{operation} #{database_name}"
+        elsif operation
+          operation
+        elsif database_name
+          database_name
+        end
+      end
+    end
+  end
+end
--- a/helpers/mysql/test/helpers/mysql_test.rb
+++ b/helpers/mysql/test/helpers/mysql_test.rb
@@ -1,0 +1,162 @@
+# frozen_string_literal: true
+
+# Copyright The OpenTelemetry Authors
+#
+# SPDX-License-Identifier: Apache-2.0
+
+require 'test_helper'
+
+describe OpenTelemetry::Helpers::MySQL do
+  describe '.database_span_name' do
+    let(:sql) { 'SELECT * FROM users' }
+    let(:operation) { 'operation' }
+    let(:database_name) { 'database_name' }
+    let(:config) { { span_name: span_name } }
+    let(:database_span_name) { OpenTelemetry::Helpers::MySQL.database_span_name(sql, operation, database_name, config) }
+
+    describe 'when config[:span_name] is :statement_type' do
+      let(:span_name) { :statement_type }
+
+      it 'returns the statement type' do
+        assert_equal(database_span_name, 'select')
+      end
+    end
+
+    describe 'when config[:span_name] is :db_name' do
+      let(:span_name) { :db_name }
+
+      it 'returns database name' do
+        assert_equal(database_span_name, database_name)
+      end
+    end
+
+    describe 'when config[:span_name] is :db_operation_and_name' do
+      let(:span_name) { :db_operation_and_name }
+
+      it 'returns db operation and name' do
+        assert_equal(database_span_name, 'operation database_name')
+      end
+    end
+
+    describe 'when config[:span_name] does not match a case' do
+      let(:span_name) { 'something_unexpected' }
+
+      it 'returns mysql' do
+        assert_equal(database_span_name, 'mysql')
+      end
+    end
+
+    describe 'when config[:span_name] is nil' do
+      let(:span_name) { nil }
+
+      it 'returns mysql' do
+        assert_equal(database_span_name, 'mysql')
+      end
+    end
+  end
+
+  describe '.db_operation_and_name' do
+    let(:operation) { 'operation' }
+    let(:database_name) { 'database_name' }
+    let(:db_operation_and_name) { OpenTelemetry::Helpers::MySQL.db_operation_and_name(operation, database_name) }
+
+    describe 'when operation and database_name are present' do
+      it 'returns a combination of the operation and database_name' do
+        assert_equal(db_operation_and_name, 'operation database_name')
+      end
+    end
+
+    describe 'when operation is nil' do
+      let(:operation) { nil }
+
+      it 'returns database_name' do
+        assert_equal(db_operation_and_name, database_name)
+      end
+    end
+
+    describe 'when database_name is nil' do
+      let(:database_name) { nil }
+
+      it 'returns the operation name' do
+        assert_equal(db_operation_and_name, operation)
+      end
+    end
+
+    describe 'when both database_name and operation are nil' do
+      let(:database_name) { nil }
+      let(:operation) { nil }
+
+      it 'returns nil' do
+        assert_nil(db_operation_and_name)
+      end
+    end
+  end
+
+  describe '.extract_statement_type' do
+    let(:sql) { 'SELECT * FROM users' }
+    let(:extract_statement_type) { OpenTelemetry::Helpers::MySQL.extract_statement_type(sql) }
+
+    describe 'when it finds a match' do
+      it 'returns the query name' do
+        assert_equal('select', extract_statement_type)
+      end
+    end
+
+    describe 'when sql contains invalid byte sequences' do
+      let(:sql) { "SELECT * from users where users.id = 1 and users.email = 'test@test.com\255'" }
+
+      it 'extracts the statement' do
+        assert_equal('select', extract_statement_type)
+      end
+    end
+
+    describe 'when sql contains unknown query statement' do
+      let(:sql) { 'DESELECT 1' }
+
+      # nil sets the span name to 'mysql'
+      it 'returns nil' do
+        assert_nil(extract_statement_type)
+      end
+    end
+
+    describe 'when sql contains multiple query statements' do
+      let(:sql) { 'EXPLAIN SELECT 1' }
+
+      it 'extracts the statement type that begins the query' do
+        assert_equal('explain', extract_statement_type)
+      end
+    end
+
+    describe 'when sql with marginalia-style prepended comments includes a query statement' do
+      let(:sql) do
+        "/*action='update',application='TrilogyTest',controller='users'*/ SELECT `users`.* FROM `users` WHERE `users`.`id` = 1 LIMIT 1"
+      end
+
+      it 'does not capture the query statement within the comment' do
+        assert_equal('select', extract_statement_type)
+      end
+    end
+
+    describe 'when sql is nil' do
+      let(:sql) { nil }
+
+      it 'returns nil' do
+        assert_nil(extract_statement_type)
+      end
+    end
+
+    describe 'when an error is raised' do
+      it 'logs a message' do
+        result = nil
+        OpenTelemetry::TestHelpers.with_test_logger do |log_stream|
+          OpenTelemetry::Common::Utilities.stub(:utf8_encode, ->(_) { raise 'boom!' }) do
+            extract_statement_type
+
+            assert_nil(result)
+            assert_match(/Error extracting/, log_stream.string)
+          end
+        end
+      end
+    end
+  end
+end
--- a/helpers/sql-obfuscation/CHANGELOG.md
+++ b/helpers/sql-obfuscation/CHANGELOG.md
@@ -1,0 +1,1 @@
+# Release History: opentelemetry-helpers-sql-obfuscation
--- a/helpers/sql-obfuscation/lib/opentelemetry/helpers.rb
+++ b/helpers/sql-obfuscation/lib/opentelemetry/helpers.rb
@@ -1,0 +1,14 @@
+# frozen_string_literal: true
+
+# Copyright The OpenTelemetry Authors
+#
+# SPDX-License-Identifier: Apache-2.0
+
+require 'opentelemetry/helpers/sql_obfuscation'
+
+module OpenTelemetry
+  # The helpers module contains functionality shared across multiple
+  # instrumentation libraries
+  module Helpers
+  end
+end
--- a/helpers/sql-obfuscation/lib/opentelemetry/helpers/sql_obfuscation.rb
+++ b/helpers/sql-obfuscation/lib/opentelemetry/helpers/sql_obfuscation.rb
@@ -1,0 +1,139 @@
+# frozen_string_literal: true
+
+# Copyright The OpenTelemetry Authors
+#
+# SPDX-License-Identifier: Apache-2.0module OpenTelemetry
+
+require 'opentelemetry-common'
+
+module OpenTelemetry
+  module Helpers
+    #
+    # This module contains SQL obfuscation behavior to share with
+    # instrumentation for specific database adapters.
+    # The class uses code from: https://github.com/newrelic/newrelic-ruby-agent/blob/1fca78cc7a087421ad58088d8bea72c0362bc62f/lib/new_relic/agent/database/obfuscation_helpers.rb
+    #
+    # To use this in your instrumentation, the `Instrumentation` class for
+    # your gem must contain configuration options for:
+    #  * `:db_statement`
+    #    Example:
+    #    `option :db_statement, default: :include, validate: %I[omit include obfuscate]`
+    #  * `:obfuscation_limit`
+    #    Example:
+    #    `option :obfuscation_limit, default: 2000, validate: :integer`
+    #
+    # If you want to add support for a new adapter, update the following
+    # constants to include keys for your adapter:
+    #  * DIALECT_COMPONENTS
+    #  * CLEANUP_REGEX
+    # You must also add a new constant that uses `generate_regex` with your
+    # adapter's dialect components that is named like
+    # `<ADAPTER>_COMPONENTS_REGEX`, such as: `MYSQL_COMPONENTS_REGEX`.
+    #
+    # @api public
+    module SqlObfuscation
+      module_function
+
+      # From: https://github.com/newrelic/newrelic-ruby-agent/blob/1fca78cc7a087421ad58088d8bea72c0362bc62f/lib/new_relic/agent/database/obfuscation_helpers.rb
+      COMPONENTS_REGEX_MAP = {
+        single_quotes: /'(?:[^']|'')*?(?:\\'.*|'(?!'))/,
+        double_quotes: /"(?:[^"]|"")*?(?:\\".*|"(?!"))/,
+        dollar_quotes: /(\$(?!\d)[^$]*?\$).*?(?:\1|$)/,
+        uuids: /\{?(?:[0-9a-fA-F]\-*){32}\}?/,
+        numeric_literals: /-?\b(?:[0-9]+\.)?[0-9]+([eE][+-]?[0-9]+)?\b/,
+        boolean_literals: /\b(?:true|false|null)\b/i,
+        hexadecimal_literals: /0x[0-9a-fA-F]+/,
+        comments: /(?:#|--).*?(?=\r|\n|$)/i,
+        multi_line_comments: %r{(?:\/\*.*?\*\/)}m,
+        oracle_quoted_strings: /q'\[.*?(?:\]'|$)|q'\{.*?(?:\}'|$)|q'\<.*?(?:\>'|$)|q'\(.*?(?:\)'|$)/
+      }.freeze
+
+      DIALECT_COMPONENTS = {
+        default: COMPONENTS_REGEX_MAP.keys,
+        mysql: %i[single_quotes double_quotes numeric_literals boolean_literals
+                  hexadecimal_literals comments multi_line_comments],
+        postgres: %i[single_quotes dollar_quotes uuids numeric_literals
+                     boolean_literals comments multi_line_comments],
+        sqlite: %i[single_quotes numeric_literals boolean_literals hexadecimal_literals
+                   comments multi_line_comments],
+        oracle: %i[single_quotes oracle_quoted_strings numeric_literals comments
+                   multi_line_comments],
+        cassandra: %i[single_quotes uuids numeric_literals boolean_literals
+                      hexadecimal_literals comments multi_line_comments]
+      }.freeze
+
+      PLACEHOLDER = '?'
+
+      # We use these to check whether the query contains any quote characters
+      # after obfuscation. If so, that's a good indication that the original
+      # query was malformed, and so our obfuscation can't reliably find
+      # literals. In such a case, we'll replace the entire query with a
+      # placeholder.
+      CLEANUP_REGEX = {
+        default: %r{'|"|\/\*|\*\/},
+        mysql: %r{'|"|\/\*|\*\//},
+        postgres: %r{'|\/\*|\*\/|\$(?!\?)/},
+        sqlite: %r{'|\/\*|\*\//},
+        cassandra: %r{'|\/\*|\*\//},
+        oracle: %r{'|\/\*|\*\//}
+      }.freeze
+
+      # @api private
+      def generate_regex(dialect)
+        components = DIALECT_COMPONENTS[dialect]
+        Regexp.union(components.map { |component| COMPONENTS_REGEX_MAP[component] })
+      end
+
+      DEFAULT_COMPONENTS_REGEX = generate_regex(:default)
+      MYSQL_COMPONENTS_REGEX = generate_regex(:mysql)
+      POSTGRES_COMPONENTS_REGEX = generate_regex(:postgres)
+      SQLITE_COMPONENTS_REGEX = generate_regex(:sqlite)
+      CASSANDRA_COMPONENTS_REGEX = generate_regex(:cassandra)
+      ORACLE_COMPONENTS_REGEX = generate_regex(:oracle)
+
+      # This is a SQL obfuscation utility intended for use in database adapter instrumentation.
+      #
+      # @param sql [String] The SQL to obfuscate.
+      # @param obfuscation_limit [optional Integer] The maximum length of an obfuscated sql statement.
+      # @param adapter [optional Symbol] the type of database adapter calling the method. `:default`, `:mysql` and `:postgres` are supported.
+      # @return [String] The SQL query string where the values are replaced with "?". When the sql statement exceeds the obufscation limit
+      #  the first matched pair from the SQL statement will be returned, with an appended truncation message. If trunaction is unsuccessful,
+      #  a string describing the error will be returned.
+      #
+      # @api public
+      def obfuscate_sql(sql, obfuscation_limit: 2000, adapter: :default)
+        regex = case adapter
+                when :mysql
+                  MYSQL_COMPONENTS_REGEX
+                when :postgres
+                  POSTGRES_COMPONENTS_REGEX
+                else
+                  DEFAULT_COMPONENTS_REGEX
+                end
+
+        # Original MySQL UTF-8 Encoding Fixes:
+        # https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/160
+        # https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/345
+        sql = OpenTelemetry::Common::Utilities.utf8_encode(sql, binary: true)
+        return truncate_statement(sql, regex, obfuscation_limit) if sql.size > obfuscation_limit
+
+        sql = sql.gsub(regex, PLACEHOLDER)
+        return 'Failed to obfuscate SQL query - quote characters remained after obfuscation' if CLEANUP_REGEX[adapter].match(sql)
+
+        sql
+      rescue StandardError => e
+        OpenTelemetry.handle_error(message: 'Failed to obfuscate SQL', exception: e)
+      end
+
+      # @api private
+      def truncate_statement(sql, regex, limit)
+        first_match_index = sql.index(regex)
+        truncation_message = "SQL truncated (> #{limit} characters)"
+        return truncation_message unless first_match_index
+
+        truncated_sql = sql[..first_match_index - 1]
+        "#{truncated_sql}...\n#{truncation_message}"
+      end
+    end
+  end
+end
--- a/helpers/sql-obfuscation/opentelemetry-helpers-sql-obfuscation.gemspec
+++ b/helpers/sql-obfuscation/opentelemetry-helpers-sql-obfuscation.gemspec
@@ -6,16 +6,16 @@
 
 lib = File.expand_path('lib', __dir__)
 $LOAD_PATH.unshift(lib) unless $LOAD_PATH.include?(lib)
-require 'opentelemetry/instrumentation/pg/version'
+require 'opentelemetry/helpers/sql_obfuscation/version'
 
 Gem::Specification.new do |spec|
-  spec.name        = 'opentelemetry-instrumentation-pg'
-  spec.version     = OpenTelemetry::Instrumentation::PG::VERSION
+  spec.name        = 'opentelemetry-helpers-sql-obfuscation'
+  spec.version     = OpenTelemetry::Helpers::SqlObfuscation::VERSION
   spec.authors     = ['OpenTelemetry Authors']
   spec.email       = ['cncf-opentelemetry-contributors@lists.cncf.io']
 
-  spec.summary     = 'PG (PostgreSQL) instrumentation for the OpenTelemetry framework'
-  spec.description = 'PG (PostgreSQL) instrumentation for the OpenTelemetry framework'
+  spec.summary     = 'SQL Obfuscation Instrumentation Helpers for the OpenTelemetry framework'
+  spec.description = 'SQL Obfuscation Instrumentation Helpers for the OpenTelemetry framework'
   spec.homepage    = 'https://github.com/open-telemetry/opentelemetry-ruby-contrib'
   spec.license     = 'Apache-2.0'
 
@@ -25,27 +25,20 @@ Gem::Specification.new do |spec|
   spec.require_paths = ['lib']
   spec.required_ruby_version = '>= 3.0'
 
-  spec.add_dependency 'opentelemetry-api', '~> 1.0'
-  spec.add_dependency 'opentelemetry-helpers-sql-obfuscation', '~> 0.1.0'
-  spec.add_dependency 'opentelemetry-instrumentation-base', '~> 0.22.1'
+  spec.add_dependency 'opentelemetry-common', '~> 0.20'
 
-  spec.add_development_dependency 'activerecord'
-  spec.add_development_dependency 'appraisal', '~> 2.5'
   spec.add_development_dependency 'bundler', '~> 2.4'
   spec.add_development_dependency 'minitest', '~> 5.0'
-  spec.add_development_dependency 'opentelemetry-sdk', '~> 1.1'
   spec.add_development_dependency 'opentelemetry-test-helpers', '~> 0.3'
-  spec.add_development_dependency 'pg', '>= 1.1.0'
-  spec.add_development_dependency 'pry'
-  spec.add_development_dependency 'pry-byebug' unless RUBY_ENGINE == 'jruby'
-  spec.add_development_dependency 'rubocop', '~> 1.60.1'
-  spec.add_development_dependency 'rubocop-performance', '~> 1.20'
-  spec.add_development_dependency 'simplecov', '~> 0.17.1'
+  spec.add_development_dependency 'rake', '~> 13.0'
+  spec.add_development_dependency 'rubocop', '~> 1.57.2'
+  spec.add_development_dependency 'rubocop-performance', '~> 1.19.1'
   spec.add_development_dependency 'yard', '~> 0.9'
+  spec.add_development_dependency 'yard-doctest', '~> 0.1.6'
 
   if spec.respond_to?(:metadata)
     spec.metadata['changelog_uri'] = "https://rubydoc.info/gems/#{spec.name}/#{spec.version}/file/CHANGELOG.md"
-    spec.metadata['source_code_uri'] = 'https://github.com/open-telemetry/opentelemetry-ruby-contrib/tree/main/instrumentation/pg'
+    spec.metadata['source_code_uri'] = 'https://github.com/open-telemetry/opentelemetry-ruby-contrib/tree/main/helpers/sql-obfuscation'
     spec.metadata['bug_tracker_uri'] = 'https://github.com/open-telemetry/opentelemetry-ruby-contrib/issues'
     spec.metadata['documentation_uri'] = "https://rubydoc.info/gems/#{spec.name}/#{spec.version}"
   end
--- a/helpers/sql-obfuscation/test/helpers/sql_obfuscation_test.rb
+++ b/helpers/sql-obfuscation/test/helpers/sql_obfuscation_test.rb
@@ -1,0 +1,108 @@
+# frozen_string_literal: true
+
+# Copyright The OpenTelemetry Authors
+#
+# SPDX-License-Identifier: Apache-2.0
+
+# This file is distributed under New Relic's license terms.
+# See https://github.com/newrelic/newrelic-ruby-agent/blob/main/LICENSE for complete details.
+
+require_relative '../test_helper'
+
+class SqlObfuscationTest < Minitest::Test
+  def test_named_arg_defaults_obfuscates
+    sql = "SELECT * from users where users.id = 1 and users.email = 'test@test.com'"
+    expected = 'SELECT * from users where users.id = ? and users.email = ?'
+    result = OpenTelemetry::Helpers::SqlObfuscation.obfuscate_sql(sql)
+
+    assert_equal(expected, result)
+  end
+
+  def test_obfuscation_limit_truncates_query_after_first_match
+    sql = "SELECT * from users where users.id = 1 and users.email = 'test@test.com'"
+    expected = "SELECT * from users where users.id = ...\nSQL truncated (> 42 characters)"
+    result = OpenTelemetry::Helpers::SqlObfuscation.obfuscate_sql(sql, obfuscation_limit: 42)
+
+    assert_equal(expected, result)
+  end
+
+  def test_obfuscation_limit_truncates_when_query_not_encoded_with_utf8
+    sql = "SELECT * from 😄 where users.id = 1 and users.😄 = 'test@test.com'"
+    expected = "SELECT * from  where users.id = ...\nSQL truncated (> 42 characters)"
+    result = OpenTelemetry::Helpers::SqlObfuscation.obfuscate_sql(sql, obfuscation_limit: 42)
+
+    assert_equal(expected, result)
+  end
+
+  def test_non_utf_8_encoded_string_obfuscates_with_mysql
+    sql = "SELECT * from users where users.id = 1 and users.email = 'test@test.com\255'"
+    expected = 'SELECT * from users where users.id = ? and users.email = ?'
+    result = OpenTelemetry::Helpers::SqlObfuscation.obfuscate_sql(sql, adapter: :mysql)
+
+    assert_equal(expected, result)
+  end
+
+  def test_non_utf_8_encoded_string_obfuscates_with_postgres
+    sql = "SELECT * from users where users.id = 1 and users.email = 'test@test.com\255'"
+    expected = 'SELECT * from users where users.id = ? and users.email = ?'
+    result = OpenTelemetry::Helpers::SqlObfuscation.obfuscate_sql(sql, adapter: :postgres)
+
+    assert_equal(expected, result)
+  end
+
+  def test_statement_with_emoji_encodes_utf_8_and_obfuscates
+    sql = "SELECT * from users where users.id = 1 and users.email = 'test@😄.com'"
+    expected = 'SELECT * from users where users.id = ? and users.email = ?'
+    result = OpenTelemetry::Helpers::SqlObfuscation.obfuscate_sql(sql)
+
+    assert_equal(expected, result)
+  end
+
+  # The following tests and their corresponding fixture are based on code from
+  # the New Relic Ruby agent.
+  # source: https://github.com/newrelic/newrelic-ruby-agent/blob/cb72bb5fab3fb318613421c86863a5ccdd2ff250/test/new_relic/agent/database/sql_obfuscation_test.rb
+
+  FAILED_TO_OBFUSCATE_MESSAGE = 'Failed to obfuscate SQL query - quote characters remained after obfuscation'
+
+  def build_failure_message(statement, dialect, acceptable_outputs, actual_output)
+    msg = +"Failed to obfuscate #{dialect} query correctly.\n"
+    msg << "Input:    #{statement}\n"
+    if acceptable_outputs.size == 1
+      msg << "Expected: #{acceptable_outputs.first}\n"
+    else
+      msg << "Acceptable outputs:\n"
+      acceptable_outputs.each do |output|
+        msg << "          #{output}\n"
+      end
+    end
+    msg << "Actual:   #{actual_output}\n"
+    msg
+  end
+
+  def self.load_fixture
+    data = File.read("#{Dir.pwd}/test/fixtures/sql_obfuscation.json")
+    JSON.parse(data)
+  end
+
+  load_fixture.each do |test_case|
+    name = test_case['name']
+    query = test_case['sql']
+    acceptable_outputs = test_case['obfuscated']
+    dialects = test_case['dialects']
+
+    # If the entire query is obfuscated because it's malformed, we use a
+    # placeholder message instead of just '?', so add that to the acceptable
+    # outputs.
+    acceptable_outputs << FAILED_TO_OBFUSCATE_MESSAGE if test_case['malformed']
+
+    dialects.each do |dialect|
+      define_method("test_sql_obfuscation_#{name}_#{dialect}") do
+        actual_obfuscated = OpenTelemetry::Helpers::SqlObfuscation.obfuscate_sql(query, adapter: dialect.to_sym)
+        message = build_failure_message(query, dialect, acceptable_outputs, actual_obfuscated)
+
+        assert_includes(acceptable_outputs, actual_obfuscated, message)
+      end
+    end
+  end
+  ## End New Relic tests
+end
--- a/instrumentation/mysql2/lib/opentelemetry/instrumentation/mysql2/patches/client.rb
+++ b/instrumentation/mysql2/lib/opentelemetry/instrumentation/mysql2/patches/client.rb
@@ -9,58 +9,27 @@ module OpenTelemetry
     module Mysql2
       module Patches
         # Module to prepend to Mysql2::Client for instrumentation
-        module Client # rubocop:disable Metrics/ModuleLength
-          QUERY_NAMES = [
-            'set names',
-            'select',
-            'insert',
-            'update',
-            'delete',
-            'begin',
-            'commit',
-            'rollback',
-            'savepoint',
-            'release savepoint',
-            'explain',
-            'drop database',
-            'drop table',
-            'create database',
-            'create table'
-          ].freeze
-
-          QUERY_NAME_RE = Regexp.new("^(#{QUERY_NAMES.join('|')})", Regexp::IGNORECASE)
-
-          # From: https://github.com/newrelic/newrelic-ruby-agent/blob/0235b288d85b8bc795bdc1a24621dd9f84cfef45/lib/new_relic/agent/database/obfuscation_helpers.rb#L9-L34
-          COMPONENTS_REGEX_MAP = {
-            single_quotes: /'(?:[^']|'')*?(?:\\'.*|'(?!'))/,
-            double_quotes: /"(?:[^"]|"")*?(?:\\".*|"(?!"))/,
-            numeric_literals: /-?\b(?:[0-9]+\.)?[0-9]+([eE][+-]?[0-9]+)?\b/,
-            boolean_literals: /\b(?:true|false|null)\b/i,
-            hexadecimal_literals: /0x[0-9a-fA-F]+/,
-            comments: /(?:#|--).*?(?=\r|\n|$)/i,
-            multi_line_comments: %r{\/\*(?:[^\/]|\/[^*])*?(?:\*\/|\/\*.*)}
-          }.freeze
-
-          MYSQL_COMPONENTS = %i[
-            single_quotes
-            double_quotes
-            numeric_literals
-            boolean_literals
-            hexadecimal_literals
-            comments
-            multi_line_comments
-          ].freeze
-
+        module Client
           def query(sql, options = {})
             attributes = client_attributes
             case config[:db_statement]
             when :include
               attributes[SemanticConventions::Trace::DB_STATEMENT] = sql
             when :obfuscate
-              attributes[SemanticConventions::Trace::DB_STATEMENT] = obfuscate_sql(sql)
+              attributes[SemanticConventions::Trace::DB_STATEMENT] =
+                OpenTelemetry::Helpers::SqlObfuscation.obfuscate_sql(
+                  sql, obfuscation_limit: config[:obfuscation_limit], adapter: :mysql
+                )
             end
             tracer.in_span(
-              database_span_name(sql),
+              OpenTelemetry::Helpers::MySQL.database_span_name(
+                sql,
+                OpenTelemetry::Instrumentation::Mysql2.attributes[
+                  SemanticConventions::Trace::DB_OPERATION
+                ],
+                database_name,
+                config
+              ),
               attributes: attributes.merge!(OpenTelemetry::Instrumentation::Mysql2.attributes),
               kind: :client
             ) do
@@ -69,57 +38,6 @@ module OpenTelemetry
           end
 
           private
-
-          def obfuscate_sql(sql)
-            if sql.size > config[:obfuscation_limit]
-              first_match_index = sql.index(generated_mysql_regex)
-              truncation_message = "SQL truncated (> #{config[:obfuscation_limit]} characters)"
-              return truncation_message unless first_match_index
-
-              truncated_sql = sql[..first_match_index - 1]
-              "#{truncated_sql}...\n#{truncation_message}"
-            else
-              obfuscated = OpenTelemetry::Common::Utilities.utf8_encode(sql, binary: true)
-              obfuscated = obfuscated.gsub(generated_mysql_regex, '?')
-              obfuscated = 'Failed to obfuscate SQL query - quote characters remained after obfuscation' if detect_unmatched_pairs(obfuscated)
-              obfuscated
-            end
-          rescue StandardError => e
-            OpenTelemetry.handle_error(message: 'Failed to obfuscate SQL', exception: e)
-            'OpenTelemetry error: failed to obfuscate sql'
-          end
-
-          def generated_mysql_regex
-            @generated_mysql_regex ||= Regexp.union(MYSQL_COMPONENTS.map { |component| COMPONENTS_REGEX_MAP[component] })
-          end
-
-          def detect_unmatched_pairs(obfuscated)
-            # We use this to check whether the query contains any quote characters
-            # after obfuscation. If so, that's a good indication that the original
-            # query was malformed, and so our obfuscation can't reliably find
-            # literals. In such a case, we'll replace the entire query with a
-            # placeholder.
-            %r{'|"|\/\*|\*\/}.match(obfuscated)
-          end
-
-          def database_span_name(sql)
-            case config[:span_name]
-            when :statement_type
-              extract_statement_type(sql)
-            when :db_name
-              database_name
-            when :db_operation_and_name
-              op = OpenTelemetry::Instrumentation::Mysql2.attributes[SemanticConventions::Trace::DB_OPERATION]
-              name = database_name
-              if op && name
-                "#{op} #{name}"
-              elsif op
-                op
-              elsif name
-                name
-              end
-            end || 'mysql'
-          end
 
           def database_name
             # https://github.com/brianmario/mysql2/blob/ca08712c6c8ea672df658bb25b931fea22555f27/lib/mysql2/client.rb#L78
@@ -149,13 +67,6 @@ module OpenTelemetry
 
           def config
             Mysql2::Instrumentation.instance.config
-          end
-
-          def extract_statement_type(sql)
-            QUERY_NAME_RE.match(sql) { |match| match[1].downcase } unless sql.nil?
-          rescue StandardError => e
-            OpenTelemetry.logger.debug("Error extracting sql statement type: #{e.message}")
-            nil
           end
         end
       end
--- a/instrumentation/trilogy/opentelemetry-instrumentation-trilogy.gemspec
+++ b/instrumentation/trilogy/opentelemetry-instrumentation-trilogy.gemspec
@@ -26,6 +26,8 @@ Gem::Specification.new do |spec|
   spec.required_ruby_version = '>= 3.0'
 
   spec.add_dependency 'opentelemetry-api', '~> 1.0'
+  spec.add_dependency 'opentelemetry-helpers-mysql', '~> 0.1.0'
+  spec.add_dependency 'opentelemetry-helpers-sql-obfuscation', '~> 0.1.0'
   spec.add_dependency 'opentelemetry-instrumentation-base', '~> 0.22.1'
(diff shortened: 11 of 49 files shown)


ericmustin:
> Would you prefer the obfuscation and MySql refactors to be in their own commits/PRs? (I have a branch on standby with separate commits, so it's easy to pivot)

I don't really care whatever is easier is fine, so happy to leave as is.

> What do you think about this pattern? These names? Is there a different design you'd like me to try?

Setting aside the question of where this stuff lives, (_i think *Base* is fine practically speaking but if ppl have opinions i probably wont die on this hill_), few questions i have.

1. it looks like PG obfuscation hasn't been abstracted yet, is that just due to this being WIP(_ie you're planning to do that or that you're waiting for a resolution on your first Q_), or is there a reason it's exempted?

2. we have a bit of loose coupling of configuration options and these helper methods. For example, we refer to   `config[:obfuscation_limit]` within the helper methods and treat it as an integer, and within the specific instrumentation's we ensure to validate these config options (see example: https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/4f4a6b0bc2073ad30744dc02c94b90fc1d8ad8e7/instrumentation/pg/lib/opentelemetry/instrumentation/pg/instrumentation.rb#L29), but because one portion of this logic lives in the instrumentation and the other in the base helper methods, it's possible for these to drift, or in the case of a new config option being added or a new db instrumentation being added which leverages these helpers, it's possible for the config option to be missed. 

    I don't know if there's an elegant way to handle this really, or even whether we should try to abstract out these config options into a universal helper _place_ (_it's entirely possible that there would be valid reasons we may want to enforce different defaults for different db libs, for example_), and frankly it feels like something we can solve by just being thoughtful maintainers and ensuring we keep the config options in sync when accepting prs/changes/etc, but just bringing it up, non-blocking imo but open to hearing what others think or if we can add some defensiveness here to help _future us_ cc @arielvalentin @ahayworth .

3. if we add some defensiveness here wrt my comment above, some tests might be nice for that, but beside that I don't really think it matters, practically speaking it would be hard to merge a change without exercising the full `-contrib`'s instrumentation test suite so the failure's will still get picked up _eventually_ even if we only make a change to _base_, so not too worried.

kaylareopelle:
> it looks like PG obfuscation hasn't been abstracted yet, is that just due to this being WIP(ie you're planning to do that or that you're waiting for a resolution on your first Q), or is there a reason it's exempted?

Thanks for catching this! This was just an oversight on my part. I was so focused on the identical code in mysql2/trilogy that I didn't realize the code for PG, though structured differently, behaves the same.

kaylareopelle:
As far as configs go, I think I'd prefer to keep them defined in the instrumentation library rather than a helper, even though we'll need to make sure they're there for the methods to work. 

We could try to make a new class for the SQL instrumentation to inherit from and raise an error if the config, or a method that accesses a memoized config, isn't defined. If we go that approach, we might want to change the current refactor a bit.

instrumentation/base/lib/opentelemetry/instrumentation/helpers/obfuscation_helper.rb
```diff
@@ -0,0 +1,53 @@
...
+
+            truncated_sql = sql[..first_match_index - 1]
+            return "#{truncated_sql}...\n#{truncation_message}"
+          end
+
+          # From:
+          # https://github.com/newrelic/newrelic-ruby-agent/blob/9787095d4b5b2d8fcaf2fdbd964ed07c731a8b6b/lib/new_relic/agent/database/obfuscator.rb
+          # https://github.com/newrelic/newrelic-ruby-agent/blob/9787095d4b5b2d8fcaf2fdbd964ed07c731a8b6b/lib/new_relic/agent/database/obfuscation_helpers.rb
+          # PG UTF-8 checks need more work
+          obfuscated = defined?(::PG) ? sql : OpenTelemetry::Common::Utilities.utf8_encode(sql, binary: true)
```
kaylareopelle, line 41: The obfuscation instrumentation is the same for all instrumented database adapters with the exception of this UTF-8 encoding case. I spent some time trying to apply the MySQL adapter code to PG, but couldn't get the tests to pass, so I went with a conditional instead.

kaylareopelle:
@ericmustin - This is ready for another look! I'll clean up the commits after feedback. I didn't make any changes to the configs, but I would be happy to. Maybe we can chat during the SIG tomorrow?

instrumentation/mysql2/lib/opentelemetry/instrumentation/mysql2/patches/client.rb
```diff
@@ -9,47 +9,8 @@ module Instrumentation
...
-            single_quotes
-            double_quotes
-            numeric_literals
-            boolean_literals
-            hexadecimal_literals
-            comments
-            multi_line_comments
-          ].freeze
+        module Client
+          include OpenTelemetry::Instrumentation::Helpers::MysqlHelper
```
robertlaurin, line 13: I'm reluctant to use includes on classes that are patching other classes. We have a helper including a helper which in turn gets included here. My concern is the easy of potentially having any of those modules have method names that collide with the class being patched, maybe not initially with this PR, but as time progresses. 

Instead I'd be more inclined for us to require the helper and explicitly reference the methods like `OpenTelemetry::Instrumentation::Helpers::MysqlHelper. generated_regex`, it is very wordy, but it's perfectly clear where the code is coming from.

robertlaurin (commented):
+1 to extracting this code out to a common place, we've already ran into one instance where parity between trilogy and mysql2 fell out of sync because a change occurred in one place but not the other.

I'm not sure base is the correct location for it, but I'm also not convinced it is the wrong place either.

My general concern is that whenever base bumps it forces all other deps to bump with it. It is toily for releasing, but also for bumping a single instrumentation package as a consumer of these gems. 

A slightly obnoxious solution might be gems for each domain? Mysql helpers, HTTP helpers, something_else helpers.

instrumentation/base/lib/opentelemetry/instrumentation/helpers/obfuscation_helper.rb
ericmustin: lgtm

fbogsany:
> My general concern is that whenever base bumps it forces all other deps to bump with it. It is toily for releasing, but also for bumping a single instrumentation package as a consumer of these gems.
>
> A slightly obnoxious solution might be gems for each domain? Mysql helpers, HTTP helpers, something_else helpers.

Agreed. `-base` is the wrong place for these helpers. At a minimum, I'd like a `-helpers` gem (or `-common`, which is what we ended up with in the "core" otel-ruby). Factoring them out further is probably too much to ask.

instrumentation/pg/opentelemetry-instrumentation-pg.gemspec
```diff
@@ -26,7 +26,16 @@ Gem::Specification.new do |spec|
   spec.required_ruby_version = '>= 3.0'
 
   spec.add_dependency 'opentelemetry-api', '~> 1.0'
-  spec.add_dependency 'opentelemetry-instrumentation-base', '~> 0.22.1'
+  # TODO: Return base version constraint. When constraint present, following error is raised:
```
ericmustin, line 29: This TODO should probably be resolved along with those in the trilogy and mysql2 instrumentations, seems bad to drop the version constraint. I'm not sure I understand the bundler error though?

instrumentation/helpers-sql-obfuscation/lib/opentelemetry/instrumentation/helpers/sql_obfuscation.rb
```diff
@@ -0,0 +1,150 @@
...
+          dollar_quotes: /(\$(?!\d)[^$]*?\$).*?(?:\1|$)/,
+          uuids: /\{?(?:[0-9a-fA-F]\-*){32}\}?/,
+          numeric_literals: /-?\b(?:[0-9]+\.)?[0-9]+([eE][+-]?[0-9]+)?\b/,
+          boolean_literals: /\b(?:true|false|null)\b/i,
+          hexadecimal_literals: /0x[0-9a-fA-F]+/,
+          comments: /(?:#|--).*?(?=\r|\n|$)/i,
+          multi_line_comments: %r{\/\*(?:[^\/]|\/[^*])*?(?:\*\/|\/\*.*)}
+        }.freeze
+
+        DIALECT_COMPONENTS = {
```
ericmustin, line 47: the more i look at this code, the more i continue to think this represents what really ought to be a Specification effort to standardize OpenTelemetry's treatment of "sql obfuscation" as a feature. Common dialect component keys ought to be part of otel semconv, and the implementation of how to accomplish that obfuscation could vary under the hood (regex might not be the most performant approach in some languages, etc), but it would be possible to do a matrix-like comparison of language-sdk's + collector's "obfuscation abilities" on sql(or other _things_, mongodb redis etc), and the varying concerns wrt PII related compliance things that different end users of different language sdks should need to worry about, and possibly, later on in their telemetry pipeline, apply some additional redaction or more fine grained ingestion or indexing approach to their data.

not blocking or anything, just was thinking about this and wanted to socialize it.

instrumentation/helpers-sql-obfuscation/lib/opentelemetry/instrumentation/helpers/sql_obfuscation.rb
```diff
@@ -0,0 +1,150 @@
...
+        #   Please include these keys in your instrumentation's configuration.
+        # @param adapter [Symbol] the type of database adapter calling the method. `:mysql` and `:postgres` are supported.
+        # @return [String] If successful, the SQL query string where the values are replaced with "?". If unsuccessful,
+        #  a string with a failure message.
+        #
+        # @api public
+        def self.obfuscate_sql(sql, config, adapter)
+          return sql unless config[:db_statement] == :obfuscate
+
+          regex = const_get("#{adapter.upcase}_COMPONENTS_REGEX")
```
ericmustin, line 108: seems like we've sacrificed some performance on the hot path for the sake of future extensibility, if i'm understanding this code correctly? Haven't benchmarked any of this ofc and it's possible i'm wrong bc i ruby bad, but at a glance we seem to be passing in a symbol, then massaging it to an upcased string, whereas a hard coded case statement could avoid this. Do you see a drawback to that approach?

instrumentation/pg/opentelemetry-instrumentation-pg.gemspec
kaylareopelle: I totally agree with you. I don't want to drop the version constraint and won't consider this done until I can find a fix for this issue. My hope was that this had been seen before in the repo and someone would know the right fix, but it sounds like that might not be the case! I might need to go back to the drawing board on how to create the new apps to see if I can stop it from happening.

instrumentation/helpers-sql-obfuscation/lib/opentelemetry/instrumentation/helpers/sql_obfuscation.rb
kaylareopelle: I need to run perf tests on this too. IIRC, case statements are very slow so that's why I didn't go straight to that approach. I'll run some tests and add them to this thread.

instrumentation/helpers-sql-obfuscation/lib/opentelemetry/instrumentation/helpers/sql_obfuscation.rb
```diff
@@ -0,0 +1,150 @@
...
+          uuids: /\{?(?:[0-9a-fA-F]\-*){32}\}?/,
+          numeric_literals: /-?\b(?:[0-9]+\.)?[0-9]+([eE][+-]?[0-9]+)?\b/,
+          boolean_literals: /\b(?:true|false|null)\b/i,
+          hexadecimal_literals: /0x[0-9a-fA-F]+/,
+          comments: /(?:#|--).*?(?=\r|\n|$)/i,
+          multi_line_comments: %r{\/\*(?:[^\/]|\/[^*])*?(?:\*\/|\/\*.*)}
+        }.freeze
+
+        DIALECT_COMPONENTS = {
+          fallback: COMPONENTS_REGEX_MAP.keys,
```
ericmustin, line 48: Is the idea that this would be something a _future_ instrumentation could refer to as a sort've "catchall" or "default" adapter symbol without needing to update the sql_obfuscation helper gem? I'm fine with that as an idea, but I would maybe want to call it `all` or `default`, and then perhaps make `self.obfuscate_sql` use that as it's default value unless specified. wdyt? Also fine with just dropping this if that's too convoluted.

instrumentation/helpers-sql-obfuscation/lib/opentelemetry/instrumentation/helpers/sql_obfuscation.rb
```diff
@@ -0,0 +1,150 @@
...
+
+          # Original MySQL UTF-8 Encoding Fixes:
+          # https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/160
+          # https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/345
+          OpenTelemetry::Common::Utilities.utf8_encode(sql, binary: true)
+        end
+
+        # @api private
+        def self.detect_unmatched_pairs(obfuscated, adapter)
+          CLEANUP_REGEX[adapter].match(obfuscated)
```
ericmustin, line 145: any decisions wrt `fallback` as a key on `DIALECT_COMPONENTS` should be mimic'd on `CLEANUP_REGEX`. Rn if an instrumentation passes in `fallback` as a value this will throw an error

instrumentation/helpers-sql-obfuscation/lib/opentelemetry/instrumentation/helpers/sql_obfuscation.rb
kaylareopelle: Yes, that's the idea. I love the name change suggestion (`fallback` is the NR name) and using this as the default value.

Do you think I should add a default value for the config hash? Then someone could just pass SQL to the API, though their mileage may vary.

instrumentation/trilogy/opentelemetry-instrumentation-trilogy.gemspec
```diff
@@ -26,7 +26,16 @@ Gem::Specification.new do |spec|
...
+  # Commented out because the following error is thrown while bundling:
+  # Because opentelemetry-instrumentation-base >= 0.22.1, < 0.23.A could not be found in source at `../base`
+  #   and every version of opentelemetry-instrumentation-trilogy depends on opentelemetry-instrumentation-base ~> 0.22.1,
+  #   opentelemetry-instrumentation-trilogy cannot be used.
+  # So, because Gemfile depends on opentelemetry-instrumentation-trilogy >= 0,
+  #   version solving has failed.
+  spec.add_dependency 'opentelemetry-instrumentation-base' # , '~> 0.22.1'
+  spec.add_dependency 'opentelemetry-instrumentation-helpers-mysql'
+  spec.add_dependency 'opentelemetry-instrumentation-helpers-sql-obfuscation'
   spec.add_dependency 'opentelemetry-semantic_conventions', '>= 1.8.0'
```
kaylareopelle, line 39: @ericmustin  - Do you know why Trilogy requires semantic_conventions, but MySql2 doesn't? As far as I can tell, the `semantic_conventions` gem isn't being used in Trilogy. However, I'm still pretty new to the way OTel organizes things and could be missing something.

ericmustin (changes requested):
Sorry for the delay here @kaylareopelle . I had a few nits but largely speaking this looks really close.

The outstanding issues, unless i'm missing something:
- bundler / version constraint TODOs, I don't totally grok the error you're seeing but afaik we shouldn't need to drop the `opentelemetry-instrumentation-base` version constraints on mysql2/trilogy/pg. can try to dig in more here if it ends up being the only thing blocking.
- had a few suggestions to make sure we dont have a performance regression on the obfuscation hot path. It might be good to get another set of eyes here too.

instrumentation/helpers-mysql/lib/opentelemetry/instrumentation/version.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+# Copyright The OpenTelemetry Authors
+#
+# SPDX-License-Identifier: Apache-2.0
+
+module OpenTelemetry
+  module Instrumentation
+    VERSION = '0.0.1'
```
kaylareopelle, line 9: @ericmustin  - What version should these gems start at? When I skimmed through some of the other libraries on RubyGems, I noticed there isn't a consistent starting value. `0.1.0` feels maybe more appropriate than `0.0.1`. Thoughts?

instrumentation/helpers-mysql/lib/opentelemetry/instrumentation/helpers/mysql.rb
```diff
@@ -0,0 +1,41 @@
...
+
+# Copyright The OpenTelemetry Authors
+#
+# SPDX-License-Identifier: Apache-2.0module OpenTelemetry
+
+module OpenTelemetry
+  module Instrumentation
+    module Helpers
+      # Helpers for MySQL instrumentation libraries, like mysql2 and trilogy
+      class MySQL
```
kaylareopelle, line 11: @ericmustin  - Is `instrumentation` the correct parent folder for these helper gems? Since propagators and instrumentation are on a peer level, I'm wondering if we should add something like that for the helper gems too. However, these helpers seem like they'd only be used by instrumentation gems, so maybe this makes sense. It's a minor thing, but I want to make sure I'm establishing a good pattern.

instrumentation/trilogy/opentelemetry-instrumentation-trilogy.gemspec
ericmustin: offhand i have no idea...checking

instrumentation/helpers-sql-obfuscation/README.md
```diff
@@ -0,0 +1,66 @@
...
+## License
+
+The `opentelemetry-instrumentation-helpers-sql-obfuscation` gem is distributed under the Apache 2.0 license. See [LICENSE][license-github] for more information.
+
+[new-relic-obfuscation-helpers]: https://github.com/newrelic/newrelic-ruby-agent/blob/96e7aca22c1c873c0f5fe704a2b3bb19652db68e/lib/new_relic/agent/database/obfuscation_helpers.rb
+[repo-github]: https://github.com/open-telemetry/opentelemetry-ruby
+[license-github]: https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/main/LICENSE
+[ruby-sig]: https://github.com/open-telemetry/community#ruby-sig
+[community-meetings]: https://github.com/open-telemetry/community#community-meetings
+[ruby-gitter]: https://gitter.im/open-telemetry/opentelemetry-ruby
```
kaylareopelle, line 66: @ericmustin - Is Ruby gitter still the preferred place to chat? I would've expected Slack, but I don't see it on the other READMEs.

instrumentation/helpers-sql-obfuscation/lib/opentelemetry/instrumentation/helpers/sql_obfuscation.rb
ericmustin: sure, it's possible a case statement has poor performance here, but it just seems like calling `.upcase` on a Symbol (https://ruby-doc.org/core-2.6.1/Symbol.html#method-i-upcase) is avoidable here, which under the hood is also calling `.to_s` on the Symbol, which would add a String instance allocation on every invocation iiuc?

instrumentation/pg/opentelemetry-instrumentation-pg.gemspec
ericmustin: can surface this with the broader SIG, seems like i'm probably just missing something trivial.

kaylareopelle: Sounds good, thank you. I saw this after I posted about it in the #otel-ruby channel. Here's the Slack link: https://cloud-native.slack.com/archives/C01NWKKMKMY/p1688672784020679

instrumentation/helpers-mysql/lib/opentelemetry/instrumentation/helpers/mysql.rb
ericmustin: yea it's a good question i think the pattern in https://github.com/open-telemetry/opentelemetry-ruby is for these to live at the peer level, ala `common`, but idk if we want to have like a `common/helpers` folder with both these in it or something like that. So tbh i'm kinda fine with `instrumentation` for now but mostly because I think someone else will have a stronger opinion.

instrumentation/helpers-sql-obfuscation/README.md
ericmustin: good catch, yea gitter is dead afaik long live slack, we can handle this change as a bulk change in a separate pr tho

kaylareopelle: cool, I'll make an issue for it.

kaylareopelle: https://github.com/open-telemetry/opentelemetry-ruby-contrib/issues/547

instrumentation/helpers-mysql/lib/opentelemetry/instrumentation/version.rb
ericmustin: good question, intuition says `0.1.0`. will check

instrumentation/helpers-sql-obfuscation/lib/opentelemetry/instrumentation/helpers/sql_obfuscation.rb
kaylareopelle: Your senses are correct! A case statement would be much more performant in this scenario and yields considerably fewer object allocations. 

Here's a gist with the test I ran: https://gist.github.com/kaylareopelle/ad910fd920401587fa0c4c79bbf7c98f

instrumentation/helpers-sql-obfuscation/lib/opentelemetry/instrumentation/helpers/sql_obfuscation.rb
kaylareopelle: Default adapter issue fixed in: 8096551319f684cf850901dbc12100dee43c8694 

Lmk about the config hash. I'm leaning toward no default there so an error is raised if the configs are missing.

instrumentation/helpers-sql-obfuscation/lib/opentelemetry/instrumentation/helpers/sql_obfuscation.rb
kaylareopelle: Fixed in: 8096551319f684cf850901dbc12100dee43c8694

instrumentation/helpers-sql-obfuscation/lib/opentelemetry/instrumentation/helpers/sql_obfuscation.rb
kaylareopelle: Fixed in: 8096551319f684cf850901dbc12100dee43c8694

instrumentation/mysql2/lib/opentelemetry/instrumentation/mysql2/patches/client.rb
kaylareopelle: There's still some work to do to get this PR ready to go, but the shift to classes and independent helper gems has been made!

instrumentation/helpers-mysql/lib/opentelemetry/instrumentation/helpers/mysql.rb
ericmustin: slack thread https://cloud-native.slack.com/archives/C01NWKKMKMY/p1688738760694089?thread_ts=1688672784.020679&cid=C01NWKKMKMY

instrumentation/helpers-mysql/lib/opentelemetry/instrumentation/version.rb
ericmustin: yea i think `0.1.0` is the standard for initial releases here, see recent `grape` instrumentation initial release for reference https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/394/files#diff-655c41c1beb01010148b54474ec3d54fb2f06d1c856b3fe9d7b8f59cfd1cff8cR11

ericmustin (commented):
i think everything has been updated here and my previous comments are addressed, with the exception of where to put these gems and the bundler errors. will add a point to the agenda to ensure both these are surfaced and discussed at tuesday SIG, thank u for quick turnaround @kaylareopelle apologies again on delay

arielvalentin:
@kaylareopelle I mentioned this in the CNCF slack but if you are amenable to it, I would like to host a mob session to look at the bundler errors together. Please share dates/times you are available in CNCF.

instrumentation/helpers-sql-obfuscation/test/test_helper.rb
```diff
@@ -0,0 +1,14 @@
...
+
+# Copyright The OpenTelemetry Authors
+#
+# SPDX-License-Identifier: Apache-2.0
+
+require 'bundler/setup'
+Bundler.require(:default, :development, :test)
+
+SimpleCov.start
+SimpleCov.minimum_coverage 85
```
kaylareopelle, line 11: @ericmustin - I added SimpleCov because it was present in the `base` test helper. However, it doesn't seem like it's working in either scenario. 

For the new helper gems, I'm getting `0 / 0 LOC (100%) covered.` 
`base` shows `5 / 5 LOC (100%) covered.`, with coverage only for the instrumentation test itself. 

At New Relic, we had to put a decent chunk of time in to get SimpleCov to work with our funky testing setup. Something similar may need to be done here. 

I'm leaning toward pulling SimpleCov out of the new helper gems for now since it's not working. wdyt?

.github/dependabot.yml
```diff
@@ -177,3 +177,11 @@ updates:
   directory: "/resource_detectors"
   schedule:
     interval: weekly
+- package-ecosystem: bundler
```
kaylareopelle, line 180: @ericmustin  - Could you double-check the GitHub actions to make sure I've added the new gems to the appropriate places?

instrumentation/helpers-sql-obfuscation/lib/opentelemetry/instrumentation/helpers/sql_obfuscation.rb
ericmustin: we should look to avoid raising an error outside of initialization, but i think it would be good to raise an error at init time if possible

instrumentation/helpers-sql-obfuscation/test/test_helper.rb
ericmustin: i dont think we are using SimpleCov in practice for anything, that's fine with me.

helpers/mysql/CHANGELOG.md
```diff
@@ -0,0 +1,5 @@
+# Release History: opentelemetry-helpers-mysql
+
+### v0.1.0 / 2023-07-13
+
+* Initial release.
```
kaylareopelle, line 5: Is the CHANGELOG automatically generated? I put this here as a boilerplate.

instrumentation/helpers-mysql/lib/opentelemetry/instrumentation/helpers/mysql.rb
kaylareopelle: We ultimately decided to move these into a helpers directory, outside of instrumentation. The names for the new gems are `opentelemetry-helpers-<type>`.

instrumentation/pg/opentelemetry-instrumentation-pg.gemspec
kaylareopelle: The resolution was to add Gemfile entries in a `:test` block for the new helpers.

helpers/mysql/CHANGELOG.md
arielvalentin: Yes the `CHANGELOG` is auto generated by the release scripts in the `.toys` directory

instrumentation/helpers-sql-obfuscation/lib/opentelemetry/instrumentation/helpers/sql_obfuscation.rb
kaylareopelle: Instead of raising an error, I decided to do a bit of refactoring so that the whole config hash wasn't required: fd2fa4b0ee585624fb596b823d18d2b475c54f05 

This drops the need for `:db_statement` and provides a default value for `:obfuscation_limit` to make sure the `.obfuscate_sql` method can run with only a SQL statement.

ericmustin:
Taking a look today

ericmustin (approved):
i think this lgtm, i like the choices around safety wrt config hash options in instrumentations that rely on these helpers. Approved. Awesome contribution @kaylareopelle 

Some non blocking things

- @fbogsany @robertlaurin this is going to hit code paths y'all use. Just a heads up, would be helpful to have your stamp here, and second set of eyes to catch anything i missed either wrt the repo structure (which i think we've resolved but was an open question earlier) or any performance regressions that might've slipped past me.
- @arielvalentin we should probably figure out the cadence to release these gems in, i suppose we'd release the helpers _first_, then release the others which would rely on them, is that right?
- not blocking, but do we have an example of this *actually working* locally somewhere on a demo setup etc? Might be a nice little confidence check, we've been burned a few times recently by bad releases that could've been caught by just making sure we see a trace the way we expect in a UI somewhere. I can try to build everything locally and provide a screenshot if not.

kaylareopelle:
@ericmustin, @arielvalentin - Thanks for the end-to-end example suggestion! 

I hadn't run one yet. Obfuscation worked for pg:

`"db.statement"=>"SELECT ? AS a, ? AS b, ? AS c"`

Here's the full console output: 
```
1253467db6f8:/app/instrumentation/pg/example$ bundle exec ruby pg.rb
I, [2023-07-13T17:39:46.691508 #29]  INFO -- : Instrumentation: OpenTelemetry::Instrumentation::PG was successfully installed with the following options {:peer_service=>nil, :db_statement=>:obfuscate, :obfuscation_limit=>2000}
#<struct OpenTelemetry::SDK::Trace::SpanData
 name="SELECT postgres",
 kind=:client,
 status=#<OpenTelemetry::Trace::Status:0x00007f145a004db0 @code=1, @description="">,
 parent_span_id="\x00\x00\x00\x00\x00\x00\x00\x00",
 total_recorded_attributes=9,
 total_recorded_events=0,
 total_recorded_links=0,
 start_timestamp=1689269986716068890,
 end_timestamp=1689269986719235862,
 attributes=
  {"db.system"=>"postgresql",
   "db.user"=>"postgres",
   "db.name"=>"postgres",
   "net.transport"=>"ip_tcp",
   "net.peer.name"=>"postgres",
   "net.peer.ip"=>"172.18.0.3",
   "net.peer.port"=>5432,
   "db.operation"=>"SELECT",
   "db.statement"=>"SELECT ? AS a, ? AS b, ? AS c"},
 links=nil,
 events=nil,
 resource=
  #<OpenTelemetry::SDK::Resources::Resource:0x00007f145e93d6a8
   @attributes=
    {"service.name"=>"unknown_service",
     "process.pid"=>29,
     "process.command"=>"pg.rb",
     "process.runtime.name"=>"ruby",
     "process.runtime.version"=>"3.2.2",
     "process.runtime.description"=>"ruby 3.2.2 (2023-03-30 revision e51014f9c0) [x86_64-linux-musl]",
     "telemetry.sdk.name"=>"opentelemetry",
     "telemetry.sdk.language"=>"ruby",
     "telemetry.sdk.version"=>"1.3.0"}>,
 instrumentation_scope=#<struct OpenTelemetry::SDK::InstrumentationScope name="OpenTelemetry::Instrumentation::PG", version="0.25.2">,
 span_id="\xA6\xCB\xE7Q\xFE\x14<\xB5",
 trace_id="\x16\xBF\xF9)Q\xCD@\x93\r\xFF\x8F(\xB9\xCDg\x16",
 trace_flags=#<OpenTelemetry::Trace::TraceFlags:0x00007f1459ba3af0 @flags=1>,
 tracestate=#<OpenTelemetry::Trace::Tracestate:0x00007f1459bcebd8 @hash={}>>
["1", "2", nil]  
```

The MySQL2 example is causing some issues for me. I think I don't have the right table set up. If someone else already has that environment set up, I'd appreciate you running it for me! 

Is this sufficient, or should I try to hook the example up to a non-console exporter?

kaylareopelle:
Also, I went ahead and rebased down to one commit, so we're one step closer to merging when ready.

fallwith (approved):
This provides a good foundation for modularity. Nicely done.

helpers/mysql/lib/opentelemetry/helpers/mysql.rb
```diff
@@ -0,0 +1,83 @@
...
+        when :db_operation_and_name
+          db_operation_and_name(operation, database_name)
+        end || 'mysql'
+      end
+
+      # @api private
+      def self.extract_statement_type(sql)
+        QUERY_NAME_REGEX.match(sql) { |match| match[1].downcase } unless sql.nil?
+      rescue StandardError => e
+        OpenTelemetry.logger.debug("Error extracting sql statement type: #{e.message}")
```
arielvalentin, line 67: Was this a debug error in other places? 

Should this use `handle_error` instead?

helpers/sql-obfuscation/test/helpers/sql_obfuscation_test.rb
```diff
@@ -0,0 +1,60 @@
...
+      it 'returns a failure message if unmatched pairs are present' do
+        assert_equal(OpenTelemetry::Helpers::SqlObfuscation::UNMATCHED_PAIRS_FAILURE_MESSAGE, obfuscate_sql)
+      end
+    end
+
+    describe 'when the adapter is mysql' do
+      let(:adapter) { :mysql }
+      let(:sql) { "SELECT * from users where users.id = 1 and users.email = 'test@test.com\255'" }
+
+      it 'encodes utf8' do
```
arielvalentin, line 55: This test seems to indicate the opposite of what I expect. This is what happens when it encounters an error encoding UTF-8 no?

Is there a positive test case example where it successfully encodes SQL with even multi-byte strings like emojis?

arielvalentin (commented):
The refactoring looks good! I have a question about adding a test for UTF-8 encoding. I will take this for a test run using example files. 

Thank you for your patience and sharing your talent with us.

helpers/mysql/lib/opentelemetry/helpers/mysql.rb
```diff
@@ -0,0 +1,83 @@
+# frozen_string_literal: true
+
+# Copyright The OpenTelemetry Authors
+#
+# SPDX-License-Identifier: Apache-2.0module OpenTelemetry
+
+module OpenTelemetry
+  module Helpers
+    # This class contains for MySQL instrumentation libraries, like mysql2 and
```
fbogsany, line 9:
```suggestion
    # This class contains helpers for MySQL instrumentation libraries, like mysql2 and
```

arielvalentin:
@kaylareopelle this looks great so far! 

I did a little exploratory testing and realized that we do not have working example files for some of these gems. I will add them in a separate PR but everything is working as I expect.

arielvalentin:
@kaylareopelle totally missed this call for help:

> The MySQL2 example is causing some issues for me. I think I don't have the right table set up. If someone else already has that environment set up, I'd appreciate you running it for me!

Same for me. The example seemed to never work right so I used the `information_schema` instead... e.g. 

```ruby
# trilogy
client.query("SELECT * from information_schema.INNODB_TABLES; /**DÃ©**/").each do |row|
  puts row
end
```

```ruby
# mysql2
client.query("SELECT * from information_schema.INNODB_TABLES; /**DÃ©**/").each do |row|
  puts row
end
```

helpers/mysql/lib/opentelemetry/helpers/mysql.rb
kaylareopelle: In MySQL2, it was a debug-level log. For Trilogy, it was an error-level log. I'm happy to use`handle_error` instead.

helpers/sql-obfuscation/lib/opentelemetry/helpers/sql_obfuscation.rb
```diff
@@ -0,0 +1,147 @@
...
+        regex = case adapter
+                when :mysql
+                  MYSQL_COMPONENTS_REGEX
+                when :postgres
+                  POSTGRES_COMPONENTS_REGEX
+                else
+                  DEFAULT_COMPONENTS_REGEX
+                end
+
+        return truncate_statement(sql, regex, obfuscation_limit) if sql.size > obfuscation_limit
```
arielvalentin, line 113: Looking at this again. Is there a use case where the truncate statement fails because we have not encoded the SQL as utf8?

helpers/sql-obfuscation/opentelemetry-helpers-sql-obfuscation.gemspec
```diff
@@ -0,0 +1,44 @@
...
+  spec.homepage    = 'https://github.com/open-telemetry/opentelemetry-ruby-contrib'
+  spec.license     = 'Apache-2.0'
+
+  spec.files = Dir.glob('lib/**/*.rb') +
+               Dir.glob('*.md') +
+               ['LICENSE', '.yardopts']
+  spec.require_paths = ['lib']
+  spec.required_ruby_version = '>= 3.0'
+
+  spec.add_dependency 'opentelemetry-sdk', '~> 1.0'
```
arielvalentin, line 28: This should be an API dependency and must not depend on the SDK. 

`common` and `semenatic_conventions` are also ok.

arielvalentin (changes requested):
Sorry @kaylareopelle I looked at this again with fresh eyes and see a few issues

helpers/sql-obfuscation/test/helpers/sql_obfuscation_test.rb
kaylareopelle: @arielvalentin  - Thank you so much for noticing this. The UTF-8 encoding was not working at all for the new SQL Obfuscation helper because `opentelemetry-sdk` wasn't a dependency, so `OpenTelemetry::Common` could not be found.

This fix allowed postgres to be correctly encoded with UTF-8 as well! 

See commit: cbf148a623884e9d74d9e885355dae6a84062fee 

Let me know what you think. Once it looks good to you, I'll rebase the commit.

helpers/sql-obfuscation/lib/opentelemetry/helpers.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+# Copyright The OpenTelemetry Authors
+#
+# SPDX-License-Identifier: Apache-2.0
+
+require 'opentelemetry-sdk'
```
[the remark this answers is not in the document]
kaylareopelle: This was just added! No worries. Good to know. I'll find another way.

helpers/mysql/lib/opentelemetry/helpers/mysql.rb
kaylareopelle: @arielvalentin  - This has been replaced with a `handle_error` call: f46b1f7d9b8a1e073c200afa5f6676f379b2a34a 

Let me know if there are any further changes you'd like me to make. I'll rebase the commit after I get your approval.

helpers/sql-obfuscation/opentelemetry-helpers-sql-obfuscation.gemspec
kaylareopelle: I'll update this to depend on `common`, that's the specific package we need.

helpers/sql-obfuscation/lib/opentelemetry/helpers/sql_obfuscation.rb
kaylareopelle: Great question. I'll test this case and get back to you.

helpers/mysql/lib/opentelemetry/helpers/mysql.rb
arielvalentin: > In MySQL2, it was a debug-level log. For Trilogy, it was an error-level log.

Good old copy/paste modify poorly. _Spoiler alert: IT WAS ME_

helpers/sql-obfuscation/opentelemetry-helpers-sql-obfuscation.gemspec
kaylareopelle: Fixed in: 891227424747422b236d2cc91ba1a419ec4ff21c

helpers/sql-obfuscation/lib/opentelemetry/helpers/sql_obfuscation.rb
kaylareopelle: After looking into it, the `truncate_statement` call wouldn't fail due to a lack of UTF-8 encoding. However, the errors the UTF-8 encoding method is trying to prevent could still occur downstream.

I updated the code to UTF-8 encode before truncation and then ran `gsub` after the obfuscation limit was checked. This changes the original functionality slightly, but it's a bit more robust to protect against UTF-8 errors.

See: 2805b6dc8f10012bce8afb794dc8e376ca30cc44

kaylareopelle: Hmm... it looks like this change has a few issues with MySql2 and Trilogy tests. Since we're changing the behavior of the original obfuscation code here, would you be open to breaking this code into a bugfix and including it in a separate release? I kind of like the idea of keeping things as consistent as possible with the original obfuscation code for the first release.

kaylareopelle: Ah, actually the error was introduced when I switched the `OpenTelemetry.logger.debug` call for the `OpenTelemetry.handle_error` call. Not quite sure what's going on here... I'll investigate.

helpers/mysql/lib/opentelemetry/helpers/mysql.rb
```diff
@@ -0,0 +1,84 @@
...
+          db_operation_and_name(operation, database_name)
+        end || 'mysql'
+      end
+
+      # @api private
+      def self.extract_statement_type(sql)
+        QUERY_NAME_REGEX.match(sql) { |match| match[1].downcase } unless sql.nil?
+      rescue StandardError => e
+        OpenTelemetry.handle_error(message: "#{EXTRACT_STATEMENT_FAILURE_MESSAGE}: #{e.message}", exception: e)
+        "OpenTelemetry Error: #{EXTRACT_STATEMENT_FAILURE_MESSAGE}"
```
kaylareopelle, line 69: @arielvalentin - Calling `OpenTelemetry.handle_error` here will raise the error after it's reported. This causes the span to have a `nil` name value. This is the behavior I'm seeing when I run the tests, but I might be misinterpreting the impact of using `.handle_error`.

When we call `OpenTelemetry.logger.<level>`, the return value is `true`, which allows an error to be logged, but does not stop the span from getting the default name of `'mysql'`.

I think it's preferable to get a span with a generic name over raising an error in an application. 

Do you see any issues with how I've added `.handle_error` that might be interfering with the span naming process? If this is indeed the behavior, would you be open to reverting the change and using `OpenTelemetry.logger.error` instead?

helpers/sql-obfuscation/lib/opentelemetry/helpers/sql_obfuscation.rb
arielvalentin: That is because we are handling the error so the last line of the method call is the standard error instead of a value you intended on returning. 

```ruby
def example(x)
  if x == 1
    x
  else
    raise "this is an error"
  end
  rescue StandardError => e
    "Unexpected #{e.message}"
end

puts example(1)
puts example(2)
```



```console
🤘 ruby return_last_line.rb
1
Unexpected this is an error
``` 

So in our case the method is returning the result of `handle_error` or `debug`

helpers/mysql/lib/opentelemetry/helpers/mysql.rb
kaylareopelle: @arielvalentin  - After thinking through this a little more, I think the first release of these helper gems should include the original behavior. 

`handle_error` changes the behavior by raising the error and preventing span creation with the default name of `mysql`.

We can update the gem to use `handle_error` here instead in a future PR, and I'm happy to add a comment and make a ticket to address this work in the future.

 Are you comfortable with this approach?

arielvalentin: Sure.

helpers/sql-obfuscation/lib/opentelemetry/helpers/sql_obfuscation.rb
arielvalentin: Either way I think the result of this message should be the `OBFUSCATION_FAILURE_MESSAGE` when a StandardError occurs.

kaylareopelle:
I've addressed the latest round of feedback and am ready for another review! 

For the error handler call in the MySQL helper, I've created #609

helpers/mysql/lib/opentelemetry/helpers/mysql.rb
```diff
@@ -0,0 +1,85 @@
...
+        'savepoint',
+        'release savepoint',
+        'explain',
+        'drop database',
+        'drop table',
+        'create database',
+        'create table'
+      ].freeze
+
+      QUERY_NAME_REGEX = Regexp.new("^(#{QUERY_NAMES.join('|')})", Regexp::IGNORECASE)
```
arielvalentin, line 39: Regexp are immutable right? I don't recall

helpers/sql-obfuscation/lib/opentelemetry/helpers/sql_obfuscation.rb
```diff
@@ -0,0 +1,151 @@
...
+        first_match_index = sql.index(regex)
+        truncation_message = "SQL truncated (> #{limit} characters)"
+        return truncation_message unless first_match_index
+
+        truncated_sql = sql[..first_match_index - 1]
+        "#{truncated_sql}...\n#{truncation_message}"
+      end
+
+      # @api private
+      def self.utf8_encode_sql(sql)
```
arielvalentin, line 138: Is this used in more than one place? If it isn't consider inlining it because it will reduce the number of method dispatches as well as keep from adding an additional line to the call stack.

helpers/sql-obfuscation/lib/opentelemetry/helpers/sql_obfuscation.rb
```diff
@@ -0,0 +1,151 @@
...
+        # Original MySQL UTF-8 Encoding Fixes:
+        # https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/160
+        # https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/345
+        OpenTelemetry::Common::Utilities.utf8_encode(sql, binary: true)
+      end
+
+      # @api private
+      def self.detect_unmatched_pairs(obfuscated, adapter)
+        CLEANUP_REGEX[adapter].match(obfuscated)
```
arielvalentin, line 147: Similar observation about inlining the method

arielvalentin (approved):
Thanks again!

Going to wait in mysql2 instrumentation authors to provide a review before merging. 

Thanks @kaylareopelle !!!

helpers/mysql/lib/opentelemetry/helpers/mysql.rb
kaylareopelle: Nice question! According to Rubocop, they're immutable starting in Ruby 3.0: https://docs.rubocop.org/rubocop/cops_style.html#styleredundantfreeze

I'm happy to remove the freeze in a future PR. It's used here because that's how it was written in the original code.

helpers/sql-obfuscation/lib/opentelemetry/helpers/sql_obfuscation.rb
```diff
@@ -0,0 +1,151 @@
...
+        sql = utf8_encode_sql(sql)
+        return truncate_statement(sql, regex, obfuscation_limit) if sql.size > obfuscation_limit
+
+        sql = sql.gsub(regex, PLACEHOLDER)
+        return UNMATCHED_PAIRS_FAILURE_MESSAGE if detect_unmatched_pairs(sql, adapter)
+
+        sql
+      rescue StandardError => e
+        OpenTelemetry.handle_error(message: OBFUSCATION_FAILURE_MESSAGE, exception: e)
+        "OpenTelemetry error: #{OBFUSCATION_FAILURE_MESSAGE}"
```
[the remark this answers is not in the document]
kaylareopelle: This is another case where the text is copied from the original code. I agree that the constant is probably enough context. I'll clean this up in a follow-up PR.

helpers/sql-obfuscation/lib/opentelemetry/helpers/sql_obfuscation.rb
kaylareopelle: Good point. `.utf8_encode_sql` and `.detect_unmatched_pairs` both are not used in more than one place. I'll make that change.

helpers/sql-obfuscation/lib/opentelemetry/helpers/sql_obfuscation.rb
kaylareopelle: This method has been inlined!

helpers/sql-obfuscation/lib/opentelemetry/helpers/sql_obfuscation.rb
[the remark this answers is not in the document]
kaylareopelle: I've added this to #609

helpers/mysql/lib/opentelemetry/helpers/mysql.rb
arielvalentin: Is there a freeze? I was asking because I didn't see one but thank you for answering my question!

helpers/mysql/lib/opentelemetry/helpers/mysql.rb
```diff
@@ -0,0 +1,85 @@
...
+          db_operation_and_name(operation, database_name)
+        end || 'mysql'
+      end
+
+      # @api private
+      def self.extract_statement_type(sql)
+        QUERY_NAME_REGEX.match(sql) { |match| match[1].downcase } unless sql.nil?
+      rescue StandardError => e
+        # TODO: Replace with `handle_error` Issue #609
+        OpenTelemetry.logger.error("#{EXTRACT_STATEMENT_FAILURE_MESSAGE}: #{e.message}")
```
fbogsany, line 69: There's not really any value in extracting the prefix here to a constant. I'd rather just have the text directly here (it isn't referenced anywhere else as far as I can tell).

Can we address the `TODO` now or is there a larger issue blocking it?

helpers/mysql/lib/opentelemetry/helpers/mysql.rb
fbogsany, line 39: There's a frustrating corner case with this regex that would be good to fix (doesn't have to be done here - I'll do it if you don't): this regex assumes queries start with the operation, and don't contain preceding whitespace or comments. In a lot of our production code, this assumption is wrong (we have "marginalia" comments that contain useful telemetry that'll be captured in slow query logs, for example). Consequently, most of our MySQL spans are simply named `mysql` 😞.

helpers/mysql/lib/opentelemetry/helpers/mysql.rb
kaylareopelle: Thanks for the feedback, @fbogsany! I removed the constant and took care of the work for #609 in the MySQL and SQL Obfuscation gems. 

See: f69eb9ce478fa5627b4c7bea9d07332cc4b00f6b 

Initially, I wanted to put that work in a separate PR to make the first version of the new helper gems identical in behavior to the code extracted out of the instrumentation gems. However, with a closer look, I realized there were already other behavior changes. 

I'm happy to back out the changes and put them in a separate PR if you'd prefer that approach.

helpers/mysql/lib/opentelemetry/helpers/mysql.rb
kaylareopelle: That sounds very frustrating! I have a potential fix here: c694e556d67b5ad02d7a9604d11bc9d2c5284472 

Now, the statement type will be set to the first matched item in `QUERY_NAMES`, regardless of where it appears in the SQL string. I used a marginalia comment to test the change.

helpers/mysql/test/helpers/mysql_test.rb
```diff
@@ -93,6 +93,18 @@
       end
     end
 
+    describe 'when sql has marginalia-style prepended comments' do
+      let(:sql) do
+        '/*application:web,controller:blob,action:show,correlation_id:01EZVMR923313VV44ZJDJ7PMEZ,' \
```
arielvalentin, line 98: @fbogsany This may put us in a situation where developers unintentionally add PII to the marginalia comments and it is ignored by the obfuscator. 

Though it is not our total responsibility to handle ever case we should make it clear that SQL comments are going to be ignored.

kaylareopelle: @arielvalentin - This is a great case to consider. [I've added a test to the SQL Obfuscation helper](https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/529/files#diff-35551bfb3ffadde6dfff04ab16be87da9efbef7e9b82f10cd8104a141535c475R35-R47) to account for it in c28e297c 

In MySQL2 and Trilogy, the original SQL is used to get the query name (ex. 'select') and to obfuscate the statement. This leads me to believe the comments will be obfuscated even if the query name is not plucked from the beginning of the statement.

arielvalentin: 😅 OK I am totally wrong then. glad to be.

helpers/mysql/lib/opentelemetry/helpers/mysql.rb
kaylareopelle: There was a little more work needed here to pass the test cases in mysql2 and trilogy: f50f16b6358cfd0843ca9cb9f1d691ffb78dd324

helpers/mysql/test/helpers/mysql_test.rb
fbogsany: @kaylareopelle is correct - the obfuscation code will obfuscate comments (all comments are replaced with `?`). There is a corner case I need to address: if we choose not to obfuscate because the query is too large (perhaps because of comments 😞), we truncate the query instead, but the remaining prefix may contain PII. The "fix" may be to apply obfuscation to the truncated query, but that may run into "unmatched quotes". I can't quite recall how these things play together - I'll 👀 this more closely today.

helpers/mysql/lib/opentelemetry/helpers/mysql.rb
fbogsany: ❤️ Thank you! I'll review again today.

helpers/mysql/lib/opentelemetry/helpers/mysql.rb
```diff
@@ -0,0 +1,87 @@
...
+        'begin',
+        'commit',
+        'create database',
+        'create table',
+        'delete',
+        'drop database',
+        'drop table',
+        'explain',
+        'insert',
+        'prepare statement',
```
kaylareopelle, line 32: @xuan-cao-swi - 'prepare statement' is now included in the list of MySQL query names.

kaylareopelle: This has been removed. @xuan-cao-swi let me know that the prepare method will need to be instrumented to make this work successfully. This work should be taken up in a separate PR. See: https://github.com/open-telemetry/opentelemetry-ruby-contrib/issues/670

helpers/mysql/test/helpers/mysql_test.rb
kaylareopelle: @fbogsany - This is an excellent callout. I prefer to address this issue in a separate PR. Would you be open to that approach?

helpers/sql-obfuscation/CHANGELOG.md
```diff
@@ -0,0 +1,5 @@
+# Release History: opentelemetry-helpers-sql-obfuscation
```
scbjans, line 1: For initial Gem releases; only this line should be included. Below lines should be removed; since it will be added by the release tooling

helpers/sql-obfuscation/test/helpers/sql_obfuscation_test.rb
```diff
@@ -0,0 +1,102 @@
...
+    end
+
+    describe 'when sql has marginalia-style prepended comments' do
+      let(:sql) do
+        '/*application:web,controller:blob,action:show,correlation_id:01EZVMR923313VV44ZJDJ7PMEZ,' \
+          'endpoint_id:Projects::BlobController#show*/ SELECT "routes".* FROM "routes" WHERE "routes"' \
+          '."source_id" = 75 AND "routes"."source_type" = \'Namespace\' LIMIT 1'
+      end
+
+      let(:expected) { '? SELECT ?.* FROM ? WHERE ?.? = ? AND ?.? = ? LIMIT ?' }
```
fbogsany, line 42: This was ... unexpected. Makes sense, but annoying if any generated SQL queries quote table or column names. Does AR do this or is this just an invented (but valid) example?

helpers/mysql/test/helpers/mysql_test.rb
```diff
@@ -0,0 +1,147 @@
...
+  end
+
+  describe '.db_operation_and_name' do
+    let(:operation) { 'operation' }
+    let(:database_name) { 'database_name' }
+    let(:db_operation_and_name) { OpenTelemetry::Helpers::MySQL.db_operation_and_name(operation, database_name) }
+
+    describe 'when operation and database_name are present' do
+      it 'returns a combination of the operation and database_name' do
+        assert_equal(db_operation_and_name, "#{operation} #{database_name}")
```
fbogsany, line 65: Presumably this?
```suggestion
        assert_equal(db_operation_and_name, "operation database_name")
```

helpers/mysql/test/helpers/mysql_test.rb
```diff
@@ -0,0 +1,147 @@
...
+    end
+
+    describe 'when database_name is nil' do
+      let(:database_name) { nil }
+
+      it 'returns the operation name' do
+        assert_equal(db_operation_and_name, operation)
+      end
+    end
+  end
```
fbogsany, line 84: When both `database_name` and `operation` are `nil`?

helpers/mysql/test/helpers/mysql_test.rb
```diff
@@ -0,0 +1,147 @@
...
+        assert_nil(extract_statement_type)
+      end
+    end
+
+    describe 'when sql has marginalia-style prepended comments' do
+      let(:sql) do
+        '/*application:web,controller:blob,action:show,correlation_id:01EZVMR923313VV44ZJDJ7PMEZ,' \
+          'endpoint_id:Projects::BlobController#show*/ SELECT "routes".* FROM "routes" WHERE "routes"' \
+          '."source_id" = 75 AND "routes"."source_type" = \'Namespace\' LIMIT 1'
+      end
```
fbogsany, line 118: I played around with this a bit. While the example above works as expected, the result is incorrect if one of the query names is in the Marginalia comment. E.g.:
```ruby
irb(main):001:1*       QUERY_NAMES = [
irb(main):002:1*         'set names',
irb(main):003:1*         'select',
irb(main):004:1*         'insert',
irb(main):005:1*         'update',
irb(main):006:1*         'delete',
irb(main):007:1*         'begin',
irb(main):008:1*         'commit',
irb(main):009:1*         'rollback',
irb(main):010:1*         'savepoint',
irb(main):011:1*         'release savepoint',
irb(main):012:1*         'explain',
irb(main):013:1*         'drop database',
irb(main):014:1*         'drop table',
irb(main):015:1*         'create database',
irb(main):016:1*         'create table'
irb(main):017:0>       ].freeze
=> 
["set names",
...
irb(main):018:0>       QUERY_NAME_REGEX = Regexp.new("\\b(#{QUERY_NAMES.join('|')})\\b", Regexp::IGNORECASE)
=> /\b(set names|select|insert|update|delete|begin|commit|rollback|savepoint|release savepoint|explain|drop database|drop table|create da...
irb(main):019:0* sql = '/*application:web,controller:blob,action:show,correlation_id:01EZVMR923313VV44ZJDJ7PMEZ,' \
irb(main):020:0* 'endpoint_id:Projects::BlobController#show*/ SELECT "routes".* FROM "routes" WHERE "routes"' \
irb(main):021:0>           '."source_id" = 75 AND "routes"."source_type" = \'Namespace\' LIMIT 1'
=> "/*application:web,controller:blob,action:show,correlation_id:01EZVMR923313VV44ZJDJ7PMEZ,endpoint_id:Projects::BlobController#show*/ S...
irb(main):022:0> QUERY_NAME_REGEX.match(sql) { |match| match[1].downcase } 
=> "select"
irb(main):023:0* sql = '/*application:web,controller:insert,action:show,correlation_id:01EZVMR923313VV44ZJDJ7PMEZ,' \
irb(main):024:0* 'endpoint_id:Projects::BlobController#show*/ SELECT "routes".* FROM "routes" WHERE "routes"' \
irb(main):025:0>           '."source_id" = 75 AND "routes"."source_type" = \'Namespace\' LIMIT 1'
=> "/*application:web,controller:insert,action:show,correlation_id:01EZVMR923313VV44ZJDJ7PMEZ,endpoint_id:Projects::BlobController#show*/...
irb(main):026:0> QUERY_NAME_REGEX.match(sql) { |match| match[1].downcase } 
=> "insert"
irb(main):027:0> 
```

helpers/mysql/lib/opentelemetry/helpers/mysql.rb
```diff
@@ -0,0 +1,86 @@
...
+        'savepoint',
+        'release savepoint',
+        'explain',
+        'drop database',
+        'drop table',
+        'create database',
+        'create table'
+      ].freeze
+
+      QUERY_NAME_REGEX = Regexp.new("\\b(#{QUERY_NAMES.join('|')})\\b", Regexp::IGNORECASE)
```
fbogsany, line 40: I suspect we need to match and ignore (multi-line) comments at the start of the query. I.e. these components:
```ruby
        comments: /(?:#|--).*?(?=\r|\n|$)/i,
        multi_line_comments: %r{\/\*(?:[^\/]|\/[^*])*?(?:\*\/|\/\*.*)}
```

kaylareopelle: Good point. Updated in: 82584bae3c9132846fdd593a124dc680b770af28

helpers/mysql/test/helpers/mysql_test.rb
kaylareopelle: Great catch. I've updated the tests to include this case. It's fixed in: 82584bae3c9132846fdd593a124dc680b770af28

helpers/sql-obfuscation/CHANGELOG.md
kaylareopelle: Good catch! Fixed in: bb37c867f04e2e011939eea9ba8f92c1e1258db6

helpers/mysql/test/helpers/mysql_test.rb
kaylareopelle: Thank you! Added here: 3f0bcb4af5efcb60daf3f33881856bb21fdaf50b

helpers/mysql/lib/opentelemetry/helpers/mysql.rb
```diff
@@ -40,7 +40,7 @@ class MySQL
       # Ignore query names that might appear in comments prepended to the
       # statement.
       PREPENDED_COMMENTS_REGEX = %r{/\*(?:[^/]|/[^*])*?(?:\*/|/\*.*)}m
-      QUERY_NAME_REGEX = Regexp.new("(?:#{PREPENDED_COMMENTS_REGEX})?.*\\b(#{QUERY_NAMES.join('|')})\\b.*", Regexp::IGNORECASE)
+      QUERY_NAME_REGEX = Regexp.new("(?:#{PREPENDED_COMMENTS_REGEX})?.*?\\b(#{QUERY_NAMES.join('|')})\\b.*", Regexp::IGNORECASE)
```
fbogsany, line 43: This seems overly permissive. I think we want to:
1. match against the start of the SQL
2. accept initial whitespace
3. allow arbitrary whitespace (only) between any comment and the query name.

From informal testing, this seems to meet those requirements:
```suggestion
      QUERY_NAME_REGEX = Regexp.new("^\s*(?:#{PREPENDED_COMMENTS_REGEX})?\s*\\b(#{QUERY_NAMES.join('|')})\\b.*", Regexp::IGNORECASE)
```

helpers/sql-obfuscation/lib/opentelemetry/helpers/sql_obfuscation.rb
```diff
@@ -42,29 +42,22 @@ class SqlObfuscation
         boolean_literals: /\b(?:true|false|null)\b/i,
         hexadecimal_literals: /0x[0-9a-fA-F]+/,
         comments: /(?:#|--).*?(?=\r|\n|$)/i,
-        multi_line_comments: %r{\/\*(?:[^\/]|\/[^*])*?(?:\*\/|\/\*.*)}
+        multi_line_comments: %r{\/\*(?:[^\/]|\/[^*])*?(?:\*\/|\/\*.*)}m,
```
fbogsany, line 45: Why have we switched to a multiline match here? Actually, I find the last part of the original regex confusing - it basically matches `/*` followed by anything, eagerly, so in the context of the rest of the regex, we'd have `/* /*` and the regex would eat the rest of the input.

I feel like this regex is cleaner, simpler and more correct: `%r{(?:\/\*.*?\*\/)}m`. Is there a corner case this misses?

kaylareopelle: Since the regex map referenced "multi_line_comments" in the name, I thought a multiline match would be an appropriate addition. I don't mind removing it though if you disagree. 

I couldn't find a corner case that your simpler regex missed. I'll update the code to use that instead.

kaylareopelle: Updated in 7b900745

helpers/sql-obfuscation/lib/opentelemetry/helpers/sql_obfuscation.rb
```diff
@@ -0,0 +1,137 @@
...
+    # If you want to add support for a new adapter, update the following
+    # constants to include keys for your adapter:
+    #  * DIALECT_COMPONENTS
+    #  * CLEANUP_REGEX
+    # You must also add a new constant that uses `generate_regex` with your
+    # adapter's dialect components that is named like
+    # `<ADAPTER>_COMPONENTS_REGEX`, such as: `MYSQL_COMPONENTS_REGEX`.
+    #
+    # @api public
+    class SqlObfuscation
```
arielvalentin, line 34: if these are never instantiated perhaps this can be a module with the methods listed under `module_function`

helpers/sql-obfuscation/test/helpers/sql_obfuscation_test.rb
```diff
@@ -0,0 +1,108 @@
...
+    msg << "Actual:   #{actual_output}\n"
+    msg
+  end
+
+  def self.load_fixture
+    data = File.read("#{Dir.pwd}/test/fixtures/sql_obfuscation.json")
+    JSON.parse(data)
+  end
+
+  load_fixture.each do |test_case|
```
arielvalentin, line 87: 💖 Thank you!!!

helpers/mysql/lib/opentelemetry/helpers/mysql.rb
```diff
@@ -40,7 +40,7 @@ class MySQL
       # Ignore query names that might appear in comments prepended to the
       # statement.
       PREPENDED_COMMENTS_REGEX = %r{/\*(?:[^/]|/[^*])*?(?:\*/|/\*.*)}m
```
fbogsany, line 42:
```suggestion
      PREPENDED_COMMENTS_REGEX = %r{(?:\/\*.*?\*\/)}m
```
This matches the `multi_line_comments` update.

helpers/sql-obfuscation/lib/opentelemetry/helpers/sql_obfuscation.rb
kaylareopelle: Good idea! I'm still wrapping my head around when to choose `module_function` over `extend self`. 

It looks like the repo has more uses of `extend self`(17) than `module_function`(3). Also, [Rubocop defaults to `module_function` over `extend self`.](https://docs.rubocop.org/rubocop/cops_style.html#stylemodulefunction) I know they have slightly different behavior too. 

What's your reasoning for using `module_function` instead of `extend self`?

arielvalentin: > This is the main difference between extend self and module_function. The latter one really creates a module function by restricting access to the included method while extend self still allows access to the included method.

https://medium.com/rubycademy/is-module-function-really-the-same-as-extend-self-ac1e96a1cda0

arielvalentin: I don't have a strong preference but just pointing out that if our intent is to treat these like global helpers and not treated as mixins, then module_function makes that more explicit. 

If we want more flexibility in composition, then we should keep using extend self

kaylareopelle: Thanks, Ariel! That's very helpful!

kaylareopelle: Updated in: 4e8aea4ae22972757ff1952bf3eea0ad9ba5b301

fbogsany (approved):
This LGTM. Thanks for this @kaylareopelle. Sorry for the lengthy review process - the PR and scope was quite large!

robertlaurin (approved):

arielvalentin:
@kaylareopelle I am going to do some exploratory testing before merging!

Thanks everyone for the thorough reviews.

PR #538, opened 2023-06-30
Propagate context from async producer

plantfansam:
I'd like to marinate on this one for a couple days, but I think that any `"... send"` spans generated from `ruby-kafka`'s `AsyncProducer`'s [`do_loop` method](https://github.com/zendesk/ruby-kafka/blob/master/lib/kafka/async_producer.rb#L233) will be root spans. This is because we do not explicitly pass context between `AsyncProducer#perform` and `Producer#perform`. To demonstrate, consider the traces generated by this code excerpted from `ruby_kafka.rb`:

```ruby
# Add message to producer buffer

async_producer = kafka.async_producer(delivery_threshold: 1, delivery_interval: 0.001)

tracer.in_span("hello async") do |span|
  async_producer.produce('async example', topic: topic, headers: { foo: :bar }, create_time: Time.now)
end

producer.deliver_messages

# Process messages
count = 0
kafka.each_message(topic: topic) do |message|
  count += 1
  break if count == 2 # we only generate 2 messages
end
```

#### "Hello async" parent span has no child spans

Screenshot 2023-07-06 at 2 01 32 PM

#### "... send" span is parent of consumer span, but does not point to "Hello async" as its parent

Screenshot 2023-07-06 at 2 06 39 PM

### Traces for async message production with new patch

image

### Background

Some background on how `AsyncProducer` works (AFAICT): the `AsyncProducer` [has a background thread](https://github.com/zendesk/ruby-kafka/blob/master/lib/kafka/async_producer.rb#L158-L165) that watches an instance variable called `@queue`. `#produce` shoves stuff onto `@queue`, which the background thread handles (i.e. [calls `@producer.produce`](https://github.com/zendesk/ruby-kafka/blob/master/lib/kafka/async_producer.rb#L267)).

### Strategy

This PR attempts to get the correct parent on `"... send"` spans by:

1. Injecting the current trace context into `headers` during `AsyncProducer#produce`
2. Trying extract the trace context from that header in `Producer#produce` and use it as the parent for the `...send` span (we only do this if the current span is `INVALID`, indicating that `Producer#produce` is running in the context of `do_loop`).

We could, alternatively, patch `AsyncProducer#producer` to push the current OpenTelemetry context onto `@queue` [when it's adding messages to them](https://github.com/zendesk/ruby-kafka/blob/master/lib/kafka/async_producer.rb#L117), then pop that off the `@queue` in `do_loop`, but then we'd have to overwrite the method signature for `perform` which I don't really want to do. It almost feels appropriate to use a propagator to pass context between these two methods, but maybe that's Friday afternoon talking.

Thoughts? Feelings? Warnings?


arielvalentin:
@plantfansam do you have a sample app and/or screenshots of this in action?

plantfansam:
@arielvalentin I'll get some screenshots and/or try to set up an example app today.

plantfansam:
@arielvalentin I updated the PR with a `docker-compose.yml` with instructions and added some screenshots/better explanation to PR description! Let me know if anything's confusing...

arielvalentin:
@plantfansam we have a docker-compose file in the top level directory. 

Is there a reason not to add the example directory there? https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/main/docker-compose.yml

instrumentation/ruby_kafka/test/opentelemetry/instrumentation/ruby-kafka/patches/producer_test.rb
```diff
@@ -71,5 +72,27 @@
...
+
+    it 'propagates context when tracing async produce calls' do
+      tracer.in_span('wat') do |sp|
+        async_producer.produce('hello async', topic: async_topic)
+        async_producer.deliver_messages
+
+        # Wait for the async calls to produce spans
+        wait_for(error_message: 'Max wait time exceeded for async producer') { EXPORTER.finished_spans.size.positive? }
+
+        _(spans.first.parent_span_id).must_equal(sp.context.span_id)
```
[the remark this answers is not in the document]
plantfansam: 5388cc6e6be3059eeaf95485e08f93f82ab6d07d

instrumentation/ruby_kafka/lib/opentelemetry/instrumentation/ruby_kafka/patches/producer.rb
```diff
@@ -17,9 +17,18 @@ def produce(value, topic:, key: nil, headers: {}, partition: nil, partition_key:
               'messaging.destination_kind' => 'topic'
             }
 
-            tracer.in_span("#{topic} send", attributes: attributes, kind: :producer) do
-              OpenTelemetry.propagation.inject(headers)
-              super
+            # If context is unset, try to inject headers injected by async producer
+            ctx = if OpenTelemetry::Trace.current_span == OpenTelemetry::Trace::Span::INVALID
```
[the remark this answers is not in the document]
plantfansam: I think so: 00cd9e8.

Now, logic is: if there's already a span recording when we produce, use that as context; otherwise try to extract context from headers.

There's maybe an edge case where the active span when #producer runs has been sampled out, but produce is being called with headers: with a trace context indicating a sampled in parent. Should we (info-level) log a warning in that circumstance?

instrumentation/ruby_kafka/test/opentelemetry/instrumentation/ruby-kafka/patches/producer_test.rb
```diff
@@ -71,5 +72,27 @@
...
+      span_id = 'b7ad6b7169203331'
+      tracer.in_span('wat') do |sp|
+        producer.produce('hello', topic: topic, headers: { 'traceparent' => "00-0af7651916cd43dd8448eb211c80319c-#{span_id}-01" })
+        producer.deliver_messages
+
+        _(spans.first.parent_span_id).must_equal(sp.context.span_id)
+      end
+    end
+
+    it 'propagates context when tracing async produce calls' do
```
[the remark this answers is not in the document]
plantfansam: I think yes: https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/538/files#diff-e991caa49d063bc36d94f59c730df0c156844b97bdc439cd02e1cd30dd410883R100. Not sure if you had something else in mind.

plantfansam:
> Is there a reason not to add the example directory there? https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/main/docker-compose.yml

Not conceptually, no, I just can't validate it with `docker-compose` (long story). `podman-compose` doesn't like for me to run `podman-compose run` against services that have a `depends_on:` block for whatever reason. I think the config would look like this:

```yaml
  ex-instrumentation-ruby_kafka:
    <<: *base
    working_dir: /app/instrumentation/ruby_kafka/example
    depends_on:
      - kafka
      - jaeger
   ...
   jaeger:
    image: jaegertracing/all-in-one
    ports:
      - "16686:16686"
      - "4318:4318" # open up this port for OTLP export
```

instrumentation/ruby_kafka/test/opentelemetry/instrumentation/ruby-kafka/patches/producer_test.rb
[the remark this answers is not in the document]
arielvalentin: How did I miss that? Thanks for pointing that out 🤦🏼

arielvalentin:
> > Is there a reason not to add the example directory there? https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/main/docker-compose.yml
> 
> Not conceptually, no, I just can't validate it with `docker-compose` (long story). `podman-compose` doesn't like for me to run `podman-compose run` against services that have a `depends_on:` block for whatever reason. I think the config would look like this:

Does that mean you are not able to use the other examples with podman?

plantfansam:
> > > Is there a reason not to add the example directory there? https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/main/docker-compose.yml
> > 
> > 
> > Not conceptually, no, I just can't validate it with `docker-compose` (long story). `podman-compose` doesn't like for me to run `podman-compose run` against services that have a `depends_on:` block for whatever reason. I think the config would look like this:
> 
> Does that mean you are not able to use the other examples with podman?

Yeah, when run from the root of the repo, that is the case (e.g. `podman-compose run ex-instrumentation-bunny`). `podman-compose up` works fine, so running instrumentation specs that rely on dependent services is no big deal.

arielvalentin:
Ok that's good to know. Would you mind writing up an issue for that?

arielvalentin:
Calling @robbkidd for exploratory testing!

plantfansam:
I'd like to add a few more tests to this this morning to ensure that all args passed properly from `async_producer#produce` to `producer#produce` across all supported ruby versions.

instrumentation/ruby_kafka/test/opentelemetry/instrumentation/ruby-kafka/patches/async_producer_test.rb
```diff
@@ -0,0 +1,57 @@
...
+        expect(producer).to have_received(:produce).with(
+          'hello',
+          key: "wat",
+          partition: 1,
+          partition_key: "ok",
+          create_time: create_time,
+          topic: topic,
+          headers: {
+            "foo" => "bar",
+            "traceparent" => "00-#{span.context.hex_trace_id}-#{span.context.hex_span_id}-01"
```
arielvalentin, line 52: What about tracestate? Should that be included?

plantfansam: This is very reasonable, OTOH we're just calling `OpenTelemetry.propagation.inject(options[:headers])` so ...maybe it's ok to just check for `traceparent`? Once we start checking for individual headers we're kinda unit testing `#inject`. Happy to go either way 🤷

arielvalentin: Ok. I need to bookmark this for myself and take a look at the spec, I.e. is tracestate a default propagator? 🤔

arielvalentin: Ok so this must be some configuration in the test suite we're missing?

plantfansam: > Ok so this must be some configuration in the test suite we're missing?

[We don't inject tracestate if it's unset in the propagator](https://github.com/open-telemetry/opentelemetry-ruby/blob/9e6d8772a5516454dd96eb29ba1d46689be1b025/api/lib/opentelemetry/trace/propagation/trace_context/text_map_propagator.rb#L31).

arielvalentin: Thank you!

instrumentation/ruby_kafka/lib/opentelemetry/instrumentation/ruby_kafka/patches/producer.rb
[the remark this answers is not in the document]
robbkidd: > Should we (info-level) log a warning in that circumstance?

What would the info log message say?

robbkidd:
Well, I like that child spans are connected to parents now.

I was able to run the example with docker-compose from the root of the repo.

* I updated the exporter endpoint in example/ruby_kafka.rb to `http://otelcol:4318/v1/traces`.
* I added the following service to root docker-compose.yml:

```yaml
  ex-instrumentation-ruby_kafka:
    <<: *base
    image: ruby:3.1-slim # protobuf is upset with the musl environment of *base's ruby alpine image
    depends_on:
      - kafka
    working_dir: /app/instrumentation/ruby_kafka/example
```
* Start it all up with `dcupd ex-instrumentation-ruby_kafka otelcol jaeger`
* Run the demo with `dce ex-instrumentation-ruby_kafka bash -c "bundle && bundle exec ruby ruby_kafka.rb"`

And I see complete traces in local Jaeger:

Screenshot 2023-07-10 at 5 18 09 PM

plantfansam:
I have updated the PR to have the synchronous `producer` always trust the headers it receives. My contention is that if someone is injecting headers into `produce`, they know what they're doing. It still looks fine in dev:

image

image

instrumentation/ruby_kafka/lib/opentelemetry/instrumentation/ruby_kafka/patches/producer.rb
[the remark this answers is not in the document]
plantfansam: Sorry, I totally missed this @robbkidd! I've changed the code a bit since I made this comment so I don't think a log is necessary...

instrumentation/ruby_kafka/lib/opentelemetry/instrumentation/ruby_kafka/patches/producer.rb
```diff
@@ -17,9 +17,14 @@ def produce(value, topic:, key: nil, headers: {}, partition: nil, partition_key:
               'messaging.destination_kind' => 'topic'
             }
 
-            tracer.in_span("#{topic} send", attributes: attributes, kind: :producer) do
-              OpenTelemetry.propagation.inject(headers)
-              super
+            # If trace context is present in headers, extract and use it as parent. The spec mandates that extract
+            # return the current context (Context.current) in Ruby, so this is a noop if propagation headers are absent.
+            ctx = OpenTelemetry.propagation.extract(headers)
```
arielvalentin, line 22: Now I'm second guessing. 

Is the _synchronous_ producer injecting the context into the as well? 

Is there no case where it should use the thread local context?

fbogsany: Comment covers this case (perhaps not totally clearly).

```ruby
ctx = OpenTelemetry.propagation.extract(headers)
```

is equivalent to:
```ruby
ctx = OpenTelemetry.propagation.extract(headers, context: Context.current)
```

On the sync `produce` path, unless the caller explicitly passed in `headers` that match the propagation format (there's a test for this case), the `headers` won't contain any trace context, so `ctx` will be `Context.current` after this call. Which means the `OpenTelemetry::Context.with_current(ctx)` below is a no-op - the `Context.current` will be whatever it was before the `extract` above.

It would be nice to have a test that makes this explicit.

plantfansam: Yeah, this implementation is perhaps a bit of inside baseball. I will add a test and also try to clarify the comment.

plantfansam: I've added the missing test here 1c6ba39688b705f050127163383447f8d1245ed9

instrumentation/ruby_kafka/lib/opentelemetry/instrumentation/ruby_kafka/patches/producer.rb
```diff
@@ -17,9 +17,15 @@ def produce(value, topic:, key: nil, headers: {}, partition: nil, partition_key:
               'messaging.destination_kind' => 'topic'
             }
 
-            tracer.in_span("#{topic} send", attributes: attributes, kind: :producer) do
-              OpenTelemetry.propagation.inject(headers)
-              super
+            # If trace context is present in headers, extract and use it as parent. If there is _no_ trace context key
+            # in the headers, OpenTelemetry.propagation.extract will return an unmodified copy of the the current
+            # Thread's context, so lines 24 and 25 preserve the correct Thread-local context.
```
fbogsany, line 22:
```suggestion
            # Thread's context, so the next two lines preserve the correct Thread-local context.
```

fbogsany (approved):

instrumentation/ruby_kafka/lib/opentelemetry/instrumentation/ruby_kafka/patches/producer.rb
plantfansam: @arielvalentin lmk if you want to re- 👀 before merge! Thanks for all your attention on the back-and-forth!

Issue #544, opened 2023-07-05
Sidekiq instrumentation not creating child spans with SDK 1.2.1


bolandoboy:
Hi! I'm using the Sidekiq instrumentation with the option `propagation_style: :child`. This works fine with `opentelemetry-sdk` 1.2.0 but fails when upgrading to 1.2.1.

"Fails" here means that spans are not exported, tested with a minimal example using the `console` exporter.

Tracking down the problem, I think it all comes down to the same root cause of https://github.com/open-telemetry/opentelemetry-ruby-contrib/issues/535

1. `opentelemetry-sdk` 1.2.1 introduced this check in `OpenTelemetry::SDK::Trace::Tracer#start_span` ([MR here](https://github.com/open-telemetry/opentelemetry-ruby/pull/1412)):

https://github.com/open-telemetry/opentelemetry-ruby/blob/1159ec04aa62168c6a841273401b71c325fa5986/sdk/lib/opentelemetry/sdk/trace/tracer.rb#L32

2. In the Sidekiq instrumentation, unless you use the option `trace_processor_process_one: true`, the context is set as `untraced`:

https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/cc244b9a1fff8bb2c79be90dd6bafd1a4212a426/instrumentation/sidekiq/lib/opentelemetry/instrumentation/sidekiq/patches/processor.rb#L16

3. By the time the `process` span is going to be created, `extracted_context` is `untraced`. (Please note that for the default `propagation_style: :link` a new root span with its own context is created, and that works)

https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/cc244b9a1fff8bb2c79be90dd6bafd1a4212a426/instrumentation/sidekiq/lib/opentelemetry/instrumentation/sidekiq/middlewares/server/tracer_middleware.rb#L33

4.  So for child spans `tracer.in_span` is used, which ends up calling the SDK `start_span` from point 1, which follow the `untraced` branch and, long story short, ends up creating a `OpenTelemetry::Trace::Span` instead of an `OpenTelemetry::SDK::Trace::Span`

`OpenTelemetry::Trace::Span` does nothing on `finish`, and spans are silently dropped:

https://github.com/open-telemetry/opentelemetry-ruby/blob/1159ec04aa62168c6a841273401b71c325fa5986/api/lib/opentelemetry/trace/span.rb#L154

[...]

Issue #535, opened 2023-06-30
Sidekiq integration can't be combined with other integrations


kevinnio:
Hi. I'm new to `opentelemetry-ruby` so I may be missing something. I tried to make this work for a while, I looked around in the repo's issues, everything to no avail so now I'm asking here.

Whenever I include the `sidekiq` instrumentation in my `configure` block inside my rails app spans from other instrumentations are no longer generated, or at least not visible in the console anymore.

Let's ilustrate the issue by starting with only `PG`:

```ruby
OpenTelemetry::SDK.configure do |c|
  c.service_name = 'MyApp'
  c.use 'OpenTelemetry::Instrumentation::PG'
end
```

`OTEL_TRACES_EXPORTER` is set to `console`.

```ruby
# app/models/my_model.rb
class MyModel < ApplicationModel
end

# app/jobs/my_job.rb
class MyJob 
  include Sidekiq::Job
  
  def perform
    MyModel.find_or_create_by!(id: 1)
  end
end

# in a rails console
MyJob.perform_async
```

Let's say I boot up sidekiq. As expected, I can see a bunch of PG traces in the console.

Now, let's add the `Sidekiq`:

```ruby
OpenTelemetry::SDK.configure do |c|
  c.service_name = 'MyApp'
  c.use 'OpenTelemetry::Instrumentation::PG'
  c.use 'OpenTelemetry::Instrumentation::Sidekiq'
end
```

Restart `sidekiq` and run the same job again. Now, only one single Sidekiq trace is visible in the console. The PG ones are missing. I expected to see one parent Sidekiq span and a bunch of PG child ones.

I tried with multiple instrumentations, not only `PG`. So far, this always happens.

I created an example app so you can replicate the issue: https://github.com/kevinnio/opentel-test. Please let me know if you need additional info.

`bundle info`:
```
* opentelemetry-sdk (1.2.1)
* opentelemetry-exporter-otlp (0.25.0)
* opentelemetry-instrumentation-pg (0.25.2)
* opentelemetry-instrumentation-sidekiq (0.24.1)
* rails (7.0.6)
* sidekiq (7.1.2)
* pg (1.5.3)
```

arielvalentin:
Thanks for creating the repo to help recreate this issue!

I do not have much experience using Sidekiq and I was wondering if you see this problem when running the sample application and export the spans to local collector. Do you see the spans then?

Here is an example to get you started:

[...]

PR #545, opened 2023-07-06
Remove patch for Sidekiq::Processor#process_one

kevinnio:
According to the discussion in #535, we can better match spec by avoiding the `sidekiq::processor#process_one` span altogether. This fixes the issues described in #535 and #544.

#### With `propagation_style: :link`
image

image

#### With `propagation_style: :child`
image

image


robertlaurin (approved):
I think this is fine to remove, but be aware it was added originally to silence background single span traces for redis.

So if you don't configure your redis to disable single span redis traces your overall trace and span volume may increase unexpectedly.

stevenharman:
> be aware it was added originally to silence background single span traces for redis.

@robertlaurin Could you explain what that means from the perspective of a Sidekiq user? And which [Redis instrumentation options](https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/cc244b9a1fff8bb2c79be90dd6bafd1a4212a426/instrumentation/redis/lib/opentelemetry/instrumentation/redis/instrumentation.rb#L22-L24) would we need to "disable single span redis traces"?

robbkidd:
I've run this PR on a minimal Rails+Sidekiq reproduction app and can confirm that the current changes work:

* `prop_style: :link` :: works as before
* `prop_style: :child` :: worker job processing spans (`default process`) appear as children of the job producer span (`default send`)

Screenshot 2023-07-06 at 5 09 42 PM

robbkidd:
@stevenharman If this PR merges in its current state, users of Sidekiq with otel-instrumentation-redis enabled will see a span from the Redis instrumentation for every time a Sidekiq worker polls for jobs. For example, this 10 minutes of the minimal reproduction app running a single Sidekiq worker process:

Screenshot 2023-07-06 at 5 07 46 PM

That's _a lot_ of spans that are not high value.

The "untraced" behavior flagged on/off by the option being removed in this PR was a knob provided by the Sidekiq instrumentation to silence any spans generated by Sidekiq's frequent use of Redis.

instrumentation/sidekiq/lib/opentelemetry/instrumentation/sidekiq/instrumentation.rb
```diff
@@ -32,7 +32,6 @@ class Instrumentation < OpenTelemetry::Instrumentation::Base
         option :trace_launcher_heartbeat,    default: false, validate: :boolean
         option :trace_poller_enqueue,        default: false, validate: :boolean
         option :trace_poller_wait,           default: false, validate: :boolean
-        option :trace_processor_process_one, default: false, validate: :boolean
```
robbkidd, line 35: Instead of removal, do we have a pattern for deprecating options? Even if we still remove the patch with this PR, maybe this option hangs around to emit a warning at initialization if it has been set?

kevinnio: I personally haven't seen any pattern for deprecation in this repo, but if there is I'm more than willing to use it instead of removing this option.

robbkidd:
Chatting about this at today's SIG meeting, I proposed that I explore a way to keep the instrumentation around process_one along with its untraced knob and to see if there is a way to prevent the "untracedness" of the _job polling_ context from leaking into the _job running_ context. I intend to get on that in the next day or so.

arielvalentin:
@robbkidd Do you any updates on this for us?

robbkidd:
I don't have any updates, but I'll get back on this.

kevinnio:
@robbkidd May I suggest a patch like this?

```ruby
def process_one
  if instrumentation_config[:trace_processor_process_one]
    attributes = {}
    attributes[SemanticConventions::Trace::PEER_SERVICE] = instrumentation_config[:peer_service] if instrumentation_config[:peer_service]
    tracer.in_span('Sidekiq::Processor#process_one', attributes: attributes) { super }
  else
    super
  end
end

def fetch
  if instrumentation_config[:trace_processor_process_one]
    super
  else
    OpenTelemetry::Common::Utilities.untraced { super }
  end
end
```

kevinnio:
@arielvalentin @robbkidd If it's ok with you, I'll submit a new PR with my suggested patch so we keep the `:trace_processor_process_one`, avoid redis noise and still show traces from inside jobs. I'll close this one for now.

arielvalentin:
Thank you @kevinnio for your patience with this.

PR #546, opened 2023-07-06
Add compatibility with sidekiq 7.0

kevinnio:
This adds a new mock loader that allows for testing against sidekiq 7.0. No actual code changes were needed to pass all the tests.

This also adds the relevant appraisal.

Included is a commit that removes some appraisals since they were effectively targeting the same sidekiq version.


arielvalentin (approved):

arielvalentin:
@kevinnio looks like the test suite never finishes and hangs at some point. 

Would you mind taking a look?

arielvalentin (changes requested):
Waiting on tear suite to complete successfully

robertlaurin (approved):

kaylareopelle:
Hi @kevinnio, Sidekiq 7.0 introduced a change to the middleware. Essentially, you're supposed to include some modules now in your middleware classes.

* Intro PR: https://github.com/sidekiq/sidekiq/issues/5371 
* Docs: https://github.com/sidekiq/sidekiq/blob/main/docs/middleware.md 

Did you look into adding this middleware? It might not be necessary for OTel's instrumentation. At New Relic, we needed it to keep our tests passing, but I don't quite recall the reason why. We [added it conditionally](https://github.com/newrelic/newrelic-ruby-agent/blob/dev/lib/new_relic/agent/instrumentation/sidekiq/client.rb#L7) to keep the instrumentation backward compatible. 

This _might_ be why the redis/sidekiq tests have some issues? It looks like `config.redis` is called [in the test helper](https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/5182c22ced6fb793249fd867af1a39d01809c3c0/instrumentation/sidekiq/test/test_helper.rb#L40-L46). I think that's only [accessible in 7.0 with the new middleware.](https://github.com/sidekiq/sidekiq/blob/main/lib/sidekiq/middleware/modules.rb)

kevinnio:
Hi, @kaylareopelle. Yes, I am aware of that new middleware. [This new middleware](https://github.com/sidekiq/sidekiq/blob/main/lib/sidekiq/middleware/modules.rb#L4) is currently only a shortcut for some common used methods. We don't call any of those methods in the instrumentation, so we don't need it right now. Nevertheless, I can include it just in case they change its behavior in future releases.

I'll also take a look at why the specs are not passing.

kevinnio:
Pushed a fix for the specs.

kevinnio:
@arielvalentin Specs are now fixed.

arielvalentin (approved):

PR #559, opened 2023-07-13
ci: group appraisal run output

robbkidd:
For some reason, GitHub Actions isn't using the name of a step in a
composite for the output group title. It uses the first line of the run
script. So repeat the name in the script.

Wrap each appraisal run for a gem in a `::group::title` to collapse the
output for a particular appraisement. Failures remain expanded because
the step stops there, before the `::endgroup::` is echoed to the console.

Maybe this will help output readability in CI?


robbkidd:
80s/90s Robb would like to apply more ASCII art to the collapsable group names …

```
⏵ Run # -=[ Install dependencies and generate appraisals ]=-
```

… and 00s+ Robb would like to liberally sprinkle emoji in there … . I'll stop here, though, unless there is interest in more eye candy in the CI output. Eye candy is _useful_ in this case, I think, in breaking up the wall of plain text that is the current CI output, but I know not every one is a fan.

robbkidd:
Yea, I couldn't resist emoji. New commit added to see how they'd feel.

🗳️ I like them. 😁

robbkidd (commented):
Yea, I couldn't resist emoji. New commit added to see how they'd feel.

🗳️:+1: I like them. 😁

arielvalentin (approved):
This is much nicer. Thank you.

robbkidd:
A sample of CI output changes. 

Will anyone miss the single, long-form Wall of Text generated by the appraisal runs? This PR does 2 things:

1. adds a comment as the first line of a step script so that what appears after `⏵  Run …` is distinct and meaningful
   a. … and rubs emoji on those lines to make them stand out when skimming
2. wraps each individual appraisal execution in a collapsable `::group::`

We can have 1. without 2. if there are folks who prefer not to expand for details of successful runs. 

| Before | After |
|--------|------|
| Screenshot 2023-07-13 at 3 42 30 PM | Screenshot 2023-07-13 at 3 41 58 PM |

PR #562, opened 2023-07-14
Add db.user to Trilogy spans

fbogsany:
This adds `db.user` to spans generated by Trilogy instrumentation. This is a [recommended connection-level attribute](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/database/database-spans.md#connection-level-attributes). It is useful for distinguishing read-only vs writer connections, for example.

### DIFF
--- a/instrumentation/trilogy/lib/opentelemetry/instrumentation/trilogy/patches/client.rb
+++ b/instrumentation/trilogy/lib/opentelemetry/instrumentation/trilogy/patches/client.rb
@@ -67,6 +67,7 @@ module OpenTelemetry
             }
 
             attributes[::OpenTelemetry::SemanticConventions::Trace::DB_NAME] = database_name if database_name
+            attributes[::OpenTelemetry::SemanticConventions::Trace::DB_USER] = database_user if database_user
             attributes[::OpenTelemetry::SemanticConventions::Trace::PEER_SERVICE] = config[:peer_service] unless config[:peer_service].nil?
             attributes['db.mysql.instance.address'] = @connected_host if defined?(@connected_host)
 
@@ -129,6 +130,10 @@ module OpenTelemetry
 
           def database_name
             connection_options[:database]
+          end
+
+          def database_user
+            connection_options[:username]
           end
 
           def tracer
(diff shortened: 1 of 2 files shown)


instrumentation/trilogy/lib/opentelemetry/instrumentation/trilogy/patches/client.rb
```diff
@@ -131,6 +132,10 @@ def database_name
             connection_options[:database]
           end
 
+          def database_user
+            connection_options[:username]
```
plantfansam, line 136: I think this is here https://github.com/trilogy-libraries/trilogy/blob/d7da1848c50bca69bccaadc3eed1b51bf8e1da72/contrib/ruby/ext/trilogy-ruby/cext.c#L444

plantfansam (approved):

instrumentation/trilogy/lib/opentelemetry/instrumentation/trilogy/patches/client.rb
```diff
@@ -131,6 +132,10 @@ def database_name
             connection_options[:database]
           end
 
+          def database_user
```
robertlaurin, line 135: Is it reasonable to be concerned with adding very generic helper method names to our instrumentation, do we worry about overriding an existing or potentially to be added method in the future?

fbogsany: It is reasonable to be concerned. I worry about all the helper methods here, but I convinced myself I'm not making the situation worse. I would like to find a cleaner way to implement these helpers in future.

plantfansam: I think this is reasonable and potentially solvable with `def _otel_database_user` or some such thing.

fbogsany: I plan to tackle this in a follow-on PR. I opened #563 to track it.

PR #565, opened 2023-07-14
ci: correct the appraised versions of redis

robbkidd:
Correcting some copy-pasta. Not adding 5.x because of deprecations yet to be addressed.

### DIFF
--- a/instrumentation/redis/Appraisals
+++ b/instrumentation/redis/Appraisals
@@ -1,23 +1,31 @@
 # frozen_string_literal: true
 
+# test against latest 4.x
+appraise 'redis-4.x' do
+  gem 'redis', '~> 4.0'
+end
+
+# and known previous minors
+appraise 'redis-4.7' do
+  gem 'redis', '~> 4.7.0'
+end
+
 appraise 'redis-4.6' do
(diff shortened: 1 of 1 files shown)


arielvalentin (commented):
Are all of these minor versions receiving maintainer updates?

robbkidd:
If this goes like it did in my dev environment, we'll see deprecations on `redis.queue` used in the test suite starting with v4.6.

robbkidd:
> Are all of these minor versions receiving maintainer updates?

I doubt it. I don't think there are maintained minor branches. I did the range as a first commit to so we could all see where things get weird if they get weird. For v4, I bet we can just ride with the ~> 4.x and dial the change in this PR down to just that prior to merge.

Adding v5 failed the test suite. I didn't pick that work up yet.

robbkidd:
You know what would make [this CI output](https://github.com/open-telemetry/opentelemetry-ruby-contrib/actions/runs/5558967974/jobs/10154611321?pr=565#step:4:70) nicer to look at? [Rollups and emoji](https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/559). 😀

PR #573, opened 2023-07-20
fix(que): Correctly set bulk_enqueue job options

j-asser:
Merge tags with the `job_options` provided to the `bulk_enqueue` method, not the one passed to `enqueue`, since job options cannot actually be passed to `enqueue` method inside `bulk_enqueue`. This resulted in `job_options` being ignored when using the bulk method.


arielvalentin:
@j-asser Thanks for your contribution!

Once you have signed the EasyCLA we will begin the process of reviewing this PR.

instrumentation/que/lib/opentelemetry/instrumentation/que/patches/que_job.rb
```diff
@@ -46,7 +46,7 @@ def enqueue(*args, job_options: {}, **arg_opts)
                 # In Que version 2.1.0 `bulk_enqueue` was introduced and in order
                 # for it to work, we must pass `job_options` to `bulk_enqueue` instead of enqueue.
                 if gem_version >= Gem::Version.new('2.1.0') && Thread.current[:que_jobs_to_bulk_insert]
-                  Thread.current[:que_jobs_to_bulk_insert][:job_options] = job_options.merge(tags: tags)
+                  Thread.current[:que_jobs_to_bulk_insert][:job_options] = Thread.current[:que_jobs_to_bulk_insert][:job_options].merge(tags: tags)
```
arielvalentin, line 49: Please add an automated test to ensure we do not introduce regressions in the future.

arielvalentin (commented):

instrumentation/que/lib/opentelemetry/instrumentation/que/patches/que_job.rb
j-asser: There already was a test for this behaviour but the test itself had an issue, where `job_options` were passed to `.enqueue` not `.job_enqueue`. Fixed in f4d5c4b

https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/f4d5c4b3f190b354f99dd41ff65db486beec4351/instrumentation/que/test/opentelemetry/instrumentation/que_test.rb#L413C13-L415C16

instrumentation/que/lib/opentelemetry/instrumentation/que/patches/que_job.rb
```diff
@@ -46,7 +46,8 @@ def enqueue(*args, job_options: {}, **arg_opts)
                 # In Que version 2.1.0 `bulk_enqueue` was introduced and in order
                 # for it to work, we must pass `job_options` to `bulk_enqueue` instead of enqueue.
                 if gem_version >= Gem::Version.new('2.1.0') && Thread.current[:que_jobs_to_bulk_insert]
-                  Thread.current[:que_jobs_to_bulk_insert][:job_options] = job_options.merge(tags: tags)
+                  Thread.current[:que_jobs_to_bulk_insert][:job_options] = bulk_job_options(tags)
```
arielvalentin, line 49: Similarly, could you issue dig with safe navigation? I also think it would be ok to in-line the helper method.

Despite our gratitude and desire for clean code, the less methods we dispatch to will result in better overall instrumentation performance and the smaller the stack trace when applications raise exceptions.

instrumentation/que/lib/opentelemetry/instrumentation/que/patches/que_job.rb
```diff
@@ -61,6 +62,12 @@ def enqueue(*args, job_options: {}, **arg_opts)
               end
             end
 
+            def bulk_job_options(tags)
+              Thread.current[:que_jobs_to_bulk_insert][:job_options].merge(tags: tags) do |_, a, b|
+                a.is_a?(Array) && b.is_a?(Array) ? a + b : b
```
arielvalentin, line 67: If I recall correctly plus will allocate a new array object. Please consider using  [concat](https://docs.ruby-lang.org/en/3.2/Array.html#method-i-concat), which will append the elements from one array to an existing array:

```suggestion
                a.is_a?(Array) && b.is_a?(Array) ? a.concat( 
b) : b
```

arielvalentin (approved):
Some nits but I think this makes sense to me.

instrumentation/que/lib/opentelemetry/instrumentation/que/patches/que_job.rb
```diff
@@ -61,6 +62,12 @@ def enqueue(*args, job_options: {}, **arg_opts)
               end
             end
 
+            def bulk_job_options(tags)
+              Thread.current[:que_jobs_to_bulk_insert][:job_options].merge(tags: tags) do |_, a, b|
```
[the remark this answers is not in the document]
j-asser: Thanks for the feedback! 
I made the changes in [4fdc4a1](https://github.com/open-telemetry/opentelemetry-ruby-contrib/commit/4fdc4a1ff2ddaaa0f2f209c464bf1f46cf80f96a) where possible. 
Unfortunately `dig` could not be used with `Thread.current` which is an array and rubocop wasn't happy with a single parameter `dig` use. Inlining the helper method required disabling the rubocop checks for `CyclomaticComplexity` and `PerceivedComplexity`.

indrekj (approved):

arielvalentin:
@j-asser looks like you have a rubocop linter error. Once you have corrected it I'll merge.

Thanks @indrekj for the review!

Issue #570, opened 2023-07-19
Return value of PG::Connection#column_values() is a PG::Result, not Array, with OpenTelemetry::Instrumentation::PG in use


Capncavedan:
Without `OpenTelemetry::SDK` configured to use `OpenTelemetry::Instrumentation::PG`, the return value of `PG::Connection#column_values` is an Array containing bare strings (at least when selecting a text/string column).

With `OpenTelemetry::Instrumentation::PG` in use, the return value is a `PG::Result` object, and calling `to_a` on it returns an array of hashes, each with the column header as key and column/row value as value.

[PG method documentation](https://deveiate.org/code/pg/PG/Result.html#method-i-column_values) states:

```
column_values( n ) → array
Returns an Array of the values from the nth column of each tuple in the result.
```

**Runtime**

Operating system details: macOS Ventura 13.4.1 (c)
RUBY_ENGINE: MRI
RUBY_VERSION: 3.2.2
RUBY_DESCRIPTION: ruby 3.2.2 (2023-03-30 revision e51014f9c0) [x86_64-darwin22]

**Gem versions**
pg (1.5.3)
opentelemetry-exporter-otlp (0.26.0)
opentelemetry-instrumentation-pg (0.25.2)
opentelemetry-sdk (1.3.0)

**Reproduction**

```rb
require 'pg'
require 'opentelemetry/sdk'
require 'opentelemetry/instrumentation/pg'
require 'opentelemetry-exporter-otlp'

if ENV['INSTRUMENT_PG'].to_s == 'true'
  OpenTelemetry::SDK.configure { |c| c.use 'OpenTelemetry::Instrumentation::PG' }
end

connection = PG.connect(dbname: ENV['NAME'], user: ENV['USER'], password: ENV['PASS'], host: ENV['HOST'], port: ENV['PORT'])

sql = "SELECT table_name FROM information_schema.tables WHERE table_schema = 'public' AND table_type = 'BASE TABLE'"

subject = connection.exec(sql) { |result| result.column_values(0) }

puts subject.class.name
puts subject.first.class.name
```

Run with e.g. `NAME=database HOST=localhost PORT=5432 USER=user PASS=password ruby ./repro.rb` against a database with at least one table in it, you get:
```
Array
String
```

Run with e.g. `INSTRUMENT_PG=true NAME=database HOST=localhost PORT=5432 USER=user PASS=password ruby ./repro.rb` against a database with at least one table in it, you get:
```
[...]

PR #574, opened 2023-07-20
fix: Pass block explicitly in `define_method` calls for PG instrumentation query methods

Capncavedan:
https://github.com/open-telemetry/opentelemetry-ruby-contrib/issues/570

"As a result of the fact blocks are captured by closures they cannot be passed implicitly to methods created with `define_method`; they must instead be passed explicitly using the `define_method(meth) { |args, &block| ... }` syntax."
from https://banisterfiend.wordpress.com/2010/11/06/behavior-of-yield-in-define_method/

This PR changes the "exec-ish" and "exec-prepared-ish" methods to all optionally handle an explicit block, with a new separate test file specifically for these PG::Connection patches.

This PR does not change the "prepare-ish" methods as those don't seem to return anything anyway.

### DIFF
--- a/instrumentation/pg/test/opentelemetry/instrumentation/pg/patches/connection_test.rb
+++ b/instrumentation/pg/test/opentelemetry/instrumentation/pg/patches/connection_test.rb
@@ -5,7 +5,6 @@
 # SPDX-License-Identifier: Apache-2.0
 
 require 'test_helper'
-require 'active_record'
 require 'pg'
 
 require_relative '../../../../../lib/opentelemetry/instrumentation/pg'
--- a/instrumentation/pg/test/opentelemetry/instrumentation/pg/patches/connection_test.rb
+++ b/instrumentation/pg/test/opentelemetry/instrumentation/pg/patches/connection_test.rb
@@ -1,0 +1,78 @@
+# frozen_string_literal: true
+
+# Copyright The OpenTelemetry Authors
+#
+# SPDX-License-Identifier: Apache-2.0
+
+require 'test_helper'
+require 'active_record'
+require 'pg'
+
+require_relative '../../../../../lib/opentelemetry/instrumentation/pg'
+require_relative '../../../../../lib/opentelemetry/instrumentation/pg/patches/connection'
+
+describe OpenTelemetry::Instrumentation::PG::Patches do
+  let(:instrumentation) { OpenTelemetry::Instrumentation::PG::Instrumentation.instance }
+  let(:config) { {} }
+  let(:host) { ENV.fetch('TEST_POSTGRES_HOST', '127.0.0.1') }
+  let(:port) { ENV.fetch('TEST_POSTGRES_PORT', '5432') }
+  let(:user) { ENV.fetch('TEST_POSTGRES_USER', 'postgres') }
+  let(:dbname) { ENV.fetch('TEST_POSTGRES_DB', 'postgres') }
+  let(:password) { ENV.fetch('TEST_POSTGRES_PASSWORD', 'postgres') }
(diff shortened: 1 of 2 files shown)


arielvalentin:
@Capncavedan thanks again for your help!

Please review and sign the Easy CLA so that we may start the PR review process.

Capncavedan:
@arielvalentin done

arielvalentin:
Thank you. Did you happen to have the ability to test this in a sample app? 

If so would you be able to share some screen shots of the fix in action?

Capncavedan:
@arielvalentin Only in the repro script so far.  This bug manifested up in a background process of ours, and we patched it to process its Postgres queries differently.

Apologies if this is already clear - the bug isn't affecting telemetry, rather one specific way of issuing Postgres queries using the PG gem.

instrumentation/pg/test/opentelemetry/instrumentation/pg/patches/connection_test.rb
```diff
@@ -0,0 +1,78 @@
+# frozen_string_literal: true
+
+# Copyright The OpenTelemetry Authors
+#
+# SPDX-License-Identifier: Apache-2.0
+
+require 'test_helper'
+require 'active_record'
```
arielvalentin, line 8: Where is `active_record` being used in these tests?

instrumentation/pg/test/opentelemetry/instrumentation/pg/patches/connection_test.rb
```diff
@@ -0,0 +1,78 @@
...
+require 'pg'
+
+require_relative '../../../../../lib/opentelemetry/instrumentation/pg'
+require_relative '../../../../../lib/opentelemetry/instrumentation/pg/patches/connection'
```
arielvalentin, line 12: Are these requires necessary? i.e. is the `test_helper` and bundler not requiring these already?

instrumentation/pg/test/opentelemetry/instrumentation/pg/patches/connection_test.rb
```diff
@@ -0,0 +1,78 @@
...
+
+describe OpenTelemetry::Instrumentation::PG::Patches do
```
arielvalentin, line 14: ✨ Thank you for adding tests for this!

arielvalentin (commented):
Thank you for this contribution!

Please update your commit messages to use conventional commits: https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/main/CONTRIBUTING.md#use-conventional-commit-messages

instrumentation/pg/test/opentelemetry/instrumentation/pg/patches/connection_test.rb
arielvalentin: NVM Looks like they are... 😕

kaylareopelle (commented):
This looks good to me once `require 'active_record'` is removed from the tests.

Capncavedan:
Fantastic. I’ll incorporate feedback early next week (I’m out this week) and update the PR. -DanOn Jul 25, 2023, at 14:12, kaylareopelle ***@***.***> wrote:﻿
@kaylareopelle commented on this pull request.

This looks good to me once require 'active_record' is removed from the tests.

plantfansam (approved):
I think this makes sense! Thank you for the tests ❤️

arielvalentin (commented):
Forgot to leave my review. I did some exploratory testing using the `example/pg.rb` script and things look good!

Thanks for your contribution. I will get a release out shortly.

Capncavedan:
Thanks @arielvalentin !

PR #587, opened 2023-07-28
fix: Resource Detectors Container release

arielvalentin:
We missed this during the review but it seems that the naming conventions did not quite match up.

This change will ensure we are able to break out individual resource detector gems from the monolithic gem.

cc: @scbjans @mwear


ericmustin (approved):

scbjans:
Thanks @arielvalentin! Happy that we now comply with conventions. I'll work on splitting the other ones into seperate gems somewhere next week now the structure is clear.

PR #590, opened 2023-07-31
chore: rename opentelemetry-resource-detector-container gem for symmetry

scbjans:
Renames the (currently unreleased) `opentelemetry-resource_detectors-container` gem to `opentelemetry-resource-detector-container`.
This is to achieve symmetry with e.g. [JS resource detector](https://github.com/open-telemetry/opentelemetry-js-contrib/tree/main/detectors/node/opentelemetry-resource-detector-container) and [Python resource detector](https://github.com/open-telemetry/opentelemetry-python-contrib/tree/main/resource/opentelemetry-resource-detector-container)

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -178,6 +178,6 @@ updates:
   schedule:
     interval: weekly
 - package-ecosystem: bundler
-  directory: "/resource_detectors/container"
+  directory: "/resources/container"
   schedule:
     interval: weekly
--- a/.github/workflows/ci-contrib-canary.yml
+++ b/.github/workflows/ci-contrib-canary.yml
@@ -59,7 +59,7 @@ jobs:
       matrix:
         gem:
           - resource_detectors
-          - resource_detectors-container
+          - resource-detector-container
         os:
           - ubuntu-latest
           - macos-latest
--- a/.github/workflows/ci-contrib.yml
+++ b/.github/workflows/ci-contrib.yml
@@ -55,7 +55,7 @@ jobs:
       matrix:
         gem:
           - resource_detectors
-          - resource_detectors-container
+          - resource-detector-container
         os:
           - ubuntu-latest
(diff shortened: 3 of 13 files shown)


arielvalentin (commented):
I think that we should keep the Detectors module plural to maintain symmetry with the existing gems as opposed to following naming conventions in other languages. 

I believe the changelog must also be left as an empty file in order for the release scripts to be able to generate the initial release properly. That is what prevented the initial release attempt.

scbjans:
Thanks for your quick review.
Changelog [should indeed be empty for a new gem](https://github.com/dazuma/toys/blob/c22362c7c206866997b52590d00823e511c7863e/.toys/release/.lib/release_requester.rb#L196) 

Regarding the singular/plural form of the `Detectors` module:
my take was that we're going to split off the other resource detectors into separate gems as well. Eventually, that would render the `Detectors` (plural) module empty; which allows for symmetry with other languages; and also some symmetry between `instrumentation` gems and `resource-detector` gems (both singular) within this repo as well.

My idea was to:
* prepare PR's for the other resource detectors to split off into their own gem
* a PR for a new version of the current `opentelemetry-resource_detectors` gem which would log a message upon initialization of the GCP or Azure Resource Detector, stating that they will be available in their own respective gems from now on
* a PR with the removal of the current all-encompassing `opentelemetry-resource_detectors`

To me, that seems to be the 'cleanest' and most consistent I can think of right now.
LMK if I should make `Detectors` plural; or if I should proceed with above steps. I'm happy either way

arielvalentin (approved):

arielvalentin (changes requested):
I approved prematurely. The CHANGELOG must only contain the `Release History` title. No version numbers of dates may be present.

scbjans:
> I approved prematurely. The CHANGELOG must only contain the `Release History` title. No version numbers of dates may be present.

addressed in 2367e329423d7764fd674b2165c13c26915fa38d

PR #591, opened 2023-07-31
fix: Add code semconv attributes

arielvalentin:
When using the default queue name it is often difficult to correlate the job that is performing the work.

Adding code semconv attributes makes it possible to use semconv span naming, but also identify spans by class name via `code.namespace`.


arielvalentin:
Thanks @plantfansam I think I have some issues to deal with w/r/t rubocop rules

PR #602, opened 2023-08-02
chore(deps): Upgrade rdkafka for tests

arielvalentin:
~Version 0.12 regularly segfaults due to race conditions.~

~Rather than include it in the matrix this change updates the test suite to test against the minimum supported version of 0.13 as well as an open ended appraisal to avoid situations where we fall behind on tests.~

Certain conditions cause v0.12 to segfault. I have addressed some of those issues in https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/606 but there are other cases unrelated to our incorrect usage of the gem in tests that also lead to segfaults. v0.13 should be more stable.


arielvalentin:
IMMEDIATELY SEGFAULTS

```
  7 runs, 40 assertions, 0 failures, 0 errors, 0 skips
  Segmentation fault (core dumped)
  rake aborted!
  Command failed with status (139)
  /opt/hostedtoolcache/Ruby/3.2.1/x64/bin/bundle:25:in `load'
  /opt/hostedtoolcache/Ruby/3.2.1/x64/bin/bundle:25:in `<main>'
  Tasks: TOP => test
  (See full trace by running task with --trace)
```

plantfansam (approved):

instrumentation/rdkafka/Appraisals
```diff
@@ -4,6 +4,10 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 
-appraise 'rdkafka-0.12.x' do
-  gem 'rdkafka', '~> 0.12.0'
+appraise 'rdkafka-0.13.x' do
+  gem 'rdkafka', '~> 0.13.0'
+end
```
arielvalentin, line 9:
```suggestion
appraise 'rdkafka-0.12.x' do
  gem 'rdkafka', '~> 0.12.0'
end

appraise 'rdkafka-0.13.x' do
  gem 'rdkafka', '~> 0.13.0'
end
```

PR #605, opened 2023-08-02
ci: Run Latest Gem Versions

arielvalentin:
This change updates the canary builds to only run against the latest version of gems.

This will give us a heads up when newer versions of a gem are released and are not covered by appraisals.


simi (approved):
:+1:

plantfansam (approved):

Issue #119, opened 2022-08-23
rdkafka test suite segfaults


arielvalentin:
The rdkafka test suite occasionally segfaults when run under Ruby 3.0. I would like to collect some data for the `appsignal` team.  

```console
Run cd "./instrumentation/rdkafka"
>> BUNDLE_GEMFILE=/home/runner/work/opentelemetry-ruby-contrib/opentelemetry-ruby-contrib/instrumentation/rdkafka/gemfiles/rdkafka_0.10.0.gemfile bundle exec rake test
Run options: --seed 63960

# Running:

......

Finished in 9.036991s, 0.6639 runs/s, 3.8730 assertions/s.

6 runs, 35 assertions, 0 failures, 0 errors, 0 skips
Segmentation fault (core dumped)
rake aborted!
Command failed with status (139)
/opt/hostedtoolcache/Ruby/3.0.4/x64/bin/bundle:23:in `load'
/opt/hostedtoolcache/Ruby/3.0.4/x64/bin/bundle:23:in `<main>'
Tasks: TOP => test
(See full trace by running task with --trace)
Error: Process completed with exit code 1.
```

Examples:

https://github.com/open-telemetry/opentelemetry-ruby-contrib/runs/7964713533?check_suite_focus=true#step:5:244
https://github.com/open-telemetry/opentelemetry-ruby-contrib/runs/7346131748?check_suite_focus=true#step:5:244

CC: https://github.com/appsignal/rdkafka-ruby/pull/211

thijsc:
I opened an issue for this: https://github.com/appsignal/rdkafka-ruby/issues/216

PR #606, opened 2023-08-02
ci: Fix rdkafka tests

arielvalentin:
Per this comment from @mensfeld

> found this: https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/main/instrumentation/rdkafka/test/opentelemetry/instrumentation/rdkafka/patches/consumer_test.rb
>
> It clearly lacks some finalization of work. consumers are not being closed and they leak out of those specs.
>
> https://github.com/appsignal/rdkafka-ruby/issues/216#issuecomment-1662279215

cc: @andyfleming

Fixes https://github.com/open-telemetry/opentelemetry-ruby-contrib/issues/119

### DIFF
--- a/instrumentation/rdkafka/.rubocop.yml
+++ b/instrumentation/rdkafka/.rubocop.yml
@@ -1,5 +1,7 @@
 inherit_from: ../../.rubocop.yml
 
+Lint/SuppressedException:
+  Enabled: false
 Metrics/ParameterLists:
   Enabled: true
   Exclude:
--- a/instrumentation/rdkafka/test/opentelemetry/instrumentation/rdkafka/patches/consumer_test.rb
+++ b/instrumentation/rdkafka/test/opentelemetry/instrumentation/rdkafka/patches/consumer_test.rb
@@ -8,7 +8,6 @@
 
 require 'test_helper'
 require 'securerandom'
-require 'pry'
 
 require_relative '../../../../../lib/opentelemetry/instrumentation/rdkafka'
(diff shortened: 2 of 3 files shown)


mensfeld:
I'm not saying this is the only cause but this is definitely one of them. If it continues to crash, I will look more.

mensfeld:
> We will keep you informed of any other issues we run into.

Anytime :)

> Now that I have your attention... It would be great to get your feedback on the code and if there may be better ways for us to integrate with rdkafka instead of using a monkey patching strategy.

That **really** depends on what you aim to instrument. rdkafka and underlying librdkafka are **hard** to instrument due to their async nature. I spent months on building reliable instrumentation pipelines for WaterDrop and Karafka to make sure that all errors are propagated and that appropriate metrics are being published.

The general stance on rdkafka-ruby from the maintainers is that it aims to be a base building block for more complex software (like Karafka) and it is suppose to be a binding to librdkafka to expose its APIs. What that means is that your patches should be ok and should be stable because those APIs should not change.

I do don't think that in near future we will come up with any instrumentation framework because of what I said. It would be outside of scope of this library because with instrumentation, some assumptions need to be made.

> Please let me know if you or anyone at app-signal would be available to help us with maintaining this instrumentation.

You can count me in.

robertlaurin (approved):

plantfansam:
🎉

mensfeld:
Great work guys! FYI if you ever encounter Ruby VM crashes please ping me immediately. Those bindings are subtle and I do my best to handle all the things gracefully but you know ;) people make mistakes

PR #615, opened 2023-08-08
Ensure Sidekiq TracerMiddleware is in expected order

stevenharman:
This change will ensure the middleware is put at the front of the Client and Server middleware chains, by default. This is a change from the existing behavior which will add them to the end of the chain, and that position is determined entirely by when the SDK is configured. In a Rails app, that usually means it depends on the name of the `config/initializers/*.rb` file `OpenTelemetry::SDK.configure` is called from.

I'm not sure if we should consider this a breaking change or not. I suppose _maybe_ as someone could have accidentally been depending on the order the middleware was kinda-sorta-randomly being installed. Thoughts?

### DIFF
--- a/instrumentation/sidekiq/lib/opentelemetry/instrumentation/sidekiq/instrumentation.rb
+++ b/instrumentation/sidekiq/lib/opentelemetry/instrumentation/sidekiq/instrumentation.rb
@@ -67,7 +67,7 @@ module OpenTelemetry
         def add_client_middleware
           ::Sidekiq.configure_client do |config|
             config.client_middleware do |chain|
-              chain.add Middlewares::Client::TracerMiddleware
+              chain.prepend Middlewares::Client::TracerMiddleware
             end
           end
         end
@@ -75,16 +75,16 @@ module OpenTelemetry
         def add_server_middleware
           ::Sidekiq.configure_server do |config|
             config.client_middleware do |chain|
-              chain.add Middlewares::Client::TracerMiddleware
+              chain.prepend Middlewares::Client::TracerMiddleware
             end
             config.server_middleware do |chain|
-              chain.add Middlewares::Server::TracerMiddleware
+              chain.prepend Middlewares::Server::TracerMiddleware
             end
           end
 
           if defined?(::Sidekiq::Testing) # rubocop:disable Style/GuardClause
             ::Sidekiq::Testing.server_middleware do |chain|
-              chain.add Middlewares::Server::TracerMiddleware
+              chain.prepend Middlewares::Server::TracerMiddleware
             end
           end
         end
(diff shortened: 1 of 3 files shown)


instrumentation/sidekiq/CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Release History: opentelemetry-instrumentation-sidekiq
 
+### Unrelease
+
+* FIXED: Ensure OpenTelemetry's Sidekiq `TracerMiddleware` is added to the front of the Client and Server middleware chains.
+
```
arielvalentin, line 6: We auto generate release notes from git commits as long as you follow Conventional Commits, so you do not have to manually set these. 

https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/main/CONTRIBUTING.md#use-conventional-commit-messages

instrumentation/sidekiq/lib/opentelemetry/instrumentation/sidekiq/instrumentation.rb
```diff
@@ -67,24 +67,24 @@ def patch_on_startup
         def add_client_middleware
           ::Sidekiq.configure_client do |config|
             config.client_middleware do |chain|
-              chain.add Middlewares::Client::TracerMiddleware
+              chain.prepend Middlewares::Client::TracerMiddleware
```
arielvalentin, line 70: How do we prevent this from regressing? Is there a automated test you would be able to submit with this to ensure that the module is the first in the ancestry?

How do we ensure it is always the first module in the list?

arielvalentin (changes requested):

instrumentation/sidekiq/CHANGELOG.md
stevenharman: Ack. Updated the commit message and rebased this away.

instrumentation/sidekiq/lib/opentelemetry/instrumentation/sidekiq/instrumentation.rb
stevenharman: > How do we prevent this from regressing? Is there a automated test you would be able to submit with this to ensure that the module is the first in the ancestry?

Because the mechanism of actually invoking the `install` block, which is what causes us to do the "install", depends on the machinery of a single `OpenTelemetry::SDK`, I'm not sure we can write a fully integrated test. But if we're OK trusting `OpenTelemetry::Instrumentation::Base::install` and `#install` to be stable and a public API, then yeah, I think I can work something out. Though I expect it's going to require some nasty `instance_variable_set`'s to reset the Sidekiq configuration state between tests.

> How do we ensure it is always the first module in the list?

In the general case, I'm not sure we can as we don't have a way to know when everyone is done configuring the Sidekiq middleware chains. In Rails-land, _maybe_ we could use a Railtie to wire up an after initialize hook to check. But even then, I'm not sure we'd want to as it's conceivable that someone has a use case where they need something else going even higher in the chain. Generic exception handling middleware (Bugsnag, Rollbar, etc… come to mind).

stevenharman: OK, I've added some tests - including some hackery to work with Sidekiq 5, 6, and 7. Phew!

arielvalentin (approved):

PR #617, opened 2023-08-10
fix: ensure that transaction name property is used, rather than self

MattAlp:
~~This ensures that if a class is dynamically instantiated (singleton class / metaclass) such as by the [Blazer](https://github.com/ankane/blazer/blob/ade703d022bc98c2bd1a1259010daab3043fd102/lib/blazer/adapters/sql_adapter.rb#L12) gem, the transaction name string is used rather than something like `#<Class:0x00007fc635f69cf0>(FooBar)`.~~

This ensures that if a class is dynamically instantiated (singleton class / metaclass) such as by the Blazer gem, the transaction name string is used for the code namespace rather than setting the span name to something like #<Class:0x00007fc635f69cf0>(FooBar)

### DIFF
--- a/instrumentation/active_record/lib/opentelemetry/instrumentation/active_record/patches/transactions_class_methods.rb
+++ b/instrumentation/active_record/lib/opentelemetry/instrumentation/active_record/patches/transactions_class_methods.rb
@@ -19,7 +19,7 @@ module OpenTelemetry
           # Contains ActiveRecord::Transactions::ClassMethods to be patched
           module ClassMethods
             def transaction(**options, &block)
-              tracer.in_span("#{self}.transaction") do
+              tracer.in_span('ActiveRecord.transaction', attributes: { 'code.namespace' => name }) do
                 super
               end
             end
--- a/instrumentation/active_record/test/instrumentation/active_record/patches/transactions_class_methods_test.rb
+++ b/instrumentation/active_record/test/instrumentation/active_record/patches/transactions_class_methods_test.rb
@@ -19,14 +19,33 @@ describe OpenTelemetry::Instrumentation::ActiveRecord::Patches::TransactionsClas
     it 'traces' do
       User.transaction { User.create! }
 
-      transaction_span = spans.find { |s| s.name == 'User.transaction' }
+      transaction_span = spans.find { |s| s.attributes['code.namespace'] == 'User' }
       _(transaction_span).wont_be_nil
     end
 
     it 'traces base transactions' do
       ActiveRecord::Base.transaction { User.create! }
 
-      transaction_span = spans.find { |s| s.name == 'ActiveRecord::Base.transaction' }
+      transaction_span = spans.find { |s| s.name == 'ActiveRecord.transaction' }
+      _(transaction_span).wont_be_nil
+    end
+
+    it 'traces dynamically created transaction classes' do
+      klass = Class.new(User) do
+        def self.name
+          'Klass'
+        end
+      end
+      klass.transaction { klass.create! }
+
+      transaction_span = spans.find { |s| s.attributes['code.namespace'] == 'Klass' }
+      _(transaction_span).wont_be_nil
+    end
+
+    it 'records transaction name as code namespace' do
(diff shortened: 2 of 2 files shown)


arielvalentin:
@MattAlp thank you for your contribution! 

Please review and sight the EasyCLA as well as add an automated test to mitigate any regressions in functionality.

instrumentation/active_record/lib/opentelemetry/instrumentation/active_record/patches/transactions_class_methods.rb
```diff
@@ -19,7 +19,7 @@ class << base
           # Contains ActiveRecord::Transactions::ClassMethods to be patched
           module ClassMethods
             def transaction(**options, &block)
-              tracer.in_span("#{self}.transaction") do
+              tracer.in_span("#{name}.transaction") do
```
plantfansam, line 22: Do you have a link to where this method is defined?

arielvalentin: I think this is `Class.name` in Ruby core since this module is a mixin for the Class methods

arielvalentin: Yeah... part of me wants to do something like 

`Span{name: activerecord.transaction, code.namespace: MyModel`

MattAlp: @robertlaurin?

robertlaurin: >The "namespace" within which `code.function` is defined. Usually the qualified class or module name, such that `code.namespace` + some separator + `code.function` form a unique identifier for the code unit.

From https://opentelemetry.io/docs/specs/otel/trace/semantic_conventions/span-general/#source-code-attributes

Seems like the right attribute. 

From a high level of readability in a trace viewer, seeing the more specific "#{Model}.transaction" has some ergonomic properties to it. 

From a general cardinality perspective around span naming, this is less desirable. In practice the amount of models that exist is rather bounded even if there are a lot of them in any particular application.

Where this gets a bit tricky is that the blazer adapter is appending the object id to the adapter name, which basically makes this unbounded cardinality... which is unfortunate for what we're doing here.

I talked with MattAlp about this recently, and I've gone back and forth since then. I think what I would like to see here is to implement Ariels suggestion of 

>Span{name: activerecord.transaction, code.namespace: MyModel}

Long term this is the direction I'd like to see.

plantfansam: Wow, what have I done??? 😛 

I think [@arielvalentin's suggestion](https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/617/files#r1290451630) makes sense 😄

ericmustin:
A wild @MattAlp appears!

MattAlp:
Hiya Eric

instrumentation/active_record/test/instrumentation/active_record/patches/transactions_class_methods_test.rb
```diff
@@ -19,14 +19,33 @@
...
       ActiveRecord::Base.transaction { User.create! }
 
-      transaction_span = spans.find { |s| s.name == 'ActiveRecord::Base.transaction' }
+      transaction_span = spans.find { |s| s.name == 'ActiveRecord.transaction' }
+      _(transaction_span).wont_be_nil
+    end
+
+    it 'traces dynamically created transaction classes' do
+      klass = Class.new(User) do
+        def self.name
```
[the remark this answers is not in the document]
MattAlp: We could.  I would imagine that we would want to keep it nil or empty, which would then surface in the observability tools / data and let folks decide if it's ok that this is happening

robertlaurin (approved):

PR #618, opened 2023-08-10
fix: improve GraphQL tracing compatibility with other tracers

swalkinshaw:
All of the graphql-ruby built-in tracers use the common [`PlatformTrace` module](https://github.com/rmosolgo/graphql-ruby/blob/15cf025a2fd34518529c95054a10d4be1985ff22/lib/graphql/tracing/platform_trace.rb). OpenTelemetry's tracer does not use it directly, but uses the same underlying instance variable and method names.

This can cause conflicts because the new tracing API mixes in trace modules into a single class. This means OpenTelemetry's trace module could override other ones (or vice versa).

This fix avoids conflicts entirely by using the `opentelemetry` prefix for some methods and instance variables instead of the ones used in `PlatformTrace`.

An alternative solution would be to refactor this implementation to actually use `PlatformTrace`. I chose against it some reason originally but I don't remember why 😅 I figured this is just easier for now.

cc @rmosolgo

### DIFF
--- a/instrumentation/graphql/lib/opentelemetry/instrumentation/graphql/tracers/graphql_trace.rb
+++ b/instrumentation/graphql/lib/opentelemetry/instrumentation/graphql/tracers/graphql_trace.rb
@@ -15,9 +15,9 @@ module OpenTelemetry
         module GraphQLTrace # rubocop:disable Metrics/ModuleLength
           def initialize(trace_scalars: false, **_options)
             @trace_scalars = trace_scalars
-            @platform_field_key_cache = Hash.new { |h, k| h[k] = platform_field_key(k) }
-            @platform_authorized_key_cache = Hash.new { |h, k| h[k] = platform_authorized_key(k) }
-            @platform_resolve_type_key_cache = Hash.new { |h, k| h[k] = platform_resolve_type_key(k) }
+            @_otel_field_key_cache = Hash.new { |h, k| h[k] = _otel_field_key(k) }
+            @_otel_authorized_key_cache = Hash.new { |h, k| h[k] = _otel_authorized_key(k) }
+            @_otel_resolve_type_key_cache = Hash.new { |h, k| h[k] = _otel_resolve_type_key(k) }
             super
           end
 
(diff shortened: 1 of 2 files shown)


swalkinshaw:
Going to write a test for this case

instrumentation/graphql/lib/opentelemetry/instrumentation/graphql/tracers/graphql_trace.rb
```diff
@@ -155,7 +155,7 @@ def trace_field?(field)
             end
           end
 
-          def platform_field_key(field)
+          def opentelemetry_field_key(field)
```
arielvalentin, line 158: I believe what we have settled on is prefixing private methods in patches or mixins with `_otel` so I will request that you use the same naming pattern, e.g. 

```suggestion
          def _otel_field_key(field)
```

instrumentation/graphql/lib/opentelemetry/instrumentation/graphql/tracers/graphql_trace.rb
```diff
@@ -15,9 +15,9 @@ module Tracers
         module GraphQLTrace # rubocop:disable Metrics/ModuleLength
           def initialize(trace_scalars: false, **_options)
             @trace_scalars = trace_scalars
-            @platform_field_key_cache = Hash.new { |h, k| h[k] = platform_field_key(k) }
-            @platform_authorized_key_cache = Hash.new { |h, k| h[k] = platform_authorized_key(k) }
-            @platform_resolve_type_key_cache = Hash.new { |h, k| h[k] = platform_resolve_type_key(k) }
+            @opentelemetry_field_key_cache = Hash.new { |h, k| h[k] = opentelemetry_field_key(k) }
```
arielvalentin, line 18: similarly for internal attribute names. 
```suggestion
            @_otel_field_key_cache = Hash.new { |h, k| h[k] = opentelemetry_field_key(k) }
```

instrumentation/graphql/lib/opentelemetry/instrumentation/graphql/tracers/graphql_trace.rb
swalkinshaw: 👍 done

arielvalentin (approved):

PR #625, opened 2023-08-19
feat: Add ability to configure propagation style per job/queue in Sidekiq instrumentation

kirillkaiumov:
Currently, propagation style can be configured for the whole app and it will be applied to all jobs. If `propagation_style` is set to `child`, then it's convenient to see a trace where the two "producer" and "consumer" spans are shown next to each other and there is a little (usually several ms) between them. 
However, in some cases, we schedule jobs to be performed over a longer period, for example in several days. Then if  `propagation_style` is set to `child` then visually the trace will be unreadable because there will be a huge gap between the spans where nothing happens.
This PR adds a new `propagation_style_per_job` option that allows dynamically choose propagation style per job or queue.

Example:
```ruby
c.use 'OpenTelemetry::Instrumentation::Sidekiq',
  propagation_style: :child,
  propagation_style_per_job: lambda { |job_class_str, _queue|
    :link if job_class_str == 'SomeJob'
  }
```

### DIFF
--- a/instrumentation/sidekiq/lib/opentelemetry/instrumentation/sidekiq/instrumentation.rb
+++ b/instrumentation/sidekiq/lib/opentelemetry/instrumentation/sidekiq/instrumentation.rb
@@ -29,6 +29,7 @@ module OpenTelemetry
 
         option :span_naming,                 default: :queue, validate: %I[job_class queue]
         option :propagation_style,           default: :link,  validate: %i[link child none]
+        option :propagation_style_per_job,   default: nil,   validate: :callable
         option :trace_launcher_heartbeat,    default: false, validate: :boolean
         option :trace_poller_enqueue,        default: false, validate: :boolean
         option :trace_poller_wait,           default: false, validate: :boolean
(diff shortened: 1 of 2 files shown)


kirillkaiumov:
@arielvalentin hey, I've opened the PR with working implementation but wanted to first get a feedback before moving further with tests, etc. I tested the changes manually and they work well. Please let me know if anything should be adjusted. Thanks!

arielvalentin:
👋🏼 @kirillkaiumov thank you for your submission. 

We require that you review and sign the EasyCLA before we accept any PRs. When you have a moment please do so and we can start iterating on reviews for this PR.

arielvalentin:
We discussed this a bit during the Ruby SIG this on 2023-08-22 and we did not come to a consensus on how or if we want to allow for per worker propagation configurations. 

@robbkidd @no-reply @bolandoboy @pollosp have been looking at the updates to from the messaging SIG and will be working on aligning the instrumentations to ensure they are spec complaint. Although I think it overlaps with the use cases described in the messaging spec, it does not prescribe how to configure automatic instrumentations or the granularity of instrumentation config. 

That being said I will defer to the folks I mentioned above to get this input on a design direction here. 

Please stay tuned

kirillkaiumov:
@robbkidd @no-reply @bolandoboy @pollosp hey! It's been 3 weeks already. Any feedback on the proposal that you can already provide?

PR #639, opened 2023-09-01
refactor: split resource_detectors into their own gems.

scbjans:
This refactoring results in the current resource detectors split into their own separate gems (as is already the case for the [Container Resource Detector](https://github.com/open-telemetry/opentelemetry-ruby-contrib/tree/main/resources/container)).

The current all-encompassing gem will still be available for now. When someone still uses it; a WARN log message will be shown stating that the new gems should be used onwards.

In a subsequent PR the all-encompassing gem (`opentelemetry-resource_detectors`) will be removed.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -178,6 +178,14 @@ updates:
   schedule:
     interval: weekly
 - package-ecosystem: bundler
+  directory: "/resources/azure"
+  schedule:
+    interval: weekly
+- package-ecosystem: bundler
   directory: "/resources/container"
+  schedule:
+    interval: weekly
+- package-ecosystem: bundler
+  directory: "/resources/google_cloud_platform"
   schedule:
     interval: weekly
--- a/.github/workflows/ci-contrib-canary.yml
+++ b/.github/workflows/ci-contrib-canary.yml
@@ -64,7 +64,9 @@ jobs:
       matrix:
         gem:
           - resource_detectors
+          - resource-detector-azure
           - resource-detector-container
+          - resource-detector-google_cloud_platform
         os:
           - ubuntu-latest
           - macos-latest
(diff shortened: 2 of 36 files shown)


arielvalentin (commented):
Thanks for doing this!

Consider using `post_install_message` and let folks know about this change when they are installing gems: https://guides.rubygems.org/specification-reference/#post_install_message

May I also confirm that we are keeping the all-in-one resources detectors gem or are we going to phase out the all-in-one? 

If we are keeping the all-in-one gem, will that reference the other gems once they are published? 

If so then these should appear ahead of the all-in-one gem in the releases.yml

scbjans:
Thanks for pointing me to `post_install_message`; I've added it in https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/639/commits/5b8a3775a37aa7fb607f8cc748d98b5842320bb8

From my perspective we'll completely remove the `opentelemetry-resource_detectors` gem. The current resource detectors in that gem are already mutually exclusive; hence I don't think it'll make sense to have an all-in-one gem for resource detectors.

We'll first need a new release of the `opentelemetry-resource_detectors` gem, which contains the deprecation warnings. I'll create a follow-up PR after that to completely remove the all-in-one gem, if agreed.

arielvalentin (approved):

arielvalentin:
Let's leave the source in place all in one gem in place for now so folks can find it when they navigate to this repo.

Once this PR is merged and gems are released, then the follow up will be to remove the all-in-one from CI and the release process.

arielvalentin:
@scbjans Mind taking a look at the merge conflicts? Once you have those resolved I will merge.

arielvalentin:
@scbjans It looks like the merge commit did not work as I expected. 

Would you mind rebasing [upstream main](https://github.com/open-telemetry/opentelemetry-ruby-contrib/tree/main) onto your branch?

PR #642, opened 2023-09-03
perf: Reduce Object allocation

arielvalentin:
The support subscriber was allocating multiple arrays per valid attribute combination, as well as attempted to compact the array to remove `nil` values for each invalid attribute, then converted the arrays into a hash.

This change removes the intermediate arrays and instead uses a hash to store the attributes, which means there isn't a need to compact the accumulating variable.


plantfansam (approved):
Nice!

Issue #561, opened 2023-07-13
Align messaging instrumentation operation names with semantic conventions


plantfansam:
Per [the specification](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/messaging.md#operation-names), operation names around messaging systems should be `publish`, `receive`, and `process`

Grepping the repo, I see `send` spans in:

- [ ]  [activejob](https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/912b3b97ddd68a61cfc7f749ffdf10c31e83afaa/instrumentation/active_job/lib/opentelemetry/instrumentation/active_job/patches/active_job_callbacks.rb#L17)
- [ ]  [aws sdk](https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/22796cb8141b3100c6ce94c25a15e31df3313464/instrumentation/aws_sdk/lib/opentelemetry/instrumentation/aws_sdk/handler.rb#L98)
- [ ] [bunny](https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/38ff1eacad26c4a803940f3849eb2d2e0c1ab4b9/instrumentation/bunny/lib/opentelemetry/instrumentation/bunny/patch_helpers.rb#L19)
- [ ]  [delayed job](https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/8f40eda20e0257bc740508b4b50758285d93e895/instrumentation/delayed_job/lib/opentelemetry/instrumentation/delayed_job/plugins/tracer_plugin.rb#L23)
- [ ]  [queue](https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/6e89c92f189bc6e187da06ea2af4e38531b93601/instrumentation/que/lib/opentelemetry/instrumentation/que/patches/que_job.rb#L57)
- [ ]  [racecar](https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/7b87ce557ed13ad80d135348050a64042d423165/instrumentation/racecar/lib/opentelemetry/instrumentation/racecar/patches/consumer.rb#L22)
- [ ]  [rdkafka](https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/a84067bb6c8404a4c784b7291e16985ab859010d/instrumentation/rdkafka/lib/opentelemetry/instrumentation/rdkafka/patches/producer.rb#L22)
[...]

PR #648, opened 2023-09-04
fix(instrumentation): Align messaging instrumentation operation names

michal-kazmierczak:
Related to https://github.com/open-telemetry/opentelemetry-ruby-contrib/issues/561

This PR renames spans related to the client-side operation of creating a job from `* send` to `* publish` in various queuing instrumentations.

The diff might seem sizeable but I believe the changes are very repetitive thus it was better to bundle it together.


arielvalentin:
Thank you for contributing to the project! 

Please review and sign the Easy CLA and then we will bring the process of reviewing this PR.

arielvalentin:
@michal-kazmierczak We probably want to give our end users a heads up because the span name is changing, so I am going to treat this like a Breaking Change, i.e. merge the commits with `fix!`

arielvalentin (approved):

PR #650, opened 2023-09-05
fix: access rdkafka message headers with strings instead of symbols

Ruto8:
For context, we're using both `rdkafka-ruby` and `opentelemetry-ruby-contrib` and we get a lot of warnings like this in the logs:

```
rdkafka deprecation warning: header access with Symbol key :traceparent treated as a String. Please change your code to use String keys to avoid this warning. Symbol keys will break in version 1.
```

I tracked it down to this [line from `rdkafka-ruby`](https://github.com/appsignal/rdkafka-ruby/blob/70e86510ef7903ca0ea041f79bae74c973184545/lib/rdkafka/consumer/headers.rb#L10-L11), where it seems they have deprecated accessing headers with Symbols.

In this PR, I removed the specific getter set to `symbol_key_getter` for the rdkafka consumer patch. This is also similar to what is done with [ruby-kafka](https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/e37d5afb3a274a3f487700da5f15a99ee1befd3a/instrumentation/ruby_kafka/lib/opentelemetry/instrumentation/ruby_kafka/patches/consumer.rb#L26).

I believe the getter will default to `text_map_getter`, so we don't have to explicitly set it here.


arielvalentin:
@Ruto8 Would you kindly take a look at this failing test?

https://github.com/open-telemetry/opentelemetry-ruby-contrib/actions/runs/6086303974/job/16513507151?pr=650

Ruto8:
> @Ruto8 Would you kindly take a look at this failing test?

@arielvalentin thank you for pointing this out. Unfortunately, the tests pass locally for me. I'm able to access the span, as expected. Can there be some differences between CI and local setup here? Is it possible it's a flaky test?

arielvalentin:
Did you try running the test with the seed option that's failing?

`Run options: --seed 35683`

Or running the test in a loop to see if it's non-deterministic? Something like

```bash

for i in {1..20}; do
  bundle exec rake test
done

```

PR #659, opened 2023-09-07
Remove opentelemetry-resource_detectors all-in-one gem

scbjans:
Follow-up PR to https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/639

Now that `opentelemetry-resource_detectors` is split into dedicated resource detector gems, _and_ the latest gem is marked as deprecated; this gem can be removed.

Do we want to keep the sources in place for people to still find? Or should we just delete the sources? ('d prefer the latter to keep things clean; if someone is really interested they can always get it from history)


arielvalentin:
I would like to wait at least a week before merging this.

arielvalentin (approved):

arielvalentin:
@scbjans would you mind resolving the conflicts? Once you do that I will merge this PR.

PR #661, opened 2023-09-07
fix(bunny): headers property needs to exist

maciekm:
bunny instrumentation errors when rabbitmq messages don't have `headers` property, e.g. published by an app that isn't instrumented yet.

```
E, [2023-09-07T12:01:06.688026 #1] ERROR -- #<Bunny::Session:0x29d88 hidden@hidden:5672, vhost=hidden, addresses=[hidden:5672]>: Uncaught exception from consumer #<Bunny::Consumer:101200 @channel_id=1 @queue=hidden @consumer_tag=hidden>: #<NoMethodError: undefined method `[]' for nil:NilClass 

carrier[key]
^^^^^> @ /gems/ruby/3.1.0/gems/opentelemetry-api-1.2.2/lib/opentelemetry/context/propagation/text_map_getter.rb:16:in `get'
```

you can reproduce the issue by spiting the [example](https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/main/instrumentation/bunny/example/bunny.rb) file into a publisher without instrumentation and a consumer with.

### DIFF
--- a/instrumentation/bunny/test/test_helper.rb
+++ b/instrumentation/bunny/test/test_helper.rb
@@ -19,4 +19,5 @@ OpenTelemetry::SDK.configure do |c|
   c.error_handler = ->(exception:, message:) { raise(exception || message) }
   c.logger = Logger.new($stderr, level: ENV.fetch('OTEL_LOG_LEVEL', 'fatal').to_sym)
   c.add_span_processor SPAN_PROCESSOR
+  c.propagators = [OpenTelemetry::Trace::Propagation::TraceContext.text_map_propagator]
 end
--- a/instrumentation/bunny/lib/opentelemetry/instrumentation/bunny/patch_helpers.rb
+++ b/instrumentation/bunny/lib/opentelemetry/instrumentation/bunny/patch_helpers.rb
@@ -35,6 +35,7 @@ module OpenTelemetry
         def self.extract_context(properties)
           # use the receive span as parent context
           parent_context = OpenTelemetry.propagation.extract(properties[:tracer_receive_headers])
+          return [parent_context, nil] if properties[:headers].nil?
 
           # link to the producer context
           producer_context = OpenTelemetry.propagation.extract(properties[:headers])
--- a/instrumentation/bunny/test/opentelemetry/instrumentation/bunny/patch_helpers_test.rb
+++ b/instrumentation/bunny/test/opentelemetry/instrumentation/bunny/patch_helpers_test.rb
@@ -1,0 +1,36 @@
+# frozen_string_literal: true
+
+# Copyright The OpenTelemetry Authors
+#
+# SPDX-License-Identifier: Apache-2.0
+
+require_relative '../../../../lib/opentelemetry/instrumentation/bunny/patch_helpers'
+
+describe OpenTelemetry::Instrumentation::Bunny::PatchHelpers do
+  let(:properties) do
+    {
+      headers: {
+        'traceparent' => '00-eab67ae26433f603121bd5674149d9e1-2007f3325d3cb6d6-01'
+      },
+      tracer_receive_headers: {
+        'traceparent' => '00-cd52775b3cb38931adf5fa880f890c25-cddb52a470027489-01'
+      }
+    }
+  end
+
+  describe '.extract_context' do
+    it 'returns the parent context with links when headers from producer exists' do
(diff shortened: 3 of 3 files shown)


instrumentation/bunny/lib/opentelemetry/instrumentation/bunny/version.rb
```diff
@@ -7,7 +7,7 @@
 module OpenTelemetry
   module Instrumentation
     module Bunny
-      VERSION = '0.21.0'
```
arielvalentin, line 10: Our release processes manages versions. Please revert this change.

arielvalentin (changes requested):

instrumentation/bunny/lib/opentelemetry/instrumentation/bunny/patches/channel.rb
```diff
@@ -18,6 +18,8 @@ def basic_get(queue, opts = { manual_ack: false })
 
               return [delivery_info, properties, payload] unless delivery_info
 
+              properties[:headers] ||= {}
```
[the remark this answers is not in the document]
maciekm: @arielvalentin thanks for reviewing, I've now switched to a guard clause where it's being extracted.

still trying to figure out how to test that, any suggestions?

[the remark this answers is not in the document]
arielvalentin: I am not familiar with this instrumentation, Bunny, or RabbitMQ so I don't have any advice for you. 

Any suggestions @open-telemetry/ruby-contrib-approvers

instrumentation/bunny/test/test_helper.rb
```diff
@@ -19,4 +19,5 @@
   c.error_handler = ->(exception:, message:) { raise(exception || message) }
   c.logger = Logger.new($stderr, level: ENV.fetch('OTEL_LOG_LEVEL', 'fatal').to_sym)
   c.add_span_processor SPAN_PROCESSOR
+  c.propagators = [OpenTelemetry::Trace::Propagation::TraceContext.text_map_propagator]
```
maciekm, line 22: while trying to test I noticed it is only an issue when a single propagator is configured:
```
  1) Error:
OpenTelemetry::Instrumentation::Bunny::PatchHelpers::.extract_context#test_0002_returns the parent context with no links when headers from producer not present:
NoMethodError: undefined method `[]' for nil:NilClass

          carrier[key]
                 ^^^^^
    .../gems/opentelemetry-api-1.2.2/lib/opentelemetry/context/propagation/text_map_getter.rb:16:in `get'
```
when more than one is used, e.g. the default `OTEL_PROPAGATORS=baggage,tracecontext` then it's only a warning:
```
.D, [2023-09-08T12:52:54.081854 #82070] DEBUG -- : Error extracting W3C baggage: undefined method `[]' for nil:NilClass

          carrier[key]
                 ^^^^^
W, [2023-09-08T12:52:54.082129 #82070]  WARN -- : Error in CompositePropagator#extract undefined method `[]' for nil:NilClass

          carrier[key]
                 ^^^^^
```

instrumentation/bunny/lib/opentelemetry/instrumentation/bunny/patches/channel.rb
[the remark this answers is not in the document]
maciekm: @arielvalentin test added

instrumentation/bunny/test/test_helper.rb
arielvalentin: @open-telemetry/ruby-maintainers Anything stand out to you about this w/r/t the number of propagators being configured?

maciekm: @arielvalentin could we get this one or https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/664 merged? Both fix the bug.

arielvalentin: I'll give this another look tomorrow. 

I'm not a bunny user so I was trying to get the attention of the original contributors to have a look since they may have additional insights and concerns. 

Thanks again for your contribution and patience.

arielvalentin (approved):
@maciekm I have not had much success in tracking down or getting feedback from the original contributors.

Would you be willing to take over as the maintainer of this instrumentation?

Also would you be able to look into why this instrumentation has multiple sets of Trace Context Propagation Carries in the properties i.e. `headers` and `tracer_receive_headers`?

PR #665, opened 2023-09-15
Micro optimization: build Hash w/ `{}`

jbourassa:
For cases where all keys are present, create the Hash with all its keys with `{...}` instead of building it incrementally with multiple `Hash[]` calls.

It really is a micro-optimization, but `execute_field` is on the hot path.

Bench:
```ruby

require "benchmark/ips"

Benchmark.ips do |x|
  x.report("[]=") do
    hash = {}
    hash["a"] = 1
    hash["b"] = 2
    hash["c"] = 3
  end

  x.report("{}") do
    { "a" => 1, "b" => 2, "c" => 3 }
  end

  x.compare
```

Result:

```
Warming up --------------------------------------
                  []=  577.660k i/100ms
                  {}     1.438M i/100ms
                  []=     5.809M (± 2.9%) i/s -     29.461M in   5.076542s
                  {}     15.373M (± 0.3%) i/s -     77.651M in   5.051278s

Comparison:
                  {}=: 15372655.4 i/s
                  []:   5808733.8 i/s - 2.65x  slower
```

### DIFF
--- a/instrumentation/graphql/lib/opentelemetry/instrumentation/graphql/tracers/graphql_trace.rb
+++ b/instrumentation/graphql/lib/opentelemetry/instrumentation/graphql/tracers/graphql_trace.rb
@@ -75,10 +75,11 @@ module OpenTelemetry
             platform_key = _otel_execute_field_key(field: field)
             return super unless platform_key
 
-            attributes = {}
-            attributes['graphql.field.parent'] = field.owner&.graphql_name
-            attributes['graphql.field.name'] = field.graphql_name
-            attributes['graphql.lazy'] = false
+            attributes = {
+              'graphql.field.parent' => field.owner&.graphql_name,
+              'graphql.field.name' => field.graphql_name,
+              'graphql.lazy' => false
+            }
 
             tracer.in_span(platform_key, attributes: attributes, &block)
           end
@@ -87,10 +88,11 @@ module OpenTelemetry
             platform_key = _otel_execute_field_key(field: field)
             return super unless platform_key
 
-            attributes = {}
-            attributes['graphql.field.parent'] = field.owner&.graphql_name
-            attributes['graphql.field.name'] = field.graphql_name
-            attributes['graphql.lazy'] = true
+            attributes = {
+              'graphql.field.parent' => field.owner&.graphql_name,
+              'graphql.field.name' => field.graphql_name,
+              'graphql.lazy' => true
+            }
 
             tracer.in_span(platform_key, attributes: attributes, &block)
           end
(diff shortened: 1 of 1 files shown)


instrumentation/graphql/lib/opentelemetry/instrumentation/graphql/tracers/graphql_trace.rb
```diff
@@ -75,10 +75,11 @@ def execute_field(field:, query:, ast_node:, arguments:, object:, &block)
             platform_key = _otel_execute_field_key(field: field)
```
jbourassa, line 75: Unrelated to this exact diff but is there a reason we can't cache this, like it's done in other fields? I see it depends on `field.trace`, `field.type`, `@trace_scalars`, all of which _look_ stable.

arielvalentin: That would be better answered by @swalkinshaw or @rmosolgo

swalkinshaw: The actual key lookup is cached within the method, but yeah `trace_field?` is called each time and I think it could be cached 👍

arielvalentin: While I have your eyes on this @swalkinshaw... would you mind providing a review?

swalkinshaw (approved):
This was carried over from the other file I ported where the individual assigment was necessary: https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/d77b700e6be06dec0cf44c39a603cc4f36966582/instrumentation/graphql/lib/opentelemetry/instrumentation/graphql/tracers/graphql_tracer.rb#L88-L107

So this makes sense anyway even if it wasn't faster!

instrumentation/graphql/lib/opentelemetry/instrumentation/graphql/tracers/graphql_trace.rb
```diff
@@ -87,10 +88,11 @@ def execute_field_lazy(field:, query:, ast_node:, arguments:, object:, &block)
...
             return super unless platform_key
 
-            attributes = {}
-            attributes['graphql.field.parent'] = field.owner&.graphql_name
-            attributes['graphql.field.name'] = field.graphql_name
-            attributes['graphql.lazy'] = true
+            attributes = {
+              'graphql.field.parent' => field.owner&.graphql_name,
```
arielvalentin, line 92: Something I did not notice before is that this line and `owner` may be `nil`.

That will result in a noisy error handling when trying to add the attributes to the span: 

https://github.com/open-telemetry/opentelemetry-ruby/blob/37e225ac63fef4617fc78bf1f638417eeb6a6e91/sdk/lib/opentelemetry/sdk/internal.rb#L57

A `compact!`, or equivalent, would be appreciated before sending this along to `in_span`

arielvalentin (approved):

PR #677, opened 2023-09-28
feat!(active_job): Use ActiveSupport instead of patches

arielvalentin:
This PR changes the `ActiveJob` instrumentation to use `ActiveSupport::Notifications` instead of using Monkey Patches.

## What's new?

1. The instrumentation now generates new internal spans for most of the ActiveJob notifications like `discard` and `retry_stopped`. 
2. ActiveJob specific attributes are now namespaced under the `messaging.active_job.*`

### Example Parent-Child propagation 

image

image

### Example Link Propagation

image

image

See https://github.com/open-telemetry/opentelemetry-ruby-contrib/issues/218


arielvalentin:
Now that I have rewritten these to use AS Notifications I have been made aware of a few idiosyncratic things in rails 6 that makes it difficult to maintain parity with 7.

In Rails 6, `perform_start` is invoked _prior_ to `perform` https://github.com/rails/rails/blob/v6.1.7.6/activejob/lib/active_job/instrumentation.rb#L13 

The callbacks are called _before_ the instrumentation events are invoked so they are not included in the timing 
https://github.com/rails/rails/blob/v6.1.7.6/activejob/lib/active_job/execution.rb#L47

This is not ideal but in order to maintain _correctness_ with 6.1 I will change this instrumentation to work with Rails 7+ and leave the original implementation in place for Rails 6.1.

Hopefully we have more luck with other framework components.

kaylareopelle (commented):
I'm still getting familiar with this code, so I will continue reviewing, but here's some low-level first-pass feedback.

instrumentation/active_job/Appraisals
```diff
@@ -4,14 +4,8 @@
...
-  gem 'activejob', '~> 6.0.0'
-end
-
-appraise 'activejob-6.1' do
-  gem 'activejob', '~> 6.1.0'
-end
-
-appraise 'activejob-7.0' do
-  gem 'activejob', '~> 7.0.0'
+%w[6.0.0 6.1.0 7.0.0 7.1.0].each do |version|
```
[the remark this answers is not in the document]
arielvalentin: Just keeping that's been merged thus far. The 6.0 was in a separate PR so I wanted to keep this the same until landing the other PR

instrumentation/active_job/lib/opentelemetry/instrumentation/active_job.rb
```diff
@@ -11,6 +11,39 @@ module OpenTelemetry
   module Instrumentation
     # Contains the OpenTelemetry instrumentation for the ActiveJob gem
     module ActiveJob
```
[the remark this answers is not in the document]
arielvalentin: Yes!

I was of thinking of making a generic "ingress" span Context that would be instrumentation agnostic that way any child span is able to access the ingress span and enrich it.

So instead of saying `Rack.current_span` or `ActiveJob.current_span` it would be `Ingress.current_span`.

instrumentation/active_job/lib/opentelemetry/instrumentation/active_job/handlers.rb
```diff
@@ -0,0 +1,74 @@
...
+
+          tracer = OpenTelemetry.tracer_provider.tracer(ActiveJob.name, ActiveJob::VERSION)
+          mapper = Mappers::Attribute.new
+          config = ActiveJob::Instrumentation.instance.config
+          parent_span_provider = OpenTelemetry::Instrumentation::ActiveJob
+
+          # TODO, use delegation instead of inheritance
+          default_handler = Handlers::Default.new(tracer, parent_span_provider, mapper, config)
+          enqueue_handler = Handlers::Enqueue.new(tracer, parent_span_provider, mapper, config)
+          perform_handler = Handlers::Perform.new(tracer, parent_span_provider, mapper, config)
```
fbogsany, line 49: Our general pattern in instrumentation is to always reach for the `tracer` from the instrumentation _when we need it_ rather than caching it in various helpers. I think the deviation from that code pattern here is the trigger for https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/688

arielvalentin: Yessir! This was intentional on my part in the hopes of starting a new tread 😄 

I wanted to use dependency injection for everything that the handlers needed instead of using a service locator or a mix of the two. 

Though we are hashing out some details in #688, do you think having this asymmetry is problematic?

bensheldon (commented):
I reviewed this PR, it looks correct. I'm excited to see more usage of Active Support Instrumentation. I'm assuming that once a few more components are instrumented, the common behavior will be extracted out into helpers and eventually slim this down.

btw, I'm the author of GoodJob; I also submitted the [initial Active Job instrumentation for `dd-trace-rb`](https://github.com/DataDog/dd-trace-rb/pull/1639) so I'm comparing and contrasting my experience implementing that to this code. 

Of things I looked at:

* I doublechecked the instrumentation names against my own lists, to make sure they were comprehensive.
* I read through all the version-specific errata and couldn't think of any other weirdness (all the real weird weirdness I know of is in Rails <= 5.2, which isn't supported anyway)
* I read through the tests and thought they poked at it in a good realistic / integrated fashion.

simi:
@bensheldon is there anything GJ specific to instrument in separate instrumentation?

bensheldon:
>  is there anything GJ specific to instrument in separate instrumentation?

I don't believe so, or anything that exists currently in GoodJob. The `*.good_job` events are all enumerated in [this file](https://github.com/bensheldon/good_job/blob/main/lib/good_job/log_subscriber.rb).

Thinking about backend queue adapters generally there's usually an operation or two _before_ `perform.active_job` that does the actual dequeuing operation. For example in GoodJob, within an execution thread there is a database query to the database to dequeue a job record, and then that record is passed into  Active Job which then deserializes and performs it. Relatedly, I'm not sure if GlobalID deserialization happens as part of `perform.active_job` or before it.  (GoodJob doesn't currently have instrumentation hooks at that foundational level... though it should!)

instrumentation/active_job/lib/opentelemetry/instrumentation/active_job/handlers.rb
arielvalentin: @fbogsany did you see my comment about this in the other issue?

Is this a deal breaker?

arielvalentin: Given the results of the conversation in #688, I have changed the handlers to use the service locator pattern and look up the tracer from the Instrumentation singleton to creating spans.

arielvalentin:
@open-telemetry/ruby-contrib-approvers Hello friends. I could use your help to get this instrumentation reviewed. 

Is there any objection to merging this PR?

instrumentation/active_job/lib/opentelemetry/instrumentation/active_job/mappers/attribute.rb
```diff
@@ -0,0 +1,41 @@
...
+            job = payload.fetch(:job)
+
+            otel_attributes = {
+              'code.namespace' => job.class.name,
+              'messaging.system' => job.class.queue_adapter_name,
+              'messaging.destination' => job.queue_name,
+              'messaging.message.id' => job.job_id,
+              'rails.active_job.execution.counter' => job.executions.to_i,
+              'rails.active_job.provider_job_id' => job.provider_job_id.to_s,
+              'rails.active_job.priority' => job.priority, # this can be problematic. Programs may use invalid attributes for priority.
```
xuan-cao-swi, line 29: Would adding to_s (e.g. job.priority.to_s) make sense to overcome invalid attributes for priority?

instrumentation/active_job/lib/opentelemetry/instrumentation/active_job/instrumentation.rb
```diff
@@ -64,12 +64,13 @@ def gem_version
...
         def require_dependencies
           require_relative 'patches/base'
-          require_relative 'patches/active_job_callbacks'
+          require_relative 'handlers'
         end
 
         def patch_activejob
-          ::ActiveJob::Base.prepend(Patches::Base)
-          ::ActiveJob::Base.prepend(Patches::ActiveJobCallbacks)
+          ::ActiveJob::Base.prepend(Patches::Base) unless ::ActiveJob::Base.ancestors.include?(Patches::Base)
```
xuan-cao-swi, line 71: I assume this is for activejob < 6.1? Is this still required for rails 6.1? Will there be double span if on activejob 6.0 (since it both prepend and subscribe)

instrumentation/active_job/lib/opentelemetry/instrumentation/active_job/mappers/attribute.rb
arielvalentin: `NilClass` will render an empty string. Do we want the key to always be present even if the value is `nil`?

instrumentation/active_job/lib/opentelemetry/instrumentation/active_job/instrumentation.rb
arielvalentin: This is required for all versions of `ActiveJob` unfortunately and only introduces patches for context propagation. It no longer contains any code that creates spans. 

ActiveJob does not have a concept of "headers" that we can use to propagate the context so we have to add a custom serializer that propagates the trace context. 

Thinking about this a bit, I am not sure why we need this check here. 🤔 Did I do this only for unit tests?

instrumentation/active_job/lib/opentelemetry/instrumentation/active_job/mappers/attribute.rb
robbkidd: What values could appear here that would be invalid? Invalid by whose rules, semantic convention†, OTLP format for attributes, or something else?

† I haven't found any, I think because a "job" namespace has yet to be defined.

arielvalentin: From https://opentelemetry.io/docs/specs/otel/common/#attribute

> An Attribute is a key-value pair, which MUST have the following properties:
> 
> - The attribute key MUST be a non-null and non-empty string.
> - The attribute value is either:
>   - A primitive type: string, boolean, double precision floating point (IEEE 754-1985) or signed 64 bit integer.
>   - An array of primitive type values. The array MUST be homogeneous, i.e., it MUST NOT contain values of different types.

Developers may set any value to `priority`. I ran into an issue where some teams were setting the value to a symbol e.g. `:low`, `:high`, and then doing custom processing of the `priority` values instead of using what was built into ActiveJob itself. 

The `priority` value was then being checked by the SDK, which prohibits `Symbol` types and would result in `OpenTelemetry.handle_error` being called for an invalid parameter.

instrumentation/active_job/README.md
```diff
@@ -30,6 +30,67 @@ OpenTelemetry::SDK.configure do |c|
...
+
+`perform.active_job` events do not include timings for `ActiveJob` callbacks therefore time spent in `before` and `after` hooks will be missing
+
+`ActiveJob::Base#executions` start at `1`.
+
+### ActiveJob 7+
+
+`perform.active_job` no longer includes exceptions handled using `rescue_from` in the payload.
+
+In order to preseve this behavior you will have to update the span yourself, e.g.
```
robbkidd, line 79:
```suggestion
In order to preserve this behavior you will have to update the span yourself, e.g.
```

instrumentation/active_job/lib/opentelemetry/instrumentation/active_job/mappers/attribute.rb
arielvalentin: > † I haven't found any, I think because a "job" namespace has yet to be defined.

Bike shedding time. I do not think there are special Job semantics that really differ from messaging semantics, specifically for ActiveJob anyway. 

When I think of "job" processing on the other hand I think of `CronJob` as a concrete example of a function that runs on an interval and may or may not produce asynchronous events for processing. The "producer" in this case is the schedular and the event only has a timestamp that executes the work and the `consumer` is the script that is being executed. The semantics for me don't seem to line up cleanly with messaging because there isn't a concept of a "destination" or a "message" to process really.      

ActiveJob on the other hand is an asynchronous event processing system, whose specification describes the use of a single publisher (producer) and worker (consumer) per event. Though you could probably write an adapter that fanouts the event to _multiple destinations_ and to me this OOTB functionality the same semantics as point-to-point messaging, that is one producer-consumer with a dedicated queue (inline, in memory, Redis, postgresql, etc...).

It's unclear to me if attributes like `priorty` are generic for messaging systems, so my first pass at this was to use framework specific semconv. 

I asked about whether or not upstream semconv for framework specific attributes was welcome... see this discussion: 

https://cloud-native.slack.com/archives/C041APFBYQP/p1695733952824799

TL;DR each language should maintain its own set of semconv for framework specific components. 

All that being said, I just took another look at the messaging spec[^1].

> All attributes that are specific for a messaging system SHOULD be populated in messaging.{system} namespace. Attributes that describe a message, a destination, a consumer, or a batch of messages SHOULD be populated under the corresponding namespace:

Now I am second guessing my decision to use a `rails.active_job` namespace in this PR. 

Based on spec guidance does it make sense to use `messaging.active_job.destination.*` or `messaging.active_job.consumer.*` attributes?

Should we bother including these attributes at all? 

Does anyone need them?

[^1]: https://opentelemetry.io/docs/specs/semconv/messaging/messaging-spans/#attributes-specific-to-certain-messaging-systems

robbkidd: If we're using the messaging semconv, I think yes to `messaging.active_job.*`. I admit that feels weird from an "I'm a Rails dev wondering what's going on in my system" perspective, though.

robbkidd: 🤔 What do you think of shipping with `messaging.active_job.*` (no `rails.` prefix) so we've hewed as close to OTel conventions as possible? Then, if we get feedback that these are weird names from a Rails perspective on conventions, we can decide to add more attributes under different names later.

instrumentation/active_job/lib/opentelemetry/instrumentation/active_job/instrumentation.rb
xuan-cao-swi: ah, that makes sense, thanks. I thought require base will also include `active_job_callbacks.rb`, but now it looks like it doesn't. Do we still need `active_job_callbacks.rb`?

> Did I do this only for unit tests?

Yes, it seems like without unless, the test will have method redefined
```
/app/instrumentation/active_job/lib/opentelemetry/instrumentation/active_job/patches/base.rb:15: warning: method redefined; discarding old __otel_headers=
```

instrumentation/active_job/lib/opentelemetry/instrumentation/active_job/handlers/default.rb
```diff
@@ -0,0 +1,113 @@
...
+          # @param _id [String] of the event (unused)
+          # @param payload [Hash] containing job run information
+          # @return [Hash] with the span and generated context tokens
+          def finish(_name, _id, payload)
+            otel = payload.delete(:__otel)
+            span = otel&.fetch(:span)
+            tokens = otel&.fetch(:ctx_tokens)
+
+            exception = payload[:error] || payload[:exception_object]
+            on_exception(exception, span) if exception
```
robbkidd, line 61: Reduce memory allocations suggestion, part 1:
```suggestion
            on_exception((payload[:error] || payload[:exception_object]), span)
```

instrumentation/active_job/lib/opentelemetry/instrumentation/active_job/handlers/default.rb
```diff
@@ -0,0 +1,113 @@
...
+          # Handled exceptions _are not_ copied to the ingress span, but it does set the status to `Error` making it easier to know that a job has failed
+          # Unhandled exceptions bubble up to the ingress span and are recorded there.
+          #
+          # @param [Exception] exception to report as a Span Event
+          # @param [OpenTelemetry::Trace::Span] the currently active span used to record the exception and set the status
+          def on_exception(exception, span)
+            status = OpenTelemetry::Trace::Status.error(exception.message)
+            span&.record_exception(exception)
+            span&.status = status
+            @parent_span_provider.current_span.status = status
```
robbkidd, line 103: Reduce memory allocations suggestion, part 2:

```suggestion
            return unless (exception && span)
​
            span&.record_exception(exception)
            span&.status =
              @parent_span_provider.current_span.status =
                OpenTelemetry::Trace::Status.error(exception.message)
```

robbkidd (commented):
I fiddled with reducing allocating memory for local variables, suggesting here as a _non-blocking_ idea. (I didn't run a profiler or anything, just eyeballed it.)

instrumentation/active_job/example/active_job.rb
```diff
@@ -4,18 +4,85 @@
...
+  discard_on DiscardError
+
+  def perform
+    raise DiscardError, 'this job failed'
+  end
 end
 
 class TestJob < ::ActiveJob::Base
+  around_enqueue do |_job, block|
+    OpenTelemetry.tracer_provider.tracer('demo', '1.0').in_span('around_enqueue') do
```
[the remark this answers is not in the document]
arielvalentin: You're right. This is a bad example. I'll change it to make more sense.

instrumentation/active_job/lib/opentelemetry/instrumentation/active_job/handlers/default.rb
```diff
@@ -0,0 +1,113 @@
...
+          #
+          # @param span [OpenTelemetry::Trace::Span]
+          # @param tokens [Array] to unregister
+          def finish_span(span, tokens)
+            # closes the span after all attributes have been finalized
+            begin
+              if span&.recording?
+                span&.status = OpenTelemetry::Trace::Status.ok if span&.status&.code == OpenTelemetry::Trace::Status::UNSET
+                span&.finish
+              end
```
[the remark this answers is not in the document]
arielvalentin: I really was trying to be paranoid.

instrumentation/active_job/lib/opentelemetry/instrumentation/active_job/handlers/default.rb
```diff
@@ -0,0 +1,114 @@
...
+          #
+          # @param [Exception] exception to report as a Span Event
+          # @param [OpenTelemetry::Trace::Span] the currently active span used to record the exception and set the status
+          def on_exception(exception, span)
+            return unless (exception && span)
+​
+            span&.record_exception(exception)
+            span&.status =
+              @parent_span_provider.current_span.status =
+                OpenTelemetry::Trace::Status.error(exception.message)
```
arielvalentin, line 104:
```suggestion
            span.record_exception(exception)
            span.status =
              @parent_span_provider.current_span.status =
                OpenTelemetry::Trace::Status.error(exception.message)
```

instrumentation/active_job/lib/opentelemetry/instrumentation/active_job/instrumentation.rb
arielvalentin: I see. Did I not delete the old callbacks file?

instrumentation/active_job/lib/opentelemetry/instrumentation/active_job/mappers/attribute.rb
arielvalentin: > 🤔 What do you think of shipping with messaging.active_job.* (no rails. prefix) so we've hewed as close to OTel conventions as possible?
> Then, if we get feedback that these are weird names from a Rails perspective on conventions, we can decide to add more attributes under different names later.

Yes! I plan on fixing that, though I think I am going to omit some of these attributes like the `execution_count` and `scheduled_at` attributes because I am dubious about their usefulness as span attributes.

instrumentation/active_job/example/active_job.rb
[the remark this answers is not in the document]
arielvalentin: I have updated the examples to use a singleton tracer instance. Please take a look and let me know if that improves the quality of the example.

instrumentation/active_job/lib/opentelemetry/instrumentation/active_job/mappers/attribute.rb
arielvalentin: @robbkidd done! https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/677/commits/c0655ebaf0eb120d78026ee43fef33b65333cfe7

instrumentation/active_job/lib/opentelemetry/instrumentation/active_job/handlers/default.rb
```diff
@@ -0,0 +1,114 @@
...
+          # Finishes the provided spans and also detaches the associated contexts
+          #
+          # @param span [OpenTelemetry::Trace::Span]
+          # @param tokens [Array] to unregister
+          def finish_span(span, tokens)
+            # closes the span after all attributes have been finalized
+            begin
+              if span&.recording?
+                span&.status = OpenTelemetry::Trace::Status.ok if span&.status&.code == OpenTelemetry::Trace::Status::UNSET
+                span&.finish
```
arielvalentin, line 76:
```suggestion
                span.status = OpenTelemetry::Trace::Status.ok if span.status.code == OpenTelemetry::Trace::Status::UNSET
                span.finish
```

instrumentation/active_job/README.md
```diff
@@ -30,6 +30,59 @@ OpenTelemetry::SDK.configure do |c|
...
+This instrumentation generally uses [Messaging semantic conventions](https://opentelemetry.io/docs/specs/semconv/messaging/messaging-spans/) by treating job enqueuers as `producers` and workers as `consumers`.
+
+Internal spans are named using the name of the `ActiveSupport` event that was provided.
+
+Attributes that are specific to this instrumentation are recorded under `messaging.active_job.*`:
+
+| Attribute Name | Type | Notes |
+| - | - | - |
+| `messaging.active_job.provider_job_id` | String | |
+| `messaging.active_job.priority` | Integer | |
```
arielvalentin, line 63: messaging.message.id
messaging.active_job.message.priority

instrumentation/active_job/lib/opentelemetry/instrumentation/active_job/mappers/attribute.rb
robbkidd: You're the best!

instrumentation/active_job/lib/opentelemetry/instrumentation/active_job/handlers/default.rb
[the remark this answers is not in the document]
robbkidd: My recollection of the conversation about this from today's SIG meeting is that the safe navigators are noisy to read but fine for implementation, so this is not blocking commentary.

They are also probably safe to remove from the calls within the `if..end` block if you cared to remove them. If the object coming in as `span` responds to `recording?` successfully, it is highly likely to respond safely to `status`, `status.code`, and `finish`.

instrumentation/active_job/README.md
```diff
@@ -30,6 +30,64 @@ OpenTelemetry::SDK.configure do |c|
...
+
+## Semantic Conventions
+
+This instrumentation generally uses [Messaging semantic conventions](https://opentelemetry.io/docs/specs/semconv/messaging/messaging-spans/) by treating job enqueuers as `producers` and workers as `consumers`.
+
+Internal spans are named using the name of the `ActiveSupport` event that was provided.
+
+Attributes that are specific to this instrumentation are recorded under `messaging.active_job.*`:
+
+| Attribute Name | Type | Notes |
```
arielvalentin, line 60: @robbkidd Please take a look at the latest list of semconv used by this instrumentation. 

Let me know if you think I should change anything.

arielvalentin:
Up to date screenshots:

instrumentation/active_job/README.md
robbkidd: I've reviewed! 👍🏻 

I think the messaging semconv selections in this PR make the most sense in a world without job semconv.

robbkidd (approved):
👍🏻 

Message for you, sir.

kaylareopelle (approved):

instrumentation/active_job/lib/opentelemetry/instrumentation/active_job/handlers.rb
```diff
@@ -0,0 +1,73 @@
...
+
+          @subscriptions = handlers_by_pattern.map do |key, handler|
+            ::ActiveSupport::Notifications.subscribe("#{key}.active_job", handler)
+          end
+        end
+
+        # Removes Event Handler Subscriptions for ActiveJob notifications
+        # @note this method is not thread-safe and should not be used in a multi-threaded context
+        def unsubscribe
+          @subscriptions&.each { |subscriber| ActiveSupport::Notifications.unsubscribe(subscriber) }
```
lavoiesl, line 67: This needs to be `::ActiveSupport::Notifications`, otherwise it throws:
```
uninitialized constant OpenTelemetry::Instrumentation::ActiveSupport::Notifications
```

instrumentation/active_job/lib/opentelemetry/instrumentation/active_job/handlers/enqueue.rb
```diff
@@ -0,0 +1,40 @@
...
+      module Handlers
+        # Handles `enqueue.active_job` and `enqueue_at.active_job` to generate egress spans
+        class Enqueue < Default
+          def initialize(...)
+            super
+            @span_name_formatter = if @config[:span_naming] == :job_class
+                                     ->(job) { "#{job.class.name} publish" }
+                                   else
+                                     ->(job) { "#{job.queue_name} publish" }
+                                   end
```
lavoiesl, line 19: This is causing a regression in our codebase because the handler no longer honours dynamic configuration.

i.e. Changing `OpenTelemetry::Instrumentation::ActiveJob::Instrumentation.instance` after this is registered no longer affects the naming.

Arguably, we probably shouldn't rely on dynamic configs, but I wanted to point it out regardless

Same applies to `Perform`

arielvalentin: Would you be able to provide us with a PR to fix this issue?

instrumentation/active_job/lib/opentelemetry/instrumentation/active_job/handlers.rb
arielvalentin: Similarly, would you be able to provide us with a PR to address this problem?

lavoiesl: https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/1078

instrumentation/active_job/lib/opentelemetry/instrumentation/active_job/handlers/enqueue.rb
lavoiesl: https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/1079

instrumentation/active_job/lib/opentelemetry/instrumentation/active_job.rb
```diff
@@ -11,6 +11,39 @@ module OpenTelemetry
...
+      end
+
+      # Activates/deactivates the Span within the current Context, which makes the "current span"
+      # available implicitly.
+      #
+      # On exit, the Span that was active before calling this method will be reactivated.
+      #
+      # @param [Span] span the span to activate
+      # @yield [span, context] yields span and a context containing the span to the block.
+      def with_span(span)
```
lavoiesl, line 44: I think this is never called and that the context is never set.

This is causing link propagation to fail.

arielvalentin: Do you have any additional details you would like to share in an issue?

lavoiesl: Here we go! https://github.com/open-telemetry/opentelemetry-ruby-contrib/issues/1131

I was working on that, but it took me longer than expected to write up the issue. 😅

PR #116, opened 2022-08-16
chore: revert hooks for http clients. its fine.

ericmustin:
### Summary

Hooks for http clients (and servers, tho not relevant to this pr) are not in the [spec](https://github.com/open-telemetry/opentelemetry-specification) and so the decision has been made not to support them. Guidance for this sort of functionality is to bring your own instrumentation (which can be submitted it to the [OpenTelemetry Registry](https://opentelemetry.io/registry/), or open an [OTEP](https://github.com/open-telemetry/oteps) to add client hooks to the spec if you feel differently
oooohthe-claw
.

PR #678, opened 2023-09-29
feat: Allow filtering the http.target attribute

fbogsany:
This adds a config option to the `Net::HTTP` instrumentation, `filter_target`. It is a callable that takes a string and returns a string. The intent is to support filtering the `http.target` attribute to redact secrets/tokens in params. One way to use this with Rails is:

```ruby
filter = ActiveSupport::ParameterFilter.new(Rails.application.config.filter_parameters)
KV_RE   = "[^&;=]+"
PAIR_RE = %r{(#{KV_RE})=(#{KV_RE})}

filter_target: proc do |path|
  url = URI(path)
  query = url.query
  if query.empty?
    path
  else
    query = query.gsub(PAIR_RE) { |_| filter.filter($1 => $2).first.join("=") }
    "#{url.path}?#{query}"
  end
rescue URI::InvalidURIError
    path
end

OpenTelemetry::Instrumentation::Net::HTTP::Instrumentation.instance.install({filter_target: filter_target})
```

### DIFF
--- a/instrumentation/net_http/lib/opentelemetry/instrumentation/net/http/patches/instrumentation.rb
+++ b/instrumentation/net_http/lib/opentelemetry/instrumentation/net/http/patches/instrumentation.rb
@@ -23,7 +23,7 @@ module OpenTelemetry
               attributes = {
                 OpenTelemetry::SemanticConventions::Trace::HTTP_METHOD => req.method,
                 OpenTelemetry::SemanticConventions::Trace::HTTP_SCHEME => USE_SSL_TO_SCHEME[use_ssl?],
-                OpenTelemetry::SemanticConventions::Trace::HTTP_TARGET => req.path,
+                OpenTelemetry::SemanticConventions::Trace::HTTP_TARGET => filter_target(req.path),
                 OpenTelemetry::SemanticConventions::Trace::NET_PEER_NAME => @address,
                 OpenTelemetry::SemanticConventions::Trace::NET_PEER_PORT => @port
               }.merge!(OpenTelemetry::Common::HTTP::ClientContext.attributes)
@@ -83,6 +83,14 @@ module OpenTelemetry
 
             def tracer
               Net::HTTP::Instrumentation.instance.tracer
+            end
+
+            def filter_target(path)
+              if (filter = Net::HTTP::Instrumentation.instance.config[:filter_target])
+                filter.call(path)
+              else
+                path
+              end
             end
 
             def untraced?
--- a/instrumentation/net_http/lib/opentelemetry/instrumentation/net/http/instrumentation.rb
+++ b/instrumentation/net_http/lib/opentelemetry/instrumentation/net/http/instrumentation.rb
@@ -26,7 +26,11 @@ module OpenTelemetry
           #   or `Regexp` in this array, the instrumentation will not record a
           #   `kind = :client` representing the request and will not propagate
           #   context in the request.
-          option :untraced_hosts, default: [], validate: :array
+          # filter_target: a callable object that takes a path `String` and returns
+          #   a `String` to be used as the http.target attribute. The default is to
+          #   use the request path as-is.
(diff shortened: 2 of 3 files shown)


instrumentation/net_http/lib/opentelemetry/instrumentation/net/http/patches/instrumentation.rb
```diff
@@ -85,6 +85,14 @@ def tracer
               Net::HTTP::Instrumentation.instance.tracer
             end
 
+            def filter_target(path)
+              if (filter = Net::HTTP::Instrumentation.instance.config[:filter_target])
+                filter.call(path)
+              else
+                path
+              end
+            end
```
fbogsany, line 94: It is probably worthwhile protecting against the callback `raise`-ing. Should this `rescue` be more specific?
```suggestion
            rescue
              path
            end
```

plantfansam: I don't know that the library should protect users from errors thrown by their filters. It's a real long-shot, but it's conceivable that someone might even _expect_ an exception to be thrown...

ericmustin: debug log perhaps?

ericmustin (approved):
the latest version of the spec seems to have gotten away from us wrt what to name things, but this seems fine generally and in the spirit of how the current spec treats scrubbing sensitive data.   https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/http.md

every day we stray further from [hook's](https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/116) light. oh well :trollface:

fbogsany:
> every day we stray further from [hook's](https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/116) light. oh well :trollface:

You're not wrong @ericmustin. Thanks for the reminder. The example I provided above feels really gross - there's a desire to use Rails' parameter filter for HTTP client spans as well, to avoid leaking tokens, etc. but we want to avoid coupling HTTP client instrumentation to Rails. I wonder if there's a better way to do this (e.g. "Shopify should bring your own instrumentation").

fbogsany:
> the latest version of the spec seems to have gotten away from us wrt what to name things

It has evolved significantly. It currently carries a large warning saying (essentially) "don't update existing instrumentation written against earlier versions of the spec until we declare this 'done'". Maybe there's an opportunity to write new instrumentation that corrects all our mistakes :trollface:.

instrumentation/net_http/lib/opentelemetry/instrumentation/net/http/patches/instrumentation.rb
fbogsany: The [error handling guidance](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/error-handling.md#guidance) is pretty explicit:

> 1. API methods that accept external callbacks MUST handle all errors.
> 5. Beware external callbacks and overridable interfaces: Expect them to throw.

So: we should rescue and call the global error handler.

plantfansam: > The [error handling guidance](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/error-handling.md#guidance) is pretty explicit:
> 
> > 1. API methods that accept external callbacks MUST handle all errors.
> > 2. Beware external callbacks and overridable interfaces: Expect them to throw.
> 
> So: we should rescue and call the global error handler.

I stand corrected, the glorious spec is unambiguous...

fbogsany:
I'm going to close this. For various reasons, it is preferable to do this with a `SpanProcessor` or a span exporter that wraps the "real" span exporter.

PR #682, opened 2023-10-04
feat!: obfuscation for mysql2, dalli and postgresql as default option for db_statement

xuan-cao-swi:
## Description

Based on [opentelemetry-specification](https://github.com/open-telemetry/opentelemetry-specification/pull/3127) and [discussion](https://github.com/open-telemetry/opentelemetry-ruby-contrib/discussions/442), change the db.statement default sql to sanitized sql

## Test
Tested with basic test suite (e.g. `rake test`)

### DIFF
--- a/instrumentation/pg/test/opentelemetry/instrumentation/pg/instrumentation_test.rb
+++ b/instrumentation/pg/test/opentelemetry/instrumentation/pg/instrumentation_test.rb
@@ -52,7 +52,6 @@ describe OpenTelemetry::Instrumentation::PG::Instrumentation do
     let(:dbname) { ENV.fetch('TEST_POSTGRES_DB', 'postgres') }
     let(:password) { ENV.fetch('TEST_POSTGRES_PASSWORD', 'postgres') }
     let(:config) { { db_statement: :include } }
-    
     before do
       instrumentation.install(config)
     end
--- a/instrumentation/pg/test/opentelemetry/instrumentation/pg/instrumentation_test.rb
+++ b/instrumentation/pg/test/opentelemetry/instrumentation/pg/instrumentation_test.rb
@@ -51,7 +51,8 @@ describe OpenTelemetry::Instrumentation::PG::Instrumentation do
     let(:user) { ENV.fetch('TEST_POSTGRES_USER', 'postgres') }
     let(:dbname) { ENV.fetch('TEST_POSTGRES_DB', 'postgres') }
     let(:password) { ENV.fetch('TEST_POSTGRES_PASSWORD', 'postgres') }
-
+    let(:config) { { db_statement: :include } }
+    
     before do
       instrumentation.install(config)
     end
--- a/instrumentation/dalli/README.md
+++ b/instrumentation/dalli/README.md
@@ -30,6 +30,22 @@ OpenTelemetry::SDK.configure do |c|
 end
 ```
 
+### Configuration options
+
+```ruby
+OpenTelemetry::SDK.configure do |c|
+  c.use 'OpenTelemetry::Instrumentation::Dalli', {
+    # You may optionally set a value for 'peer.service', which
+    # will be included on all spans from this instrumentation:
+    peer_service: '',
+
+    # The obfuscation of query in the db.statement attribute is enabled by default.
+    # To disable, set db_statement to :include; to omit the query completely, set db_statement to :omit
+    db_statement: :include,
+  }
+end
+```
+
 ## How can I get involved?
 
 The `opentelemetry-instrumentation-dalli` gem source is [on github][repo-github], along with related gems including `opentelemetry-api` and `opentelemetry-sdk`.
--- a/instrumentation/dalli/lib/opentelemetry/instrumentation/dalli/instrumentation.rb
+++ b/instrumentation/dalli/lib/opentelemetry/instrumentation/dalli/instrumentation.rb
@@ -20,7 +20,7 @@ module OpenTelemetry
         end
 
         option :peer_service, default: nil, validate: :string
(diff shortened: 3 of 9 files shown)


ericmustin (changes requested):
Obfuscation adds a significant performance penalty.

We should either omit or include by default. The spec says to omit now if I understand correctly?

arielvalentin:
> Obfuscation adds a significant performance penalty.
> 
> We should either omit or include by default. The spec says to omit now if I understand correctly?

@ericmustin The spec says nothing about the defaults client-side sanitization, AFAIK, other than:

>  [2]: Should be collected by default only if there is sanitization that excludes sensitive information.

I interpret that to mean you should only collect it if the SQL is sanitized. If I extrapolate a bit, then it should never record any unsanitized SQL at all. 

There are some open issues about db statement sanitization in general:

- https://github.com/open-telemetry/opentelemetry-specification/issues/1106
- https://github.com/open-telemetry/opentelemetry-specification/issues/3104
- https://github.com/open-telemetry/opentelemetry-specification/issues/3594

W/R/T `db.operation` it says:

> [3]: When setting this to an SQL keyword, it is not recommended to attempt any client-side parsing of db.statement just to get this property, but it should be set if the operation name is provided by the library being instrumented. If the SQL statement has an ambiguous operation, or performs more than one operation, this value may be omitted.

Given this information I for one prefer to sanitize by default and `omit` if the driver does not have a sanitizer implemented.

arielvalentin:
> The appropriate place to do sanitization is in the collector, or more generally out of process. If your interpretation of the spec is correct(and I think it is), then that would effectively exclude the possibility of doing sql sanitization in the collector because the spec would dictate it can never be exported without being sanitized. This seems flawed to me.
> 
> Additionally we _know_ real examples of the "obfuscate" config option introducing a performance regression in applications at large enterprise organizations which use ruby. Setting it as the default would seem like a poor choice.
> 
> I would prefer omit as the default if we're left with a choice btwn omit and "conditionally obfuscate depending on the particulars of driver options", as I think users will find the lack of standard confusing.

Would it make the most sense to discussion in the spec issues? 

I think the challenge here is there are no OOTB SQL sanitizers in [contrib](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/processor). If there were, then I would totally be onboard with leaving `include`/`raw` as the default, but since there isn't I still advocate that we `obfucate`/`sanitize` by default.

cheempz:
Regarding 
> The appropriate place to do sanitization is in the collector, or more generally out of process. 

that disagrees with a recent [separate discussion](https://github.com/open-telemetry/opentelemetry.io/pull/3309#discussion_r1337501757), which is the case with our product--telemetry is exported directly to our platform and doesn't go though a collector.  

OTel [JS](https://github.com/open-telemetry/opentelemetry-js-contrib/issues/1237) and [Java](https://opentelemetry.io/docs/instrumentation/java/automatic/agent-config/#db-statement-sanitization) obfuscate/sanitize by default, so there's a bit of precedence despite the much needed spec clarification/configuration naming being discussed :)

ericmustin:
Indeed, I disagree with the spec discussion. If I'm not
Mistaken, Both github and Shopify have introduced regressions into production by enabling obfuscation in app ootb. Perhaps benchmarking would be useful here, but im dubious of defaulting to obfuscation. The chief concern of observability instrumentation should be minimizing the observer effect as much as possible, and I feel we're disregarding that for what's effectively product/vendor developer experience preferences

arielvalentin:
For context, at GH we've accepted the trade off so we obfuscate by default. It's slow but we mitigate it with the character limit setting.

ericmustin:
just circling back, if others feel `obfuscate` is the appropriate default, then I won't stand in the way here.

arielvalentin (commented):
Is this all of the data store drivers? Or do the others already have obfuscate by default?

xuan-cao-swi:
> Is this all of the data store drivers? Or do the others already have obfuscate by default?

AFAIK, I think these three gems are the last to set obfuscate to default

arielvalentin (approved):

instrumentation/pg/test/opentelemetry/instrumentation/pg/instrumentation_test.rb
```diff
@@ -51,7 +51,8 @@
     let(:user) { ENV.fetch('TEST_POSTGRES_USER', 'postgres') }
     let(:dbname) { ENV.fetch('TEST_POSTGRES_DB', 'postgres') }
     let(:password) { ENV.fetch('TEST_POSTGRES_PASSWORD', 'postgres') }
-
+    let(:config) { { db_statement: :include } }
+
```
arielvalentin, line 55:
```suggestion

```

ericmustin (approved):
So it goes

arielvalentin:
@xuan-cao-swi just one more favor to ask before I merge. 

Please update the commit messages to use conventional commits:

https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/main/CONTRIBUTING.md#use-conventional-commit-messages

This is something that I think requires a Minor release, so please using the `BREAKING CHANGE` format in the message, e.g. 

`feat!: Set db.statement option to obfuscate by default`

xuan-cao-swi:
> @xuan-cao-swi just one more favor to ask before I merge.
> 
> Please update the commit messages to use conventional commits:
> 
> https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/main/CONTRIBUTING.md#use-conventional-commit-messages
> 
> This is something that I think requires a Minor release, so please using the `BREAKING CHANGE` format in the message, e.g.
> 
> `feat!: Set db.statement option to obfuscate by default`

Changed!

PR #685, opened 2023-10-07
fix!: Drop DelayedJob ActiveRecord in Tests

arielvalentin:
The DelayedJob tests suite relied on the ActiveRecord Backend gem for testing, however incompatabilities with ActiveRecord/Rails 7.1 means would result in failures in the test suite.

It is not the responsibility of OTel Ruby Instrumentation to ensure compatability of 3rd party gems or ensure our test suite to come up with workarounds for bugs.

For this reason I have rewritten the test suite to use the Test backend that Delayed Job used to test the upstream gem.

This does introduce one change, the `created_at` attribute, which is _specific_ to the ActiveRecord backend will _no longer_ be emitted. Users that need this event should pin to earlier versions of the gem.

There is a risk here that the upstream DelayedJob gem will no longer be included in future releases, but I think that is less likely than to run into incompatabilities with future Rails versions.

See: https://github.com/open-telemetry/opentelemetry-ruby-contrib/issues/686

### DIFF
--- a/instrumentation/delayed_job/opentelemetry-instrumentation-delayed_job.gemspec
+++ b/instrumentation/delayed_job/opentelemetry-instrumentation-delayed_job.gemspec
@@ -30,8 +30,7 @@ Gem::Specification.new do |spec|
 
   spec.add_development_dependency 'appraisal', '~> 2.5'
   spec.add_development_dependency 'bundler', '~> 2.4'
-  spec.add_development_dependency 'delayed_job', '~> 4.1.0'
-  spec.add_development_dependency 'delayed_job_active_record'
+  spec.add_development_dependency 'delayed_job', '~> 4.1.7'
   spec.add_development_dependency 'minitest', '~> 5.0'
   spec.add_development_dependency 'opentelemetry-sdk', '~> 1.1'
   spec.add_development_dependency 'opentelemetry-test-helpers', '~> 0.3'
--- a/instrumentation/delayed_job/opentelemetry-instrumentation-delayed_job.gemspec
+++ b/instrumentation/delayed_job/opentelemetry-instrumentation-delayed_job.gemspec
@@ -31,7 +31,6 @@ Gem::Specification.new do |spec|
   spec.add_development_dependency 'appraisal', '~> 2.5'
   spec.add_development_dependency 'bundler', '~> 2.4'
   spec.add_development_dependency 'delayed_job', '~> 4.1.7'
-  spec.add_development_dependency 'debug'
   spec.add_development_dependency 'minitest', '~> 5.0'
   spec.add_development_dependency 'opentelemetry-sdk', '~> 1.1'
   spec.add_development_dependency 'opentelemetry-test-helpers', '~> 0.3'
--- a/instrumentation/delayed_job/opentelemetry-instrumentation-delayed_job.gemspec
+++ b/instrumentation/delayed_job/opentelemetry-instrumentation-delayed_job.gemspec
@@ -31,6 +31,7 @@ Gem::Specification.new do |spec|
   spec.add_development_dependency 'appraisal', '~> 2.5'
   spec.add_development_dependency 'bundler', '~> 2.4'
   spec.add_development_dependency 'delayed_job', '~> 4.1.7'
+  spec.add_development_dependency 'debug'
   spec.add_development_dependency 'minitest', '~> 5.0'
   spec.add_development_dependency 'opentelemetry-sdk', '~> 1.1'
   spec.add_development_dependency 'opentelemetry-test-helpers', '~> 0.3'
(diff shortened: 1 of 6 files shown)


instrumentation/delayed_job/test/test_helper.rb
```diff
@@ -24,31 +29,8 @@
...
-      end
-    end
-  end
-
-  def teardown_active_record
-    ::ActiveRecord::Base.connection.close
-  end
-end
+gem_dir = Gem::Specification.find_by_name('delayed_job').gem_dir
+require "#{gem_dir}/spec/delayed/serialization/test"
```
simi, line 33: Any reason to load this empty file?

https://github.com/collectiveidea/delayed_job/blob/v4.1.11/spec/delayed/serialization/test.rb

instrumentation/delayed_job/test/test_helper.rb
```diff
@@ -24,31 +29,8 @@
...
+require "#{gem_dir}/spec/delayed/backend/test"
```
simi, line 34: IMHO it would make sense to add PR to delayed job to move this file into lib folder since it is useful for testing depending libs. Should I try to open PR?

instrumentation/delayed_job/test/test_helper.rb
```diff
@@ -7,8 +7,13 @@
 require 'bundler/setup'
 Bundler.require(:default, :development, :test)
 
+# These are dependencies that delayed job assumes are already loaded
+# We are compensating for that here in this test... that is a smell
+# NoMethodError: undefined method `extract_options!' for [#<ActiveJobPayload:0x0000000108bf5d48>, {}]:Array
+# delayed_job-4.1.11/lib/delayed/backend/job_preparer.rb:7:in `initialize'0
+require 'active_support/core_ext/array'
```
simi, line 14: Any reason to not just require `delayed_job` itself here?

instrumentation/delayed_job/test/test_helper.rb
arielvalentin: That would be great! Thank you

instrumentation/delayed_job/test/test_helper.rb
arielvalentin: `delayed_job` is loaded by bundler therefore it is not necessary to load it twice.

instrumentation/delayed_job/test/test_helper.rb
arielvalentin: This may not be required since I am using the class instead of a symbol. I will remove it and see if tests pass. 

https://github.com/collectiveidea/delayed_job/blob/master/lib/delayed/worker.rb#L67

arielvalentin, line 33:
```suggestion
```

instrumentation/delayed_job/test/test_helper.rb
simi: if required properly, it would load this https://github.com/collectiveidea/delayed_job/blob/b66bb64437a8606a414a390531ac73108911434e/lib/delayed_job.rb#L11 and that would load https://github.com/collectiveidea/delayed_job/blob/b66bb64437a8606a414a390531ac73108911434e/lib/delayed/backend/job_preparer.rb#L1

:thinking:

instrumentation/delayed_job/test/test_helper.rb
arielvalentin: I removed this line and the tests pass.

instrumentation/delayed_job/test/test_helper.rb
arielvalentin: That is very curious isn't it! 

Let me try this again and see what I can figure out

arielvalentin: Interesting in that it fails for Rails 7.0 but not for 7.1 🤔

arielvalentin: The curious part for me here is that the backtrace clearly states its raised in JobPreparer, which is what requires the file.

arielvalentin: Explicitly requiring `delayed_job` and still get this error.

arielvalentin: https://github.com/open-telemetry/opentelemetry-ruby-contrib/actions/runs/6443662571/job/17495853381?pr=685

instrumentation/delayed_job/opentelemetry-instrumentation-delayed_job.gemspec
```diff
@@ -30,8 +30,8 @@ Gem::Specification.new do |spec|
 
   spec.add_development_dependency 'appraisal', '~> 2.5'
   spec.add_development_dependency 'bundler', '~> 2.4'
-  spec.add_development_dependency 'delayed_job', '~> 4.1.0'
-  spec.add_development_dependency 'delayed_job_active_record'
+  spec.add_development_dependency 'delayed_job', '~> 4.1.7'
+  spec.add_development_dependency 'debug'
```
arielvalentin, line 34:
```suggestion
```

instrumentation/delayed_job/test/test_helper.rb
simi: :pray: https://github.com/collectiveidea/delayed_job/pull/1195

arielvalentin: Lead time seems like it is about 3-4 months IMO https://github.com/collectiveidea/delayed_job/pull/1169

simi: definitely not blocker from my side

mwear (approved):

Issue #452, opened 2023-05-03
PG name incorrect when prefixing annotations


KaoruDev:
**Description of the bug**

When using https://github.com/basecamp/marginalia gem to prefix SQL queries with annotations, the Span name no longer contains the operation. For example, prior to prefixing I would see `SELECT my_db` now I just see `my_db` and the `operation` attribute is missing.

I suspect it's because the [extract_operation](https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/a11d8b135d9ac4c28521619dc3b4744692ae2e6e/instrumentation/pg/lib/opentelemetry/instrumentation/pg/patches/connection.rb#L94-L97) method does not take comments into consideration.

**Share details about your runtime**

Operating system details: Linux, Debian 11 LTS
RUBY_ENGINE: "ruby"
RUBY_VERSION: "3.1.3"
RUBY_DESCRIPTION: "ruby 3.1.3p185 (2022-11-24 revision 1a6b16756e) [aarch64-linux]"

**Share a simplified reproduction if possible**

```rb
require 'bundler/inline'

gemfile(true) do
  source 'https://rubygems.org'

  gem 'opentelemetry-api'
  gem 'opentelemetry-sdk'
  # gem 'opentelemetry-exporter-jaeger'
  # gem 'opentelemetry-exporter-otlp'
  # gem 'opentelemetry-exporter-zipkin'
end

require 'opentelemetry-api'
require 'opentelemetry-sdk'

span_processor = OpenTelemetry::SDK::Trace::Export::SimpleSpanProcessor.new(
  OpenTelemetry::SDK::Trace::Export::ConsoleSpanExporter.new
)

# require 'opentelemetry/exporter/jaeger'
# exporter = OpenTelemetry::Exporter::Jaeger::AgentExporter.new(max_packet_size: 9 * 1024)
# span_processor = OpenTelemetry::SDK::Trace::Export::BatchSpanProcessor.new(exporter)

# require 'opentelemetry/shopify/exporters/otlp'
# exporter = OpenTelemetry::Exporter::OTLP::Exporter.new
# span_processor = OpenTelemetry::SDK::Trace::Export::BatchSpanProcessor.new(exporter)

# require 'opentelemetry/shopify/exporters/otlp'
# exporter = OpenTelemetry::Exporter::OTLP::Exporter.new
# span_processor = OpenTelemetry::SDK::Trace::Export::BatchSpanProcessor.new(exporter)

[...]

PR #690, opened 2023-10-12
feat: Support prepend SQL comment for PG instrumentation

keiko713:
Currently, `extract_operation` method is simply extracting the first word before the space as an operation (as commented, this was borrowed from opentelemetry-js-contrib's pg instrument).

This works okay for most of queries, however, when [Active Record Query Logs](https://api.rubyonrails.org/classes/ActiveRecord/QueryLogs.html) (known as marginalia) is used with the option of `ActiveRecord::QueryLogs.prepend_comment = true`, the query will start with a query comment and this extracting logic fails, making a span name as simply database name and leaving the operation empty.

This PR adds a simple logic to ignore any prepend query comment (with using `/* comment */` way, not supporting `-- comment` type comment).
As Active Record Query Logs became part of Rails, we may see more and more folks use this likely with `prepend_comment = true` option, so it would be nice to support this with some minimal change.

Fixes https://github.com/open-telemetry/opentelemetry-ruby-contrib/issues/452


keiko713:
I'm reading the specification for what `db.operation` should be, and wondering if this regex could be considered as "client-side parsing".

https://opentelemetry.io/docs/specs/otel/trace/semantic_conventions/database/#call-level-attributes

> When setting this to an SQL keyword, it is not recommended to attempt any client-side parsing of `db.statement` just to get this property, but it should be set if the operation name is provided by the library being instrumented. If the SQL statement has an ambiguous operation, or performs more than one operation, this value may be omitted.

Given that it's a small simple one, I hope that it's okay to tweak like this (this really helps readability of the tracing so I'd love to have this personally), but I understand if that's not the case. I'd appreciate any inputs!

arielvalentin:
Thanks for your contribution.

We've got work already underway to address this problem in a separate PR https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/529

Would you be interested in helping us do some exploratory testing there?

keiko713:
Thanks for the pointer, @fbogsany ! I took a look that PR deeper and now have several questions but let me see if I can keep it short.

First, let's see if I understand that PR properly:

* The main motivation of that PR is to move shared sql behavior to helper gems (like the title said), mainly with the obfuscation feature. In that PR, it's extracted as `opentelemetry-helpers-sql-obfuscation` and both MySQL and Postgres (potentially more moving forward?) are using it to obfuscate SQL statements `db.statement`
  * This is a good refactoring, though I think this part is unrelated to my PR
* In addition to the `opentelemetry-helpers-sql-obfuscation` helper gem, that PR also adds `opentelemetry-helpers-mysql`, which is used by two MySQL clients: `mysql2` and `trilogy`
  * This helper contains the method somewhat related to my PR, [`OpenTelemetry::Helpers::MySQL.extract_statement_type`](https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/529/files#diff-d5f9a08cfc687dc67d13b9908ac79b44d5fc2356411545a68eaf8936fc2e11c8R65)
  * This `extract_statement_type` method takes a query and returns the statement type such as `select` or `update`
    * @fbogsany has feedback for the logic of this method (more accurately, the regexp the method is using) as described in [this](https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/529#discussion_r1355160854) and [this](https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/529/files/70058ed9583d03a2d0a6f913152944e769874102#r1355151796) PR comments
 
Now, coming back to my PR and the current implementation of Postgres client `pg` instrumentation:
 
 * With `pg` instrumentation, currently it's using `sql.to_s.split[0].to_s.upcase` to get a statement type (aka `db.operation`)
   * This is different from what MySQL is doing, where searching a query against the list of "query names"
   * Postgres also has the list of query names-ish, [`SQL_COMMANDS`](https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/27448c4feaccf051aeead7965c6b3519be0aeaa1/instrumentation/pg/lib/opentelemetry/instrumentation/pg/constants.rb#L14), and the statement type (operation) will only be used as a part of span name when it's in `SQL_COMMANDS` (otherwise ignored)
   
From here, my questions would be:

* Regarding the improvement of `OpenTelemetry::Helpers::MySQL.extract_statement_type` method (`QUERY_NAME_REGEX`) in #529 (to support query comment including query names), I'm happy to work on it. However, that PR is being worked on by @kaylareopelle and I'm unsure if I could step on her toe.
* #529 doesn't actually fix the issue #452 for Postgres, but we could use a similar logic MySQL is using to get a statement type with Postgres too. In that case, one option is to update #529 to have it, and the other option is to change this PR to have it. Former has the same concern as above, but what would work better for the community?


I'm sorry for many questions and a long comment! 🙈

kaylareopelle:
Hi @keiko713, I'm so sorry about the delay in responding to this PR. Thanks for your patience, your great questions, and this awesome fix.

Though the work is thematically similar to #529, as you correctly stated, the `OpenTelemetry::Helpers::MySQL.extract_statement_type` only applies to MySQL adapters. In my opinion, this PR should not be blocked by #529 and can be merged straightaway.

The MySQL helpers gem created in 529 is intended to reduce duplication between the trilogy and mysql2 instrumentation. Since we only instrument one Postgres gem, pg, we don't need to abstract this code out into a separate gem, nor should we try to combine it with the MySQL helpers gem.

I've also added functionality to #529 to make MySQL statement extraction respect Active Record Query Logs, so both MySQL and Postgres instrumentation via OpenTelemetry should now be compatible with this feature.

kaylareopelle (approved):

simi:
I don't really understand this need. Why would you enable this comment in app to remove it later during instrumentation? :thinking:

kaylareopelle:
> I don't really understand this need. Why would you enable this comment in app to remove it later during instrumentation? 🤔

@simi, the way I understand it, the comment isn't exactly being removed by the instrumentation. 

Instead, the comment is ignored when trying to identify what statement type is in the query (ex. SELECT). If the comment is not ignored, the wrong statement type will be chosen, making the span name less precise. 

The prepended comment will still be included in the statement if the user configured it to be included by using the default `:db_statement` or configuring `:db_statement` to `:include`.

simi:
@kaylareopelle ahh, thanks for explanation. Now it makes sense. I'm ok to move this forward. Once there will be universal SQL helpers gem, this can be moved in there.

instrumentation/pg/lib/opentelemetry/instrumentation/pg/patches/connection.rb
```diff
@@ -101,7 +101,9 @@ def span_attrs(kind, *args)
 
           def extract_operation(sql)
             # From: https://github.com/open-telemetry/opentelemetry-js-contrib/blob/9244a08a8d014afe26b82b91cf86e407c2599d73/plugins/node/opentelemetry-instrumentation-pg/src/utils.ts#L35
-            sql.to_s.split[0].to_s.upcase
+            # Ignores prepend comment
+            comment_regex = %r{\A\/\*.*?\*\/}m
```
simi, line 105: Would it make sense to move this into constant?

arielvalentin: I concur with @simi 

@keiko713 Please extract this regex into a constant to avoid allocating it for every query.

instrumentation/pg/lib/opentelemetry/instrumentation/pg/patches/connection.rb
```diff
@@ -111,7 +111,9 @@ def span_attrs(kind, *args)
...
+            sql.to_s.sub(comment_regex, '').split[0].to_s.upcase
```
arielvalentin, line 116: To confirm, we are only interested in removing the first set of long form comments (without newlines) like those generated in SQL Commenter format.

We are also not concerned about inline comments `-- this is a comment`

Is that correct?

kaylareopelle:
@keiko713 - I think the only thing holding this PR back is to [move the `comment_regex` to a constant.](https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/690/files/676425c5300a6c2eb70f8b61df9e654226781599#r1448277129) 

Once that's fixed, we should be clear to merge. Thanks for your contribution and for your patience with the review process!

instrumentation/pg/lib/opentelemetry/instrumentation/pg/patches/connection.rb
kaylareopelle: TIL Ruby freezes Regexes like it freezes strings. 

Running a file named `foo.rb` with the following ruby code:
```ruby
# frozen_string_literal: true

def foo
  %r{\A\/\*.*?\*\/}m.object_id
end

def bar
  'hello'.object_id
end

puts foo
puts foo
puts foo

puts bar
puts bar
puts bar
```

has this result:
```
➜  ~ ruby foo.rb
60
60
60
80
80
80
```

So I'm going to merge this in without moving it to a constant.

Issue #689, opened 2023-10-12
graphql 2.1.2 tests are failing


arielvalentin:
https://github.com/open-telemetry/opentelemetry-ruby-contrib/actions/runs/6488928643/job/17623342092

```console
1) Failure:
  OpenTelemetry::Instrumentation::GraphQL::Tracers::GraphQLTrace::#platform_trace::when platform_resolve_type is enabled#test_0003_includes attributes [/home/runner/work/opentelemetry-ruby-contrib/opentelemetry-ruby-contrib/instrumentation/graphql/test/instrumentation/graphql/tracers/graphql_trace_test.rb:269]:
  Expected nil to not be nil.
  
    2) Failure:
  OpenTelemetry::Instrumentation::GraphQL::Tracers::GraphQLTrace::#platform_trace::when platform_resolve_type is enabled#test_0001_traces .resolve_type [/home/runner/work/opentelemetry-ruby-contrib/opentelemetry-ruby-contrib/instrumentation/graphql/test/instrumentation/graphql/tracers/graphql_trace_test.rb:243]:
  Expected nil to not be nil.
  
    3) Failure:
  OpenTelemetry::Instrumentation::GraphQL::Tracers::GraphQLTrace::#platform_trace::when platform_resolve_type is enabled#test_0002_traces .resolve_type_lazy [/home/runner/work/opentelemetry-ruby-contrib/opentelemetry-ruby-contrib/instrumentation/graphql/test/instrumentation/graphql/tracers/graphql_trace_test.rb:256]:
  Expected nil to not be nil.
  
    4) Failure:
  OpenTelemetry::Instrumentation::GraphQL::Tracers::GraphQLTrace::#platform_trace::when platform_authorized is enabled with legacy naming#test_0001_traces .authorized [/home/runner/work/opentelemetry-ruby-contrib/opentelemetry-ruby-contrib/instrumentation/graphql/test/instrumentation/graphql/tracers/graphql_trace_test.rb:181]:
  Expected nil to not be nil.
  
    5) Failure:
  OpenTelemetry::Instrumentation::GraphQL::Tracers::GraphQLTrace::#platform_trace::when platform_field is enabled with legacy naming#test_0001_traces execute_field [/home/runner/work/opentelemetry-ruby-contrib/opentelemetry-ruby-contrib/instrumentation/graphql/test/instrumentation/graphql/tracers/graphql_trace_test.rb:124]:
  Expected nil to not be nil.
  
[...]

PR #691, opened 2023-10-13
Update test reset code for GraphQL-Ruby 2.1.3

rmosolgo:
I'm trying to address the test failures mentioned here: https://github.com/open-telemetry/opentelemetry-ruby-contrib/issues/689#issuecomment-1760704528

(Actually, I tried once already, over in https://github.com/rmosolgo/graphql-ruby/pull/4663, but I missed some failures because I was only running one of the test files.)


arielvalentin (approved):
Thanks again for your help!

PR #703, opened 2023-10-19
feat!(action_pack): Use ActiveSupport instead of patches

xuan-cao-swi:
## Description
Use ActiveSupport::Notification to subscribe the log from action controller, then modify the rack span. Remove the patch on function `dispatch` from metal.rb

Related to https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/638, https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/667, issue https://github.com/open-telemetry/opentelemetry-ruby-contrib/issues/218, https://github.com/open-telemetry/opentelemetry-ruby-contrib/issues/635

## Test
Added test assertion that focus on error handling. 
Tested with rails 6.1.7.1 and 7.1.* that made sure this change has consistent behaviour as original patching. 
Tested on basic test suite.


instrumentation/action_pack/lib/opentelemetry/instrumentation/action_pack/handlers/action_controller.rb
```diff
@@ -0,0 +1,61 @@
...
+          # @param _name [String] of the Event (unused)
+          # @param _id [String] of the event (unused)
+          # @param payload [Hash] containing job run information
+          # @return [Hash] the payload passed as a method argument
+          def start(_name, _id, payload)
+            rack_span = OpenTelemetry::Instrumentation::Rack.current_span
+
+            # from rails 6.1, the request will be added to payload
+            request = payload[:request]
+            request = payload[:headers].instance_variable_get(:@req) if ::ActionPack.version < Gem::Version.new('6.1.0')
```
xuan-cao-swi, line 29: This can be removed once the rails 6.0 is no longer supported

instrumentation/action_pack/lib/opentelemetry/instrumentation/action_pack/handlers/action_controller.rb
```diff
@@ -0,0 +1,61 @@
...
+
+            rack_span.name = "#{payload[:controller]}##{payload[:action]}" unless request.env['action_dispatch.exception']
```
arielvalentin, line 31: I would like for this to look like a templatized route if possible including the HTTP Method.  

Something like this:

```ruby
rake_span.name = "#{payload[:method]}.underscore} #{payload[:controller].underscore/#{payload[:action].underscore} "
```

Do you know if the rails route is included in the payload?

Also, are the controller and action missing in the event of an exception? I think I would like it to rename the span if they are present.

What happens in the event of a 404 route not found?

instrumentation/action_pack/lib/opentelemetry/instrumentation/action_pack/handlers/action_controller.rb
```diff
@@ -0,0 +1,61 @@
...
+
+          # Invoked by ActiveSupport::Notifications at the end of the instrumentation block
+          #
+          # @param _name [String] of the Event (unused)
+          # @param _id [String] of the event (unused)
+          # @param payload [Hash] containing job run information
+          # @return [Hash] the payload passed as a method argument
+          def finish(_name, _id, payload)
+            rack_span = OpenTelemetry::Instrumentation::Rack.current_span
+            rack_span.record_exception(payload[:exception_object]) if payload[:exception_object]
```
arielvalentin, line 53: what about `action_dispatch.exception`? I see you reference it in start but not record it?

instrumentation/action_pack/lib/opentelemetry/instrumentation/action_pack/handlers/action_controller.rb
xuan-cao-swi: > Do you know if the rails route is included in the payload?

Yes, but only in payload from finish (i.e. finish of the subscribe). In that case, something like
```ruby
def finish(_name, _id, payload)
  rack_span.name = "#{payload[:method]} #{payload[:controller]}/#{payload[:action]}" unless request.env['action_dispatch.exception']
end
```


> Also, are the controller and action missing in the event of an exception? I think I would like it to rename the span if they are present.

Yes, they are present; I can add them for sure. I am just uncertain about the initial intention behind `rack_span.name = "#{self.class.name}##{name}" unless request.env['action_dispatch.exception']`. I guess in original monkey patch require no exception so the `#{self.class.name}##{name}` will be valid string (e.g. not empty string).

> What happens in the event of a 404 route not found?

For 404, (e.g. curl 0.0.0.0:80/url_that_do_not_exist) there will be no modification on rack_span with this implementation.

We only subscribe the `process_action.action_controller`, and from [debug_exceptions.rb](https://github.com/rails/rails/blob/main/actionpack/lib/action_dispatch/middleware/debug_exceptions.rb#L31-L34), if the url is not found, the `Constants::X_CASCADE` will be `pass`, which will directly raise `ActionController::RoutingError` without going to process_action (I think that's because the [show_exceptions.rb](https://github.com/rails/rails/blob/main/actionpack/lib/action_dispatch/middleware/show_exceptions.rb#L31) treat it as non exception?).

instrumentation/action_pack/lib/opentelemetry/instrumentation/action_pack/handlers/action_controller.rb
xuan-cao-swi: > what about action_dispatch.exception? I see you reference it in start but not record it?

I had difficulty producing the `request.env['action_dispatch.exception']` from request.env although I set the exceptions_app. But I think `request.env['action_dispatch.exception']` is the same object as `payload[:exception_object]`.

In fact, if rename the span if controller and action are present, then there is no need to check `request.env['action_dispatch.exception']`

arielvalentin: Sounds good. I think that makes sense to me.

instrumentation/action_pack/lib/opentelemetry/instrumentation/action_pack/handlers/action_controller.rb
xuan-cao-swi: After digging more on actionpack, I think `process_action.action_controller` won't be invoke if it's 404 route not found error (based on [actionable_exceptions.rb](https://github.com/rails/rails/blob/main/actionpack/lib/action_dispatch/middleware/actionable_exceptions.rb#L16))

instrumentation/action_pack/lib/opentelemetry/instrumentation/action_pack/handlers/action_controller.rb
xuan-cao-swi: Just realized that the `action_dispatch.exception` will only appear if there is no exceptions_app configured. This `action_dispatch.exception` will be added in [show_exception](https://github.com/rails/rails/blob/main/actionpack/lib/action_dispatch/middleware/show_exceptions.rb#L37), however, [debug_exception](https://github.com/rails/rails/blob/main/actionpack/lib/action_dispatch/middleware/debug_exceptions.rb#L43) choose not to raise the exception to show_exceptions if `wrapper.show?(request)` is true.

Anyway, `action_dispatch.exception` and `payload[:exception_object]` should have same object; and if renaming the span with present of controller and action, then there is no need to check `action_dispatch.exception`.

arielvalentin: `debug_exception` is only for dev. mode right? Does `debug_exception` end up causing span leaks because it does not call `finish` on the notifications?

xuan-cao-swi: I think `debug_exception` is for all environment based on [default_middleware_stack.rb](https://github.com/rails/rails/blob/main/railties/lib/rails/application/default_middleware_stack.rb#L57-L59). 

> Does debug_exception end up causing span leaks because it does not call finish on the notifications?

I don't think so, since this actionpack instrumentation only modify the existing rack span, so rack instrumentation will make sure span won't be unfinished/leaked. Also, the notification is subscribed on `process_action`, and debug_exception is just part of the entire middleware stack, therefore, I think the notification will finish.

arielvalentin: 🤦‍♂️ of course. I keep building this mental model where this instrumentation makes its own spans

arielvalentin (approved):
Reviewed during the Ruby SIG on 2023-11-21

instrumentation/action_pack/lib/opentelemetry/instrumentation/action_pack/handlers/action_controller.rb
xuan-cao-swi: Done

instrumentation/action_pack/lib/opentelemetry/instrumentation/action_pack/handlers/action_controller.rb
```diff
@@ -0,0 +1,59 @@
...
+
+# Copyright The OpenTelemetry Authors
+#
+# SPDX-License-Identifier: Apache-2.0
+
+module OpenTelemetry
+  module Instrumentation
+    module ActionPack
+      module Handlers
+        # Action controller handler to handle the notification from ActiveSupport
```
kaylareopelle, line 11:
```suggestion
        # Action Controller handler to handle the notification from Active Support
```

instrumentation/action_pack/lib/opentelemetry/instrumentation/action_pack/handlers.rb
```diff
@@ -0,0 +1,38 @@
...
+          handlers_by_pattern = {
+            'process_action.action_controller' => Handlers::ActionController.new(config)
+          }
+
+          @subscriptions = handlers_by_pattern.map do |key, handler|
+            ::ActiveSupport::Notifications.subscribe(key, handler)
+          end
+        end
+
+        # Removes Event Handler Subscriptions for ActionController notifications
```
kaylareopelle, line 29:
```suggestion
        # Removes Event Handler Subscriptions for Action Controller notifications
```

instrumentation/action_pack/lib/opentelemetry/instrumentation/action_pack/handlers.rb
```diff
@@ -0,0 +1,38 @@
...
+        # @note this method is not thread safe and sholud not be used in a multi-threaded context
```
kaylareopelle, line 30:
```suggestion
        # @note this method is not thread-safe and should not be used in a multi-threaded context
```

instrumentation/action_pack/lib/opentelemetry/instrumentation/action_pack/handlers/action_controller.rb
```diff
@@ -0,0 +1,59 @@
...
+          # @param config [Hash] of instrumentation options
+          def initialize(config)
+            @config = config
+          end
+
+          # Invoked by ActiveSupport::Notifications at the start of the instrumentation block
+          #
+          # @param _name [String] of the Event (unused)
+          # @param _id [String] of the event (unused)
+          # @param payload [Hash] containing job run information
```
kaylareopelle, line 22:
```suggestion
          # @param payload [Hash] the payload passed as a method argument
```
This matches the description for #finish

kaylareopelle (commented):
Hi @xuan-cao-swi, nice work! I have some documentation suggestions. 

Tests using the example `trace_demonstration.ru` output essentially identical content before and after the change! 🎉

kaylareopelle (changes requested):

kaylareopelle (approved):
🎉

PR #704, opened 2023-10-20
Trilogy: instrument connect and ping

casperisfine:
These can be just as slow as a query if not slower.

### DIFF
--- a/instrumentation/trilogy/lib/opentelemetry/instrumentation/trilogy/patches/client.rb
+++ b/instrumentation/trilogy/lib/opentelemetry/instrumentation/trilogy/patches/client.rb
@@ -48,6 +48,28 @@ module OpenTelemetry
 
           FULL_SQL_REGEXP = Regexp.union(MYSQL_COMPONENTS.map { |component| COMPONENTS_REGEX_MAP[component] })
 
+          def initialize(options = {})
+            @connection_options = options # This is normally done by Trilogy#initialize
+
+            tracer.in_span(
+              'connect',
+              attributes: client_attributes.merge!(OpenTelemetry::Instrumentation::Trilogy.attributes),
+              kind: :client
+            ) do
+              super
+            end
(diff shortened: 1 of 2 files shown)


instrumentation/trilogy/lib/opentelemetry/instrumentation/trilogy/patches/client.rb
```diff
@@ -48,6 +48,28 @@ module Client # rubocop:disable Metrics/ModuleLength
 
           FULL_SQL_REGEXP = Regexp.union(MYSQL_COMPONENTS.map { |component| COMPONENTS_REGEX_MAP[component] })
 
+          def initialize(options = {})
+            @connection_options = options # This is normally done by Trilogy#initialize
```
casperisfine, line 52: This is a bit of a hack, it may make sense to refactor the code upstream a tiny bit to make it less hacky here.

E.g. do the initialization proper in `initialize` and extract the connection in something like `_connect`.

robertlaurin (approved):

fbogsany (approved):

instrumentation/trilogy/lib/opentelemetry/instrumentation/trilogy/patches/client.rb
composerinteralia: That's fine with me. We already have `initialize` calling `_initialize`, so if we do the ivar set in Ruby instead we can patch `_initialize` here.

casperisfine: So thanks to you merging https://github.com/trilogy-libraries/trilogy/pull/130, in future versions of trilogy we might be able to just decorate `_connect` and get rid of this hack.

But overall I don't think it's a big deal, just a bit fragile if upstream refactor things out.

Issue #705, opened 2023-10-21
Remove Explicit Dependencies on ActiveSupport Core Extentions


arielvalentin:
**Description of the bug**

Rails 7.1 is no longer auto-loading some core language extensions[^1], however we have a few instances of our instrumentation depending on those methods. 

We _must not_ rely on dependencies from any gem outside of the Ruby standard library. 

[^1]: https://github.com/rails/rails/commit/0170745b376acd150fec5f8cc57253cc1ffe0cf2

This is causing errors for some of our end users who are in the process of upgrading to Rails 7.1:

```console
ERROR -- : OpenTelemetry error: unexpected configuration error due to undefined method `blank?' for "eu-west-1":String - OpenTelemetry::SDK::ConfigurationError
```

I had replaced the usage of `blank?` in the `gruf` gem when adding test coverage for Rails 7.1 but it seems we still have references in at least the `grape` instrumentation.

cc: @muripic for Grape 

**Share details about your runtime**

Ruby 3.2 
Rails 7.1

muripic:
Thanks for flagging, @arielvalentin ! I'll try to submit a PR fixing this today 😃

PR #706, opened 2023-10-23
fix: Remove dependency on ActiveSupport core extensions from Grape instrumentation

muripic:
Addresses https://github.com/open-telemetry/opentelemetry-ruby-contrib/issues/705

I checked if we are using any other core extensions apart from `blank?` using this as a reference: https://guides.rubyonrails.org/active_support_core_extensions.html

Also, instead of replacing directly with `empty?`, I added an extra check for `nil?`. We shouldn't get `nil` values here, but I added it as a safeguard in case there are any changes in the Grape gem that change this behaviour.

Another case that was covered by `blank?` but not by `empty?` is pure whitespace strings, but I decided not to add this since it would be very unlikely to have path with whitespace.

### DIFF
--- a/instrumentation/grape/lib/opentelemetry/instrumentation/grape/event_handler.rb
+++ b/instrumentation/grape/lib/opentelemetry/instrumentation/grape/event_handler.rb
@@ -96,7 +96,7 @@ module OpenTelemetry
             version = endpoint.routes.first.options[:version] || ''
             prefix = endpoint.routes.first.options[:prefix]&.to_s || ''
             parts = [prefix, version] + namespace.split('/') + endpoint.options[:path]
-            parts.reject { |p| p.blank? || p.eql?('/') }.join('/').prepend('/')
+            parts.reject { |p| p.nil? || p.empty? || p.eql?('/') }.join('/').prepend('/')
           end
 
           def formatter_type(formatter)


arielvalentin (approved):

instrumentation/grape/lib/opentelemetry/instrumentation/grape/event_handler.rb
```diff
@@ -96,7 +96,7 @@ def path(endpoint)
             version = endpoint.routes.first.options[:version] || ''
             prefix = endpoint.routes.first.options[:prefix]&.to_s || ''
             parts = [prefix, version] + namespace.split('/') + endpoint.options[:path]
-            parts.reject { |p| p.blank? || p.eql?('/') }.join('/').prepend('/')
+            parts.reject { |p| p.nil? || p.empty? || p.eql?('/') }.join('/').prepend('/')
```
arielvalentin, line 99: Would safe navigation work?
```suggestion
            parts.reject { |p| p&.empty? || p.eql?('/') }.join('/').prepend('/')
```

arielvalentin (approved):

instrumentation/grape/lib/opentelemetry/instrumentation/grape/event_handler.rb
muripic: In this case, no, because `p&.empty?` returns `nil` if `p` is `nil`, which evaluates to `false` as a boolean.

```ruby
3.1.3 :001 > p = nil
 => nil
3.1.3 :002 > p&.empty?
 => nil
3.1.3 :003 > p.nil? || p.empty? || p.eql?('/')
 => true
3.1.3 :004 > p&.empty? || p.eql?('/')
 => false
```

I guess there are refactor alternatives, but even though it is a bit verbose I thought it was quite readable like this. But of course I'm open to other suggestions 😃

PR #707, opened 2023-10-25
fix: remove call to ActiveSupport::Notifications.notifier#synchronize deprecated in Rails 7.2

octokatherine:
When running Rails 7.2, we get the error 

```
NoMethodError: undefined method `synchronize' for #<ActiveSupport::Notifications::Fanout (1436 patterns)>
...opentelemetry/instrumentation/active_support/span_subscriber.rb:33:in `subscribe'
```

because the private `synchronize` API was removed here: https://github.com/rails/rails/commit/bcdeea5da7657375df21a856135ae7a66a8c3812. 

This `span_subscriber` code block was a workaround for a bug that was fixed in Rails 7.0, broken again in Rails 7.1, and will be fixed in Rails 7.1.2: https://github.com/rails/rails/pull/49784

I wrapped the workaround code in a conditional based on Rails version, since the code fails in 7.2 but will also be unneeded.


instrumentation/active_support/lib/opentelemetry/instrumentation/active_support/span_subscriber.rb
```diff
@@ -30,19 +30,21 @@ def self.subscribe(
...
-            OpenTelemetry.handle_error(
-              message: 'Unable to move OTEL ActiveSupport Notifications subscriber to the front of the notifications list which may cause incomplete traces.' \
-                       'Please report an issue here: ' \
-                       'https://github.com/open-telemetry/opentelemetry-ruby-contrib/issues/new?labels=bug&template=bug_report.md&title=ActiveSupport%20Notifications%20subscribers%20list%20is%20nil'
-            )
-          else
-            subscribers.unshift(
-              subscribers.delete(subscriber_object)
-            )
+        if Gem::Version.new(::Rails.version) < Gem::Version.new('7.2')
```
arielvalentin, line 33: Is this disabling forcing the subscriber to the from of the list just for Rails 7.2? 

Is there something available in 7.2 that allows us to force this subscriber to be inserted in the front or that not possible because the notification order non-deterministic?

simi, line 33: Not sure what's best practice, but alternative would be to check if `synchronize` method is present to make it more like feature detection instead of version detection.

octokatherine: From my understanding, you shouldn't need to force the subscriber to be first anymore after Rails 7.2, because  the new changes in Rails guarantee that if an exception occurs within a subscriber the remaining subscribers will still be run. I believe this whole code block was written to work around that problem

octokatherine: I like that idea, I'll update it

jhawthorn: @arielvalentin by design there isn't anything available in any version of Rails which allows you to do this, and it will not be added. Everyone thinks their subscriber is the most important and wants it to go first 😅.

arielvalentin (approved):
Thanks again @octokatherine I will get a release out a bit later today.

arielvalentin:
@robertlaurin @simi merging if there are no objections. 

We need some help setting up canaries against Rails edge if you are up for it @octokatherine !!!

robertlaurin (approved):

instrumentation/active_support/lib/opentelemetry/instrumentation/active_support/span_subscriber.rb
```diff
@@ -30,19 +30,21 @@ def self.subscribe(
...
-            OpenTelemetry.handle_error(
-              message: 'Unable to move OTEL ActiveSupport Notifications subscriber to the front of the notifications list which may cause incomplete traces.' \
-                       'Please report an issue here: ' \
-                       'https://github.com/open-telemetry/opentelemetry-ruby-contrib/issues/new?labels=bug&template=bug_report.md&title=ActiveSupport%20Notifications%20subscribers%20list%20is%20nil'
-            )
-          else
-            subscribers.unshift(
-              subscribers.delete(subscriber_object)
-            )
+        if ::ActiveSupport::Notifications.notifier.respond_to?(:synchronize)
```
robertlaurin, line 33: Do we want to leave any comments to our future selves that we can remove this block once everything prior to 7.2 is EoL?

octokatherine: I'm happy to add a comment!

PR #709, opened 2023-10-28
feat(rack)! Use Rack Events By Default

arielvalentin:
I have been running the `Rack::Events` based instrumentation in our production workloads since 2023-10-23.

There is no noticeable difference in performance, however applications using `Rack::BodyProxy` to stream responses will notice a slight increase in latency since it will include timings writing response output to the socket.


instrumentation/rack/example/trace_demonstration.rb
```diff
@@ -18,7 +18,7 @@
 builder = Rack::Builder.app do
   # integration should be automatic in web frameworks (like rails),
   # but for a plain Rack application, enable it in your config.ru, e.g.,
-  use OpenTelemetry::Instrumentation::Rack::Middlewares::TracerMiddleware
+  use ::Rack::Events, [OpenTelemetry::Instrumentation::Rack::Middlewares::EventHandler.new]
```
kaylareopelle, line 22: Since using the `TracerMiddleware` is still a configurable option, do you think it's valuable to mention it in one of the Rack examples?

instrumentation/rack/lib/opentelemetry/instrumentation/rack/instrumentation.rb
```diff
@@ -29,7 +29,7 @@ class Instrumentation < OpenTelemetry::Instrumentation::Base
         option :untraced_requests,        default: nil,   validate: :callable
         option :response_propagators,     default: [],    validate: :array
         # This option is only valid for applicaitons using Rack 2.0 or greater
```
kaylareopelle, line 31:
```suggestion
        # This option is only valid for applications using Rack 2.0 or greater
```

instrumentation/rails/test/instrumentation/opentelemetry/instrumentation/rails/instrumentation_test.rb
```diff
@@ -1,13 +0,0 @@
...
-# Copyright The OpenTelemetry Authors
-#
-# SPDX-License-Identifier: Apache-2.0
-
-require 'test_helper'
-
-describe OpenTelemetry::Instrumentation::Rails::Instrumentation do
-  it 'adds the rack tracing middleware' do
-    _(DEFAULT_RAILS_APP.config.middleware).must_include OpenTelemetry::Instrumentation::Rack::Middlewares::TracerMiddleware
-  end
```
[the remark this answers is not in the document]
arielvalentin: I don't think so 🤔 we'd need to break encapsulation of the Rack middleware in a sense and fetch the constructor args.

instrumentation/rack/example/trace_demonstration.rb
arielvalentin: Sure I could add more. I think the follow up PR would be to remove the middleware implementation entirely.

kaylareopelle: Got it. Let's leave it as-is then.

kaylareopelle (approved):

PR #712, opened 2023-11-06
feat(excon)! Add a connect span to excon and add more span attributes to the tracer middleware

misalcedo:
Added a span for TLS and TCP socket connections that matches the one in [net_http](https://github.com/open-telemetry/opentelemetry-ruby-contrib/tree/main/instrumentation/net_http). Also, I switched to using semantic convention constants over hard-coded strings where applicable. Lastly, I added a few more attributes to the request span.

The untraced_hosts option was added to mirror net_http and allow for selective exclusion of instrumentation.

The attributes before were: `http.host`, `http.method`, `http.scheme`, `http.target` and `peer.service`. I added peer name and port as those were also included in `net_http`.

## Testing
The tests are based off the ones found in the `net_http` instrumentation for the connect span. In writing the tests I found 2 things worth calling out:

- When disabling mocking, WebMock creates an extra span because it creates a new connection with a non-mocked middleware chain.
- Excon does not handle TLS proxies properly in that is still uses plaintext TCP on a TLS proxy because the scheme is based purely on the end host.


instrumentation/excon/lib/opentelemetry/instrumentation/excon/patches/socket.rb
```diff
@@ -0,0 +1,60 @@
...
+            tracer.in_span(span_name, attributes: attributes, kind: span_kind) do
+              super
+            end
+          end
+
+          def tracer
+            Excon::Instrumentation.instance.tracer
+          end
+
+          def untraced?
```
arielvalentin, line 45: This was inspired by the `Net::HTTP` instrumentation correct?

Since that time we have taken a different approach, which is to use an `untraced` helper, which is more consistent and reliable because it suppresses all child spans as opposed to potentially creating orphans: 

https://github.com/open-telemetry/opentelemetry-ruby/blob/d365dfeb3982f09cd62a17c6e67eb3792f56e460/common/lib/opentelemetry/common/utilities.rb#L90

Here is an example usage:

https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/7ba26160ec240f27bd763507ab1c3664eeb2107e/instrumentation/que/lib/opentelemetry/instrumentation/que/patches/poller.rb#L22

instrumentation/excon/lib/opentelemetry/instrumentation/excon/patches/socket.rb
```diff
@@ -0,0 +1,60 @@
...
+      module Patches
+        # Module to prepend to an Excon Socket for instrumentation
+        module Socket
+          private
+
+          def connect
+            return super if untraced?
+
+            if @data[:proxy]
+              conn_address = @data[:proxy][:hostname]
```
arielvalentin, line 19: Consider using `Hash#dig` https://www.rubydoc.info/stdlib/core/Hash:dig

instrumentation/excon/lib/opentelemetry/instrumentation/excon/patches/socket.rb
```diff
@@ -0,0 +1,60 @@
...
+              conn_port = @data[:proxy][:port]
+            else
+              conn_address = @data[:hostname]
+              conn_port = @port
+            end
+
+            attributes = { OpenTelemetry::SemanticConventions::Trace::NET_PEER_NAME => conn_address, OpenTelemetry::SemanticConventions::Trace::NET_PEER_PORT => conn_port }.merge!(OpenTelemetry::Common::HTTP::ClientContext.attributes)
+
+            if is_a?(::Excon::SSLSocket) && @data[:proxy]
```
arielvalentin, line 28: Are these reusable by the same socket object instance? 

Is it possible to memoize this decision in some way?

instrumentation/excon/test/opentelemetry/instrumentation/excon/instrumentation_test.rb
```diff
@@ -150,4 +151,143 @@
...
+    ensure
+      WebMock.disable_net_connect!
+    end
+
+    it 'emits a "connect" span when connecting through an non-ssl proxy' do
+      WebMock.allow_net_connect!
+
+      _(-> { Excon.get('http://localhost', proxy: 'https://proxy_user:proxy_pass@localhost') }).must_raise(Excon::Error::Socket)
+
+      _(exporter.finished_spans.size).must_equal(3)
```
arielvalentin, line 284: is there anything interesting about the other 2 spans that we would want to review/check?

instrumentation/excon/lib/opentelemetry/instrumentation/excon/middlewares/tracer_middleware.rb
```diff
@@ -71,43 +75,32 @@ def self.around_default_stack
...
-                span.status = OpenTelemetry::Trace::Status.error("Request has failed: #{datum[:error]}") if datum.key?(:error)
+              span.status = OpenTelemetry::Trace::Status.error("Request has failed: #{datum[:error]}") if datum.key?(:error)
 
-                span.finish
-                datum.delete(:otel_span)
-              end
+              span.finish
             end
           rescue StandardError => e
             OpenTelemetry.logger.debug(e.message)
```
arielvalentin, line 92: This is not related to your PR but I think these should be `handle_error` instead of `logger.debug`

instrumentation/excon/lib/opentelemetry/instrumentation/excon/middlewares/tracer_middleware.rb
```diff
@@ -71,43 +75,32 @@ def self.around_default_stack
...
-                  span.status = OpenTelemetry::Trace::Status.error unless (100..399).include?(response[:status].to_i)
-                end
+              if datum.key?(:response)
+                response = datum[:response]
+                span.set_attribute(OpenTelemetry::SemanticConventions::Trace::HTTP_STATUS_CODE, response[:status])
+                span.status = OpenTelemetry::Trace::Status.error unless (100..399).include?(response[:status].to_i)
+              end
 
-                span.status = OpenTelemetry::Trace::Status.error("Request has failed: #{datum[:error]}") if datum.key?(:error)
+              span.status = OpenTelemetry::Trace::Status.error("Request has failed: #{datum[:error]}") if datum.key?(:error)
```
arielvalentin, line 87: Also not related to your PR but do you know if `error` is an Exception object? If so then it should probably recorded using `record_exception`?

instrumentation/excon/lib/opentelemetry/instrumentation/excon/middlewares/tracer_middleware.rb
```diff
@@ -71,43 +75,32 @@ def self.around_default_stack
           private
 
           def handle_response(datum)
-            if datum.key?(:otel_span)
-              datum[:otel_span].tap do |span|
-                return span if span.end_timestamp
+            datum.delete(:otel_span)&.tap do |span|
+              return span if span.end_timestamp
```
arielvalentin, line 79: Hmm I think this should be `unless span.recording?`, which covers more than cases where it was "finished".

instrumentation/excon/lib/opentelemetry/instrumentation/excon/middlewares/tracer_middleware.rb
```diff
@@ -22,23 +22,27 @@ class TracerMiddleware < ::Excon::Middleware::Base
...
+              OpenTelemetry::SemanticConventions::Trace::NET_PEER_PORT => datum[:port]
+            }
+
+            peer_service = Excon::Instrumentation.instance.config[:peer_service]
+            attributes[OpenTelemetry::SemanticConventions::Trace::PEER_SERVICE] = peer_service if peer_service
+            attributes.merge!(OpenTelemetry::Common::HTTP::ClientContext.attributes)
+
+            datum[:otel_span] = tracer.start_span(HTTP_METHODS_TO_SPAN_NAMES[http_method], attributes: attributes, kind: :client)
+
+            OpenTelemetry::Trace.with_span(datum[:otel_span]) do
```
arielvalentin, line 44: This is not related to your PR, and I am not familiar with Excon, but I think this is not implemented the way I expect. 

I would have expected `@stack.request_call(datam)` to be nested under `with_span` to ensure that any other middlewares would also have access to the `current_span` and the correct context. 

This instrumentation is sharing the span in the datum but then _disconnecting_ the context it belongs to. You do not have to address this as part of the PR but I imagine this _may_ cause problems for us when we use it in our applications.

instrumentation/excon/test/opentelemetry/instrumentation/excon/instrumentation_test.rb
```diff
@@ -150,4 +151,143 @@
...
+      instrumentation.install(untraced_hosts: ['foobar.com', /bazqux\.com/])
+
+      stub_request(:get, 'http://example.com/body').to_return(status: 200)
+      stub_request(:get, 'http://foobar.com/body').to_return(status: 200)
+      stub_request(:get, 'http://bazqux.com/body').to_return(status: 200)
+    end
+
+    it 'does not create a span when request ignored using a string' do
+      Excon.get('http://foobar.com/body')
+      _(exporter.finished_spans.size).must_equal 0
```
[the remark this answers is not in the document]
misalcedo: Done.

instrumentation/excon/test/opentelemetry/instrumentation/excon/instrumentation_test.rb
```diff
@@ -150,4 +151,143 @@
...
+  end
+
+  # NOTE: WebMock introduces an extra HTTP request span due to the way the mocking is implemented.
+  describe '#connect' do
+    before do
+      instrumentation.install
+    end
+
+    it 'emits span on connect' do
+      WebMock.allow_net_connect!
```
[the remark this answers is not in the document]
misalcedo: They do. I followed the pattern of net http and I vaguely remember seeing that in minitest specs groups of test with `describe` and `context` don't actually do anything but change the test name. However, seems like it does work correctly for this library so I changed it.

instrumentation/excon/lib/opentelemetry/instrumentation/excon/patches/socket.rb
misalcedo: Done

instrumentation/excon/lib/opentelemetry/instrumentation/excon/patches/socket.rb
misalcedo: A socket is only ever connected once. We could potentially make this decision here instead: https://github.com/excon/excon/blob/v0.104.0/lib/excon/connection.rb#L470-L479

However, that would require a patch on the Connection as well as the Socket class. Also, memoizing it would be difficult because the inputs to the decision can change at any point so we would need to handle those updates.

instrumentation/excon/lib/opentelemetry/instrumentation/excon/patches/socket.rb
misalcedo: I switched to using untraced instead of just calling super when we connect. Do we you recommend removing the `:untraced_hosts` check as well?

instrumentation/excon/lib/opentelemetry/instrumentation/excon/middlewares/tracer_middleware.rb
misalcedo: Done

instrumentation/excon/lib/opentelemetry/instrumentation/excon/middlewares/tracer_middleware.rb
misalcedo: Yes, it is an exception: https://github.com/excon/excon/blob/v0.104.0/lib/excon/connection.rb#L308C20-L308C20

instrumentation/excon/lib/opentelemetry/instrumentation/excon/middlewares/tracer_middleware.rb
misalcedo: Done.

instrumentation/excon/lib/opentelemetry/instrumentation/excon/middlewares/tracer_middleware.rb
misalcedo: Done

instrumentation/excon/lib/opentelemetry/instrumentation/excon/middlewares/tracer_middleware.rb
misalcedo: I went ahead and did that. However, I would have expected the current span to be in context until another span took over or until the span was finished. Is there a way to start the span, let it be in context, and then finish it without requiring the code to be run in a block?

For example, the span should be in context until we finish it in the response or error call methods.

instrumentation/excon/test/opentelemetry/instrumentation/excon/instrumentation_test.rb
misalcedo: The last span is because of the way WebMock mocks requests in Excon, so we can always ignore that one. The middle span, we could check that it recorded the exception but we already did that outside of the connect context.

misalcedo: I added assertions on this anyways just to be thorough.

kaylareopelle:
Hi @misalcedo, the excon suite is raising some rubocop offenses: 
```
Offenses:

lib/opentelemetry/instrumentation/excon/middlewares/tracer_middleware.rb:50:1: C: [Correctable] Layout/TrailingWhitespace: Trailing whitespace detected.
lib/opentelemetry/instrumentation/excon/patches/socket.rb:16:13: C: [Correctable] Style/IfUnlessModifier: Favor modifier if usage when having a single-line body. Another good alternative is the usage of control flow &&/||.
            if untraced?
            ^^
test/opentelemetry/instrumentation/excon/instrumentation_test.rb:277:33: C: [Correctable] Style/StringLiterals: Prefer single-quoted strings when you don't need string interpolation or special symbols.
      assert_http_spans(scheme: "https")
                                ^^^^^^^
```

instrumentation/excon/lib/opentelemetry/instrumentation/excon/middlewares/tracer_middleware.rb
arielvalentin: Hmmm. You know what, I am second guessing now since I am not sure what the underlying behavior of the stack is. 

I think you are right and we probably do not want to use `with_span` but rather manage the context stack ourselves. 

Checkout how we do that in these instrumentations that use callback hooks and cannot be wrapped in a block:

https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/main/instrumentation/racecar/lib/opentelemetry/instrumentation/racecar/process_message_subscriber.rb#L21

https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/main/instrumentation/rack/lib/opentelemetry/instrumentation/rack/middlewares/event_handler.rb#L57

instrumentation/excon/lib/opentelemetry/instrumentation/excon/middlewares/tracer_middleware.rb
```diff
@@ -71,43 +79,35 @@ def self.around_default_stack
...
+              if datum.key?(:response)
+                response = datum[:response]
+                span.set_attribute(OpenTelemetry::SemanticConventions::Trace::HTTP_STATUS_CODE, response[:status])
+                span.status = OpenTelemetry::Trace::Status.error unless (100..399).include?(response[:status].to_i)
+              end
 
-                span.finish
-                datum.delete(:otel_span)
+              if datum.key?(:error)
+                span.status = OpenTelemetry::Trace::Status.error("Request has failed: #{datum[:error]}")
```
arielvalentin, line 92: Since we are recording the exception we probably should remove the dynamic portion of the message. I am not sure why the original author would have chosen interpolation instead of recording the error.

instrumentation/excon/lib/opentelemetry/instrumentation/excon/patches/socket.rb
arielvalentin: Not at all. How do you feel about extracting a shared module function for the `untraced_hosts` check and reusing it here?

arielvalentin:
@misalcedo check out this PR which has identified another use case where `untraced?` should be taken into consideration: https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/722/files

misalcedo:
@kaylareopelle thanks for the tip, I addressed the rubocop lints.

instrumentation/excon/lib/opentelemetry/instrumentation/excon/middlewares/tracer_middleware.rb
misalcedo: That makes sense to me. Went ahead and made this message static.

instrumentation/excon/lib/opentelemetry/instrumentation/excon/middlewares/tracer_middleware.rb
misalcedo: I went ahead and switch to calls to `OpenTelemetry::Context.attach` and `OpenTelemetry::Context.detach` based on [HTTPX](https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/02799c87f9bc9fb8af4ee90b2d704f1b213c891f/instrumentation/httpx/lib/opentelemetry/instrumentation/httpx/plugin.rb#L60) and other examples in the repo. The links you shared do the same thing, but it was a bit harder for me to follow what exactly they were doing.

instrumentation/excon/lib/opentelemetry/instrumentation/excon/patches/socket.rb
misalcedo: I added a new `OpenTelemetry::Instrumentation::Concerns::UntracedHosts` module in the base directory and used that in the instrumentation class. Then, I updated the middleware and patch to call out to the instrumentation's `untraced?(host)` method. Is this what you had in mind?

arielvalentin: My initial though was to extract it as a module in the excon instrumentation gem but it does seem like sharing it will all instrumentations would be quite useful. 

Adding features to the Base gem gets a little tricky. Any changes we make to it must be released first and then child gems needs to be updated to have min version that supports the shared attribute. 

Since that process is a bit more challenging I would like to suggest that you isolate the shared module in this gem then we can follow up with a separate PR that moves the shared logic into the Base gem.

misalcedo: Will do.

misalcedo: Done, I moved the module to the excon gem with a similar path to what it would be when in the base gem.

arielvalentin: Nice! I'll take another look this afternoon. If all goes well I'll cut a release

instrumentation/excon/lib/opentelemetry/instrumentation/excon/middlewares/tracer_middleware.rb
```diff
@@ -71,43 +77,35 @@ def self.around_default_stack
           private
 
           def handle_response(datum)
-            if datum.key?(:otel_span)
-              datum[:otel_span].tap do |span|
-                return span if span.end_timestamp
+            datum.delete(:otel_span)&.tap do |span|
+              return span unless span.recording?
```
arielvalentin, line 81: I was a little confused about it returning the span explicitly. 

This is really intended to be a short cut to exit the `handle_response` method correct?

```suggestion
              return unless span.recording?
```

instrumentation/excon/lib/opentelemetry/instrumentation/excon/middlewares/tracer_middleware.rb
```diff
@@ -71,43 +77,35 @@ def self.around_default_stack
...
-                  response = datum[:response]
-                  span.set_attribute('http.status_code', response[:status])
-                  span.status = OpenTelemetry::Trace::Status.error unless (100..399).include?(response[:status].to_i)
-                end
-
-                span.status = OpenTelemetry::Trace::Status.error("Request has failed: #{datum[:error]}") if datum.key?(:error)
+              if datum.key?(:response)
+                response = datum[:response]
+                span.set_attribute(OpenTelemetry::SemanticConventions::Trace::HTTP_STATUS_CODE, response[:status])
+                span.status = OpenTelemetry::Trace::Status.error unless (100..399).include?(response[:status].to_i)
```
arielvalentin, line 86: Nit: Is there any performance gain if we extracted this range as a Constant?

arielvalentin (approved):
Thanks for this contribution 🎉 

I have a few nits in there but otherwise LGTM!

instrumentation/excon/lib/opentelemetry/instrumentation/excon/middlewares/tracer_middleware.rb
misalcedo: I believe this returned the span because [Span#finish](https://www.rubydoc.info/gems/opentelemetry-sdk/OpenTelemetry/SDK/Trace/Span#finish-instance_method) used to be the last expression in this method and thus a span would normally be returned. Your change makes sense to me as this method should have a void return type.

instrumentation/excon/lib/opentelemetry/instrumentation/excon/middlewares/tracer_middleware.rb
misalcedo: I whipped together a quick benchmark. The answer is it doesn't matter.
I ran the benchmark with and without GC and the results were the same.

## Benchmark
```
require "benchmark"

CONSTANT = (100..399)

n = 10_000_000

GC.disable if ENV.include?("DISABLE_GC")

Benchmark.bmbm do |x|
  x.report("constant") do
    n.times do
      CONSTANT.include?(200) && !CONSTANT.include?(500)
    end
  end

  x.report("inline") do
    n.times do
      (100..399).include?(200) && !(100..399).include?(500)
    end
  end
end
```

## Results
### With GC
```
Rehearsal --------------------------------------------
constant   1.923770   0.025237   1.949007 (  1.989501)
inline     1.867091   0.025007   1.892098 (  1.996068)
----------------------------------- total: 3.841105sec

               user     system      total        real
constant   1.900203   0.027615   1.927818 (  1.980085)
inline     1.850816   0.021425   1.872241 (  1.909520)
```

### Without GC
```
Rehearsal --------------------------------------------
constant   1.870138   0.014792   1.884930 (  1.896819)
inline     1.826463   0.018247   1.844710 (  2.032935)
----------------------------------- total: 3.729640sec

               user     system      total        real
constant   1.851058   0.014311   1.865369 (  1.878788)
inline     1.793849   0.014889   1.808738 (  1.822874)
```

arielvalentin: Thanks for humoring me and ultimately answering my question :)

instrumentation/excon/lib/opentelemetry/instrumentation/excon/middlewares/tracer_middleware.rb
```diff
@@ -71,43 +77,35 @@ def self.around_default_stack
...
```
arielvalentin, line 86:
```suggestion
                span.status = OpenTelemetry::Trace::Status.error unless (100..399).cover?(response[:status].to_i)
```

arielvalentin:
@misalcedo we added the rubcop-performance linters on main and now there is a conflict sorry 😿

misalcedo:
@arielvalentin I addressed the merge conflict. Seems like I just needed to change `.include?` to `.cover?` for the status range chack.

arielvalentin (approved):

PR #715, opened 2023-11-08
Removal Sinatra 1.x Appraisal

### LINKED ISSUE #364: Drop Support for Rack and Sinatra 1.x
Sinatra depends on Rack 1.x which does not support `Rack::Events` API. 

- [ ] Write up documentation for last supported version of Sinatra 1.x
- [ ] Upgrade minimal gemspec dependencies to Sinatra > 2

zacheryph:
Sinatra 1.x is not supported as it does not provide the Sinatra::Event API. It does not need to be appraised.

This should close #364.

This issue recommends documenting the last supported version for Sinatra 1.x, Given that the last release of Sinatra was January 2017, is this still something that is advised?

The gemspec references "sinatra", without a strict version requirement. I was going to update this but I noticed all the other instrumentation packages don't reference minimal versions for dependencies either.

### DIFF
--- a/instrumentation/rack/README.md
+++ b/instrumentation/rack/README.md
@@ -12,6 +12,19 @@ gem install opentelemetry-instrumentation-rack
 
 Or, if you use [bundler][bundler-home], include `opentelemetry-instrumentation-rack` in your `Gemfile`.
 
+### Version Compatibility
+
+Older versions of Rack are not supported by the latest version of this instrumentation. If you are using an older version of Rack and need an earlier version of this instrumentation, then consider installing and pinning the compatible gem version, e.g.:
+
+```console
+gem opentelemetry-instrumentation-rack, "<version>"
+```
+
+| Rack Version | Instrumentation Version |
+| ------------ | ----------------------- |
+| `< 2.0`      | `= 0.22.1`              |
+| `>= 2.0`     | `~> 0.23`               |
+
 ## Usage
 
 To use the instrumentation, call `use` with the name of the instrumentation:
--- a/instrumentation/sinatra/Appraisals
+++ b/instrumentation/sinatra/Appraisals
@@ -11,7 +11,3 @@ end
 appraise 'sinatra-2.x' do
   gem 'sinatra', '~> 2.1'
 end
-
-appraise 'sinatra-1.x' do
-  gem 'sinatra', '~> 1.4'
-end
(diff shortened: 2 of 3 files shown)


simi:
@zacheryph hello! Change looks good. Can you please update commit message according to https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/main/CONTRIBUTING.md#use-conventional-commit-messages?

simi (approved):

arielvalentin:
I think all we need a matrix that contains the last supported version in a markdown file.

I also agree that the gemspec should also reflect the minimum version we support. We tend to encode that at runtime but making it explicit in the gemspec certainly helps IMO.

As far as this PR is concerned, removing the tests from the suite is the first step. If you would like to also include the min version in the gemspec as part of this PR I would be grateful.

simi:
> If you would like to also include the min version in the gemspec as part of this PR I would be grateful.

https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/7ba26160ec240f27bd763507ab1c3664eeb2107e/instrumentation/sinatra/opentelemetry-instrumentation-sinatra.gemspec#L41

arielvalentin:
@simi that is bizarre. Does that mean the appraisal gem is overriding the sinatra version specified in the gemspec somehow?

simi:
> @simi that is bizarre. Does that mean the appraisal gem is overriding the sinatra version specified in the gemspec somehow?

Yes, both ways. Up (for sinatra 3) and down (for sinatra 1).

arielvalentin:
>  That is how gemspec included in Gemfile works (it is not Appraisals related) for development dependencies.

This is going to sound like I have no idea how to Ruby. Is this because appraisals are overriding development dependencies with runtime installs?

simi:
@arielvalentin so as 
deivid-rodriguez pointed me out, it is actually "bug" (https://github.com/rubygems/rubygems/pull/6014). Better to not rely on that IMHO.

zacheryph:
@arielvalentin This is getting slightly messy at this point 🤣  (also rebased to capture latest main)

after looking at dependency requirements, the bigger issue i think here is Rack instrumentation and not Sinatra specifically. I went ahead and added an excerpt to the Rack README, as well as the Sinatra README. The latest Sinatra instrumentation should technically still work since it has a dependency version on otel-inst-rack `~> 0.21`, and 0.22.1 is the latest version that will support Rack 1.x (does not introduce the Rack::Event API)

Issue #714, opened 2023-11-07
Add option for X-Ray remote sampling


gillesbergerp:
We have successfully integrated OpenTelemetry with various of our Java applications (with X-Ray as a backend). Luckily, the OpenTelemetry Java SDK supports an X-Ray remote sampler that periodically polls X-Ray for sampling rules and applies them based on resource attributes (service name/type/...).
It would be great if the Ruby SDK would add support for this functionality as well since it makes managing sampling across a fleet of applications much easier.

mwear:
Thanks for opening this issue. We've moved it to the contrib repo. Is there any chance you'd be interested in contributing this functionality?

gillesbergerp:
Definitely. I just don't know how much time I can dedicate to this. I'd also need a little guidance as to where it would best fit in. Should we start a new `sampling` package similar to `instrumentaion` and `propagator`?

arielvalentin:
@gillesbergerp yes, I think that creating a new **sampling** directory makes sense. 

I think each sampler should be encapsulated in its own gem.

gillesbergerp:
I agree, that makes total sense. I'll see if I can find some time on the weekend and will reach out if I need anything.

gillesbergerp:
X-Ray falls back to "1r/s or 5% after that" sampling rule if it cannot find any matching remote rules. I thought about making that implicit in the X-Ray sampling gem but then came to the conclusion that this could be interesting as a general option as well. At the very least, it will keep the X-Ray sampling implementation less complex. What do you think?
See https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/717

gillesbergerp:
Please see https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/719 for the X-Ray gem. We'll also need adaptions in https://github.com/gillesbergerp/opentelemetry-ruby (specifically [here](https://github.com/gillesbergerp/opentelemetry-ruby/blob/main/sdk/lib/opentelemetry/sdk/trace/tracer_provider.rb#L167)) to allow for initializing the new sampler.

PR #717, opened 2023-11-11
Hierarchical and Rate Limited Sampler Implementations

gillesbergerp:
The hierarchical sampler allows for specifying a list of samplers that will be invoked in order. It will sample a request if any of its children would.
The Rate Limited sampler on the other hand samples requests at a fixed rate.

This is a prerequisite for https://github.com/open-telemetry/opentelemetry-ruby-contrib/issues/714


sampling/hierarchical/README.md
```diff
@@ -0,0 +1,42 @@
...
+## How do I get started?
+
+Install the gem using:
+
+```
+gem install opentelemetry-sampling-hierarchical
+```
+
+Or, if you use [bundler][bundler-home], include `opentelemetry-sampling-hierarchical` in your `Gemfile`.
+
```
arielvalentin, line 24: I think it would be helpful to add example usage in the README and/or an `example` directory with a script that shows how to configure it.

sampling/hierarchical/lib/opentelemetry/sampling/hierarchical/sampler.rb
```diff
@@ -0,0 +1,37 @@
+# frozen_string_literal: true
+
+# Copyright The OpenTelemetry Authors
+#
+# SPDX-License-Identifier: Apache-2.0
+
+module OpenTelemetry
+  module Sampling
+    module Hierarchical
```
arielvalentin, line 9: Is Hierarchical Sampler a term used in X-Ray to describe this pattern? As opposed to a `Any` or `All` sampler?

sampling/hierarchical/lib/opentelemetry/sampling/hierarchical/version.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+# Copyright OpenTelemetry Authors
+#
+# SPDX-License-Identifier: Apache-2.0
+
+module OpenTelemetry
```
arielvalentin, line 7: I think the package the package naming should be symmetric with the SDK:

```ruby
OpenTelemetry::SDK::Trace::Samplers
````

Or more use the API namespace

```ruby
OpenTelemetry::Trace::Samplers
````

@open-telemetry/ruby-contrib-maintainers WDYT?

File: sampling/hierarchical/opentelemetry-sampling-hierarchical.gemspec [hunk 4]
```diff
@@ -0,0 +1,43 @@
...
+  spec.email = ['cncf-opentelemetry-contributors@lists.cncf.io']
+
+  spec.summary = 'Hierarchical Sampling Extension for the OpenTelemetry framework'
+  spec.description = 'Hierarchical Sampling Extension for the OpenTelemetry framework'
+  spec.homepage = 'https://github.com/open-telemetry/opentelemetry-ruby-contrib'
+  spec.license = 'Apache-2.0'
+
+  spec.files = Dir.glob('lib/**/*.rb') + Dir.glob('*.md') + %w[LICENSE .yardopts]
+  spec.require_paths = %w[lib]
+  spec.required_ruby_version = '>= 2.6.0'
```
arielvalentin, line 25: We do not support versions of EOL Ruby versions and target Ruby 3.0 syntax.

```suggestion
  spec.required_ruby_version = '>= 3.0.0'
```

sampling/hierarchical/test/opentelemetry/sampling/hierarchical/sampler_test.rb
```diff
@@ -0,0 +1,168 @@
...
+require('test_helper')
+
+describe(OpenTelemetry::Sampling::Hierarchical::Sampler) do
+  describe('#should_sample?') do
+    it('returns DROP if it has no samplers') do
+      sampler = OpenTelemetry::Sampling::Hierarchical::Sampler.new([])
+
+      _(
+        sampler.should_sample?(
+          trace_id: SecureRandom.uuid.to_s,
```
arielvalentin, line 16: Consider using static string literals here instead of random number generation since it is not making a decision on dynamic data.

sampling/hierarchical/test/opentelemetry/sampling/hierarchical/sampler_test.rb
```diff
@@ -0,0 +1,168 @@
...
+        ).sampled?
+      ).must_equal(false)
+    end
+
+    it('returns RECORD_AND_SAMPLE if the first sampler returns RECORD_AND_SAMPLE') do
+      trace_id = SecureRandom.uuid.to_s
+      name = SecureRandom.uuid.to_s
+
+      first_sampler = Minitest::Mock.new
+      second_sampler = Minitest::Mock.new
```
arielvalentin, line 31: Though not required, we tend to prefer using `rspec-mocks` over `minitest` mocks due to it being a bit more feature rich.

sampling/hierarchical/test/opentelemetry/sampling/hierarchical/sampler_test.rb
```diff
@@ -0,0 +1,168 @@
...
+          parent_context: nil,
+          links: [],
+          name: name,
+          kind: :internal,
+          attributes: {}
+        ).sampled?
+      ).must_equal(true)
+
+      first_sampler.verify
+      second_sampler.verify
```
arielvalentin, line 61: This will fail with an unexpected method invocation I presume?

sampling/rate_limited/lib/opentelemetry/sampling/rate_limited/sampler.rb
```diff
@@ -0,0 +1,63 @@
...
+# SPDX-License-Identifier: Apache-2.0
+
+module OpenTelemetry
+  module Sampling
+    module RateLimited
+      class Sampler
+        def initialize(credits_per_second)
+          @credits_per_second = credits_per_second
+          @balance = 0
+          @last_tick = Time.now.to_i
```
arielvalentin, line 14: Please use Process clocks for more precise time calculations:

https://ruby-doc.org/3.2.2/Process.html#method-c-clock_gettime

sampling/rate_limited/lib/opentelemetry/sampling/rate_limited/version.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+# Copyright OpenTelemetry Authors
+#
+# SPDX-License-Identifier: Apache-2.0
+
+module OpenTelemetry
+  module Sampling
+    module RateLimited
+      VERSION = '0.0.1'
```
arielvalentin, line 10:
```suggestion
      VERSION = '0.0.0'
```

sampling/rate_limited/test/opentelemetry/sampling/rate_limited/sampler_test.rb
```diff
@@ -0,0 +1,86 @@
...
+            parent_context: nil,
+            links: [],
+            name: SecureRandom.uuid.to_s,
+            kind: :internal,
+            attributes: {}
+          ).sampled?
+        ).must_equal(false)
+      end
+
+      sleep(1)
```
arielvalentin, line 70: Something to consider with tests that require sleeping is that other factors may result in non-deterministic behavior of tests like GC pauses. 

You may want to consider simulating time calculations using test doubles

sampling/hierarchical/test/opentelemetry/test_helper.rb
arielvalentin, line 1: why do you have two test helper files?

sampling/rate_limited/lib/opentelemetry/sampling/rate_limited/sampler.rb
```diff
@@ -0,0 +1,63 @@
...
+
+        # @param [String] trace_id
+        # @param [OpenTelemetry::Context] parent_context
+        # @param [Enumerable<Link>] links
+        # @param [String] name
+        # @param [Symbol] kind
+        # @param [Hash<String, Object>] attributes
+        # @return [OpenTelemetry::SDK::Trace::Samplers::Result] The sampling result.
+        def should_sample?(trace_id:, parent_context:, links:, name:, kind:, attributes:)
+          if can_spend?
```
robertlaurin, line 26: Sampling is done per span so if you run out of credits_per_second mid trace you will start dropping spans, and potentially end up with broken traces.

sampling/rate_limited/lib/opentelemetry/sampling/rate_limited/sampler.rb
```diff
@@ -0,0 +1,63 @@
...
+          end
+        end
+
+        private
+
+        # @return [Boolean]
+        def can_spend?
+          return false if @credits_per_second <= 0
+
+          @lock.synchronize do
```
robertlaurin, line 45: The sampler is set on the tracer provider so you'll be locking every time you create a span, this may not be appropriate depending on how much span volume your service is handling.

sampling/rate_limited/lib/opentelemetry/sampling/rate_limited/sampler.rb
```diff
@@ -0,0 +1,63 @@
...
+              decision: OpenTelemetry::SDK::Trace::Samplers::Decision::DROP,
+              tracestate: OpenTelemetry::Trace.current_span(parent_context).context.tracestate
+            )
+          end
+        end
+
+        private
+
+        # @return [Boolean]
+        def can_spend?
```
robertlaurin, line 42: Do we care about how this performs in the case where an app is configured with Puma, with Unicorn? Forking vs Pre-forking web servers will result in different spend rates. What's the expected outcome here?

sampling/hierarchical/README.md
gillesbergerp: Unfortunately, there is no way yet to use it. This will require an update to https://github.com/open-telemetry/opentelemetry-ruby/blob/d365dfeb3982f09cd62a17c6e67eb3792f56e460/sdk/lib/opentelemetry/sdk/trace/tracer_provider.rb#L167. I will open a PR to support the [X-Ray sampler](https://github.com/open-telemetry/opentelemetry-ruby-contrib) there as well and will adjust the README afterwards accordingly if that's fine with you

sampling/hierarchical/lib/opentelemetry-sampling-hierarchical.rb
```diff
@@ -0,0 +1,8 @@
+# frozen_string_literal: true
+
+# Copyright OpenTelemetry Authors
+#
+# SPDX-License-Identifier: Apache-2.0
+
+require('opentelemetry-api')
```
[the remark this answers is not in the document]
gillesbergerp: Oh, sorry for that. That's the pitty when switching between projects with different conventions...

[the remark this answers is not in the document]
gillesbergerp: I just checked and this is not enforced by Rubocop, it seems. Should that be changed?

sampling/hierarchical/lib/opentelemetry/sampling/hierarchical/sampler.rb
gillesbergerp: It is not but I find `Any` sampler to be a bit misleading since it doesn't imply a fixed order.

sampling/hierarchical/test/opentelemetry/sampling/hierarchical/sampler_test.rb
gillesbergerp: Yes, it would

sampling/rate_limited/lib/opentelemetry/sampling/rate_limited/sampler.rb
gillesbergerp: Wouldn't this be better solved by instantiating a `ParentBased` sampler with the `RateLimited` sampler as `root`? I'D rather avoid making these distinctions here

sampling/rate_limited/lib/opentelemetry/sampling/rate_limited/sampler.rb
gillesbergerp: Do you suggest being optimistic and not locking at all?

sampling/rate_limited/lib/opentelemetry/sampling/rate_limited/sampler.rb
gillesbergerp: This is a good point that I haven't considered. I believe that the best option is to not make any assumptions about this at all. Configuring the rate accordingly could restore the parity between forking and pre-forking servers. Should we note this in the README?

sampling/hierarchical/lib/opentelemetry-sampling-hierarchical.rb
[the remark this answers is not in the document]
arielvalentin: Yes please. 

Generally speaking, you should avoid using parenthesis for declarations and DSLs and conditional expressions unless it is absolutely necessary.

It's a difficult rule for Rubocop to enforce in some cases but it does have some examples of a rules that it enforces by default where it enforces omitting parenthesis. E.g. for method declaration if there are no arguments:

https://docs.rubocop.org/rubocop/cops_style.html#styledefwithparentheses

sampling/hierarchical/lib/opentelemetry/sampling/hierarchical/sampler.rb
arielvalentin: I guess this name is not intuitive to me. Would you be able to help me understand what makes it Hierarchical or Ranked ordering vs "If any of the items in this list return true regardless of order"?

sampling/hierarchical/lib/opentelemetry-sampling-hierarchical.rb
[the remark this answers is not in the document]
gillesbergerp: I hope I found and changed every instance

sampling/hierarchical/lib/opentelemetry/sampling/hierarchical/sampler.rb
gillesbergerp: The list of samplers is iterated left to right. If any returns a positive response, we stop and say that the span should be sampled. `Any` doesn't give you this guarantee (in my interpretation).
The order can make a difference. A contrived example would be having two samplers sampling the following spans
* Sampler `A`: one per second, only if `kind == :internal`
* Sampler `B`: one per second, only if `kind in [:server, :internal]`

Let's say the application produces two traces in the same second, one with `kind = :internal` and one with `kind = :server` (in this order). If sampler `A` was applied first, both traces would be sampled. On the other hand, if sampler `B` was first, only the first (`:internal`) trace would be sampled since `B` already spent all its credits when the second span arrives.

sampling/hierarchical/test/opentelemetry/sampling/hierarchical/sampler_test.rb
gillesbergerp: I tend to lean the other way. If I were to pass a static value, the test would only argue about this one value and nothing else

sampling/hierarchical/test/opentelemetry/sampling/hierarchical/sampler_test.rb
gillesbergerp: Oh, I didn't know. The examples I checked all used minitest. Can I resolve this point nevertheless?

sampling/hierarchical/lib/opentelemetry-sampling-hierarchical.rb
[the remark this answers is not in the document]
gillesbergerp: I was just about to do the same for the X-Ray PR and have a really hard time. I don't want to argue, but what exactly are the benefits of omitting parentheses? The left version is without a doubt more consistent and easier to parse
image

sampling/rate_limited/lib/opentelemetry/sampling/rate_limited/sampler.rb
arielvalentin: It's more that I think we want clarity on what the expectations are for the scope of the rate limiter.

Does the rate limit apply per process, per thread, or per fiber?

sampling/rate_limited/lib/opentelemetry/sampling/rate_limited/sampler.rb
arielvalentin: What do you mean by making the parent based sampler as root?

Wouldn't that make it so the rate limited sampling result is always ignored since the Hierarchical Sampler will always check the other samplers to see if it should override the result?

sampling/rate_limited/lib/opentelemetry/sampling/rate_limited/sampler.rb
arielvalentin: I don't think we have a suggestion as much as we're doing is raising concerns about the performance implications of adding another lock around span creation.

I'm not sure how much latency this introduces for applications that generate thousands of spans in a single transaction.

sampling/rate_limited/lib/opentelemetry/sampling/rate_limited/sampler.rb
gillesbergerp: Per process makes most sense, I believe. I don't see why different processes should have this knowledge of each other. It also makes the handling easier as we don't need to know the details of different servers

sampling/rate_limited/lib/opentelemetry/sampling/rate_limited/sampler.rb
gillesbergerp: Since default thread counts are fairly low in most server implementations (in Puma it's `5`, for instance for MRI). We can also provide a configuration option for that

PR #721, opened 2023-11-15
feat: add aws Lambda instrumentation

xuan-cao-swi:
This instrumentation allow user to use the handler to extract parent context from header that produce distributed tracing.

Inspired by splunk [ruby lambda layer](https://github.com/signalfx/splunk-otel-lambda/tree/main/ruby) and opentelemetry [python lambda layer](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/instrumentation/opentelemetry-instrumentation-aws-lambda/src/opentelemetry/instrumentation/aws_lambda/__init__.py). 
Ruby layer creation can be found in opentelemetry-lambda [PR](https://github.com/open-telemetry/opentelemetry-lambda/pull/918)

Sample wrapper should be added in layer creation (with lambda instrumentation).
```ruby
# wrapper.rb
require 'opentelemetry/sdk'
require 'opentelemetry/exporter/otlp'
require 'opentelemetry/instrumentation/all'

  c.service_name = '<YOUR_SERVICE_NAME>'
  c.use_all() # enables all instrumentation!
end

def otel_wrapper(event:, context:)
  otel_wrapper = OpenTelemetry::Instrumentation::AwsLambda::Handler.new()
  otel_wrapper.call_wrapped(event: event, context: context)
end
```

To test with aws lambda, you can try to follow the steps [here](https://gist.github.com/xuan-cao-swi/151a8348504b4c4fcd4019a5aa05d82a)


instrumentation/aws_lambda/opentelemetry-instrumentation-aws_lambda.gemspec
```diff
@@ -0,0 +1,50 @@
...
+
+  spec.add_development_dependency 'appraisal', '~> 2.5'
+  spec.add_development_dependency 'bundler', '~> 2.4'
+  spec.add_development_dependency 'minitest', '~> 5.0'
+  spec.add_development_dependency 'opentelemetry-sdk', '~> 1.1'
+  spec.add_development_dependency 'opentelemetry-test-helpers', '~> 0.3'
+  spec.add_development_dependency 'pry'
+  spec.add_development_dependency 'pry-byebug' unless RUBY_ENGINE == 'jruby'
+  spec.add_development_dependency 'rspec-mocks'
+  spec.add_development_dependency 'rubocop', '~> 1.56.1'
```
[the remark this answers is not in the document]
xuan-cao-swi: Yes, I have added rubocop-performance in gemspec.

instrumentation/aws_lambda/lib/opentelemetry/instrumentation/aws_lambda/handler.rb
```diff
@@ -0,0 +1,132 @@
...
+          original_handler_parts = original_handler.split('.')
+          if original_handler_parts.size == 2
+            hanlder_file, @handler_method = original_handler_parts
+          elsif original_handler_parts.size == 3
+            hanlder_file, @handler_class, @handler_method = original_handler_parts
+          else
+            OpenTelemetry.logger.warn("aws-lambda instrumentation: Invalid handler #{original_handler}, must be of form FILENAME.METHOD or FILENAME.CLASS.METHOD.")
+          end
+
+          require hanlder_file
```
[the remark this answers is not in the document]
xuan-cao-swi: Yes, for testing, my current solution is to overwrite the handler class that don't require the file but require it in `trace_demonstration.rb`. In lambda environment, the system will find the handler file automatically either is in subdirectory or root directory.

instrumentation/aws_lambda/README.md
```diff
@@ -0,0 +1,60 @@
...
+```
+
+Or, if you use [bundler][bundler-home], include `opentelemetry-instrumentation-aws_lambda` in your `Gemfile`.
+
+## Usage
+
+From Lambda Layer side, create the wrapper. More information can be found in https://github.com/open-telemetry/opentelemetry-lambda
+```ruby
+require 'opentelemetry/sdk'
+require 'opentelemetry/instrumentation/all'
```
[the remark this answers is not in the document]
xuan-cao-swi: It actually just need require opentelemetry-instrumentation-lambda for the access of handler class. I have changed the README.

xuan-cao-swi:
> @xuan-cao-swi - Are there any tests I should run using both your opentelemetry-lambda PR and this one?

I will create a detailed step to test lambda-related instrumentation.

xuan-cao-swi:
@kaylareopelle I have created a [gist](https://gist.github.com/xuan-cao-swi/151a8348504b4c4fcd4019a5aa05d82a) for the steps; but let me know if it's uncleared or failed.

instrumentation/aws_lambda/lib/opentelemetry/instrumentation/aws_lambda/version.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+# Copyright The OpenTelemetry Authors
+#
+# SPDX-License-Identifier: Apache-2.0
+
+module OpenTelemetry
+  module Instrumentation
+    module AwsLambda
+      VERSION = '0.0.1'
```
kaylareopelle, line 10:
```suggestion
      VERSION = '0.1.0'
```
Since the third column is for patches, I think moving the one to the second/minor column may make more sense.

kaylareopelle (changes requested):
@xuan-cao-swi, following up from our update in the SIG, I wasn't able to get the telemetry data into New Relic, but I don't think this needs to be a blocker for the PR. Besides that, everything worked as expected. 

I'd like to get this into the hands of users. It seems ready for a 0.x release. 

I have requests for a few more small changes

xuan-cao-swi:
> I wasn't able to get the telemetry data into New Relic, but I don't think this needs to be a blocker for the PR. Besides that, everything worked as expected.

Are you using otel [default samplers](https://github.com/open-telemetry/opentelemetry-ruby/blob/main/sdk/lib/opentelemetry/sdk/trace/samplers.rb)? If that's the case, I need to go back to check if I made some mistakes either in layer side or instrumentation side.

xuan-cao-swi:
I have added another instruction/result [here](https://gist.github.com/xuan-cao-swi/f75925a1dc290588de7e48b5edf5fe83) that I was experimenting with Jaeger on aws lambda instrumentation. 

This instrumentation can also help complete the lambda layer build on [opentelemetry-lambda](https://github.com/open-telemetry/opentelemetry-lambda)

kaylareopelle:
@xuan-cao-swi - I was able to get the lambda instrumentation to work with your updated instructions! 🎉 

The key was adding the `AWS_LAMBDA_EXEC_WRAPPER=/opt/ruby/otel-handler` environment variable.

instrumentation/aws_lambda/lib/opentelemetry/instrumentation/aws_lambda/instrumentation.rb
```diff
@@ -0,0 +1,33 @@
...
+      # Instrumentation class that detects and installs the AwsLambda instrumentation
+      class Instrumentation < OpenTelemetry::Instrumentation::Base
+        install do |_config|
+          require_dependencies
+        end
+
+        present do
+          # maybe check if ORIG_HANDLER or _HANLDER exist
+          true
+        end
```
kaylareopelle, line 19: Could we check to see if the gem is installed? Maybe with `defined?(Aws::Lambda)`? 

With the current code, if we add this to `instrumentation-all`, I think the lambda instrumentation will attempt to install every time.

instrumentation/aws_lambda/lib/opentelemetry/instrumentation/aws_lambda/instrumentation.rb
```diff
@@ -0,0 +1,33 @@
...
+
+        compatible do
+          true
+        end
```
kaylareopelle, line 23: Are you aware of any versions of the gem that aren't compatible with the instrumentation? Are there any versions that have been EOL'd by AWS that we explicitly don't want to support?

instrumentation/aws_lambda/lib/opentelemetry/instrumentation/aws_lambda/version.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+# Copyright The OpenTelemetry Authors
+#
+# SPDX-License-Identifier: Apache-2.0
+
+module OpenTelemetry
+  module Instrumentation
+    module AwsLambda
+      VERSION = '0.1.0'
```
kaylareopelle, line 10: @arielvalentin  - I think I got this wrong with the helper gems. Does toys expect `0.0.0`? (sorry about all the churn on this, @xuan-cao-swi)

CODEOWNERS
```diff
@@ -31,3 +31,5 @@ instrumentation/racecar/ @chrisholmes @open-telemetry/ruby-contrib-maintainers @
 instrumentation/rspec/ @chrisholmes @open-telemetry/ruby-contrib-maintainers @open-telemetry/ruby-contrib-approvers @fbogsany @mwear @robertlaurin @dazuma @ericmustin @arielvalentin @ahayworth @plantfansam @robbkidd @simi @kaylareopelle @xuan-cao-swi
 
 instrumentation/que/ @indrekj @open-telemetry/ruby-contrib-maintainers @open-telemetry/ruby-contrib-approvers @fbogsany @mwear @robertlaurin @dazuma @ericmustin @arielvalentin @ahayworth @plantfansam @robbkidd @simi @kaylareopelle @xuan-cao-swi
+
+instrumentation/aws_lambda/ @indrekj @open-telemetry/ruby-contrib-maintainers @open-telemetry/ruby-contrib-approvers @fbogsany @mwear @robertlaurin @dazuma @ericmustin @arielvalentin @ahayworth @plantfansam @robbkidd @simi @kaylareopelle @xuan-cao-swi
```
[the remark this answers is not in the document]
kaylareopelle: Ok, this might not be the best advice. We're chatting about it here: https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/879#discussion_r1496308460

instrumentation/aws_lambda/lib/opentelemetry/instrumentation/aws_lambda/handler.rb
```diff
@@ -0,0 +1,132 @@
...
+        # Extract context from request headers
+        def call_wrapped(event:, context:)
+          parent_context   = extract_parent_context(event)
+          span_attributes  = event['version'] == '2.0' ? v2_proxy_attributes(event) : v1_proxy_attributes(event)
+          original_handler = resolve_original_handler
+          response         = call_original_handler(event: event, context: context)
+
+          span_attributes[OpenTelemetry::SemanticConventions::Trace::HTTP_STATUS_CODE] = response['statusCode'] if response.instance_of?(Hash) && response['statusCode']
+          span_attributes[OpenTelemetry::SemanticConventions::Resource::FAAS_ID]       = context.invoked_function_arn
+          span_attributes[OpenTelemetry::SemanticConventions::Trace::FAAS_EXECUTION]   = context.aws_request_id
```
kaylareopelle, line 27: It's nice to see these attributes. Is there a spec you followed to add these specific items? I found this: https://opentelemetry.io/docs/specs/semconv/faas/

xuan-cao-swi: Not really following the spec. This code is copied from splunk and also took reference on python lambda.

instrumentation/aws_lambda/lib/opentelemetry/instrumentation/aws_lambda/instrumentation.rb
xuan-cao-swi: I think aws instrumentation doesn't rely on any external gem that require prepend on methods, so I think it should always be compatible; but please correct me if I miss something.

instrumentation/aws_lambda/lib/opentelemetry/instrumentation/aws_lambda/handler.rb
```diff
@@ -0,0 +1,132 @@
...
+
+          OpenTelemetry::Context.with_current(parent_context) do
+            span = tracer.start_span(
+              original_handler,
+              attributes: span_attributes,
+              kind: :server # Span kind MUST be `:server` for a HTTP server span
+            )
+          rescue Exception => e # rubocop:disable Lint/RescueException
```
xuan-cao-swi, line 35: Should we rescue in function level rather than in context level?

instrumentation/aws_lambda/lib/opentelemetry/instrumentation/aws_lambda/instrumentation.rb
kaylareopelle: I like that!

instrumentation/aws_lambda/lib/opentelemetry/instrumentation/aws_lambda/instrumentation.rb
kaylareopelle: You're correct, thanks for clarifying!

instrumentation/aws_lambda/lib/opentelemetry/instrumentation/aws_lambda/handler.rb
kaylareopelle: Maybe we should follow-up this PR with a few issues to bring the instrumentation into compliance? I think the key specs are: 
* [faas spans](https://opentelemetry.io/docs/specs/semconv/faas/faas-spans/)
* [aws lambda](https://opentelemetry.io/docs/specs/semconv/faas/aws-lambda/)

CODEOWNERS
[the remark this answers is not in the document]
kaylareopelle:
```suggestion
```

Since your name is already listed in the reviewers, I don't think we need this. 

Instead I'd recommend this approach: https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/879#discussion_r1496463167

instrumentation/aws_lambda/lib/opentelemetry/instrumentation/aws_lambda/handler.rb
kaylareopelle: Yes, I think that's a good call.

instrumentation/aws_lambda/lib/opentelemetry/instrumentation/aws_lambda/instrumentation.rb
xuan-cao-swi: Thanks, since we are not requiring gem like [aws-sdk-lambda](https://github.com/aws/aws-sdk-ruby/blob/version-3/gems/aws-sdk-lambda/lib/aws-sdk-lambda.rb) , so `Aws::Lambda` won't be presented.

kaylareopelle (changes requested):
Thanks for these updates, @xuan-cao-swi! I like the error handling updates you made and think the new span attributes look great. Just a few small wording change suggestions and some readme questions.

instrumentation/aws_lambda/README.md
```diff
@@ -0,0 +1,60 @@
...
+
+```
+gem install opentelemetry-instrumentation-aws_lambda
+```
+
+Or, if you use [bundler][bundler-home], include `opentelemetry-instrumentation-aws_lambda` in your `Gemfile`.
+
+## Usage
+
+From the Lambda Layer side, create the wrapper. More information can be found at https://github.com/open-telemetry/opentelemetry-lambda
```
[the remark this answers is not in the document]
xuan-cao-swi: Yes, I can create PR to add that in their README along with the PR that add wrapper to point to this instrumentation.

> If Ruby's present on the main repo README, then I think we might not need the link to the specific Ruby directory.

Agree

instrumentation/aws_lambda/README.md
```diff
@@ -0,0 +1,60 @@
...
+
+## How do I get started?
+
+Install the gem using:
+
+```
+gem install opentelemetry-instrumentation-aws_lambda
+```
+
+Or, if you use [bundler][bundler-home], include `opentelemetry-instrumentation-aws_lambda` in your `Gemfile`.
```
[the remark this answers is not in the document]
xuan-cao-swi: No, I don't think we should ask user to install this gem. The [layer](https://github.com/open-telemetry/opentelemetry-lambda/blob/main/ruby/src/layer/Gemfile) should include everything they need to instrument aws instrumentation.

p.s. we should remove install all instrumentation like [line 5](https://github.com/open-telemetry/opentelemetry-lambda/blob/main/ruby/src/layer/Gemfile#L5), and just install this aws lambda instrumentation

instrumentation/aws_lambda/lib/opentelemetry/instrumentation/aws_lambda/handler.rb
```diff
@@ -0,0 +1,172 @@
...
+        # We want to capture the error if user's handler is causing issue
+        # but our wrapper and handler shouldn't cause any issue
+        def call_wrapped(event:, context:)
+          parent_context   = extract_parent_context(event)
+          span_attributes  = otel_attributes(event, context)
+
+          response = nil
+          original_handler_error = nil
+          begin
+            response = call_original_handler(event: event, context: context)
```
cheempz, line 35: @xuan-cao-swi question: it looks like we're calling the wrapped (actual) function handler here, but it's only afterwards in line 45 that we're starting a span for the function? If that's true it is counter to what I expect an instrumenter to do--it should capture the timing of the actual traced operation, the function handler execution in this case.

instrumentation/aws_lambda/lib/opentelemetry/instrumentation/aws_lambda/handler.rb
```diff
@@ -0,0 +1,172 @@
...
+            status_code = response.is_a?(Hash) ? (response['statusCode'] || response[:statusCode]) : nil
+          rescue StandardError => e
+            original_handler_error = e
+            status_code = '500'
```
cheempz, line 39: I don't think instrumentation can assume this is an HTTP invocation trigger since there are [many trigger types](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/faas/faas-spans.md#function-trigger-type) and blindly collect a status code attribute.

instrumentation/aws_lambda/lib/opentelemetry/instrumentation/aws_lambda/handler.rb
xuan-cao-swi: Yes, after double check with python instrumentation, the span time should include the execution of wrapped function. Updated.

instrumentation/aws_lambda/lib/opentelemetry/instrumentation/aws_lambda/handler.rb
xuan-cao-swi: Updated the part that checking the type of trigger. Thanks!

instrumentation/aws_lambda/lib/opentelemetry/instrumentation/aws_lambda/handler.rb
```diff
@@ -0,0 +1,201 @@
...
+          resolve_original_handler
+        end
+
+        # We want to capture the error if user's handler is causing issue
+        # but our wrapper and handler shouldn't cause any issue
+        def call_wrapped(event:, context:)
+          parent_context = extract_parent_context(event)
+
+          span_kind = nil
+          span_kind = if event['Records'] && ['aws:sqs', 'aws:s3', 'aws:sns', 'aws:dynamodb'].include?(event['Records'].dig(0, 'eventSource'))
```
kaylareopelle, line 32: To avoid instantiating the strings every time this method is called, what do you think about defining a constant for this array? Some of the other strings might benefit from being defined as constants too.

instrumentation/aws_lambda/lib/opentelemetry/instrumentation/aws_lambda/instrumentation.rb
```diff
@@ -0,0 +1,33 @@
...
+
+        # determine if current environment is lambda by checking _HANLDER or ORIG_HANDLER
+        present do
+          (ENV.key?('_HANDLER') || ENV.key?('ORIG_HANDLER'))
+        end
+
+        compatible do
+          true
+        end
```
kaylareopelle, line 23: I think this will return `true` if it's not defined, so we might not need this here.

https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/7141afaf8b78f16098d9c754cb4fc781fd6ee06e/instrumentation/base/lib/opentelemetry/instrumentation/base.rb#L246-L250

kaylareopelle (commented):
I think this looks good, @xuan-cao-swi. I haven't been able to successfully send data with these changes by creating a new layer, but I think there's something else going on in my environment preventing it.

instrumentation/aws_lambda/lib/opentelemetry/instrumentation/aws_lambda/instrumentation.rb
xuan-cao-swi: updated!

instrumentation/aws_lambda/lib/opentelemetry/instrumentation/aws_lambda/handler.rb
xuan-cao-swi: Yes, that's good idea, updated!

instrumentation/aws_lambda/lib/opentelemetry/instrumentation/aws_lambda/handler.rb
```diff
@@ -0,0 +1,203 @@
...
+          @original_handler = ENV['ORIG_HANDLER'] || ENV['_HANDLER'] || ''
+          @handler_class    = nil
+          @handler_method   = nil
+          @handler_file     = nil
+
+          resolve_original_handler
+        end
+
+        # We want to capture the error if user's handler is causing issue
+        # but our wrapper and handler shouldn't cause any issue
```
cheempz, line 29:
```suggestion
        # Try to record and re-raise any exception from the wrapped function handler
        # Instrumentation should never raise its own exception
```

instrumentation/aws_lambda/lib/opentelemetry/instrumentation/aws_lambda/handler.rb
```diff
@@ -0,0 +1,203 @@
...
+              status_code = response['statusCode'] || response[:statusCode] if response.is_a?(Hash)
+              span.set_attribute(OpenTelemetry::SemanticConventions::Trace::HTTP_STATUS_CODE, status_code) if status_code
+            rescue StandardError => e
+              original_handler_error = e
+            ensure
+              original_response = response
+            end
+          rescue StandardError => e
+            span&.record_exception(e)
+            span&.status = OpenTelemetry::Trace::Status.error("Unhandled exception of type: #{e.class}")
```
cheempz, line 61: If I'm reading correctly this is handling an exception from instrumentation. In this case it should just be logged, not recorded. The Python example has try/catch [very specific to the calling of the wrapped function](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/fdcbbddb6c753e5e9d494ba399a5b4bcab4afc3f/instrumentation/opentelemetry-instrumentation-aws-lambda/src/opentelemetry/instrumentation/aws_lambda/__init__.py#L368-L373) (to record), then [other instrumention operations which might raise](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/fdcbbddb6c753e5e9d494ba399a5b4bcab4afc3f/instrumentation/opentelemetry-instrumentation-aws-lambda/src/opentelemetry/instrumentation/aws_lambda/__init__.py#L397-L401) are always tried and logged.

instrumentation/aws_lambda/lib/opentelemetry/instrumentation/aws_lambda/handler.rb
```diff
@@ -0,0 +1,203 @@
...
+          ensure
+            if original_handler_error
+              span&.record_exception(original_handler_error)
+              span&.status = OpenTelemetry::Trace::Status.error("Original lambda handler exception: #{original_handler_error.class}. Please check if you have the correct handler setting and code in your lambda function.")
```
cheempz, line 65: The description string is optional and I vote for either not setting it like Python implementation, or to something extracted from the actual exception similar to the JS [instrumentation](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/fe18e2fbb2a6535cb72f314fdb1550a3a4160403/plugins/node/opentelemetry-instrumentation-aws-lambda/src/instrumentation.ts#L325-L335).

cheempz (commented):
Thanks @xuan-cao-swi! Another batch of comments, I'll take a look at the collected attributes next week.

instrumentation/aws_lambda/lib/opentelemetry/instrumentation/aws_lambda/handler.rb
```diff
@@ -0,0 +1,203 @@
...
+            end
+            span&.finish
+            OpenTelemetry.tracer_provider.force_flush(timeout: @flush_timeout)
```
cheempz, line 68: Just occurred to me that we're not flushing the meter provider at end of span, which we should per the [OTel Lambda README](https://github.com/open-telemetry/opentelemetry-lambda?tab=readme-ov-file#flush-meterprovider).

cheempz (commented):
Final batch of comments from me @xuan-cao-swi ;). I decided to skip commenting on the API Gateway proxy integration attributes because the semconv specs probably need more clarification whether both the [HTTP _and_ REST style APIs](https://docs.aws.amazon.com/apigateway/latest/developerguide/http-api-vs-rest.html#http-api-vs-rest.differences.integrations) should be treated the same, something to follow-up with the OTel lambda folks.

instrumentation/aws_lambda/lib/opentelemetry/instrumentation/aws_lambda/handler.rb
xuan-cao-swi: Currently there is no stable metrics sdk therefore this meter provider force flush will be added in the future improvement.

instrumentation/aws_lambda/lib/opentelemetry/instrumentation/aws_lambda/handler.rb
xuan-cao-swi: Thanks! Updated!

instrumentation/aws_lambda/lib/opentelemetry/instrumentation/aws_lambda/handler.rb
xuan-cao-swi: Thanks! Updated!

instrumentation/aws_lambda/lib/opentelemetry/instrumentation/aws_lambda/handler.rb
```diff
@@ -0,0 +1,203 @@
...
+          attributes[OpenTelemetry::SemanticConventions::Trace::HTTP_TARGET] += "?#{event['rawQueryString']}" if event['rawQueryString']
+          attributes
+        end
+
+        # fass.trigger set to http: https://github.com/open-telemetry/semantic-conventions/blob/main/docs/faas/aws-lambda.md#api-gateway
+        # TODO: need to update Semantic Conventions for invocation_id, trigger and resource_id
+        def otel_attributes(event, context)
+          span_attributes = {}
+          span_attributes['faas.invocation_id'] = context.aws_request_id
+          span_attributes['cloud.resource_id'] = "#{context.invoked_function_arn};#{context.aws_request_id};#{context.function_name}"
```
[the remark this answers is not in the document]
xuan-cao-swi: Updated!

instrumentation/aws_lambda/lib/opentelemetry/instrumentation/aws_lambda/handler.rb
```diff
@@ -0,0 +1,203 @@
...
+          span_attributes[OpenTelemetry::SemanticConventions::Trace::AWS_LAMBDA_INVOKED_ARN] = context.invoked_function_arn
+
+          account_id = context.invoked_function_arn.split(':')[4]
```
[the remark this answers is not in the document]
xuan-cao-swi: Updated!

instrumentation/aws_lambda/lib/opentelemetry/instrumentation/aws_lambda/instrumentation.rb
```diff
@@ -0,0 +1,33 @@
...
+    module AwsLambda
+      # Instrumentation class that detects and installs the AwsLambda instrumentation
+      class Instrumentation < OpenTelemetry::Instrumentation::Base
+        install do |_config|
+          require_dependencies
+        end
+
+        present do
+          # maybe check if ORIG_HANDLER or _HANLDER exist
+          true
```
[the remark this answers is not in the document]
xuan-cao-swi: This instrumentation doesn't rely on any dependencies. It's designed specifically for the AWS Lambda environment, so if the environment variable `_HANDLER` is present, it indicates the presence of AWS Lambda, and this tool should be installed accordingly.

instrumentation/aws_lambda/README.md
```diff
@@ -0,0 +1,60 @@
...
+From the Lambda Layer side, create the wrapper. More information can be found at https://github.com/open-telemetry/opentelemetry-lambda
+```ruby
+require 'opentelemetry/sdk'
+require 'opentelemetry/instrumentation/aws_lambda'
+OpenTelemetry::SDK.configure do |c|
+  c.service_name = '<YOUR_SERVICE_NAME>'
+  c.use 'OpenTelemetry::Instrumentation::AwsLambda'
+end
+
+def otel_wrapper(event:, context:)
```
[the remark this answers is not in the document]
xuan-cao-swi: otel_wrapper is a custom script that is provided by user (mostly included in layer itself, see this [example](https://github.com/open-telemetry/opentelemetry-lambda/blob/main/python/src/otel/otel_sdk/otel_wrapper.py) from python). And in aws lambda configuration, user can specify the location of wrapper through [AWS_LAMBDA_EXEC_WRAPPER](https://docs.aws.amazon.com/lambda/latest/dg/runtimes-modify.html).

The call sequence will be: wrapper -> opentelemetry-instrumentation-aws-lambda's handler -> user's lambda function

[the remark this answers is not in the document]
xuan-cao-swi: > Do you have to do this yourself everytime in the lambda script?

No, it should be setup in wrapper script, which will be loaded in runtime from created layer.

> Would it make sense to add a creation method or command method or something?

It kind need initialize first to resolve the original function, and rest of other language seems use this convention.

instrumentation/aws_lambda/lib/opentelemetry/instrumentation/aws_lambda/handler.rb
```diff
@@ -0,0 +1,132 @@
...
+              attributes: span_attributes,
+              kind: :server # Span kind MUST be `:server` for a HTTP server span
+            )
+          rescue Exception => e # rubocop:disable Lint/RescueException
+            span&.record_exception(e)
+            span&.status = OpenTelemetry::Trace::Status.error("Unhandled exception of type: #{e.class}")
+            raise e
+          ensure
+            span&.finish
+            OpenTelemetry.tracer_provider.force_flush(timeout: @flush_timeout)
```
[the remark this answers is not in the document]
xuan-cao-swi: The lambda will close itself once the function is finished so when exporter receives the span the lambda environment maybe already closed. It's hard to control when lambda env close itself.

> they should adjust the exporter settings or use the simple span processor

They can choose whatever span processor and exporter they like in the runtime initialization through wrapper), but here it just ensure it gets flush in time.

instrumentation/aws_lambda/lib/opentelemetry/instrumentation/aws_lambda/handler.rb
```diff
@@ -0,0 +1,132 @@
...
+          span_attributes[OpenTelemetry::SemanticConventions::Resource::FAAS_ID]       = context.invoked_function_arn
+          span_attributes[OpenTelemetry::SemanticConventions::Trace::FAAS_EXECUTION]   = context.aws_request_id
+
+          OpenTelemetry::Context.with_current(parent_context) do
+            span = tracer.start_span(
+              original_handler,
+              attributes: span_attributes,
+              kind: :server # Span kind MUST be `:server` for a HTTP server span
+            )
+          rescue Exception => e # rubocop:disable Lint/RescueException
```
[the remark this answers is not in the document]
xuan-cao-swi: I think this has been updated

instrumentation/aws_lambda/README.md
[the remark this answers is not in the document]
kaylareopelle: @xuan-cao-swi  - Does this sound like the right process to have someone start using the OpenTelemetry instrumentation in their Ruby lambda?

```suggestion
## How do I get started? 

Installation of the `opentelemetry-instrumentation-aws_lambda` gem is handled by the [OpenTelemetry Lambda Layer for Ruby](https://github.com/open-telemetry/opentelemetry-lambda/tree/main/ruby). 

We do not advise installing the `opentelemetry-instrumentation-aws_lambda` gem directly into your Ruby lambda. Instead, clone the [OpenTelemetry Lambda Layer for Ruby](https://github.com/open-telemetry/opentelemetry-lambda/tree/main/ruby) and build the layer locally. Then, save it in your AWS account.
```

instrumentation/aws_lambda/README.md
```diff
@@ -0,0 +1,60 @@
...
+
+```
+gem install opentelemetry-instrumentation-aws_lambda
+```
+
+Or, if you use [bundler][bundler-home], include `opentelemetry-instrumentation-aws_lambda` in your `Gemfile`.
+
+## Usage
+
+From the Lambda Layer side, create the wrapper. More information can be found at https://github.com/open-telemetry/opentelemetry-lambda/tree/main/ruby
```
kaylareopelle, line 17:
```suggestion
From the Lambda Layer side, create the wrapper. More information can be found at https://github.com/open-telemetry/opentelemetry-lambda/tree/main/ruby

Below is an example of `ruby/src/layer/wrapper.rb`, where you can configure the layer to suit your needs before building it:
```

instrumentation/aws_lambda/README.md
```diff
@@ -0,0 +1,60 @@
...
+```ruby
```
kaylareopelle, line 18: I was kind of surprised when I took a look at the opentelemetry lambda's [ruby/layer/src](https://github.com/open-telemetry/opentelemetry-lambda/tree/main/ruby/src/layer) directory. I didn't see a `wrapper.rb` file. Is this something that users need to make? 
Is there more that needs to be done to the layer repo before both pieces are ready?

instrumentation/aws_lambda/lib/opentelemetry/instrumentation/aws_lambda/handler.rb
```diff
@@ -0,0 +1,198 @@
...
+module OpenTelemetry
+  module Instrumentation
+    module AwsLambda
+      AWS_TRIGGERS = ['aws:sqs', 'aws:s3', 'aws:sns', 'aws:dynamodb'].freeze
+
+      # Handler class that creates a span around the _HANDLER
+      class Handler
+        attr_reader :handler_method, :handler_class
+
+        # anytime when update the code in a Lambda function or change the functional configuration,
```
kaylareopelle, line 16:
```suggestion
        # anytime when the code in a Lambda function is updated or the functional configuration is changed,
```

instrumentation/aws_lambda/lib/opentelemetry/instrumentation/aws_lambda/handler.rb
```diff
@@ -0,0 +1,198 @@
...
+        # the next invocation results in a cold start; therefore these instance variable will be up-to-date
```
kaylareopelle, line 17:
```suggestion
        # the next invocation results in a cold start; therefore these instance variables will be up-to-date
```

kaylareopelle (commented):
This is looking great! I have a few, small documentation changes.

instrumentation/aws_lambda/README.md
xuan-cao-swi: Yes, I will work on creating the wrapper.rb in that directory once this pr is ready (or published).
Since wrapper.rb needs the lambda instrumentation to work, so I thought to wait for this PR.
For example, [python wrapper.rb](https://github.com/open-telemetry/opentelemetry-lambda/blob/main/python/src/otel/otel_sdk/otel_wrapper.py#L40) needs the instrumentation as well.

instrumentation/aws_lambda/README.md
xuan-cao-swi: Yes, I think it's great!

instrumentation/aws_lambda/README.md
kaylareopelle: Got it, that makes sense. Thank you!

kaylareopelle (approved):
Xuan, Amazing job with this. Thank you for your hard work! 🚀

cheempz (approved):
LGTM, thanks @xuan-cao-swi (and @kaylareopelle + @arielvalentin for the continued guidance!). Really looking forward to when we can get the metrics component into upstream for the full Lambda tracing experience.

kaylareopelle:
@xuan-cao-swi While reviewing #887, I noticed you made a few changes to `release-please`-related files. Do those changes need to be made for the lambda gem as well, or are releases different for this gem?

xuan-cao-swi:
> @xuan-cao-swi While reviewing https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/887, I noticed you made a few changes to release-please-related files. Do those changes need to be made for the lambda gem as well, or are releases different for this gem?

Yes, that would be great if can release together. I have updated the release files.

.github/workflows/ci-instrumentation-canary.yml
arielvalentin, line 1: This file was deleted on main. Please remove it from this branch.

kaylareopelle:
It looks like we're getting close, @xuan-cao-swi! Is there anything else you were hoping to accomplish before we sync with main and merge this in?

xuan-cao-swi:
> It looks like we're getting close, @xuan-cao-swi! Is there anything else you were hoping to accomplish before we sync with main and merge this in?

Hi @kaylareopelle , no, I think it should be ok. I will open new PR if anything come up (esp. when incorporating with lambda layer build on [lambda repo](https://github.com/open-telemetry/opentelemetry-lambda))

PR #727, opened 2023-11-22
chore(dev-deps): Add rubocop-performance

kaylareopelle:
This PR adds the [`rubocop-performance` gem](https://rubygems.org/search?query=rubocop-performance) to all contrib gems. It also updates the code to fulfill the linter's recommendations.


instrumentation/grape/lib/opentelemetry/instrumentation/grape/event_handler.rb
```diff
@@ -79,7 +79,7 @@ def handle_payload_exception(span, exception)
             # Only record exceptions if they were not raised (i.e. do not have a status code in Grape)
             # or do not have a 5xx status code. These exceptions are recorded by Rack.
             # See instrumentation/rack/lib/opentelemetry/instrumentation/rack/middlewares/tracer_middleware.rb#L155
-            return unless exception.respond_to?('status') && ::Rack::Utils.status_code(exception.status) < 500
+            return unless exception.respond_to?(:status) && ::Rack::Utils.status_code(exception.status) < 500
```
arielvalentin, line 82: Odd. this should not make a difference if we are using frozen string literals right?

arielvalentin (approved):

instrumentation/grape/lib/opentelemetry/instrumentation/grape/event_handler.rb
kaylareopelle: Good question! I'll do a bit of digging.

kaylareopelle: So, I don't understand why, yet, but I can confirm with `Benchmark.ips` and `Benchmark.ipsa` that it is faster even when using frozen string literals. [Here's a gist](https://gist.github.com/kaylareopelle/e8e1f04cb1c9fb399b38e27f82d390da) with the code I ran using Ruby 3.2.2 using ipsa. 

I ran it a few times and strings were always between 1.5 - 1.8 times slower.

Here's the original PR for that cop: https://github.com/rubocop/rubocop-performance/pull/276 

It sounds like it's from Polished Ruby Programming, which I have at home, so I'll see if I can learn more there.

arielvalentin: Amazing. String Literals always return the same object id whereas symbols do not. 

This surprises me so much!!!

Thank goodness for these linters.

PR #773, opened 2023-12-22
chore: Add Tests for GraphQL C Parser

arielvalentin:
Graphql 2.2 introduces an optional C implementation of the Parser as well as optimizations to the pure Ruby parser that result in changes to what is captured during instrumentation.

Users of the pure Ruby implementation will no longer see `graphql.lex` spans emitted as a part of their traces and will instead need to switch to using the C Parser in order to explicitly see time spent in the lexer. 

This PR introduces updates to the test suite to add coverage for the C Parser and changes to the Ruby implementation. 

There are no underlying changes in the instrumentation itself.

### DIFF
--- a/instrumentation/graphql/Appraisals
+++ b/instrumentation/graphql/Appraisals
@@ -24,6 +24,10 @@ appraise 'graphql-2.1' do
   gem 'graphql', '~> 2.1.8'
 end
 
+appraise 'graphql-2.2.0' do
+  gem 'graphql', '2.2.0'
+end
+
 appraise 'graphql-2.x' do
-  gem 'graphql', '>= 2.2', '< 3.0.0'
+  gem 'graphql', '>= 2.2.1', '< 3.0.0'
 end
--- a/instrumentation/graphql/Appraisals
+++ b/instrumentation/graphql/Appraisals
@@ -32,3 +32,11 @@ end
 appraise 'graphql-2.2.x' do
   gem 'graphql', '~> 2.2.1', '< 3.0.0'
 end
+
+appraise 'graphql-c_parser-latest' do
+  gem 'graphql-c_parser'
+end
+
+appraise 'graphql-latest' do
+  gem 'graphql'
+end
--- a/instrumentation/graphql/Appraisals
+++ b/instrumentation/graphql/Appraisals
@@ -4,18 +4,26 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 
-appraise 'graphql-1.13' do
+# Max compatible version of 1.x
+appraise 'graphql-1.x' do
   gem 'graphql', '~> 1.13'
 end
 
-appraise 'graphql-2.0.17' do
-  gem 'graphql', '2.0.17'
-end
-
+# A bug was introduced in 2.0.18 that was fixed in 2.0.19
 appraise 'graphql-2.0.18' do
   gem 'graphql', '2.0.18'
 end
 
+# Max compatible version of 2.0.x
+appraise 'graphql-2.0' do
+  gem 'graphql', '~> 2.0.27'
+end
+
+# Max compatible version of 2.1.x
+appraise 'graphql-2.1' do
+  gem 'graphql', '~> 2.1.8'
+end
+
 appraise 'graphql-2.x' do
-  gem 'graphql', '>= 2.0.18', '< 3.0.0'
+  gem 'graphql', '>= 2.2', '< 3.0.0'
 end
--- a/instrumentation/graphql/Appraisals
+++ b/instrumentation/graphql/Appraisals
@@ -24,10 +24,11 @@ appraise 'graphql-2.1' do
   gem 'graphql', '~> 2.1.8'
 end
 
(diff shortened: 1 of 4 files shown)


ravangen:
If you do want tests to run on both the default parser, and the [c parser](https://graphql-ruby.org/language_tools/c_parser.html), I do something similar with:

```ruby
USING_C_PARSER = T.let(false, T::Boolean)
if ENV["GRAPHQL_CPARSER"] == "1"
  require "graphql/c_parser"
  USING_C_PARSER = true
end

# When tracing, is the parser expected to call `lex` before `parse`
def trace_lex_supported?
  return @trace_lex_supported if defined?(@trace_lex_supported)

  # In GraphQL 2.2, the default parser was changed such that `lex` is no longer called
  @trace_lex_supported = USING_C_PARSER ||
    Gem::Requirement.new("< 2.2").satisfied_by?(Gem::Version.new(GraphQL::VERSION))
end
```

Where I have CI run the appraisals twice (`GRAPHQL_CPARSER=0` and `GRAPHQL_CPARSER=1`) since I wasn't sure how to put that into the appraisal itself.

arielvalentin:
Thanks for the pointers.

arielvalentin:
@ravangen @rmosolgo @swalkinshaw I added separate appraisals for the C Parser and I think I found the incantation necessary to not have to rely on setting environment variables for our test suite.

Please take a look and let me know if these changes:
 
1. Are sensible
2. Are maintainable
3. Provide the appropriate coverage for the functionality that includes the C Parser

Thank you.

instrumentation/graphql/Appraisals
```diff
@@ -4,18 +4,31 @@
...
+  gem 'graphql', '~> 2.1.8'
+end
+
+appraise 'graphql-c_parser-2.2.x' do
+  gem 'graphql', '~> 2.2.1'
+  gem 'graphql-c_parser', '~> 1.0.7'
+end
+
+appraise 'graphql-2.2.x' do
+  gem 'graphql', '~> 2.2.1', '< 3.0.0'
```
ravangen, line 33: Did you want to have one run against latest?

Perhaps could match the styles of other appraisals to the same effect.

```suggestion
# Max compatible version of 2.2.x
appraise 'graphql-2.2' do
  gem 'graphql', '~> 2.2.1'
```

ravangen (approved):

instrumentation/graphql/Appraisals
arielvalentin: Major version changes have been difficult for us to test using appraisals, while maintaining minor versions. 

We used to have canary builds that rain on unpinned versions of gems but I couldn't keep up with them so I disabled those builds.

I will however add unpinned versions of the gem to this PR and see if that puts us in a better place for the future canary work.

instrumentation/graphql/test/test_helper.rb
```diff
@@ -125,3 +125,12 @@ def uses_platform_interfaces?
...
   Gem::Version.new(GraphQL::VERSION)
 end
+
+# When tracing, is the parser expected to call `lex` before `parse`
+def trace_lex_supported?
+  return @trace_lex_supported if defined?(@trace_lex_supported)
+
+  # In GraphQL 2.2, the default parser was changed such that `lex` is no longer called
+  @trace_lex_supported = Gem::Requirement.new('< 2.2').satisfied_by?(Gem::Version.new(GraphQL::VERSION)) \
+    || (defined?(GraphQL::CParser) == 'constant')
```
ravangen, line 135: Does anything result in `require "graphql/c_parser"` such that this is `true` when the gem is installed?

arielvalentin: The test helper uses bundler require, which will load all of the declared gems by default before running any tests.

In this case the Gemfile requires the graphql-c_parser file in the gems' lib directory, which requires `graphql/c_parser`... then turtles all the way down.

arielvalentin: https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/main/instrumentation/graphql/test/test_helper.rb#L8

instrumentation/graphql/test/instrumentation/graphql/tracers/graphql_trace_test.rb
```diff
@@ -103,7 +105,7 @@
         it 'traces the provided schemas' do
           SomeOtherGraphQLAppSchema.execute('query SimpleQuery{ __typename }')
 
-          _(spans.size).must_equal(8)
+          _(spans.select { |s| s.name.start_with?('graphql.') }).wont_be(:empty?)
```
kaylareopelle, line 108: I like looking for the name rather than the quantity. Nice update.

kaylareopelle (approved):
Wow! Nice work. I learned a lot from this PR.

instrumentation/graphql/Appraisals
arielvalentin: Added here: https://github.com/github/github/discussions/302107

PR #792, opened 2024-01-04
Trilogy: only set `db.instance.id` attribute if there is a value

bensheldon:
It's possible for the `@connected_host` value to be ~unset~ `nil` because a query is made within trilogy to initially assign the `@connected_host`:

https://github.com/trilogy-libraries/trilogy/blob/3c3a3865a3896f3056b1d7c69c5244964b60937c/contrib/ruby/lib/trilogy.rb#L39-L50

This PR adds a presence check before assigning the attribute.


arielvalentin (approved):

bensheldon:
@arielvalentin I made one more tweak which was to only do a nil check. The ivar will always be defined because it's declared in `#initialize`: https://github.com/trilogy-libraries/trilogy/blob/3c3a3865a3896f3056b1d7c69c5244964b60937c/contrib/ruby/lib/trilogy.rb#L16

jhawthorn (approved):

PR #800, opened 2024-01-07
ci: Add Conventional Commit Linter

arielvalentin:
I have made too many mistakes approving and merging PRs that lack conventional commits.

This Action should mitigate those mistakes going forward.

## Failure 

image

## Success

image


.github/workflows/conventional-commits.yaml
```diff
@@ -0,0 +1,38 @@
...
+        with:
+          update-labels: false  # OPTIONAL; do not update the Pull Request labels based on the Conventional Commits information.
+          include-commits: true
+          types: |
+            chore
+            ci
+            docs
+            refactor
+            release
+            test
```
scbjans, line 38: `fix`, `feat` and `perf` are missing (of which the first two should be included by default; but it might be nice to be explicit).
Since `include-commits: true`, I expect `squash` should also be included here.

Lastly, I believe a [token](https://github.com/dev-build-deploy/commit-me/blob/main/docs/github-action.md#inputs) is needed to validate the commits associated to the PR

arielvalentin: Makes sense. I did was not certain that it was necessary to make a distinction between `perf` and `fix` and I am on the fence about whether `refactor` and `test` are useful to make distinctions from CI or `fix`. 

Do we gain any additional insights supporting the additional types?

> Since include-commits: true, I expect squash should also be included here.

Fair. I use `squash` all the time and probably do not need it to fail the linter check; though I do want it to be more strict when merging to main

> Lastly, I believe a [token](https://github.com/dev-build-deploy/commit-me/blob/main/docs/github-action.md#inputs) is needed to validate the commits associated to the PR

For some reason I thought it used a default token! Thanks for the pointer.

scbjans: > Do we gain any additional insights supporting the additional types?

I'm not sure; but I do think that since these are conventional, that we should support them.
Especially since [we encourage people to use them](https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/main/CONTRIBUTING.md#use-conventional-commit-messages), CI shouldn't fail on those.

kaylareopelle: > Especially since [we encourage people to use them](https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/main/CONTRIBUTING.md#use-conventional-commit-messages), CI shouldn't fail on those.

Great catch. I'm comfortable limiting the list to just those included in our CONTRIBUTING documentation. However, I could see `ci` and `release` being very helpful for maintainers, but not for contributors.

scbjans (approved):

kaylareopelle (approved):

.github/workflows/conventional-commits.yaml
```diff
@@ -0,0 +1,44 @@
...
+          update-labels: false  # OPTIONAL; do not update the Pull Request labels based on the Conventional Commits information.
+          include-commits: true
+          token: ${{ secrets.GITHUB_TOKEN }}
+          types: |
+            chore
+            ci
+            docs
+            feat
+            fix
+            merge
```
arielvalentin, line 39:
```suggestion
            Merge
```

.github/workflows/conventional-commits.yaml
```diff
@@ -0,0 +1,44 @@
...
+      - main
+  pull_request:
+    types:
+      - opened
+      - synchronize
+      - reopened
+      - edited
+
+permissions:
+  contents: read
```
arielvalentin, line 16:
```suggestion
  contents: write
```

.github/workflows/conventional-commits.yaml
```diff
@@ -0,0 +1,44 @@
...
+          update-labels: false  # OPTIONAL; do not update the Pull Request labels based on the Conventional Commits information.
+          include-commits: true
+          token: ${{ secrets.GITHUB_TOKEN }}
+          types: |
+            chore
+            ci
+            docs
+            feat
+            fix
+            Merge
```
arielvalentin, line 39:
```suggestion
```

arielvalentin:
@kaylareopelle @scbjans Seems like default merge commit messages are not allowed because only Squash Commits are allowed. 

That will mean we have to rebase whenever we sync branches. Do you all think that is OK?

scbjans:
> @kaylareopelle @scbjans Seems like default merge commit messages are not allowed because only Squash Commits are allowed.
> 
> It also means revert commits will need to be formatted differently.
> 
> That will mean we have to rebase whenever we sync branches. Do you all think that is OK?

@arielvalentin Apparently [Merge commits are allowed](https://github.com/dev-build-deploy/commit-me/blob/ef114af3d79e2f0a69c64a3b4b7245476bbc0709/src/repository.ts#L28), but with our current configuration we cannot set `include-commits: true` (because of the merge commits with a non-conventional commit message).

Would it be possible to [update the repository](https://docs.github.com/en/rest/repos/repos?apiVersion=2022-11-28#update-a-repository) and configure the `merge_commit_message` (sorry, can't deeplink) to `BLANK`?

IIUC a rebasing merge strategy would mean (potential) contributors need to reset their local branch upon each branch sync, that does not seem desirable to me.

arielvalentin:
I believe the docs you're referencing are for the defaults values used when merging a PR to a target branch but I'm not aware of how to set the default merge commit message when syncing a fork or branch. 

I feel like we have limited options here. I am a fan of rebasing but also am sympathetic to how tricky it can get when dealing with conflicts. 

I want to scan commits because it's what the release tooling uses at the moment but we also squash and merge to main so maybe that's enough? 

I'll disable commit scanning and see how this works purely with PR title scans.

.github/workflows/conventional-commits.yaml
```diff
@@ -0,0 +1,43 @@
...
+
+jobs:
+  validate-commits:
+    name: Conventional Commits Validation
+    runs-on: ubuntu-latest
+    steps:
+      - uses: dev-build-deploy/commit-me@v1.2.0
+        with:
+          update-labels: false  # OPTIONAL; do not update the Pull Request labels based on the Conventional Commits information.
+          include-commits: true
```
arielvalentin, line 31:
```suggestion
          include-commits: false
```

.github/workflows/conventional-commits.yaml
```diff
@@ -0,0 +1,43 @@
+name: Conventional Commits Validation
+
+on:
+  workflow_dispatch:
+  push:
+    branches:
+      - main
```
arielvalentin, line 7:
```suggestion
```

.github/workflows/conventional-commits.yaml
```diff
@@ -0,0 +1,41 @@
+name: Conventional Commits Validation
+
+on:
+  workflow_dispatch:
+  push:
```
arielvalentin, line 5:
```suggestion
```

.github/workflows/conventional-commits.yaml
```diff
@@ -0,0 +1,40 @@
...
+          types: |
+            chore
+            ci
+            docs
+            feat
+            fix
+            perf
+            refactor
+            release
+            squash
```
mwear, line 39: Can we remove `squash`?
```suggestion
```

arielvalentin:
@scbjans et.al. after removing the commit checks this does not properly validate the title of the PR. Not sure what I did wrong.

arielvalentin:
This should have failed 

image

.github/workflows/conventional-commits.yaml
```diff
@@ -0,0 +1,39 @@
...
+  pull_request:
+    types:
+      - opened
+      - synchronize
+      - reopened
+      - edited
+
+permissions:
+  contents: write
+  pull-requests: read
```
arielvalentin, line 14:
```suggestion
  pull-requests: write
```

.github/workflows/conventional-commits.yaml
```diff
@@ -0,0 +1,39 @@
...
+
+jobs:
+  validate-commits:
+    name: Conventional Commits Validation
+    runs-on: ubuntu-latest
+    steps:
+      - uses: dev-build-deploy/commit-me@v1.2.0
+        with:
+          update-labels: false  # OPTIONAL; do not update the Pull Request labels based on the Conventional Commits information.
+          include-commits: false
```
arielvalentin, line 28:
```suggestion
          include-commits: false
          include-pull-requests: true
```
Will this work?

arielvalentin:
Ok looks like I got it to fail if I use the config file. It seems to default checking pull request titles unless set via a configuration config file:

image

https://github.com/search?q=repo%3Adev-build-deploy%2Fcommit-me+includePullRequest&type=code

arielvalentin:
well it seems it breaks now for unrelated reasons I cannot seem to understand 

>   Error: Cannot read properties of undefined (reading 'breaking')

arielvalentin:
@Kevin-de-Jong any input on what may be incorrect that is causing our actions to fail?

https://github.com/open-telemetry/opentelemetry-ruby-contrib/actions/runs/7547155406/job/20546480272?pr=800

Kevin-de-Jong:
> Seems like default merge commit messages are not allowed because only Squash Commits are allowed.

I have created https://github.com/dev-build-deploy/commit-it/issues/33 (which will be followed up in `commit-me`) to better manage both `fixup` and `merge` commits.

> Ok looks like I got it to fail if I use the config file. It seems to default checking pull request titles unless set via a configuration config file:
> ...
> well it seems it breaks now for unrelated reasons I cannot seem to understand

I am looking into this :eyes:

.github/workflows/conventional-commits.yaml
Kevin-de-Jong: By default, the Convention Commits specification does not impose any restrictions on the `type` - it only requires `feat` and `fix` as these have a special purpose (in relation to SemVer).

Therefore, `CommitMe` will not impose any restrictions on the `type` unless you provide an additional list of `type`s to restrict to.

Typically, projects are using the Angular definition of types;
- `build`: Changes that affect the build system or external dependencies (example scopes: gulp, broccoli, npm)
- `ci`: Changes to our CI configuration files and scripts (examples: CircleCi, SauceLabs)
- `docs`: Documentation only changes
- `feat`: A new feature
- `fix`: A bug fix
- `perf`: A code change that improves performance
- `refactor`: A code change that neither fixes a bug nor adds a feature
- `test`: Adding missing tests or correcting existing tests

(taken from: https://github.com/angular/angular/blob/68a6a07/CONTRIBUTING.md#-commit-message-format)


> [!NOTE]
> The same principle true for the `scopes` configuration parameter (no restrictions, unless configured)

.github/workflows/conventional-commits.yaml
```diff
@@ -0,0 +1,30 @@
...
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number }}  # Ensure that only one instance of this workflow is running per Pull Request
+  cancel-in-progress: true  # Cancel any previous runs of this workflow
+
+jobs:
+  validate-commits:
+    name: Conventional Commits Validation
+    runs-on: ubuntu-latest
+    steps:
+      - uses: dev-build-deploy/commit-me@v1.3.1
```
arielvalentin, line 25:
```suggestion
      - uses: dev-build-deploy/commit-me@v1.3.1
         env:
           FORCE_COLOR: 3
```

.github/workflows/lint-conventional-commits.yaml
```diff
@@ -0,0 +1,39 @@
...
+          types: |
+            fix
+            feat
+            chore
+            ci
+            docs
+            style
+            refactor
+            perf
+            test
```
kaylareopelle, line 36:
```suggestion
            chore
            ci
            docs
            feat
            fix
            perf
            refactor
            style
            test
```

kaylareopelle (changes requested):
Are the lists of commit types in the following files supposed to be the same? 
* `.commit-me.json`
* `.github/workflows/lint-conventional-commits.yaml`
* `commitlint.config.js`

I see some slight differences. `release`, `refactor`, `revert`, `style`, and `squash` seem to be the words that are included in some of the lists, but not all.

arielvalentin:
Those difference exist because one action only validates PR titles but the other the commit messages. We don't want PR titles to allow squash since the default value of the merge commit is the PR title.

.commit-me.json
```diff
@@ -0,0 +1,4 @@
+{
+  "include-pull-requests": true,
+  "types": [ "chore", "ci", "docs", "feat", "fix", "perf", "refactor", "release", "squash", "test" ]
```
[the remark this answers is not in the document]
arielvalentin: > style and revert are both listed in https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/821, but aren't present here. Do we want to include them?

👍🏼 

> Does this get around merge commit validations, or do we need to add those too?

Yes! Where it did not support it in my initial tests, it does now. https://github.com/open-telemetry/opentelemetry-ruby-contrib/actions/runs/7614598575/job/20737243103

> Last thing, test is in this list and [it's also in the Conventional Commits Angular documentation](https://github.com/angular/angular/blob/22b96b9/CONTRIBUTING.md#type), but our [CONTRIBUTING.md file](https://github.com/angular/angular/blob/22b96b9/CONTRIBUTING.md#type) currently lists tests, plural.

👍🏼

kaylareopelle (approved):
🎉

scbjans (approved):

scbjans:
Been thinking about this a little longer. Functionally we only need PR title validation for the release tooling?
If so, maybe we should not include commits at this stage since it might break contributor workflows?
We can enable that setting later on if we want?

arielvalentin:
> Been thinking about this a little longer. Functionally we only need PR title validation for the release tooling? If so, maybe we should not include commits at this stage since it might break contributor workflows? We can enable that setting later on if we want?

I brought this up during the SIG. I do not think folks had a strong opinion either way, however during the SIG @mwear also shared a preference to using client-side git-hooks to validate individual commit messages. This would reduce friction for contributors, who would get an error message earlier in the process and avoid generating non-conventional commits. 

It would not enforce commits coming from say the PR UI suggestions but as you stated, perhaps all we need is PR title validation. That would at least mitigate issues with maintainers (mostly me) end up merging bad squashed commits.

.github/workflows/conventional-commits.yaml
```diff
@@ -0,0 +1,32 @@
...
+        env:
+          FORCE_COLOR: 3
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+          include-commits: true
```
arielvalentin, line 30:
```suggestion
          include-commits: false
```

Per @scbjans comments, it may be good to disable commit validation to avoid friction for any open PRs.

In addition to that, I would like to follow up on @mwear suggestion to install client side git-hooks before enabling commit validation in the action.

scbjans (approved):
LGTM

> In addition to that, I would like to follow up on @mwear suggestion to install client side git-hooks before enabling commit validation in the action.

Let's implement that in a separate PR.

PR #818, opened 2024-01-16
chore: Update rubocop version to 1.60.0

kaylareopelle:
Consolidate the dependabot changes to update rubocop to 1.160.0 into one PR.

This version introduces Style/ArgumentsForwarding, which attempts to replace calls to &block, *args, etc. with ...

### DIFF
--- a/instrumentation/bunny/lib/opentelemetry/instrumentation/bunny/patch_helpers.rb
+++ b/instrumentation/bunny/lib/opentelemetry/instrumentation/bunny/patch_helpers.rb
@@ -12,19 +12,19 @@ module OpenTelemetry
       # For additional details around trace messaging semantics
       # See https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/semantic_conventions/messaging.md#messaging-attributes
       module PatchHelpers
-        def self.with_send_span(channel, tracer, exchange, routing_key, &block)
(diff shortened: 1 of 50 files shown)


instrumentation/bunny/lib/opentelemetry/instrumentation/bunny/patch_helpers.rb
```diff
@@ -12,19 +12,19 @@ module Bunny
       # For additional details around trace messaging semantics
       # See https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/semantic_conventions/messaging.md#messaging-attributes
       module PatchHelpers
-        def self.with_send_span(channel, tracer, exchange, routing_key, &block)
+        def self.with_send_span(channel, tracer, exchange, routing_key, &block) # rubocop:disable Style/ArgumentsForwarding
           attributes = basic_attributes(channel, channel.connection, exchange, routing_key)
           destination = destination_name(exchange, routing_key)
 
-          tracer.in_span("#{destination} publish", attributes: attributes, kind: :producer, &block)
+          tracer.in_span("#{destination} publish", attributes: attributes, kind: :producer, &block) # rubocop:disable Style/ArgumentsForwarding
```
kaylareopelle, line 19: Since the `tracer.in_span` call has arguments that aren't forwarded, we can't replace `&block` with `...` in the method definition args nor the `tracer.in_span` args. Same is true for example below.

arielvalentin (approved):

PR #823, opened 2024-01-17
chore: Disable redundant linter

kaylareopelle:
The disable directives for Style/ArgumentsForwarding on L15 and L19 are removed, the offense is re-raised locally. When they are present, the CI notes an offense of the Lint/RedundantCopDisableDirective.

To keep both environments happy, add both the style and lint cops to the disable directive.


arielvalentin:
Strange. Is there something conflicting in the rubocop versions or the way we run it in CI vs local?

kaylareopelle:
@arielvalentin - It looks like there was a bug with `Style/ArgumentsForwarding` in rubocop 1.60.0. They've released a fix in 1.60.1. I'll change the purpose of this PR to update to the fix version.

arielvalentin (approved):

PR #824, opened 2024-01-18
docs: ✏️ Sidekiq instrumentation options

kohbis:
For users of Sidekiq instrumentation, I thought it would be nice to have a list of options and their uses.
So I added a (simple explain and format) table in the README ✏️ 
（At least no code reading required）

I improve any further explanations, items, or errors that are necessary.

If you don't need it, you can close it 🙏

PR #834, opened 2024-01-24
docs: relocate Resque config option comments to render in Yard docs

robbkidd:
As formatted markdown comments on the class, the information is made visible in the Yard-generated docs without making the options appear to be class or instance variables. Possibly instrumentation READMEs could link in [the Usage section](https://github.com/open-telemetry/opentelemetry-ruby-contrib/tree/f5f21c686aa8b463a2fb6edd39260cf2a7da7043/instrumentation/resque#usage) to the instrumentation's latest version at rubydoc.info (for example, [this here resque](https://rubydoc.info/gems/opentelemetry-instrumentation-resque/OpenTelemetry/Instrumentation/Resque)) to avoid repeating the details of config options in the README but still making the options discoverable from there.

Placing all this on the Instrumentation class is in contrast to the approach used in #824 where each option has a distinct comment.

### Advantages to class-level doc comment:

1. The option names do not [appear to be class or instance methods](https://rubydoc.info/gems/opentelemetry-instrumentation-resque/OpenTelemetry/Instrumentation/Resque/Instrumentation) on `Instrumentation` within the documentation.
2. The documentation can be included and rendered higher up the module namespace (on `OTel::Inst::Resque`) with [a single line](https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/834/files#diff-6400ed026f380d2c28fbe78a9acd4a8de98145857314a47c768e2c7e48e2ca66R12) while the doc content still lives close(ish) to the option declarations.
3. Simpler than holding Yard macros the right way.

### Disadvantages to class-level doc comment:

1. The documentation is a little removed from the option declarations.

## Screenshots!

Screenshot 2024-01-24 at 3 02 07 PM

... and scroll down a bit ...

Screenshot 2024-01-24 at 3 04 43 PM

### DIFF
--- a/instrumentation/resque/lib/opentelemetry/instrumentation/resque.rb
+++ b/instrumentation/resque/lib/opentelemetry/instrumentation/resque.rb
@@ -9,7 +9,7 @@ require 'opentelemetry-instrumentation-base'
 
 module OpenTelemetry
   module Instrumentation
-    # Contains the OpenTelemetry instrumentation for the Resque gem
+    # (see OpenTelemetry::Instrumentation::Resque::Instrumentation)
     module Resque
     end
   end
--- a/instrumentation/resque/lib/opentelemetry/instrumentation/resque/instrumentation.rb
+++ b/instrumentation/resque/lib/opentelemetry/instrumentation/resque/instrumentation.rb
@@ -7,7 +7,61 @@
 module OpenTelemetry
   module Instrumentation
     module Resque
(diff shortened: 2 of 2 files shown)


arielvalentin (approved):
Since this is going to help your customers I think that getting this merged will be ok with me, however per our slack conversations I think longer term what we want is a way to document individual options and then somehow render them in the README.

PR #837, opened 2024-01-29
feat: support context propagation through links

gwuah:
This pull request extends the rack instrumentation library to support context propagation through links. 
Several libraries have this feature, for example [que](https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/56ff830daa6b018913c717abb971dc4669a62be4/instrumentation/que/lib/opentelemetry/instrumentation/que/instrumentation.rb#L32) and [otel-http ](https://pkg.go.dev/go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp#WithPublicEndpoint) lib.

Our use case is simple, we want to connect our frontend & backend traces using links, as opposed to having the frontend be the source of the root span, as that's messing with our sampling. We have had our fork running in production for a while now, just thought to upstream it. 

I'm basically brute-forcing my way through ruby, so any sort of style suggestion is welcome.


xuan-cao-swi:
I think it should be ok as it is an option and the rationale behind it seems logical to me. 

However, I do have a few formatting suggestions, which you may consider or ignore as you see fit. 

```
if propagate_with_link&.call(request.env)
  links = prepare_span_links(parent_context)
  span = create_root_span(request, links)
else
  span = create_span(parent_context, request)
end
```

Additionally, I'm unsure whether option :propagate_with_link should be a callable or rather as an array of links (i.e. array of link is more restricted on options in my opinion).

arielvalentin (commented):
Although we already support it with URL Quantization, I am weary of adding additional extension points using code block options. 

If we moved forward with this feature, I think I would prefer if this code were implemented with additional safety measures and potentially have some fallback behavior to use the parent context. 

IIRC there is supposed to be something related to the Trace State that should include information about whether or not to _continue_ a trace or start a new one, however I am not familiar with supporting Span Links for Server Span Kinds.

Is there something in the specification that states that server span kinds should support links?

kaylareopelle:
@gwuah, what do you think about [Ariel's comment?](https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/837#pullrequestreview-1869282718)

dmathieu:
Related suggestion that would have offered the same feature, and has a similar implementation in other SDKs (Go and JS, maybe others):

https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/130

kaylareopelle:
Thanks, @arielvalentin for bringing this to the maintainers slack channel. What was the conclusion/next steps drawn from your conversation?

arielvalentin:
> Thanks, @arielvalentin for bringing this to the maintainers slack channel. What was the conclusion/next steps drawn from your conversation?

I got very limited engagement on the topic with links to existing issues, which to be honest I have not read yet: 

https://github.com/open-telemetry/opentelemetry-specification/pull/3877
https://github.com/open-telemetry/opentelemetry-specification/issues/3867

Something that is required by Golang library instrumentations is that users have to register specific endpoints to be traced or not and have more control over what handlers should ignore in the incoming trace headers and start a new trace.

That allows them more granular control of when to consider creating a _new_ trace id as opposed to what something like Rack does where currently have a single Event Handler injected for all requests. 

A lot of routing is also handled by the frameworks themselves and tend not to rely on Rack figuring out what needs to process a request.

zacheryph:
Other instrumentation already use `propagation_style`, is there any reason this could not be done vs a callable proc?

gwuah:
@zacheryph 

>it's a callable because we want to dynamically enable/disable this feature, based on stuff in headers, etc.

PR #853, opened 2024-02-07
fix: return nil for non-existant key in AwsSdk::MessageAttributeGetter

yoheyk:
Since `tracestate` is [optional](https://github.com/open-telemetry/opentelemetry-ruby/blob/opentelemetry-sdk/v1.4.0/api/lib/opentelemetry/trace/propagation/trace_context/text_map_propagator.rb#L31), it can be omitted.

However, current implementation of `AwsSdk::MessageAttributeGetter` raises an exception [here](https://github.com/open-telemetry/opentelemetry-ruby/blob/opentelemetry-sdk/v1.4.0/api/lib/opentelemetry/trace/propagation/trace_context/text_map_propagator.rb#L52) when it is omitted.

Same as [TextMapGetter](https://github.com/open-telemetry/opentelemetry-ruby/blob/opentelemetry-sdk/v1.4.0/api/test/opentelemetry/context/propagation/text_map_getter_test.rb#L23-L25), it should return nil for non-existant key.


instrumentation/aws_sdk/lib/opentelemetry/instrumentation/aws_sdk/message_attributes.rb
```diff
@@ -29,7 +29,8 @@ def self.set(carrier, key, value)
       #   OpenTelemetry.propagation.extract(message, getter: MessageAttributeGetter)
       class MessageAttributeGetter
         def self.get(carrier, key)
-          carrier[key][:string_value] if carrier[key][:data_type] == 'String'
+          message_attribute = carrier[key]
```
xuan-cao-swi, line 32: Looks good to me, just a little nit.
Would `return unless carrier[key]` introduce less code change?

yoheyk: it is less code change but since it is doing `carrier[key]` multiple times, it seems to be a little slower:

```ruby
def get_with_variable(carrier, key)
  message_attribute = carrier[key]
  message_attribute[:string_value] if message_attribute && message_attribute[:data_type] == 'String'
end

def get_with_return(carrier, key)
  return unless carrier[key]

  carrier[key][:string_value] if carrier[key][:data_type] == 'String'
end

carrier = { 'traceparent' => { data_type: 'String', string_value: 'tp' } }

Benchmark.ips do |x|
  x.report("get_with_variable") { get_with_variable(carrier, 'traceparent') }
  x.report("get_with_return") { get_with_return(carrier, 'traceparent') }

  x.compare!
end
```
```
   get_with_variable      4.959M (± 7.8%) i/s -     24.754M in   5.053304s
     get_with_return      4.337M (± 2.0%) i/s -     21.870M in   5.045064s

Comparison:
   get_with_variable:  4959117.6 i/s
     get_with_return:  4336852.7 i/s - 1.14x  slower
```

but either way is fine. let me know if I should change the code

arielvalentin:
@yoheyk Thank you for your contribution. 

We are currently looking for help maintaining AWS related instrumentation. 

Would you be interested?

arielvalentin (approved):

instrumentation/aws_sdk/lib/opentelemetry/instrumentation/aws_sdk/message_attributes.rb
xuan-cao-swi: Thanks for putting up the comparison. I am ok with better performance.

xuan-cao-swi (approved):

yoheyk:
Thanks for merging and releasing so quickly!

> We are currently looking for help maintaining AWS related instrumentation.
> 
> Would you be interested?

@arielvalentin yeah, I think I can help to some extent.

Issue #670, opened 2023-09-21
mysql2: database client span(s) not recorded when using prepared statement with mysql2


xuan-cao-swi:
**Description of the bug**

When prepared statements are enabled, the SQL execution sequence follows a `prepare` and `execute` pattern. The existing mysql2 instrumentation does not include the prepare function from mysql2::client, so it fails to record the span during the statement's preparation phase.

I was checking the rails code on activerecord's [mysql2](https://github.com/rails/rails/blob/main/activerecord/lib/active_record/connection_adapters/mysql2/database_statements.rb#L21-L39), and found the function split the function call based on whether the statement is prepared or not prepared.

```ruby
def internal_exec_query(sql, name = "SQL", binds = [], prepare: false, async: false) # :nodoc:
  if without_prepared_statement?(binds)
    execute_and_free(sql, name, async: async) do |result|
      ...
    end
  else
    exec_stmt_and_free(sql, name, binds, cache_stmt: prepare, async: async) do |_, result|
      ...
    end
  end
end
```
This function `exec_stmt_and_free` will call two function: `conn.prepare(sql)` and `stmt.execute(*type_casted_binds)`

However, there is no preprend the function `prepare` in opentelemetry::instrumentation::mysql2 (this `prepare` function is directly from mysql2 C code here https://github.com/brianmario/mysql2/blob/master/ext/mysql2/client.c#L1554)

To prove my guess, I prepended `prepare` in opentelemetry::instrumentation::mysql2 patches/client, and got the statement in span
```ruby
def prepare(sql)
  ...
  tracer.in_span(
    database_span_name(sql),
    attributes: attributes.merge!(OpenTelemetry::Instrumentation::Mysql2.attributes),
    kind: :client
  ) do
    super(sql)
  end
end
```
If user wants full statement (i.e. without any obfuscation), then we need to preprend not only the mysql2:client, but also the mysql2:statement since the `stmt.execute(*type_casted_binds)` is from statement, which opentelemetry::instrumentation::mysql2 is lacking right now.

Or even better, just prepend mysql2:statement's `execute` (haven't test it), which will record the full sql

**Share details about your runtime**

Operating system details: Linux, Ubuntu 22.04 LTS
RUBY_ENGINE: "ruby"
RUBY_VERSION: "3.1.1"
RUBY_DESCRIPTION: "ruby 3.1.1p18 (2022-02-18 revision 53f5fc4236) [arm64-darwin21]"

[...]

PR #862, opened 2024-02-09
feat: instrument mysql2 prepare statement

xuan-cao-swi:
### Description

This PR aims to enhance mysql2 instrumentation by capturing prepare statement execution, akin to the [pg instrumentation](https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/opentelemetry-instrumentation-pg/v0.27.1/instrumentation/pg/lib/opentelemetry/instrumentation/pg/patches/connection.rb#L30-L37).

In contrast to the pg instrumentation, this implementation for mysql2 focuses solely on instrumenting prepare statements without extending the instrumentation to [execute statements](https://github.com/brianmario/mysql2/blob/0.5.5/lib/mysql2/statement.rb#L3). The rationale behind this approach is that the mysql2 client passes only the binding values to the execute function, offering minimal informative value, which is why the execute function is not instrumented. Additionally, another reason for this distinction is the way active_record executes these [two types of function executions](https://github.com/rails/rails/blob/v7.1.3/activerecord/lib/active_record/connection_adapters/mysql2/database_statements.rb#L123-L128) separately.

### Reproduce
database.yml
```
default: &default
  adapter: mysql2
  pool: <%= ENV.fetch("RAILS_MAX_THREADS") { 5 } %>
  timeout: 5000

development:
  <<: *default
  host: sample-mysql
  port: 3306
  database: classicmodels
  username: root
  password: admin
  prepared_statements: true
```  
Execute the query:
```ruby
Customer.where("contactLastName like ?", 'Schmitt')
# OR
stmt = "SELECT contactLastName from customers where contactLastName like ?"
binds = [[nil, 'Schmitt']]
result = ActiveRecord::Base.connection.exec_query stmt, 'SQL', binds, prepare: true
```
Sample span
```
#<struct OpenTelemetry::SDK::Trace::SpanData name="select", kind=:client, status=#<OpenTelemetry::Trace::Status:0x0000ffff7f562618 @code=1, @description="">, parent_span_id="A\x92\xE9S\xCC\x0E\xFD\xA8", total_recorded_attributes=5, total_recorded_events=0, total_recorded_links=0, start_timestamp=1707448523002822050, end_timestamp=1707448523003312883, attributes={"db.system"=>"mysql", "net.peer.name"=>"sinatra-mysql", "net.peer.port"=>"3306", "db.name"=>"classicmodels", "db.statement"=>"SELECT `customers`.* FROM `customers` WHERE (contactLastName like ?) ORDER BY `customers`.`customerNumber` ASC LIMIT ?"}, links=nil, events=nil, resource=#<OpenTelemetry::SDK::Resources::Resource:0x0000ffff7f206dc0 @attributes={"service.name"=>"unknown_service", "process.pid"=>805, "process.command"=>"/usr/local/bundle/bin/rackup", "process.runtime.name"=>"ruby", "process.runtime.version"=>"3.1.0", "process.runtime.description"=>"ruby 3.1.0p0 (2021-12-25 revision fb4df44d16) [aarch64-linux]", "telemetry.sdk.name"=>"opentelemetry", "telemetry.sdk.language"=>"ruby", "telemetry.sdk.version"=>"1.4.0"}>, instrumentation_scope=#<struct OpenTelemetry::SDK::InstrumentationScope name="OpenTelemetry::Instrumentation::Mysql2", version="0.26.1">, span_id="\xAF\xDC\f\xC0\xEA_\xD9W", trace_id="\x92]'\x18\xDB4J5[\xCC\a\xE2\xA7\x8E\xD0\xAF", trace_flags=#<OpenTelemetry::Trace::TraceFlags:0x0000ffff7f441e78 @flags=1>, tracestate=#<OpenTelemetry::Trace::Tracestate:0x0000ffff7f28ebf8 @hash={}>>
```

Related to issue https://github.com/open-telemetry/opentelemetry-ruby-contrib/issues/670

### DIFF
--- a/instrumentation/mysql2/lib/opentelemetry/instrumentation/mysql2/patches/client.rb
+++ b/instrumentation/mysql2/lib/opentelemetry/instrumentation/mysql2/patches/client.rb
@@ -14,12 +14,9 @@ module OpenTelemetry
         # Module to prepend to Mysql2::Client for instrumentation
         module Client
           def query(sql, options = {})
-            attributes = client_attributes
-            attributes.merge!(span_attributes(sql))
-
             tracer.in_span(
-              span_name(sql),
-              attributes: attributes.merge!(OpenTelemetry::Instrumentation::Mysql2.attributes),
+              _otel_span_name(sql),
+              attributes: _otel_span_attributes(sql),
               kind: :client
             ) do
               super(sql, options)
@@ -27,12 +24,9 @@ module OpenTelemetry
           end
 
           def prepare(sql)
-            attributes = client_attributes
-            attributes.merge!(span_attributes(sql))
-
             tracer.in_span(
-              span_name(sql),
-              attributes: attributes.merge!(OpenTelemetry::Instrumentation::Mysql2.attributes),
+              _otel_span_name(sql),
+              attributes: _otel_span_attributes(sql),
               kind: :client
             ) do
               super(sql)
@@ -41,19 +35,19 @@ module OpenTelemetry
 
           private
 
-          def span_name(sql)
+          def _otel_span_name(sql)
             OpenTelemetry::Helpers::MySQL.database_span_name(
               sql,
               OpenTelemetry::Instrumentation::Mysql2.attributes[
                 SemanticConventions::Trace::DB_OPERATION
               ],
-              database_name,
+              _otel_database_name,
               config
             )
           end
 
-          def span_attributes(sql)
-            attributes = {}
+          def _otel_span_attributes(sql)
+            attributes = _otel_client_attributes
             case config[:db_statement]
             when :include
               attributes[SemanticConventions::Trace::DB_STATEMENT] = sql
@@ -63,15 +57,18 @@ module OpenTelemetry
                   sql, obfuscation_limit: config[:obfuscation_limit], adapter: :mysql
                 )
             end
+
+            attributes.merge!(OpenTelemetry::Instrumentation::Mysql2.attributes)
+            attributes.compact!
             attributes
           end
 
-          def database_name
+          def _otel_database_name
             # https://github.com/brianmario/mysql2/blob/ca08712c6c8ea672df658bb25b931fea22555f27/lib/mysql2/client.rb#L78
             (query_options[:database] || query_options[:dbname] || query_options[:db])&.to_s
           end
 
-          def client_attributes
+          def _otel_client_attributes
             # The client specific attributes can be found via the query_options instance variable
             # exposed on the mysql2 Client
             # https://github.com/brianmario/mysql2/blob/ca08712c6c8ea672df658bb25b931fea22555f27/lib/mysql2/client.rb#L25-L26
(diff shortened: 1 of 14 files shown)


instrumentation/mysql2/lib/opentelemetry/instrumentation/mysql2/patches/client.rb
```diff
@@ -15,6 +15,45 @@ module Patches
...
+              attributes: attributes.merge!(OpenTelemetry::Instrumentation::Mysql2.attributes),
+              kind: :client
+            ) do
+              super(sql)
+            end
+          end
+
+          private
+
+          def span_name(sql)
```
arielvalentin, line 44:
```suggestion
          def _otel_span_name(sql)
```
Something that we haven't done yet and probably should is adopt a naming convention for private methods we add via monkey patches to a class.

We should start prefixing all private methods with `_otel` going forward.

instrumentation/mysql2/lib/opentelemetry/instrumentation/mysql2/patches/client.rb
```diff
@@ -15,6 +15,45 @@ module Patches
         module Client
           def query(sql, options = {})
             attributes = client_attributes
+            attributes.merge!(span_attributes(sql))
+
+            tracer.in_span(
+              span_name(sql),
```
arielvalentin, line 21:
```suggestion
              _otel_span_name(sql),
```

instrumentation/mysql2/lib/opentelemetry/instrumentation/mysql2/patches/client.rb
```diff
@@ -15,6 +15,45 @@ module Patches
...
+              super(sql, options)
+            end
+          end
+
+          def prepare(sql)
+            attributes = client_attributes
+            attributes.merge!(span_attributes(sql))
+
+            tracer.in_span(
+              span_name(sql),
```
arielvalentin, line 34:
```suggestion
              _otel_span_name(sql),
```

instrumentation/mysql2/lib/opentelemetry/instrumentation/mysql2/patches/client.rb
```diff
@@ -15,6 +15,45 @@ module Patches
...
+              sql,
+              OpenTelemetry::Instrumentation::Mysql2.attributes[
+                SemanticConventions::Trace::DB_OPERATION
+              ],
+              database_name,
+              config
+            )
+          end
+
+          def span_attributes(sql)
```
arielvalentin, line 55: If you're able to update this as well please

arielvalentin (approved):
Small suggestion that doesn't have to be a part of this PR but I think we should adopt going forward.

instrumentation/mysql2/lib/opentelemetry/instrumentation/mysql2/patches/client.rb
xuan-cao-swi: Thanks for the review, I have changed all the private method with `_otel`, but let me if some of the private method is not required with `_otel`.

arielvalentin (approved):
Thank you for this contribution!

PR #863, opened 2024-02-09
fix: dup string if frozen in trilogy query

plantfansam:
Why do this? We may want to propagate context into sql strings that arrive at the trilogy client frozen.

This PR `dup`s frozen strings if we're not using the noop propagator in trilogy instrumentation. [The specification requires that propagators mutate the carrier](https://github.com/open-telemetry/opentelemetry-specification/blob/fa5eca7aa9d1fc5e11d09e9aec1e67aa2c0afb1f/specification/context/api-propagators.md?plain=1#L73), so we should not do this in the Vitess SqlQuerySetter, which properly does not mutate frozen strings.


instrumentation/trilogy/lib/opentelemetry/instrumentation/trilogy/patches/client.rb
```diff
@@ -50,6 +50,7 @@ def query(sql)
               ),
               kind: :client
             ) do |_span, context|
+              sql = sql.dup unless !sql.frozen? || propagator.is_a?(OpenTelemetry::Instrumentation::Trilogy::NoopPropagator)
```
plantfansam, line 53: I think we could do 

```ruby
sql = +sql unless propagator.is_a?(OpenTelemetry::Instrumentation::Trilogy::NoopPropagator)
```

But I think the syntax is slightly obscure ([docs](https://ruby-doc.org/3.2.2/String.html#method-i-2B-40)). Curious for others' thoughts.

```ruby
irb(main):001:0> x = "foo".freeze
=> "foo"
irb(main):002:0> x.gsub!("f", "o")
Traceback (most recent call last):
        5: from /usr/bin/irb:23:in `<main>'
        4: from /usr/bin/irb:23:in `load'
        3: from /Library/Ruby/Gems/2.6.0/gems/irb-1.0.0/exe/irb:11:in `<top (required)>'
        2: from (irb):2
        1: from (irb):2:in `gsub!'
FrozenError (can't modify frozen String)
irb(main):003:0> x = +x
=> "foo"
irb(main):004:0> x.gsub!("f", "o")
=> "ooo"
irb(main):005:0> x
=> "ooo"
```

arielvalentin: Are we mutating the string to avoid additional allocations?

instrumentation/trilogy/lib/opentelemetry/instrumentation/trilogy/patches/client.rb
```diff
@@ -50,6 +50,7 @@ def query(sql)
...
               propagator.inject(sql, context: context)
               super(sql)
```
arielvalentin, line 55: Should this `freeze` the string again before sending it off it the one it received was already frozen?

plantfansam: 👍 9ea834e06bc48d59eca272b9c4ebeb2518bca56d

instrumentation/trilogy/lib/opentelemetry/instrumentation/trilogy/patches/client.rb
plantfansam: Yeah, sorry, got a little too steeped in my own context... Updated pr description with:

> Why do this? We may want to propagate context into sql strings that arrive at the trilogy client frozen.

Does that clarify things?

fbogsany: I think it may be cleaner at this point to not install a noop propagator and instead check `unless propagator.nil?` here.

plantfansam: I agree; will make this change

instrumentation/trilogy/test/opentelemetry/instrumentation/trilogy/instrumentation_test.rb
```diff
@@ -376,13 +376,14 @@
...
+        sql = 'SELECT * from users where users.id = 1 and users.email = "test@test.com"'.freeze # rubocop:disable Style/RedundantFreeze
+        propagator = double('propagator')
+        allow(propagator).to receive(:inject)
+        allow(client).to receive(:propagator).and_return(propagator)
         expect do
           client.query(sql)
         end.must_raise Trilogy::Error
-        encoded = Base64.strict_encode64("{\"uber-trace-id\":\"#{span.hex_trace_id}:#{span.hex_span_id}:0:1\"}")
-        _(sql).must_equal "/*VT_SPAN_CONTEXT=#{encoded}*/#{original_sql}"
+        expect(propagator).to have_received(:inject).with(sql, context: instance_of(OpenTelemetry::Context)).once
```
arielvalentin, line 386: How does this test ensure that a mutable string is what is passed in to the propagator?

How does it ensure what is returned is a frozen string that was potentially changed?

plantfansam: Yeah, great point - I'm going to take another pass at specs.

plantfansam: The tests ended up pretty gnarly on this 😳  - see e84b23d2fc13f7d7a2e5965066be871f5ede314e

Why gnarly? Based on my observation (i.e. not sourcediving, take it with a grain of salt, etc.), rspec-mocks appears to capture a _reference_ to the arguments passed when you use methods like `foo.receive(:inject).with(...)`. This means that if a method mutates an argument, rspec-mocks will only know about the argument's value at assertion time (vs. at call time). To get around this, I used `receive(:inject).and_wrap_original` to make assertions about frozenness etc. I plan to take another look at this tomorrow when my eyes are less glazed over 😛

arielvalentin (approved):

instrumentation/trilogy/lib/opentelemetry/instrumentation/trilogy/instrumentation.rb
```diff
@@ -61,10 +45,10 @@ def configure_propagator(config)
           propagator = config[:propagator]
           @propagator = case propagator
                         when 'vitess' then fetch_propagator(propagator, 'OpenTelemetry::Propagator::Vitess')
-                        when 'none', nil then NoopPropagator.new
+                        when 'none', nil
+                          nil
```
fbogsany, line 49:
```suggestion
                        when 'none', nil then nil
```
for consistency (otherwise remove the `then` in the `when` above).

instrumentation/trilogy/lib/opentelemetry/instrumentation/trilogy/patches/client.rb
```diff
@@ -50,7 +50,13 @@ def query(sql)
               ),
               kind: :client
             ) do |_span, context|
-              propagator.inject(sql, context: context)
+              if propagator
+                dup = sql.frozen?
+                sql = sql.dup if dup
+                propagator.inject(sql, context: context)
+                sql.freeze if dup
+              end
```
fbogsany, line 58: This is kinda messy and hard to follow. It'd be clearer to:
```suggestion
              if propagator
                injectable_sql = +sql
                propagator.inject(injectable_sql, context: context)
                sql = -injectable_sql if sql.frozen?
              end
```
or
```suggestion
              if propagator && sql.frozen?
                sql = +sql
                propagator.inject(sql, context: context)
                sql.freeze
              elsif propagator
                propagator.inject(sql, context: context)
              end
```

fbogsany (approved):

PR #872, opened 2024-02-17
feat: Add support gruf 2.19

AS-AlStar:
Update appraise for gruf 2.19 and old versions

### DIFF
--- a/instrumentation/gruf/Appraisals
+++ b/instrumentation/gruf/Appraisals
@@ -4,10 +4,7 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 
-appraise 'gruf-2.15.1' do
-  gem 'gruf', '~> 2.15.1'
-end
-
-appraise 'gruf-2.16.1' do
-  gem 'gruf', '~> 2.16.1'
+appraise 'gruf-2.19.0' do
+  gem 'gruf', '~> 2.19.0'
+  gem 'grpc', '>= 1.61.0'
 end
--- a/instrumentation/gruf/Appraisals
+++ b/instrumentation/gruf/Appraisals
@@ -4,7 +4,14 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 
-appraise 'gruf-2.19.0' do
+appraise 'gruf-2.17' do
+  gem 'gruf', '~> 2.17.0'
+end
+
+appraise 'gruf-2.18' do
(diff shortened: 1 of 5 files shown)


instrumentation/gruf/lib/opentelemetry/instrumentation/gruf/version.rb
```diff
@@ -7,7 +7,7 @@
 module OpenTelemetry
   module Instrumentation
     module Gruf
-      VERSION = '0.1.1'
+      VERSION = '0.1.2'
```
arielvalentin, line 10:
```suggestion
      VERSION = '0.1.1'
```
Our release process manages gem versions. No need to change this manually

instrumentation/gruf/Appraisals
```diff
@@ -4,10 +4,7 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 
-appraise 'gruf-2.15.1' do
```
arielvalentin, line 7: This means we lose test coverage for versions prior to 2.19.

Is that ok? Are you looking to drop support for older versions?

arielvalentin (changes requested):
This means we won't have test coverage for versions earlier than 2.19

instrumentation/gruf/lib/opentelemetry/instrumentation/gruf/version.rb
AS-AlStar: ok

instrumentation/gruf/Appraisals
AS-AlStar: I added the support of the last three versions of Gruf, it seems that there is no need to indicate the dependence of the GRPC version

arielvalentin: thank you!

arielvalentin:
Please remove the following to re-enable the tests in CI: 

https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/a005a987c74f1dd0958aae9e4d47ad84cf34a611/.github/workflows/ci-instrumentation.yml#L55
https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/a005a987c74f1dd0958aae9e4d47ad84cf34a611/.github/workflows/ci-instrumentation-canary.yml#L64

arielvalentin (approved):

PR #884, opened 2024-02-21
fix: suppress deprecation warning in Rdkafka Instrumentation

yoheyk:
### Background

Observing the following warning with rdkafka `0.13.0` and above:

https://github.com/open-telemetry/opentelemetry-ruby-contrib/actions/runs/7980847049/job/21791353080#step:4:224

>    rdkafka deprecation warning: header access with Symbol key :traceparent treated as a String. Please change your code to use String keys to avoid this warning. Symbol keys will break in version 1.

This warning is coming from here:
https://github.com/karafka/rdkafka-ruby/blob/v0.14.0/lib/rdkafka/consumer/headers.rb#L10-L11

### Changes summary

Use `Context::Propagation.text_map_getter` with rdkafka `0.13.0` and above.

if we can drop support for rdkafka `0.12.0`, it's going to be much simpler
https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/c95e91909912fede6dad07bffa15213c087a677d/instrumentation/rdkafka/opentelemetry-instrumentation-rdkafka.gemspec#L39


instrumentation/rdkafka/lib/opentelemetry/instrumentation/rdkafka/patches/consumer.rb
```diff
@@ -74,6 +74,14 @@ def extract_message_key(key)
           rescue Encoding::UndefinedConversionError
             nil
           end
+
+          def getter
+            if Gem::Version.new(::Rdkafka::VERSION) >= Gem::Version.new('0.13.0')
```
arielvalentin, line 79: Is there a way to check/set this only once when the gem is activated as opposed to each time a message is processed?

arielvalentin (commented):
Thank you for your continued support here! 

I have a question about potentially reducing the number of checks that I left in line.

instrumentation/rdkafka/lib/opentelemetry/instrumentation/rdkafka/patches/consumer.rb
yoheyk: we could memoize like 661cd674e1668b62092a25239c44c5b2c4ddecff or do you think it's better to use a constant like this?
```ruby
          GETTER = if Gem::Version.new(::Rdkafka::VERSION) >= Gem::Version.new('0.13.0')
              Context::Propagation.text_map_getter
            else
              OpenTelemetry::Common::Propagation.symbol_key_getter
            end
          private_constant :GETTER
```

(btw we still need to support rdkafka `0.12.0`, right?)

yoheyk: fyi, ` 0.12.0` is released on June 17, 2022, and the latest version of rdkafka is `0.15.1` as of now

https://rubygems.org/gems/rdkafka/versions

instrumentation/rdkafka/lib/opentelemetry/instrumentation/rdkafka/patches/consumer.rb
```diff
@@ -74,6 +74,14 @@ def extract_message_key(key)
           rescue Encoding::UndefinedConversionError
             nil
           end
+
+          def getter
+            @getter ||= if Gem::Version.new(::Rdkafka::VERSION) >= Gem::Version.new('0.13.0')
```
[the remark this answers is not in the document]
yoheyk: how about this? 21e4a43f1824c7c0fa17a0675c53adee93ed9d2e

arielvalentin (approved):

PR #389, opened 2023-03-19
feat!: Drop support for EoL Ruby 2.7

arielvalentin:
https://www.ruby-lang.org/en/downloads/branches/

> Ruby 2.7
> status: security maintenance
> release date: 2019-12-25
> normal maintenance until: 2022-04-01
> EOL: 2023-03-31 (expected)

Fixes https://github.com/open-telemetry/opentelemetry-ruby-contrib/issues/363

PR #885, opened 2024-02-23
chore: drop dependency on ruby2_keywords

Earlopain:
Now that Ruby below 2.7 is not supported (#389, https://github.com/open-telemetry/opentelemetry-ruby-contrib/commit/38b083aa41b93bcc8f900cd65bc922c97fe26e5d) this is not needed anymore

### DIFF
--- a/instrumentation/active_record/lib/opentelemetry/instrumentation/active_record/instrumentation.rb
+++ b/instrumentation/active_record/lib/opentelemetry/instrumentation/active_record/instrumentation.rb
@@ -51,9 +51,6 @@ module OpenTelemetry
         end
 
         def require_dependencies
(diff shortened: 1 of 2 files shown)


Earlopain:
The `ruby2_keywords` methods are natively defined in ruby versions after 2.6, all this dependency does is define empty shims for compatibility with older versions (see https://github.com/ruby/ruby2_keywords/blob/3882dac4c9bc784183d7d4cf7374edd56017746f/lib/ruby2_keywords.rb)

Removing these from existing code would change behaviour in some way and I can't say what that entails. It would be better kept for a separate PR.

xuan-cao-swi (approved):

Issue #293, opened 2023-01-26
ActionMailer instrumentation


arielvalentin:
Convert Datadog instrumentation for ActionMailer to OpenTelemetry.

See

[Datadog porting guide](https://github.com/open-telemetry/opentelemetry-ruby/blob/master/adapters/datadog-porting-guide.md)
[Semantic Conventions](https://github.com/open-telemetry/opentelemetry-specification/tree/master/specification/trace/semantic_conventions)
[Datadog implementation](https://github.com/DataDog/dd-trace-rb/tree/a7c8aa5d81b67d136665dfb86d2345b783fb2290/lib/datadog/tracing/contrib/action_mailer)

kaylareopelle:
Closed by #887 and #984

PR #887, opened 2024-02-23
feat: add instrumentation support for action mailer

xuan-cao-swi:
## Description

Adding support for instrumenting action mailer using active support notification. This instrumentation is modified based on action_view.

Related: https://github.com/open-telemetry/opentelemetry-ruby-contrib/issues/293

## Test

Sample span_data

```
#<struct OpenTelemetry::SDK::Trace::SpanData
 name="action_mailer deliver",
 kind=:internal,
 status=#<OpenTelemetry::Trace::Status:0x0000ffff8eb50908 @code=1, @description="">,
 parent_span_id="\x00\x00\x00\x00\x00\x00\x00\x00",
 total_recorded_attributes=7,
 total_recorded_events=0,
 total_recorded_links=0,
 start_timestamp=1713299681034888674,
 end_timestamp=1713299681044902841,
 attributes=
  {"email.message_id"=>"661ee0e1893d_16bba442389@e985aff9a500.mail",
   "email.subject"=>"Welcome to OpenTelemetry!",
   "email.x_mailer"=>"TestMailer",
   "email.to.address"=>["test_mailer@otel.org"],
   "email.from.address"=>["no-reply@example.com"],
   "email.cc.address"=>["cc@example.com"],
   "email.bcc.address"=>["bcc@example.com"]},
 links=nil,
 events=nil,
 resource=
  #<OpenTelemetry::SDK::Resources::Resource:0x0000ffff85aec2e0
   @attributes=
    {"service.name"=>"unknown_service",
     "process.pid"=>363,
     "process.command"=>"trace_request_demonstration.ru",
     "process.runtime.name"=>"ruby",
     "process.runtime.version"=>"3.2.2",
     "process.runtime.description"=>"ruby 3.2.2 (2023-03-30 revision e51014f9c0) [aarch64-linux-musl]",
     "telemetry.sdk.name"=>"opentelemetry",
     "telemetry.sdk.language"=>"ruby",
     "telemetry.sdk.version"=>"1.4.1"}>,
 instrumentation_scope=#<struct OpenTelemetry::SDK::InstrumentationScope name="OpenTelemetry::Instrumentation::ActionMailer", version="0.1.0">,
 span_id="}\x1C\x91C\e\x88\x00v",
 trace_id="\xBF\xC0\xAD%\xBA\xF8\xAD\x974\xD0\x1A \xE2=F\xDB",
 trace_flags=#<OpenTelemetry::Trace::TraceFlags:0x0000ffff883d0030 @flags=1>,
 tracestate=#<OpenTelemetry::Trace::Tracestate:0x0000ffff883f3ee0 @hash={}>>
```


arielvalentin:
I'll take a closer look later this week. Apologies for waiting so long to review.

instrumentation/action_mailer/lib/opentelemetry/instrumentation/action_mailer/railtie.rb
```diff
@@ -0,0 +1,32 @@
...
+      ].freeze
+
+      # This Railtie sets up subscriptions to relevant ActionMailer notifications
+      class Railtie < ::Rails::Railtie
+        config.after_initialize do
+          ::OpenTelemetry::Instrumentation::ActiveSupport::Instrumentation.instance.install({})
+
+          SUBSCRIPTIONS.each do |subscription_name|
+            config = ActionMailer::Instrumentation.instance.config
+            ::OpenTelemetry::Instrumentation::ActiveSupport.subscribe(
```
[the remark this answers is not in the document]
xuan-cao-swi: By omitting the email address, like omitting `:to`, `:from`, `:bcc`, `:cc`?
For semantic convention, I can only find the message_id from opentelemetry-ruby `MESSAGE_ID`. I didn't find anything that's related in spec.

instrumentation/action_mailer/lib/opentelemetry/instrumentation/action_mailer/instrumentation.rb
```diff
@@ -0,0 +1,40 @@
...
+
+        present do
+          defined?(::ActionMailer)
+        end
+
+        compatible do
+          gem_version >= MINIMUM_VERSION
+        end
+
+        option :disallowed_notification_payload_keys, default: [],  validate: :array
```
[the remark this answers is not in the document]
xuan-cao-swi: Do you mean like we change the logic of [span_subscriber](https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/opentelemetry-instrumentation-active_support/v0.5.1/instrumentation/active_support/lib/opentelemetry/instrumentation/active_support/span_subscriber.rb#L105) so it can reflect to all instrumentations that depend on it?

instrumentation/action_mailer/lib/opentelemetry/instrumentation/action_mailer/instrumentation.rb
```diff
@@ -0,0 +1,40 @@
...
+        option :notification_payload_transform,       default: nil, validate: :callable
```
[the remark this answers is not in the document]
xuan-cao-swi: same comment from previous one?

instrumentation/action_mailer/lib/opentelemetry/instrumentation/action_mailer/instrumentation.rb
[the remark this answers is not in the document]
arielvalentin: yes. Again I do not think that needs to be done here.

instrumentation/action_mailer/lib/opentelemetry/instrumentation/action_mailer/instrumentation.rb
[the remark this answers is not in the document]
arielvalentin: same. We can address this in the future.

instrumentation/action_mailer/lib/opentelemetry/instrumentation/action_mailer/railtie.rb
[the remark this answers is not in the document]
xuan-cao-swi: > By omitting the email address, like omitting :to, :from, :bcc, :cc?

I have updated the instrumentation that have modifying the options if email_address is omit; but not sure if it's best practice.

> rails.action_mailer.message.id

Do you mean like introduce a new semconv (i.e. RAILS_ACTION_MAILER_MESSAGE_ID=...)?

instrumentation/action_mailer/README.md
arielvalentin, line 1: Please add some details about what attributes are published. 

Take a look at the ActiveJob Readme for an example.

instrumentation/action_mailer/README.md
```diff
@@ -0,0 +1,71 @@
...
+end
+```
+
+## Examples
+
+Example usage can be seen in the `./example/trace_request_demonstration.ru` file [here](https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/main/instrumentation/action_mailer/example/trace_request_demonstration.ru)
+
+## Known issues
+
+ActionMailer instrumentation uses ActiveSupport notifications and in the case when a subscriber raises in start method an unclosed span would break successive spans ends. Example:
```
arielvalentin, line 41: Is this still a problem? I thought this was fixed?

instrumentation/action_mailer/lib/opentelemetry/instrumentation/action_mailer/railtie.rb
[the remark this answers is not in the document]
arielvalentin: Yes, I do mean introduce new semconv like I did for active job.

It may be something to ask in the semantic conventions SIG to see what others think.

instrumentation/action_mailer/lib/opentelemetry/instrumentation/action_mailer/instrumentation.rb
```diff
@@ -0,0 +1,63 @@
...
+          if ActionMailer::Instrumentation.instance.config[:notification_payload_transform].nil?
+            mask_email = lambda do |payload|
+              payload[:mail].gsub!(/(From:\s+|To:\s+|Cc:\s+)[^\s]+/, '\1[redacted]')
+              payload
+            end
+          else
+            original_callable = ActionMailer::Instrumentation.instance.config[:notification_payload_transform]
+            mask_email = lambda do |payload|
+              original_callable.call(payload)
+              payload[:mail].gsub!(/(From:\s+|To:\s+|Cc:\s+)[^\s]+/, '\1[redacted]')
```
arielvalentin, line 50: Is this the full contents of the email message? 

If it is, we should not include this at all. It would be the equivalent of recording the contents of an HTTP response body.

instrumentation/action_mailer/lib/opentelemetry/instrumentation/action_mailer/railtie.rb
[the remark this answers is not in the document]
arielvalentin: The problem is I have no idea what would be good names for this.

instrumentation/action_mailer/README.md
xuan-cao-swi: Thanks, updated

instrumentation/action_mailer/README.md
xuan-cao-swi: Updated

instrumentation/action_mailer/lib/opentelemetry/instrumentation/action_mailer/instrumentation.rb
xuan-cao-swi: Yes, the problem is that when omit the email address, the attributes one can be excluded, but the content still have the email address

```
"Date: Fri, 23 Feb 2024 15:34:28 +0000\r\n" +
    "From: no-reply@example.com\r\n" +
    "To: test_mailer@otel.org\r\n" +
    "Message-ID: <65d8bb04c68d8_15bcc45779@4d4fcd81426e.mail>\r\n" +
    "Subject: Welcome to OpenTelemetry!\r\n" +
    "Mime-Version: 1.0\r\n" +
...
```
With gsub:
```
"Date: Fri, 23 Feb 2024 15:34:28 +0000\r\n" +
    "From: [redacted]\r\n" +
    "To: [redacted]\r\n" +
    "Message-ID: <65d8bb04c68d8_15bcc45779@4d4fcd81426e.mail>\r\n" +
    "Subject: Welcome to OpenTelemetry!\r\n" +
    "Mime-Version: 1.0\r\n" +
```

xuan-cao-swi: > Is this the full contents of the email message?

> If it is, we should not include this at all. It would be the equivalent of recording the contents of an HTTP response body.

Will remove the mail part.

instrumentation/action_mailer/lib/opentelemetry/instrumentation/action_mailer/instrumentation.rb
```diff
@@ -0,0 +1,79 @@
...
+          payload['email.subject']    = payload[:subject]
+          payload['email.x_mailer']   = payload[:mailer]
+          payload['email.to.address'] = payload[:to]
+          payload['email.from.address'] = payload[:from]
+          payload['email.cc.address'] = payload[:cc]
+          payload['email.bcc.address'] = payload[:bcc]
+          payload['email.delivery_timestamp'] = payload[:date]
+          payload['email.origination_timestamp'] = payload[:date]
+
+          payload.delete_if { |item| item.instance_of?(Symbol) }
```
arielvalentin, line 70: Why delete symbols? Would making them strings satisfy the API?

instrumentation/action_mailer/lib/opentelemetry/instrumentation/action_mailer/instrumentation.rb
```diff
@@ -0,0 +1,79 @@
...
+        # email attribute key convention is obtained from: https://www.elastic.co/guide/en/ecs/8.11/ecs-email.html
+        def transform_payload(payload)
+          payload['email.message_id'] = payload[:message_id]
+          payload['email.subject']    = payload[:subject]
+          payload['email.x_mailer']   = payload[:mailer]
+          payload['email.to.address'] = payload[:to]
+          payload['email.from.address'] = payload[:from]
+          payload['email.cc.address'] = payload[:cc]
+          payload['email.bcc.address'] = payload[:bcc]
+          payload['email.delivery_timestamp'] = payload[:date]
```
arielvalentin, line 67: Should this be the timestamp of the event and not the date of the email? Are those values different?

Does rails attempt to resend the email if it fails?

instrumentation/action_mailer/lib/opentelemetry/instrumentation/action_mailer/instrumentation.rb
```diff
@@ -0,0 +1,79 @@
...
+          ::ActionMailer.version
+        end
+
+        def resolve_email_address
+          return unless ActionMailer::Instrumentation.instance.config[:email_address] == :omit
+
+          ActionMailer::Instrumentation.instance.config[:disallowed_notification_payload_keys] += ['email.to.address', 'email.from.address', 'email.cc.address', 'email.bcc.address']
+        end
+
+        def ecs_mail_convention
```
arielvalentin, line 43: Can we set this at install instead of checking with watch email?

instrumentation/action_mailer/lib/opentelemetry/instrumentation/action_mailer/instrumentation.rb
```diff
@@ -0,0 +1,79 @@
...
+        option :notification_payload_transform,       default: nil, validate: :callable
+        option :email_address,                        default: :omit, validate: %I[omit include]
+
+        private
+
+        def gem_version
+          ::ActionMailer.version
+        end
+
+        def resolve_email_address
```
arielvalentin, line 37: Same here. Can this be evaluated at install time?

instrumentation/action_mailer/lib/opentelemetry/instrumentation/action_mailer/instrumentation.rb
xuan-cao-swi: If not removing the symbol, they will show in span attributes, which they contain the same value as `email.***` attributes. 

active_support notification expose all key/value from payload unless there is key in `disallowed_notification_payload_keys`, but that will have to include a lot of keys; so I just remove them here.

instrumentation/action_mailer/lib/opentelemetry/instrumentation/action_mailer/instrumentation.rb
arielvalentin: it is and I am dumb.

instrumentation/action_mailer/lib/opentelemetry/instrumentation/action_mailer/instrumentation.rb
arielvalentin: 🤦🏼 it is already

instrumentation/action_mailer/lib/opentelemetry/instrumentation/action_mailer/instrumentation.rb
arielvalentin: I see. Do you think it will be too expensive to allocate a new hash here?

instrumentation/action_mailer/lib/opentelemetry/instrumentation/action_mailer/instrumentation.rb
xuan-cao-swi: > Should this be the timestamp of the event and not the date of the email? Are those values different?

Yes, I wasn't clear on how mail gem generate the date before, after looking more deeper, 

By ECS definition:

Key | Value
-- | --
email.origination_timestamp |  The date and time the email message was composed. Many email clients will fill in this value automatically when the message is sent by a user.
email.delivery_timestamp | The date and time when the email message was received by the service or client.

The date value from [mail gem](https://github.com/mikel/mail/blob/master/lib/mail/message.rb#L352) is auto-generated when mail object is created (current time). So I think the `payload[:date]` value should be under origination_timestamp.

> Does rails attempt to resend the email if it fails?

I assume the failure of operation? I think Action Mailer in Ruby on Rails does not automatically attempt to resend an email if the sending process fails, but it will raise error (e.g. network issues, authentication errors)

instrumentation/action_mailer/lib/opentelemetry/instrumentation/action_mailer/instrumentation.rb
xuan-cao-swi: Do you mean calling this function inside `install do |_config| ... end` block?

instrumentation/action_mailer/lib/opentelemetry/instrumentation/action_mailer/instrumentation.rb
xuan-cao-swi: You mean `delete_if`? I don't think `delete_if` create new hash but correct me if I am wrong.

arielvalentin: The `transform_payload` should return a new hash.

The problem I think is that this is mutating the payload, which will impact other downstream subscribers that will be expecting those fields:

https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/8336decf0e28712c98bb010744e677c0d02f6f95/instrumentation/active_support/lib/opentelemetry/instrumentation/active_support/span_subscriber.rb#L81

xuan-cao-swi: Updated!

instrumentation/action_mailer/lib/opentelemetry/instrumentation/action_mailer/instrumentation.rb
```diff
@@ -0,0 +1,84 @@
...
+
+        def ecs_mail_convention
+          if ActionMailer::Instrumentation.instance.config[:notification_payload_transform].nil?
+            transform_attributes = ->(payload) { transform_payload(payload) }
+          else
+            original_callable = ActionMailer::Instrumentation.instance.config[:notification_payload_transform]
+            transform_attributes = lambda do |payload|
+              new_payload = transform_payload(payload)
+              user_payload = original_callable.call(new_payload)
+              if user_payload.instance_of?(Hash)
```
arielvalentin, line 51: what are we protecting from here? The user passing in an invalid transformer that does not return a hash?

Should this actually be in the `ActiveSupport::Notifications` instrumentation instead?

instrumentation/action_mailer/lib/opentelemetry/instrumentation/action_mailer/instrumentation.rb
```diff
@@ -0,0 +1,84 @@
...
+            'email.message_id' => payload[:message_id],
+            'email.subject' => payload[:subject],
+            'email.x_mailer' => payload[:mailer],
+            'email.to.address' => payload[:to],
+            'email.from.address' => payload[:from],
+            'email.cc.address' => payload[:cc],
+            'email.bcc.address' => payload[:bcc],
+            'email.origination_timestamp' => payload[:date]
+          }
+          new_payload.compact!
```
arielvalentin, line 74: This is a micro optimization but checking if the values are `nil` before adding them to the hash is generally faster than using `compact!`

https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/693#discussion_r1360669417

arielvalentin (approved):

instrumentation/action_mailer/lib/opentelemetry/instrumentation/action_mailer/instrumentation.rb
xuan-cao-swi: > The user passing in an invalid transformer that does not return a hash?  

To ensure user's callable return hash, otherwise the active_support part will report error.

> Should this actually be in the ActiveSupport::Notifications instrumentation instead?

It could be.

Also, just want to pitch the idea of restriction of mutating original payload here:

When user provide callable function for this instrumentation, they can only play with the transformed payload that is generated from `transform_payload()`, not the original payload. This avoid user mutating original payload in this instrumentation. 

They can still mutate the payload if they subscribe the same subscription pattern (i.e. `deliver.action_mailer`) but that's up to user.

instrumentation/action_mailer/test/opentelemetry/instrumentation/action_mailer/instrumentation_test.rb
```diff
@@ -0,0 +1,34 @@
...
+  it 'has #name' do
+    _(instrumentation.name).must_equal 'OpenTelemetry::Instrumentation::ActionMailer'
+  end
+
+  it 'has #version' do
+    _(instrumentation.version).wont_be_nil
+    _(instrumentation.version).wont_be_empty
+  end
+
+  describe '#install' do
```
kaylareopelle, line 21: Would you be open to adding some tests for the logic in `resolve_email_address`, `ecs_mail_convention`, and `transform_payload`? Since they're private methods, I know that's not always a common practice, but I'm concerned about not having coverage for their logic.

instrumentation/action_mailer/lib/opentelemetry/instrumentation/action_mailer/instrumentation.rb
```diff
@@ -0,0 +1,82 @@
...
+        def transform_payload(payload)
+          new_payload = {}
+          new_payload['email.message_id'] = payload[:message_id] if payload[:message_id]
+          new_payload['email.subject'] = payload[:subject] if payload[:subject]
+          new_payload['email.x_mailer'] = payload[:mailer] if payload[:mailer]
+          new_payload['email.to.address'] = payload[:to] if payload[:to]
+          new_payload['email.from.address'] = payload[:from] if payload[:from]
+          new_payload['email.cc.address'] = payload[:cc] if payload[:cc]
+          new_payload['email.bcc.address'] = payload[:bcc] if payload[:bcc]
+          new_payload['email.origination_timestamp'] = payload[:date] if payload[:date]
```
kaylareopelle, line 72: It looks like the original payload used symbols as keys, but the new payload is using strings. Is that intended? Would there be a performance benefit by making the strings symbols, ex. `:'email.message_id'`?

kaylareopelle (changes requested):
Thank you for writing this, Xuan!

instrumentation/action_mailer/lib/opentelemetry/instrumentation/action_mailer/instrumentation.rb
arielvalentin: These are used to populate the span so they must be strings as spans do not allow symbol key attributes

kaylareopelle: Got it, thanks @arielvalentin! What about making them constants?

arielvalentin: To avoid "magic variables"?

instrumentation/action_mailer/example/trace_request_demonstration.ru
```diff
@@ -0,0 +1,63 @@
...
+# SPDX-License-Identifier: Apache-2.0
+
+require 'bundler/inline'
+
+gemfile(true) do
+  source 'https://rubygems.org'
+
+  gem 'rails'
+  gem 'opentelemetry-sdk'
+  gem 'stringio', '3.0.4'
```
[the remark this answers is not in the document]
xuan-cao-swi: for my env (ruby 3.2.2 (2023-03-30 revision e51014f9c0) [x86_64-linux-musl]), it asks 3.0.4. I guess it's env varied.

instrumentation/action_mailer/README.md
```diff
@@ -0,0 +1,100 @@
...
+
+If only want to hide certain field for email address:
+```ruby
+OpenTelemetry::SDK.configure do |c|
+  c.use 'OpenTelemetry::Instrumentation::ActionMailer', { email_address: :include, disallowed_notification_payload_keys: ['email.to.address'] }
+end
+```
+
+
+## Active Support Instrumentation
```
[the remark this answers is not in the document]
xuan-cao-swi: updated!

instrumentation/action_mailer/lib/opentelemetry/instrumentation/action_mailer/instrumentation.rb
```diff
@@ -0,0 +1,82 @@
...
+        option :email_address,                        default: :omit, validate: %I[omit include]
+
+        private
+
+        def gem_version
+          ::ActionMailer.version
+        end
+
+        def resolve_email_address
+          return unless ActionMailer::Instrumentation.instance.config[:email_address] == :omit
```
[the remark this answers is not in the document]
xuan-cao-swi: Updated!

instrumentation/action_mailer/test/opentelemetry/instrumentation/action_mailer/instrumentation_test.rb
xuan-cao-swi: Yes, I will updated the test case

kaylareopelle: Thank you!

instrumentation/action_mailer/lib/opentelemetry/instrumentation/action_mailer/instrumentation.rb
```diff
@@ -0,0 +1,82 @@
...
+          defined?(::ActionMailer)
+        end
+
+        compatible do
+          gem_version >= MINIMUM_VERSION
+        end
+
+        option :disallowed_notification_payload_keys, default: [], validate: :array
+        option :notification_payload_transform,       default: nil, validate: :callable
+        option :email_address,                        default: :omit, validate: %I[omit include]
```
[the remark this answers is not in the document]
kaylareopelle: I created an issue to tackle this: https://github.com/open-telemetry/opentelemetry-ruby-contrib/issues/969

instrumentation/action_mailer/lib/opentelemetry/instrumentation/action_mailer/instrumentation.rb
kaylareopelle: I was thinking string allocations, but I don't think this is going to be run very many times, so I don't think we need to worry about it here.

kaylareopelle (approved):
Approved as-is, but I have a few documentation-related suggestions you're welcome to apply or ignore. 

Thanks, Xuan!

xuan-cao-swi:
> could you either add this gem to [the rails instrumentation gemspec](https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/main/instrumentation/rails/opentelemetry-instrumentation-rails.gemspec) as part of this PR or open an issue about this?

Yes, I created an issue that is tagged on this PR

instrumentation/action_mailer/README.md
```diff
@@ -0,0 +1,102 @@
...
+
+This instrumentation relies entirely on `ActiveSupport::Notifications` and registers a custom Subscriber that listens to relevant events to report as spans.
+
+See the table below for details of what [Rails Framework Hook Events](https://guides.rubyonrails.org/active_support_instrumentation.html#action-mailer) are recorded by this instrumentation:
+
+| Event Name | Creates Span? | Notes |
+| - | - | - |
+| `deliver.action_mailer` | :white_check_mark: | Creates an span with kind `internal` and email content and status|
+| `process.action_mailer` | :x: | Lack of useful info so ignored |
```
mrsimo, line 48: Not sure whether to open a new issue, but wanted to try here first.

We had added instrumentation for `ActionMailer` ourselves using `OpenTelemetry::Instrumentation::ActiveSupport.subscribe`, but had added `process.action_mailer` as well. The span itself doesn't have interesting attributes but it helps wrap everything necessary to render the email in a common span. We find this very useful. For example:

image

Helps see how long it takes to render the email vs actually deliver it:

image

Would you be open to adding this as well?

arielvalentin: We generally add functionality based on our user need and in many cases we tend to add features that we (maintainers) regularly use in production. 

We are happy to review any contributions you'd like to submit to the community and look forward to your help.

PR #899, opened 2024-03-05
feat: Skip  Trilogy`ping` span with option `ping_enabled`

### LINKED ISSUE #842: Trilogy instrumentation should make root `ping` spans opt-in
Trilogy can generate _quite a lot_ of `ping` spans that have no parent (or child). AFAICT these spans come from background workers that are booting up, and don't add a ton of value. IMO we should default to _not_ tracing `ping` if it doesn't have a parent and allow a config option to opt in.

TonyCTHsu:
closes #842 

Trilogy: Add an configuration option `ping_enabled` to skip `ping` instrumentation


instrumentation/trilogy/lib/opentelemetry/instrumentation/trilogy/instrumentation.rb
```diff
@@ -28,6 +28,7 @@ class Instrumentation < OpenTelemetry::Instrumentation::Base
         option :span_name, default: :statement_type, validate: %I[statement_type db_name db_operation_and_name]
         option :obfuscation_limit, default: 2000, validate: :integer
         option :propagator, default: nil, validate: :string
+        option :ping_enabled, default: true, validate: :boolean
```
xuan-cao-swi, line 31: Looks good to me. I just have two questions based on the original issue
> IMO we should default to not tracing ping if it doesn't have a parent and allow a config option to opt in.

1. Should this option be false as default?
2. I'm not familiar with mysql2 ping operation, so I'm not sure in what case the ping operation will have a parent span. Potentially, a scenario involving a rack span as the parent, with other spans acting as child spans if it's a Rails or Sinatra application invoking ActiveRecord. Do you have any recommendations for addressing cases where a parent span exists and enabling the ping accordingly?

TonyCTHsu: Addressed with: https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/899/commits/8049f2c3752ad90b0053ef3cd6b72c77deb5bdd5

> 1. Should this option be false as default?

I set it as `true` originally because that would result in no data lost and no configuration changes for existing user. I don't mind the option being `false`. However, I would like to learn whether such change should be consider as a breaking change (major in SemVer2) or behavior change (minor in SemVer2)?

> so I'm not sure in what case the ping operation will have a parent span. 

I can imagine someone is building a `/healthcheck` endpoint that implements pinging databases. I am not familiar with OTel's public interfaces but I tried to simulate such scenario in test with
```ruby
OpenTelemetry.tracer_provider.tracer.in_span('parent') do
  client.ping
end
```
Let me know if I am missing something or wrong about this.

xuan-cao-swi: > I set it as true originally because that would result in no data lost and no configuration changes for existing user.

Make sense. Maybe ask it in next SIG meeting (also breaking vs behavior change).

>  I would feel strange to see the ping span generated if I configured ping_enabled as false (Which also means that where is no way to silence it entirely.)

Agree. I think even document it still confuse people. I suggest either removing the condition of `if ping doesn't have parent span ...` (i.e. disable all if set false)  or make the option name more specific. 

cc: @plantfansam

instrumentation/trilogy/lib/opentelemetry/instrumentation/trilogy/patches/client.rb
```diff
@@ -26,6 +26,8 @@ def initialize(options = {})
           end
 
           def ping(...)
+            return super unless OpenTelemetry::Trace.current_span.context.valid? || config[:ping_enabled]
```
arielvalentin, line 29: Is this something that we can mixin during initialization instead of checking every time this method is invoked?

kaylareopelle:
👋  Hi @TonyCTHsu! It looks like there are two comments in the PR blocking further review:
* [Comment #1](https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/899#discussion_r1524085656)
* [Comment #2](https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/899#discussion_r1530460791)

Let us know if you have any follow-up questions!

arielvalentin:
I think we had a similar discussion where we decided to reject a similar PR for PostgreSQL

@robertlaurin suggested using custom samplers instead: 

https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/955

PR #912, opened 2024-03-22
fix: Propagate response attributes on Faraday::Error

CGA1123:
The current `Faraday` middleware won't report `http.status_code` in
cases where an error is raised via another middleware (such as
`:raise_error` or `:json`).

This adds a special case for subclasses of `Faraday::Error` which have
response information available, so that `http.status_code` & the span
status are still added to the span.


arielvalentin:
Thanks for the contribution @CGA1123 !

CGA1123:
Thanks for the _speeeeeedy_ turn-around on this @arielvalentin 🙇‍♂️

PR #915, opened 2024-03-26
fix: instrumentation/active_record: add `:allow_retry` option to `find_by_sql` patch

andrewn617:
Rails 7.2 (rails/rails@eabcff2) introduces the `:allow_retry` option for the `find_by_sql` method, so we need to add it to this patch to maintain compatibility. The patch doesn't use the method's args, so we can just use `(...)` to avoid this problem.

~To check the Rails version, I had to add rails as a development dependency. I added it in the gemspec and set the minimum version to 6.1 following the pattern I saw in the other libraries in this repo ([for example](https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/main/instrumentation/action_pack/opentelemetry-instrumentation-action_pack.gemspec#L37)).~

### DIFF
--- a/instrumentation/active_record/lib/opentelemetry/instrumentation/active_record/patches/querying.rb
+++ b/instrumentation/active_record/lib/opentelemetry/instrumentation/active_record/patches/querying.rb
@@ -18,7 +18,7 @@ module OpenTelemetry
 
           # Contains ActiveRecord::Querying to be patched
           module ClassMethods
-            def find_by_sql(sql, binds = [], preparable: nil, &block)
+            def find_by_sql(...)
               tracer.in_span("#{self}.find_by_sql") do
                 super
               end


dmathieu:
Do we need the condition?
If we provide the new option no matter the rails version, and someone is using a version lower than 7.2, `super` will fail just as it would if the instrumentation wasn't there.

arielvalentin:
> Do we need the condition?
> 
> If we provide the new option no matter the rails version, and someone is using a version lower than 7.2, `super` will fail just as it would if the instrumentation wasn't there.

I think we do. We must honor the monkey patched class method signatures.

Without that condition, the side effect will be changing the public API to add an option that does not exist. Though it will be backward compatible it will confuse end users and if they disable the instrumentation it will break their application.

I am not keen on the if statements and would probably prefer mixing in separate modules depending on the gem versions but I'd like the original author @robertlaurin to weigh in on that.

instrumentation/active_record/lib/opentelemetry/instrumentation/active_record/patches/querying.rb
```diff
@@ -18,9 +18,17 @@ class << base
...
           module ClassMethods
-            def find_by_sql(sql, binds = [], preparable: nil, &block)
-              tracer.in_span("#{self}.find_by_sql") do
-                super
+            if Gem::Version.new(Rails::VERSION::STRING) >= Gem::Version.new("7.2.0.alpha")
+              def find_by_sql(sql, binds = [], preparable: nil, allow_retry: false, &block)
+                tracer.in_span("#{self}.find_by_sql") do
+                  super
+                end
+              end
```
arielvalentin, line 26: On second thought we could capture all args and forward them instead of declaring them since they are not in use right?
```suggestion
              def find_by_sql(...)
                tracer.in_span("#{self}.find_by_sql") do
                  super
                end
              end
```

andrewn617: Done

instrumentation/active_record/lib/opentelemetry/instrumentation/active_record/patches/querying.rb
```diff
@@ -18,7 +18,7 @@ class << base
 
           # Contains ActiveRecord::Querying to be patched
           module ClassMethods
-            def find_by_sql(sql, binds = [], preparable: nil, &block)
+            def find_by_sql(...)
```
arielvalentin, line 21: Are there other methods here that need the same treatment?

kaylareopelle (commented):
Thank you, @andrewn617! Are there existing tests that cover this change?

instrumentation/active_record/lib/opentelemetry/instrumentation/active_record/patches/querying.rb
andrewn617: No it does not look like it.

andrewn617:
> Thank you, @andrewn617! Are there existing tests that cover this change?

@kaylareopelle I think now that we are just using `(...)` there is nothing to test, since it is just ruby functionality to forward all the args to when calling `super`.

arielvalentin (approved):

arielvalentin:
Seems to me like we need to do something similar for these to future proof API changes:

https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/00703640ef59799bfcca0e24dad3e38bb17e1be9/instrumentation/active_record/lib/opentelemetry/instrumentation/active_record/patches/persistence_class_methods.rb

arielvalentin:
@andrewn617 I did miss @kaylareopelle comment about adding code coverage. 

Would you kindly add a test that uses `allow_retry` as a parameter to ensure we do not introduce a regression in the future?

arielvalentin:
hmmm actually this is edge rails so we cannot really at the moment.

PR #925, opened 2024-04-03
fix: fix markdown header

jamesrwhite:
Fixes markdown header rendering in racecar docs


kaylareopelle (approved):
Good eye! Thank you, @jamesrwhite!

kaylareopelle:
@jamesrwhite - Thank you for formatting your commit in the style of conventional commits! 

The PR title is what gets included as the commit in the final changelog, so we need the PR titles to also use conventional commits. Could you update your title to use that style?

PR #927, opened 2024-04-05
test: Temporary fix for missing OpenStruct

arielvalentin:
It looks like a transitive dependency removed `require 'ostruct'`, which results in an `uninitialized Constant` error. 

It may be this change to `rake` that the GraphQL gem was dependent on: https://github.com/ruby/rake/commit/a27bb8e10786efa4776cc3ecfe12490e1ea106a7

This PR temporarily requires `OpenStruct` until the issue is fixed in the Gem: https://github.com/rmosolgo/graphql-ruby/issues/4900


arielvalentin:
aahh looks like a fix was released here https://github.com/rmosolgo/graphql-ruby/pull/4901

PR #933, opened 2024-04-13
chore: Updates HTTP attributes

andyfleming:
Based on https://opentelemetry.io/docs/specs/semconv/http/http-spans/#http-server-semantic-conventions and https://github.com/open-telemetry/semantic-conventions/blob/main/model/trace/http.yaml

(Tests still need updated)

### DIFF
--- a/instrumentation/rack/lib/opentelemetry/instrumentation/rack/middlewares/tracer_middleware.rb
+++ b/instrumentation/rack/lib/opentelemetry/instrumentation/rack/middlewares/tracer_middleware.rb
@@ -124,13 +124,14 @@ module OpenTelemetry
 
           def request_span_attributes(env:)
             attributes = {
-              'http.method' => env['REQUEST_METHOD'],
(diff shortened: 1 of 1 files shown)


dmathieu:
Note that the 1.21.0 semantic conventions recommend the use of an environment variable to make this transition smoother.
https://github.com/open-telemetry/semantic-conventions/blob/v1.21.0/docs/http/README.md

Not all SDKs are following this recommendation (it's a SHOULD, not a MUST). But IMHO, there should be an explicit decision on whether ruby should follow the recommendation or not.

andyfleming:
@dmathieu Hadn't seen that. Thanks! I will take a look at updating it to support configuration via `OTEL_SEMCONV_STABILITY_OPT_IN`.

kaylareopelle:
Hi @andyfleming! Are you still working on this? Would you like me to put a `keep` label on this PR?

PR #954, opened 2024-04-27
fix: Cancel duplicate builds

arielvalentin:
This change cancels any pending builds when someone pushes changes before the tests are complete


robbkidd (approved):
Save the Planet

PR #965, opened 2024-05-07
fix: rack event baggage handling

robertlaurin:
Fixes extracted baggage entries not being attached to the opentelemetry context.

### DIFF
--- a/instrumentation/rack/lib/opentelemetry/instrumentation/rack/middlewares/event_handler.rb
+++ b/instrumentation/rack/lib/opentelemetry/instrumentation/rack/middlewares/event_handler.rb
@@ -42,7 +42,7 @@ module OpenTelemetry
         class EventHandler
           include ::Rack::Events::Abstract
 
-          TOKEN_KEY = 'otel.context.token'
+          RACK_OTEL_CONTEXT_KEY = 'otel.rack.context.token'
           GOOD_HTTP_STATUSES = (100..499)
 
           # Creates a server span for this current request using the incoming parent context
@@ -58,7 +58,7 @@ module OpenTelemetry
             span = create_span(parent_context, request)
             span_ctx = OpenTelemetry::Trace.context_with_span(span, parent_context: parent_context)
             rack_ctx = OpenTelemetry::Instrumentation::Rack.context_with_span(span, parent_context: span_ctx)
-            request.env[TOKEN_KEY] = OpenTelemetry::Context.attach(rack_ctx)
+            request.env[RACK_OTEL_CONTEXT_KEY] = [OpenTelemetry::Context.attach(rack_ctx), rack_ctx]
           rescue StandardError => e
             OpenTelemetry.handle_error(exception: e)
           end
@@ -194,10 +194,11 @@ module OpenTelemetry
           end
 
           def detach_context(request)
-            return nil unless request.env[TOKEN_KEY]
+            return nil unless request.env[RACK_OTEL_CONTEXT_KEY]
 
-            OpenTelemetry::Trace.current_span.finish
-            OpenTelemetry::Context.detach(request.env[TOKEN_KEY])
+            token, rack_ctx = request.env[RACK_OTEL_CONTEXT_KEY]
+            OpenTelemetry::Trace.current_span(rack_ctx).finish
+            OpenTelemetry::Context.detach(token)
           rescue StandardError => e
             OpenTelemetry.handle_error(exception: e)
           end
--- a/instrumentation/rack/lib/opentelemetry/instrumentation/rack/middlewares/event_handler.rb
+++ b/instrumentation/rack/lib/opentelemetry/instrumentation/rack/middlewares/event_handler.rb
@@ -42,7 +42,7 @@ module OpenTelemetry
         class EventHandler
           include ::Rack::Events::Abstract
 
-          RACK_OTEL_CONTEXT_KEY = 'otel.rack.context.token'
+          OTEL_TOKEN_AND_SPAN = 'otel.rack.token_and_span'
           GOOD_HTTP_STATUSES = (100..499)
 
           # Creates a server span for this current request using the incoming parent context
@@ -58,7 +58,7 @@ module OpenTelemetry
             span = create_span(parent_context, request)
             span_ctx = OpenTelemetry::Trace.context_with_span(span, parent_context: parent_context)
             rack_ctx = OpenTelemetry::Instrumentation::Rack.context_with_span(span, parent_context: span_ctx)
-            request.env[RACK_OTEL_CONTEXT_KEY] = [OpenTelemetry::Context.attach(rack_ctx), rack_ctx]
+            request.env[OTEL_TOKEN_AND_SPAN] = [OpenTelemetry::Context.attach(rack_ctx), span]
           rescue StandardError => e
             OpenTelemetry.handle_error(exception: e)
           end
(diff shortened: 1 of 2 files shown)


instrumentation/rack/lib/opentelemetry/instrumentation/rack/middlewares/event_handler.rb
```diff
@@ -56,7 +56,9 @@ def on_start(request, _)
 
             parent_context = extract_remote_context(request)
             span = create_span(parent_context, request)
-            request.env[TOKENS_KEY] = register_current_span(span)
+            span_ctx = OpenTelemetry::Trace.context_with_span(span, parent_context: parent_context)
+            rack_ctx = OpenTelemetry::Instrumentation::Rack.context_with_span(span, parent_context: span_ctx)
+            request.env[TOKEN_KEY] = OpenTelemetry::Context.attach(rack_ctx)
```
robertlaurin, line 61: We don't need to attach multiple context objects just a single context that captures:
- The extracted context (this is where we get our baggage)
- The current span context
- The current rack span context

The `*.context_with_span` calls make use of the [set_value](https://github.com/open-telemetry/opentelemetry-ruby/blob/main/api/lib/opentelemetry/context.rb#L148-L157) context interface which always returns a new ctx object.

So instead of layering contexts, we're building a single one here that contains all the entries needed when we start a rack request.

robertlaurin:
To fix the untraced endpoint we'll need to update the utilities helper to allow for a [non block structured](https://github.com/open-telemetry/opentelemetry-ruby/blob/main/common/lib/opentelemetry/common/utilities.rb#L90-L100) way of setting the untraced ctx, as the key for the entry [is currently private.](https://github.com/open-telemetry/opentelemetry-ruby/blob/main/common/lib/opentelemetry/common/utilities.rb#L13-L14)

instrumentation/rack/lib/opentelemetry/instrumentation/rack/middlewares/event_handler.rb
```diff
@@ -191,11 +193,11 @@ def request_span_attributes(env)
...
 
-          def detach_contexts(request)
-            request.env[TOKENS_KEY]&.reverse_each do |token|
-              OpenTelemetry::Context.detach(token)
-              OpenTelemetry::Trace.current_span.finish
-            end
+          def detach_context(request)
+            return nil unless request.env[TOKEN_KEY]
+
+            OpenTelemetry::Trace.current_span.finish
```
fbogsany, line 199: The lack of symmetry (with `on_start`'s use of `Foo.context_with_span`) threw me quite a bit here. Nothing to be done - just noting my confusion. Basically down to `on_start` managing contexts explicitly, whereas here we use the implicit "current context".

robertlaurin: I wonder if it would make sense to store a reference to the rack span in the rack request env so we can close it later, rather than relying on in the implicit `current_span`. 

With the current approach we're vulnerable to a scenario where some misconfigured or buggy instrumentation opening a span, setting the context, and not closing it... right? It's probably not likely, but still plausible.

instrumentation/rack/lib/opentelemetry/instrumentation/rack/middlewares/event_handler.rb
arielvalentin: TIL

instrumentation/rack/lib/opentelemetry/instrumentation/rack/middlewares/event_handler.rb
arielvalentin: If the Rack span is not the current span at this point then it means the context token is out of sync also and we will likely get a detach error. 

If this tracked the token and the `Context` together as a tuple then it would fetch the Rack span from it instead and ensure it is closing the right span and the token/Context would be the same.

If something went wrong with a different implicit "current" context not being `detached`, it would still result in a detach error but at least the Rack span would be the one we are trying to finish. 
e.g. 

```ruby
return nil unless request.env[RACK_OTEL_CONTEXT_KEY]

[token, rack_ctx] = request.env[RACK_OTEL_CONTEXT_KEY]

OpenTelemetry::Trace.current_span(rack_ctx).finish
OpenTelemetry::Context.detach(token)
```

robertlaurin: I think that makes sense to me, see https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/965/commits/0b666b995053662ad8c72a462478f6defa20a75c

arielvalentin (approved):

instrumentation/rack/lib/opentelemetry/instrumentation/rack/middlewares/event_handler.rb
```diff
@@ -56,7 +56,9 @@ def on_start(request, _)
 
             parent_context = extract_remote_context(request)
             span = create_span(parent_context, request)
-            request.env[TOKENS_KEY] = register_current_span(span)
+            span_ctx = OpenTelemetry::Trace.context_with_span(span, parent_context: parent_context)
+            rack_ctx = OpenTelemetry::Instrumentation::Rack.context_with_span(span, parent_context: span_ctx)
+            request.env[RACK_OTEL_CONTEXT_KEY] = [OpenTelemetry::Context.attach(rack_ctx), rack_ctx]
```
fbogsany, line 61: Why do we need both the token and the context here? 👀 Oh, because you need to fetch the span from the context to finish it, and use the token to detach. Can we either document the content of `request.env[RACK_OTEL_CONTEXT_KEY]` or rename the key to something that better describes the content?

fbogsany: Naive question: do we need the context? Could we just use the span? I.e.
```ruby
            rack_ctx = OpenTelemetry::Instrumentation::Rack.context_with_span(span, parent_context: parent_context)
            token = OpenTelemetry::Context.attach(rack_ctx)
            request.env[OTEL_TOKEN_AND_SPAN] = [token, span]
```

instrumentation/rack/lib/opentelemetry/instrumentation/rack/middlewares/event_handler.rb
```diff
@@ -191,11 +193,12 @@ def request_span_attributes(env)
...
-            request.env[TOKENS_KEY]&.reverse_each do |token|
-              OpenTelemetry::Context.detach(token)
-              OpenTelemetry::Trace.current_span.finish
-            end
+          def detach_context(request)
+            return nil unless request.env[RACK_OTEL_CONTEXT_KEY]
+
+            token, rack_ctx = request.env[RACK_OTEL_CONTEXT_KEY]
+            OpenTelemetry::Trace.current_span(rack_ctx).finish
+            OpenTelemetry::Context.detach(token)
```
fbogsany, line 201: If we store the token and span instead:
```suggestion
            return nil unless request.env[OTEL_TOKEN_AND_SPAN]

            token, span = request.env[OTEL_TOKEN_AND_SPAN]
            span.finish
            OpenTelemetry::Context.detach(token)
```

instrumentation/rack/lib/opentelemetry/instrumentation/rack/middlewares/event_handler.rb
robertlaurin: We can just use the span https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/965/commits/dfcbfb257e2c59965e2f0ee191c523fe019cf0bc

instrumentation/rack/lib/opentelemetry/instrumentation/rack/middlewares/event_handler.rb
robertlaurin: 👍 https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/965/commits/dfcbfb257e2c59965e2f0ee191c523fe019cf0bc

fbogsany (approved):
👍 LGTM

instrumentation/rack/lib/opentelemetry/instrumentation/rack/middlewares/event_handler.rb
arielvalentin: My suggestion for using the context may be speculative generality based on the rest of the OTel API always wanting or requiring a context (usually defaulting to `current`). 

No objections from me. If we need the context in the future we can change this.

Issue #962, opened 2024-05-03
[inst-rack] rack_events + untraced_endpoints will still emit child spans of those endpoint


robbkidd:
**Description of the bug**

There's a difference in behavior for untraced endpoints when using middleware vs the newer implementation using Rack events.

* middleware: wraps call with the untraced helper from Utilities which sets untraced in the context used by child spans
* rack events: early returns from Rack's on_start and on_end handlers but does not change the context, so child spans get traced

I owe more details here! Opening as a placeholder for me to add more later.

robertlaurin:
https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/965#issuecomment-2099275533

robertlaurin:
https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/968

PR #968, opened 2024-05-08
fix: Untrace entire request

robertlaurin:
Depends on common 0.21.0 for https://github.com/open-telemetry/opentelemetry-ruby/pull/1634
closes: https://github.com/open-telemetry/opentelemetry-ruby-contrib/issues/962

### DIFF
--- a/instrumentation/rack/lib/opentelemetry/instrumentation/rack/middlewares/event_handler.rb
+++ b/instrumentation/rack/lib/opentelemetry/instrumentation/rack/middlewares/event_handler.rb
@@ -43,7 +43,6 @@ module OpenTelemetry
           include ::Rack::Events::Abstract
 
           OTEL_TOKEN_AND_SPAN = 'otel.rack.token_and_span'
-          UNTRACED_CTX_TOKEN = 'otel.rack.untraced_ctx_token'
           GOOD_HTTP_STATUSES = (100..499)
 
           # Creates a server span for this current request using the incoming parent context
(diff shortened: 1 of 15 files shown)


instrumentation/rack/lib/opentelemetry/instrumentation/rack/middlewares/event_handler.rb
```diff
@@ -52,7 +53,11 @@ class EventHandler
           # @param [Rack::Response] This is nil in practice
           # @return [void]
           def on_start(request, _)
-            return if untraced_request?(request.env)
+            if untraced_request?(request.env)
+              ctx = OpenTelemetry::Common::Utilities.untraced
+              request.env[UNTRACED_CTX_TOKEN] = OpenTelemetry::Context.attach(ctx)
```
fbogsany, line 58: Why do we need another context (attached) and another token for this? We should add a `context = OpenTelemetry::Context.current` parameter to `extract_remote_context` and pass in the untraced context if necessary. I.e.:
```ruby
parent_context = if untraced_request?(request.env)
    extract_remote_context(request, OpenTelemetry::Common::Utilities.untraced)
  else
    extract_remote_context(request)
  end
```

fbogsany (approved):
LGTM 👍

instrumentation/rack/lib/opentelemetry/instrumentation/rack/middlewares/event_handler.rb
```diff
@@ -52,9 +52,12 @@ class EventHandler
           # @param [Rack::Response] This is nil in practice
           # @return [void]
           def on_start(request, _)
-            return if untraced_request?(request.env)
+            parent_context = if untraced_request?(request.env)
```
arielvalentin, line 55: Is this relying on start_span to look for the untraced context as opposed to seething the current context to untraced without an active span?

robertlaurin: Yeah it's seeing that it's parent context is untraced so it will create a non recording span, and we have a lot of `recording?` checks in this instrumentation as is so it seemed like a simpler implementation than what I had in this commit https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/968/commits/a8548efd6ba065ce27da5bcf9d21504efa40b3c1

arielvalentin: Understood. Thanks again for fixing this!

arielvalentin (approved):

robertlaurin:
Bumping all the common deps here https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/968/commits/c8144bc3783a732a85424d6aa64c2b58e5ea9be5

PR #983, opened 2024-05-14
feat: Ruby Logger instrumentation 

### LINKED ISSUE #668: Add instrumentation for Ruby Logger leveraging Log Records
The [opentelemetry-ruby repo](https://github.com/open-telemetry/opentelemetry-ruby/projects/3) has WIP logs signal support. Leveraging this work, add instrumentation for Ruby's Logger library to create Log Records whenever Ruby's logger emits a log.

Example from New Relic: https://github.com/newrelic/newrelic-ruby-agent/blob/dev/lib/new_relic/agent/instrumentation/logger/instrumentation.rb 

This should include the ability to exclude certain loggers from instrumentation. If we instrument Ruby's logger, we may want to exclude ActiveSupport Logger to prevent double recording, as seen in [New Relic's instrumentation.](https://github.com/newrelic/newrelic-ruby-agent/blob/dev/lib/new_relic/agent/instrumentation/active_support_logger/instrumentation.rb)

kaylareopelle:
### Description
This an [OpenTelemetry logs bridge](https://opentelemetry.io/docs/specs/otel/logs/bridge-api/) for Ruby's standard [Logger library](https://github.com/ruby/logger).

It also includes patches to `ActiveSupport::Logger.broadcast` and the `ActiveSupport::BroadcastLogger` to emit only one log record for a broadcast.

@khushijain21 is a co-author of this PR and contributed functionality as part of her [LFX mentorship](https://lfx.linuxfoundation.org/tools/mentorship/) with OpenTelemetry in 2024.

Closes #668


instrumentation/all/lib/opentelemetry/instrumentation/all.rb
```diff
@@ -25,7 +23,9 @@
 require 'opentelemetry-instrumentation-faraday'
 require 'opentelemetry-instrumentation-grape'
 require 'opentelemetry-instrumentation-graphql'
+require 'opentelemetry-instrumentation-gruf'
 require 'opentelemetry-instrumentation-http_client'
+require 'opentelemetry-instrumentation-logger'
```
kaylareopelle, line 28: Should the Logger bridge be part of `instrumentation/all`? I didn't realize the difference between a "bridge" and an "instrumentation" until after this had been worked out. We could create a separate directory at a peer level with instrumentation named "bridge(s?)" and put the code there. 

If someone knows of an "instrumentation" definition within OTel documentation, that might help.

I'm sure there's other pros/cons, but this is what I have off the top of my head:

**Pros to having it in instrumentation/all:** 
* Easier for people to adopt if they're already installing their instrumentation via all
* One less CI workflow to create, along with all the other structure required for another grouping (ex. `processor`, `propagator`)


**Cons**: 
* The specs reference a ["Bridge API"](https://opentelemetry.io/docs/specs/otel/logs/bridge-api/), so we could be breaking from that concept by grouping this with instrumentation

instrumentation/logger/Appraisals
```diff
@@ -0,0 +1,18 @@
...
+#   gem 'rails', '~> 6.0.0'
+# end
+
+# appraise 'rails-6.1' do
+#   gem 'rails', '~> 6.1.0'
+# end
+
+# appraise 'rails-7.0' do
+#   gem 'rails', '~> 7.0.0'
+# end
```
kaylareopelle, line 18: I haven't been able to get Appraisal to work with gems installed from git source. This might just be due to my limited knowledge of Appraisal. We may need to wait until the opentelemetry-logs-related gems are released before we can re-instate these.

instrumentation/logger/Gemfile
```diff
@@ -0,0 +1,22 @@
...
+# then target specific versions in the Appraisals file.
+
+gemspec
+
+group :test do
+  gem 'opentelemetry-instrumentation-base', path: '../base'
+  gem 'opentelemetry-api', git: 'https://github.com/kaylareopelle/opentelemetry-ruby', branch: 'log-record-processor3', glob: 'api/*.gemspec'
+  gem 'opentelemetry-logs-api', git: 'https://github.com/kaylareopelle/opentelemetry-ruby', branch: 'log-record-processor3', glob: 'logs_api/*.gemspec'
+  gem 'opentelemetry-logs-sdk', git: 'https://github.com/kaylareopelle/opentelemetry-ruby', branch: 'log-record-processor3', glob: 'logs_sdk/*.gemspec'
+  gem 'opentelemetry-sdk', git: 'https://github.com/kaylareopelle/opentelemetry-ruby', branch: 'log-record-processor3', glob: 'sdk/*.gemspec'
```
kaylareopelle, line 21: Current workaround to test against the WIP SDK/etc. code.

instrumentation/logger/lib/opentelemetry/instrumentation.rb
```diff
@@ -0,0 +1,19 @@
...
+# SPDX-License-Identifier: Apache-2.0
+
+# OpenTelemetry is an open source observability framework, providing a
+# general-purpose API, SDK, and related tools required for the instrumentation
+# of cloud-native software, frameworks, and libraries.
+#
+# The OpenTelemetry module provides global accessors for telemetry objects.
+# See the documentation for the `opentelemetry-api` gem for details.
+module OpenTelemetry
+  # Instrumentation should be able to handle the case when the library is not installed on a user's system.
```
kaylareopelle, line 14:
```suggestion
```

instrumentation/logger/lib/opentelemetry/instrumentation/logger/patches/active_support_logger.rb
```diff
@@ -0,0 +1,24 @@
...
+
+# Copyright The OpenTelemetry Authors
+#
+# SPDX-License-Identifier: Apache-2.0
+
+module OpenTelemetry
+  module Instrumentation
+    module Logger
+      module Patches
+        # Patches for the ActiveSupport::Logger class included in Rails
```
kaylareopelle, line 11: What's this all about? Well, when this instrumentation is used with Rails, we get double the logs emitted because of how Rails broadcasts the logs. This workaround is similar to [what we did in New Relic's Ruby agent.](https://github.com/newrelic/newrelic-ruby-agent/pull/1019) 

I'm open to feedback and better solutions.

An additional workaround still needs to be developed for the [Rails 7.1 broadcast logger](https://github.com/newrelic/newrelic-ruby-agent/pull/2252).

instrumentation/logger/lib/opentelemetry/instrumentation/logger/patches/logger.rb
```diff
@@ -0,0 +1,60 @@
...
+            return formatted_message if skip_instrumenting?
+
+            OpenTelemetry.logger_provider.logger(
+              name: OpenTelemetry::Instrumentation::Logger::NAME,
+              version: OpenTelemetry::Instrumentation::Logger::VERSION
+            ).on_emit(
+              severity_text: severity,
+              severity_number: severity_number(severity),
+              timestamp: datetime,
+              body: msg # New Relic uses formatted_message here. This also helps us with not recording progname, because it is included in the formatted message by default. Which seems more appropriate?
```
kaylareopelle, line 26: Formatted message is what you're used to seeing in stdout or your logs. By default, it includes severity, timestamp, process.pid, and progname in addition to the message.

I don't see a clear answer from the [Logs Data Model entry on `body`](https://opentelemetry.io/docs/specs/otel/logs/data-model/#field-body).

instrumentation/all/lib/opentelemetry/instrumentation/all.rb
xuan-cao-swi: I think it's ok to include it in all since it has the standard instrumentation structure. Although, instrumentation `present do ... end` that check if the instrumentation is little bit loose in my opinion. Maybe also check if `::OpenTelemetry::SDK::Logs` present?

instrumentation/logger/Appraisals
xuan-cao-swi: Have you tried something like:
```
gem 'opentelemetry-logs-api', github: 'open-telemetry/opentelemetry-ruby', branch: 'main', glob: 'logs_api/opentelemetry-logs-api.gemspec'
gem 'opentelemetry-logs-sdk', github: 'open-telemetry/opentelemetry-ruby', branch: 'main', glob: 'logs_sdk/opentelemetry-logs-sdk.gemspec'
```

instrumentation/logger/lib/opentelemetry/instrumentation/logger/instrumentation.rb
```diff
@@ -0,0 +1,41 @@
...
+    module Logger
+      # The Instrumentation class contains logic to detect and install the Logger instrumentation
+      class Instrumentation < OpenTelemetry::Instrumentation::Base
+        install do |_config|
+          require_dependencies
+          patch
+        end
+
+        present do
+          defined?(::Logger)
```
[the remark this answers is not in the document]
khushijain21: @xuan-cao-swi accommodated this feedback

instrumentation/logger/lib/opentelemetry/instrumentation/logger/patches/logger.rb
```diff
@@ -0,0 +1,58 @@
...
+            return formatted_message if skip_instrumenting?
+
+            OpenTelemetry.logger_provider.logger(
+              name: OpenTelemetry::Instrumentation::Logger::NAME,
+              version: OpenTelemetry::Instrumentation::Logger::VERSION
+            ).on_emit(
+              severity_text: severity,
+              severity_number: severity_number(severity),
+              timestamp: datetime,
+              body: msg,
```
kaylareopelle, line 26: New Relic uses `formatted_message` here. This includes both the progname and the message in the body, whereas `msg` does not include the progname. Which seems more appropriate?

tomash:
I am very interested in getting this to work with Opentelemetry. Any way I could help with this work?

kaylareopelle:
Hi @tomash! Thanks for reaching out. It's great to hear there's interest in this PR. I'd love your help!

The features for the bridge should be complete, but we need to agree on the approach and get more feedback overall.

Specifically, I could use help with: 

1. **Reviewing the code:** What are your thoughts on how OpenTelemetry logs are bridged from the Ruby Logger? Do you have suggestions on how we could improve the approach? An alternative approach has been presented in https://github.com/open-telemetry/opentelemetry-ruby/discussions/1789, would you rather see something like this? 

2. **Testing the code in an app:** Install the gem from this branch and provide feedback on how it works in your application. This is [a Gemfile](https://github.com/kaylareopelle/otel_ruby_logs_demo/blob/main/Gemfile#L31-L45) from a sample application I wrote to add the OpenTelemetry bridge for the Ruby logger to a Rails application. Give it a try and let us know how it goes.

tomash:
@kaylareopelle 

For now I have this monstrosity in Gemfile:
```
gem "opentelemetry-instrumentation-logger",
  git: "https://github.com/kaylareopelle/opentelemetry-ruby-contrib.git",
  branch: "logger-instrumentation",
  glob: "instrumentation/logger/opentelemetry-instrumentation-logger.gemspec"
  ```
  
And I can confirm it works well for our app (the logs appear in Honeycomb just as events). The only problem is that Rails logs are noisy by default and it makes "top messages" unusable (top message being "rendered profiles/_carousel.html.erb"), so I've also dropped in Lograge to produce fewer logs.

I haven't seen https://github.com/open-telemetry/opentelemetry-ruby/discussions/1789 , will evaluate and compare! I do like how lightweight and easy to read it is.

**Comment by tomash** [comment]
So I'm after a few weeks of running this PR in production for a medium-sized app with a lot of traffic, together with Lograge. Works beautifully.

I've also evaluated https://github.com/open-telemetry/opentelemetry-ruby/discussions/1789 but we have a specific setup -- using both Newrelic and Honeycomb -- and that one-screener did not produce logs to both outputs (so maybe it was intercepting too hard). It also does not use `OTEL_EXPORTER_OTLP_HEADERS` so this PR is our choice.

The only flaw I found in this one is a very specific case.
1. Due to our traffic we do head-based sampling with `OTEL_TRACES_SAMPLER=traceidratio` and `OTEL_TRACES_SAMPLER_ARG=0.01` which works great for sampling our traces.
2. Of course we **do not want log sampling** and it's all good, 100% of logs are sent out with this PR. 
3. **BUT** because of that sampling, to have proper counts and reflect trace representation, we have `OTEL_RESOURCE_ATTRIBUTES="SampleRate=100"` (a reverse of 0.01 sampling fraction).
4. This SampleRate attribute seems to be also attached to log events because for each log entry Honeycomb multiplies its occurence count by 100.

**Comment by fguillen** [comment]
Funny behaviour. If I have activated `ENV["OTEL_LOG_LEVEL"] ="debug"`. When a log is exported another DEBUG log is generated like:

```
D, [2025-05-01T22:42:17.912579 #95172] DEBUG -- : Successfully exported 1 log records
```

And it gets into an infinite loop. It may be expected. But just for you to take into consideration :)

**Comment by tomash** [comment]
@fguillen I was about to write an answer but you figured it all out in the meantime!

Sharing my slightly-different lograge.rb initializer:
```ruby

Rails.application.configure do
  config.lograge.enabled = true

  # the hotwire connect-disconnect logs are just noise
  config.lograge.ignore_actions = [
    "Turbo::StreamsChannel#subscribe",
    "Turbo::StreamsChannel#unsubscribe",
    "Hotwire::Livereload::ReloadChannel#subscribe",
    "Hotwire::Livereload::ReloadChannel#unsubscribe",
    "ApplicationCable::Connection#connect",
    "ApplicationCable::Connection#disconnect",
    "ApplicationCable::Connection#reconnect",
    "ActionCable::Connection::Base#connect",
    "ActionCable::Connection::Base#disconnect",
  ]
  config.lograge.custom_payload do |controller|
    {
      params: controller.request.filtered_parameters,
    }
  end
end
```

In opentelemetry.rb initializer we also have this line, as I saw it in this PR's example code:
```ruby
  at_exit do
    OpenTelemetry.logger_provider.shutdown
  end
```

fguillen:
hi @tomash:

> Why do you use OTEL_LOG_LEVEL instead of application-global log level?

As explained in a [previous comment](https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/983#issuecomment-2845734587), I can not set `OTEL_LOG_LEVEL=debug` and also activate `OTEL_LOGS_EXPORTER`. Because the system enters an infinite loop of noticing the debug log that has been sent, and by this, generating another debug log, and so on.

fguillen:
I am having an issue when loading the logs in Grafana (Rails -> OtelCollector -> Loki -> Grafana). I don't know if it is related to this logger plugin or by my configuration or it is the expected behaviour.

The case is that in Grafana I see the logs like this:

```
{
  "body": "{\"method\":\"GET\",\"path\":\"/admin/invitations\",\"format\":\"html\",\"controller\":\"Admin::InvitationsController\",\"action\":\"index\",\"status\":200,\"allocations\":847384,\"duration\":470.19,\"view\":351.69,\"db\":86.12,\"trace_id\":\"7d5334a61dc66894a9c6ef508529de6e\",\"span_id\":\"c02fc38d3c0b7580\",\"user_id\":\"257b4cba-d430-4db7-9406-fdd08c495571\",\"host\":\"localhost\",\"remote_ip\":null,\"params\":{},\"environment\":\"development\",\"request_id\":null}",
  "flags": 1,
  "instrumentation_scope": {
    "name": "opentelemetry-instrumentation-logger",
    "version": "0.1.0"
  },
  "resources": {
    "process.command": "bin/rails",
    "process.pid": 81016,
    "process.runtime.description": "ruby 3.2.4 (2024-04-23 revision af471c0e01) [arm64-darwin24]",
    "process.runtime.name": "ruby",
    "process.runtime.version": "3.2.4",
    "service.name": "app.playcocola.com",
    "telemetry.sdk.language": "ruby",
    "telemetry.sdk.name": "opentelemetry",
    "telemetry.sdk.version": "1.8.0"
  },
  "severity": "INFO",
  "spanid": "c02fc38d3c0b7580",
  "traceid": "7d5334a61dc66894a9c6ef508529de6e"
}
```

The message has been created in a proper JSON format, but when it arrives at Grafana, it is in the `body` value and it is stringified, and I can not do proper queries on it. 

I have tried many different `lograge` configurations but have had no luck. The `body` is always stringified.

kaylareopelle:
@xuan-cao-swi - I've looked into move the logger instrumentation into a separate `bridge` directory and have some concerns. I'd like to get your feedback.

1. If we move out of the `instrumentation` directory, should we still allow this instrumentation to depend on the `opentelemetry-instrumentation-base` gem? Currently, the logger instrumentation leverages the install/present/compatible methods provided by this gem. Eventually, we may want to add configuration options too, which would also use these APIs. Right now, nothing outside the `instrumentation` directory depends on the `base` gem. 

2. If we move the logger logic into bridges, should we have a separate registry for bridges? Personally, I'd prefer to have one registry that holds all the possible prepended modules. 

3. We'll eventually need to include creating a `logger` into instrumentation-base for AI events since it seems like they will be logs with the event_name attribute. This also reveals that an unnecessary tracer is being instantiated for the logger instrumentation because it depends on the `instrumentation-base` gem, but we could update the internals to make creating a tracer conditional. 

At the moment, I'd like to keep the logger instrumentation out of `all` at least until the logs implementation is stable, which would add some barriers to installing it accidentally.

wsmoak:
I'm interested in trying this out... if you have time could you please merge in the latest changes?

instrumentation/logger/lib/opentelemetry/instrumentation/logger/patches/logger.rb
kaylareopelle: We're going to leave this as-is for now, but can consider changing it if people want the progname/expect something different from the message.

ericmustin:
lgtm once the branch conflicts r resolved

xuan-cao-swi (approved):

robbkidd (approved):
It's GOOD!

PR #984, opened 2024-05-14
fix: add action_mailer to rails and all

xuan-cao-swi:
This PR resolves the action_mailer leftover from https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/887.
Related issue: https://github.com/open-telemetry/opentelemetry-ruby-contrib/issues/293.


instrumentation/all/lib/opentelemetry/instrumentation/all.rb
```diff
@@ -11,6 +11,7 @@
 require 'opentelemetry-instrumentation-active_job'
 require 'opentelemetry-instrumentation-active_record'
 require 'opentelemetry-instrumentation-action_view'
+require 'opentelemetry-instrumentation-action_mailer'
```
kaylareopelle, line 14: I believe this also needs to be added as a dependency in the "instrumentation-all" gem's gemspec.

xuan-cao-swi: I saw all gem's gemspec only have `spec.add_dependency 'opentelemetry-instrumentation-rails', '~> 0.30.0'`, without other rails components' instrumentation.

kaylareopelle (approved):
Thanks, @xuan-cao-swi!

PR #986, opened 2024-05-14
fix: add aws lambda to instrumentation-all

xuan-cao-swi:
This PR opens as placeholder to discuss if aws_lambda instrumentation should be included in instrumentation-all.

aws_lambda instrumentation ideally should only exist in lambda layer (see [PR 1323](https://github.com/open-telemetry/opentelemetry-lambda/pull/1323/files#diff-bf01ee00ac546ec2297efd6e71961f5d546c5b10837a1d9c31261b3773052591R5)). 

Reference from [js auto-instrumentation](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/main/metapackages/auto-instrumentations-node/src/utils.ts#L21) and [py auto-instrumentation](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/opentelemetry-contrib-instrumentations/pyproject.toml#L38).


xuan-cao-swi:
> Will the instrumentation install outside a lambda environment?

No, ideally, it should only exist in lambda env.

> If we include it as part of all does that make it easier for lambda users to access other instrumentation within their lambdas?

I haven't think about instrumentation other library inside lambda env. In my understanding, lambda env should only produce one span (not 100% certain on this).

mwear (approved):
Given that other SIGs do the same, and that this makes it easier for lambda users to get started by using instrumentation-all, I'm in favor.

PR #988, opened 2024-05-15
feat: add option for untraced commands

zacheryph:
solution to https://github.com/open-telemetry/opentelemetry-ruby-contrib/discussions/264

Adds the ability to not trace specified commands. This is in the same fashion that Rack supports with `:untraced_endpoints`, and net_http with `:untraced_hosts`

### DIFF
--- a/instrumentation/pg/README.md
+++ b/instrumentation/pg/README.md
@@ -52,8 +52,13 @@ OpenTelemetry::SDK.configure do |c|
 
     # By default, this instrumentation obfuscate/sanitize the executed SQL as the `db.statement`
     # semantic attribute. Optionally, you may disable the inclusion of this attribute entirely by
-    # setting this option to :omit or disbale sanitization the attribute by setting to :include
+    # setting this option to :omit or disable sanitization the attribute by setting to :include
     db_statement: :include,
+
+    # Add certain commands that you do not want to produce spans within traces. This can
+    # be useful if you find something like PREPARE statements too noisy in your trace data.
+    # note: these must be uppercase to match the command.
+    untraced_commands: %w[PREPARE DEALLOCATE],
   }
 end
 ```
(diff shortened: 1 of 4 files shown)


robbkidd:
The OTel Ruby contrib maintainers discussed this change at our SIG meeting last week. We are a little wary of adding more configuration surface area to the instrumentation. This isn't a "no" but a "not right now." We're looking into how much configuration is being supported in other language's common instrumentation libraries and how the new OTel config file format affects how we provide these knobs.

In the meantime, there is another option for folks willing to run an OpenTelemetry Collector. The contrib build of the Collector includes a filterprocessor that can drop spans from being sent along if they match criteria added to the collector config. Because these PG client spans are likely leaves in the trace tree (in other words, they don't tend to have descendant child spans), their absence in a trace backend won't result in broken traces.

The following snippet would configure a contrib Collector to drop spans coming from the OTel Ruby PG instrumentation for DEALLOCATE and PREPARE operations. 

```yaml
receivers:
  otlp:
    protocols:
      http:

processors:
  batch:
  filter/ignorethesedboperations:
    error_mode: ignore
    traces:
      span:
        - 'instrumentation_scope.name == "OpenTelemetry::Instrumentation::PG" and attributes["db.operation"] == "DEALLOCATE"'
        - 'instrumentation_scope.name == "OpenTelemetry::Instrumentation::PG" and attributes["db.operation"] == "PREPARE"'

exporters:
  otlp:
    endpoint: # wherever you're sending your data

service:
  pipelines:
    traces:
      receivers: [otlp]
      processors: [filter/ignorethesedboperations, batch]
      exporters: [otlp]
```

igor-rounded:
@robbkidd I'm having trouble locating the filtering processor in the contrib repository, can you please link the relevant code?

PR #990, opened 2024-05-16
feat!: Add baggage key predicate func to baggage span processor

### LINKED ISSUE #956: Baggage span processor - key predicate
This issue is to track adding a method of selecting what baggage key entries should be copied.

Feedback in the JS contrib PR was to allow a user-provided predicate function. This puts the responsibility on the user to ensure sensitive baggage keys are not copied while also not prescribing how that is determined.
- https://github.com/open-telemetry/opentelemetry-js-contrib/issues/2166

              > What about making the baggage items that are added as spans tags configurable, with the option to provide `*` for all - like here: https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/355003538a4dcf7ec58b6d9be42086c53238bdac/instrumentation/log4j/log4j-appender-2.17/library/README.md?plain=1#L101

We had a similar feedback in the .NET contrib project but thought it was more complicated than just using a set of prefixes so created an issue to continue the discussion. The plain processor that copies all baggage entries (like using `*` in your example) is likely to be accepted first.
- https://github.com/open-telemetry/opentelemetry-dotnet-contrib/issues/1695

MikeGoldsmith:
## Which problem is this PR solving?
- Closes #956 

## Short description of the changes
- Add baggage key lambda as constructor parameter to baggage span processor
- Add `ALLOW_ALL_BAGGAGE_KEYS` lambda that allows all keys
- Add unit test to verify behaviour
- Update README with examples of setting a filter

### DIFF
--- a/processor/baggage/lib/opentelemetry/processor/baggage/baggage_span_processor.rb
+++ b/processor/baggage/lib/opentelemetry/processor/baggage/baggage_span_processor.rb
@@ -42,6 +42,14 @@ module OpenTelemetry
       #     )
       #   end
       class BaggageSpanProcessor < OpenTelemetry::SDK::Trace::SpanProcessor
+        # Create a new BaggageSpanProcessor
+        # @param [Proc] keyfilter A proc that takes a key and returns true if the key should be added to the span.
+        # Default keyfilter adds all baggage entries to the span.
+        def initialize(keyfilter: nil)
+          @keyfilter = keyfilter || ->(_key) { true }
+          super
+        end
+
         # Called when a `Span` is started, adds Baggage keys/values to the span as attributes.
         #
         # @param [Span] span the `Span` that just started, expected to conform
@@ -51,7 +59,11 @@ module OpenTelemetry
         def on_start(span, parent_context)
           return unless span.respond_to?(:add_attributes) && parent_context.is_a?(::OpenTelemetry::Context)
 
-          span.add_attributes(::OpenTelemetry::Baggage.values(context: parent_context))
+          span.add_attributes(
+            ::OpenTelemetry::Baggage
+              .values(context: parent_context)
+              .select { |k, _v| @keyfilter.call(k) }
+          )
         end
 
         # Called when a Span is ended, does nothing.
--- a/processor/baggage/lib/opentelemetry/processor/baggage/baggage_span_processor.rb
+++ b/processor/baggage/lib/opentelemetry/processor/baggage/baggage_span_processor.rb
@@ -10,8 +10,8 @@ require 'opentelemetry-sdk'
 module OpenTelemetry
   module Processor
     module Baggage
-      # A regex that matches all Baggage keys.
-      ALLOW_ALL = '*'
+      # A baggage key predicate that allows all keys to be added to the span as attributes.
+      ALLOW_ALL_BAGGAGE_KEYS = ->(_) { true }
 
       # The BaggageSpanProcessor reads key/values stored in Baggage in the
       # starting span's parent context and adds them as attributes to the span.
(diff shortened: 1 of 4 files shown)


MikeGoldsmith:
Ping @robbkidd @arielvalentin - please could you take a look?

processor/baggage/lib/opentelemetry/processor/baggage/baggage_span_processor.rb
```diff
@@ -42,6 +47,16 @@ module Baggage
       #     )
       #   end
       class BaggageSpanProcessor < OpenTelemetry::SDK::Trace::SpanProcessor
+        # Create a new BaggageSpanProcessor that reads Baggage keys and values from the parent context
+        # and adds them as attributes to the span.
+        #
+        # @param [lambda] baggage_key_predicate A lambda that takes a baggage key and returns true if
+        #  the key should be added to the span as an attribute, false otherwise.
+        def initialize(baggage_key_predicate)
+          @baggage_key_predicate = baggage_key_predicate
```
xuan-cao-swi, line 56: NIT: could `ALLOW_ALL_BAGGAGE_KEYS` be default for `@baggage_key_predicate`?

e.g.
```
def initialize(baggage_key_predicate: nil)
    @baggage_key_predicate = baggage_key_predicate || ALLOW_ALL_BAGGAGE_KEYS
    ...
end
```

MikeGoldsmith: I went with a required user parameter to make it super clear to users they need to explicitly allow baggage keys they want. If we used the `ALLOW_ALL_BAGGAGE_KEYS ` predicate as the default value, it would be easy for users to unknowing include copy all baggage entires by default.

Using a required parameter helps make it more secure by default.

robbkidd: I agree with the opt-in behavior of needing to specify a predicate method to the constructor.

robbkidd (approved):

MikeGoldsmith:
Thanks for the updates @robbkidd - much appreciated.

arielvalentin (commented):
Are there any cases where endusers may want to filter based on value as opposed to key alone?

If I were writing my own predicate I would likely want to reach for an implementation that uses [`Hash#slice`](https://ruby-doc.org/core-3.0.0/Hash.html#method-i-slice) and have an explicit key list as well as potentially scrub attributes that I may be concerned has PII in it. I wouldn't be able to do that with the current implementation. 

Of course the more flexibility we provide, the more likely it will be a foot gun, so take my comments there with a grain of salt. 

_rant: wouldn't it be nice for the SDK to have the same attribute processors as the OTel collector_

In addition to that, since this predicate does not get applied to the active span, should we communicate that somewhere in docs?

MikeGoldsmith:
> Are there any cases where endusers may want to filter based on value as opposed to key alone?
> 
> If I were writing my own predicate I would likely want to reach for an implementation that uses [`Hash#slice`](https://ruby-doc.org/core-3.0.0/Hash.html#method-i-slice) and have an explicit key list as well as potentially scrub attributes that I may be concerned has PII in it. I wouldn't be able to do that with the current implementation.
> 
> Of course the more flexibility we provide, the more likely it will be a foot gun, so take my comments there with a grain of salt.

I think if you're starting to reach for a way to scrub data, your needs are more complex than what this processor offers and  you should consider creating you own. I think using the key is a nice middle ground of being aware & having some control of what's going into baggage and doing more complex / custom logic.

> _rant: wouldn't it be nice for the SDK to have the same attribute processors as the OTel collector_

Yeah, I agree it's sometimes frustrating SDKs don't have all the same processors as the Collector. Although, then we'd need to maintain them in two places.

> In addition to that, since this predicate does not get applied to the active span, should we communicate that somewhere in docs?

Sure, I can madd an extra bit to the description that newly added baggage entries are not automatically added to the active span.

arielvalentin (approved):

arielvalentin:
@MikeGoldsmith please update your branch, since I do not have permissions to do so and I will merge this once specs pass.

MikeGoldsmith:
Done - thanks @arielvalentin 🎉

PR #997, opened 2024-05-31
chore: Update Rubocop to 1.64.0

kaylareopelle:
[Rubocop 1.64.0 Release Notes](https://github.com/rubocop/rubocop/releases/tag/v1.64.0).

The main introduction is the [`Style/SuperArguments` linter.](https://msp-greg.github.io/rubocop/RuboCop/Cop/Style/SuperArguments.html)

It also lints the Vitess Propagator. It looks like this propagator is not currently being tested, so I'll add that to the CI in a separate PR.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -7,5 +7,5 @@
 source 'https://rubygems.org'
 
 gem 'rake', '~> 13.0'
-gem 'rubocop', '~> 1.62'
+gem 'rubocop', '~> 1.64.0'
 gem 'rubocop-performance', '~> 1.21.0'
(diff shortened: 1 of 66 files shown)


kaylareopelle:
Once this is merged, we can continue [the release.](https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/995)

Gemfile
```diff
@@ -7,5 +7,5 @@
 source 'https://rubygems.org'
 
 gem 'rake', '~> 13.0'
-gem 'rubocop', '~> 1.62'
+gem 'rubocop', '~> 1.64.0'
```
arielvalentin, line 10: Do you intentionally only want to allow installing bug releases going forward?

arielvalentin (approved):

Gemfile
kaylareopelle: Thanks for noticing this. I'm not sure. 

In the previous PR, I changed it from allowing only bug releases to allowing minor version updates. That caused our main build to fail. If we only allow bug releases, we are less likely to have that problem, and may have dependabot open PRs first. 

I don't feel strongly either way, they both have their pros/cons. What do you think?

PR #1004, opened 2024-06-11
fix: Add net.peer.name to ethon

arielvalentin:
The ethon instrumentation did not include the `net.peer.name` attribute like other HTTP instrumentations and was not ported from ddtrace-rb: <https://github.com/DataDog/dd-trace-rb/blob/master/lib/datadog/tracing/contrib/ethon/easy_patch.rb#L159>

This attribute is important to GitHub because we base the `peer.service` extracting the subdomain and host information from `net.peer.name`.

In order to avoid parsing the URL multiple times, I inlined the common gem `cleanse_url` function and plan to open a separate PR adding a feature to support returing URI objects instead of Strings.

### DIFF
--- a/instrumentation/ethon/lib/opentelemetry/instrumentation/ethon/patches/easy.rb
+++ b/instrumentation/ethon/lib/opentelemetry/instrumentation/ethon/patches/easy.rb
@@ -87,14 +87,31 @@ module OpenTelemetry
               'http.method' => method
             }
 
-            http_url = OpenTelemetry::Common::Utilities.cleanse_url(url)
-            instrumentation_attrs['http.url'] = http_url if http_url
+            uri = _otel_cleanse_uri(url)
+            if uri
+              instrumentation_attrs['http.url'] = uri.to_s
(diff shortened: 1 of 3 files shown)


mwear (approved):
lgtm

instrumentation/ethon/lib/opentelemetry/instrumentation/ethon/patches/easy.rb
```diff
@@ -97,6 +100,20 @@ def span_creation_attributes(method)
             )
           end
 
+          # Returns a URL string with userinfo removed.
+          #
+          # @param [String] url The URL string to cleanse.
+          #
+          # @return [String] the cleansed URL.
```
johngallagher, line 107:
```
# @return [URI] the cleansed URL.
```

arielvalentin:
cc: @lauraway this is fix we need. Releases are blocked by failing builds but I have a fix for that pending review https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/1006

Issue #1005, opened 2024-06-11
Grape Test Suite Failures


arielvalentin:
https://github.com/open-telemetry/opentelemetry-ruby-contrib/actions/runs/9468201116/job/26084025685?pr=1004

```text
 /opt/hostedtoolcache/Ruby/3.3.1/x64/lib/ruby/3.3.0/bundled_gems.rb:74:in `require': cannot load such file -- rack/auth/digest/md5 (LoadError)
  	from /opt/hostedtoolcache/Ruby/3.3.1/x64/lib/ruby/3.3.0/bundled_gems.rb:74:in `block (2 levels) in replace_require'
  	from /opt/hostedtoolcache/Ruby/3.3.1/x64/lib/ruby/gems/3.3.0/gems/grape-1.8.0/lib/grape.rb:8:in `<top (required)>'
  	from <internal:/opt/hostedtoolcache/Ruby/3.3.1/x64/lib/ruby/3.3.0/rubygems/core_ext/kernel_require.rb>:37:in `require'
  	from <internal:/opt/hostedtoolcache/Ruby/3.3.1/x64/lib/ruby/3.3.0/rubygems/core_ext/kernel_require.rb>:37:in `require'
  	from /opt/hostedtoolcache/Ruby/3.3.1/x64/lib/ruby/3.3.0/bundled_gems.rb:74:in `block (2 levels) in replace_require'
  	from /opt/hostedtoolcache/Ruby/3.3.1/x64/lib/ruby/gems/3.3.0/gems/bundler-2.5.11/lib/bundler/runtime.rb:60:in `block (2 levels) in require'
  	from /opt/hostedtoolcache/Ruby/3.3.1/x64/lib/ruby/gems/3.3.0/gems/bundler-2.5.11/lib/bundler/runtime.rb:55:in `each'
  	from /opt/hostedtoolcache/Ruby/3.3.1/x64/lib/ruby/gems/3.3.0/gems/bundler-2.5.11/lib/bundler/runtime.rb:55:in `block in require'
  	from /opt/hostedtoolcache/Ruby/3.3.1/x64/lib/ruby/gems/3.3.0/gems/bundler-2.5.11/lib/bundler/runtime.rb:44:in `each'
  	from /opt/hostedtoolcache/Ruby/3.3.1/x64/lib/ruby/gems/3.3.0/gems/bundler-2.5.11/lib/bundler/runtime.rb:44:in `require'
[...]

PR #1006, opened 2024-06-11
ci: Updates grape appraisals

arielvalentin:
Fixes https://github.com/open-telemetry/opentelemetry-ruby-contrib/issues/1005


xuan-cao-swi (approved):
Thanks!

Will we be discontinuing support for grape < 1.8.0?

PR #1012, opened 2024-06-14
feat: Add after_request hook to Rack tracer middleware

adammohammed:
The intention here is to allow the user to be able to modify the span attributes using details about the response. This PR adds an `after_request` hook to the configuration options for the tracer middleware so that the user may supply a method which can interact with the span attributes.


adammohammed:
Hmm, on second thought, I might have done this wrong. I'm using this through the `action_pack` instrumentation, which seems like I'd want the changes in the event middleware, instead? Maybe it makes more sense to just override there?

ericmustin:
Just interjecting here for the sake of saving you some time/effort, but I'm relatively sure we've already rejected PRs like this in the past, which expose arbitrary hooks. It boils down to a maintenance question. I'd suggest opening an issue to discuss or attending a SIG before you commit cycles to trying to upstream this. Cc @fbogsany @arielvalentin

adammohammed:
Good to know, will do next time, thanks!

PR #1013, opened 2024-06-16
feat: Harmonize HTTP client instrumentation

tt:
This attempts to harmonize the HTTP client instrumentation by adopting the latest semantic conventions, version 1.26.0.

Each gem currently sets a slightly different set of attributes which makes it harder to create good dashboards in an environment where more than one HTTP client is used.

My change adopts the latest semantic conventions but doesn't retire what's already there. That seems like a better fit for the next major release which will allow a seamless migration.

Notably, this pull request makes the following [HTTP client span attributes](https://opentelemetry.io/docs/specs/semconv/http/http-spans/#http-client) available:

- `http.request.method`
- `http.response.status_code`
- `server.address`
- `server.port`
- `url.full`
- `url.scheme`

There are no constants available for these yet as [the opentelemetry-semantic_conventions gem](https://rubygems.org/gems/opentelemetry-semantic_conventions) was last updated in open-telemetry/opentelemetry-ruby@159044d74ce006cdfda8bb344b1f9c2267d7f040 to version 1.10.0.

See also [this summary of changes](https://opentelemetry.io/docs/specs/semconv/http/migration-guide/#summary-of-changes) and [the list of deprecated attributes](https://opentelemetry.io/docs/specs/semconv/attributes-registry/http/#http-deprecated-attributes).

### DIFF
--- a/instrumentation/ethon/lib/opentelemetry/instrumentation/ethon/patches/easy.rb
+++ b/instrumentation/ethon/lib/opentelemetry/instrumentation/ethon/patches/easy.rb
@@ -41,6 +41,7 @@ module OpenTelemetry
                 message = return_code ? ::Ethon::Curl.easy_strerror(return_code) : 'unknown reason'
                 @otel_span.status = OpenTelemetry::Trace::Status.error("Request has failed: #{message}")
               else
+                @otel_span.set_attribute('http.response.status_code', response_code)
                 @otel_span.set_attribute(OpenTelemetry::SemanticConventions::Trace::HTTP_STATUS_CODE, response_code)
                 @otel_span.status = OpenTelemetry::Trace::Status.error unless (100..399).cover?(response_code.to_i)
               end
@@ -85,12 +86,17 @@ module OpenTelemetry
           def span_creation_attributes(method)
             instrumentation_attrs = {
               OpenTelemetry::SemanticConventions::Trace::HTTP_METHOD => method
+              'http.request.method' => method
             }
 
             uri = _otel_cleanse_uri(url)
             if uri
               instrumentation_attrs[OpenTelemetry::SemanticConventions::Trace::HTTP_URL] = uri.to_s
               instrumentation_attrs[OpenTelemetry::SemanticConventions::Trace::NET_PEER_NAME] = uri.host if uri.host
+              instrumentation_attrs['server.address'] = uri.host if uri.host
+              instrumentation_attrs['server.port'] = uri.port if uri.port
+              instrumentation_attrs['url.full'] = uri.to_s
+              instrumentation_attrs['url.scheme'] = uri.scheme if uri.scheme
             end
 
             config = Ethon::Instrumentation.instance.config
--- a/instrumentation/ethon/lib/opentelemetry/instrumentation/ethon/patches/easy.rb
+++ b/instrumentation/ethon/lib/opentelemetry/instrumentation/ethon/patches/easy.rb
@@ -41,7 +41,7 @@ module OpenTelemetry
                 message = return_code ? ::Ethon::Curl.easy_strerror(return_code) : 'unknown reason'
                 @otel_span.status = OpenTelemetry::Trace::Status.error("Request has failed: #{message}")
               else
-                @otel_span.set_attribute('http.status_code', response_code)
+                @otel_span.set_attribute(OpenTelemetry::SemanticConventions::Trace::HTTP_STATUS_CODE, response_code)
                 @otel_span.status = OpenTelemetry::Trace::Status.error unless (100..399).cover?(response_code.to_i)
               end
             ensure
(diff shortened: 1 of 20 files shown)


tt:
I saw to [use conventional commit messages](https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/main/CONTRIBUTING.md#use-conventional-commit-messages) but also that this will be squash merged. I used the style for the pull request title and not for individual commits but happy to change it if necessary.

arielvalentin (changes requested):
Thank you for your contribution but unfortunately, we cannot accept this PR in its current state.

We have not adopted 1.x semconv in our instrumentations because we want to be able to be able to generally support schema URLs[^2] as well as the `OTEL_SEMCONV_STABILITY_OPT_IN` environment variables as described in the specification:

> Warning Existing HTTP instrumentations that are using [v1.20.0 of this document](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.20.0/specification/trace/semantic_conventions/http.md) (or prior):
> 
> SHOULD NOT change the version of the HTTP or networking conventions that they emit until the HTTP semantic conventions are marked stable (HTTP stabilization will include stabilization of a core set of networking conventions which are also used in HTTP instrumentations). Conventions include, but are not limited to, attributes, metric and span names, and unit of measure.
> SHOULD introduce an environment variable OTEL_SEMCONV_STABILITY_OPT_IN in the existing major version which is a comma-separated list of values. The only values defined so far are:
> http - emit the new, stable HTTP and networking conventions, and stop emitting the old experimental HTTP and networking conventions that the instrumentation emitted previously.
> http/dup - emit both the old and the stable HTTP and networking conventions, allowing for a seamless transition.
> The default behavior (in the absence of one of these values) is to continue emitting whatever version of the old experimental HTTP and networking conventions the instrumentation was emitting previously.
> Note: http/dup has higher precedence than http in case both values are present
> SHOULD maintain (security patching at a minimum) the existing major version for at least six months after it starts emitting both sets of conventions.
> SHOULD drop the environment variable in the next major version.[^1]

Though I recognize this is extremely challenging, many of our users are relying on pre 1.0 schema name, so having the ability to support `http/dup` would be important for us to provide during a transition.

In addition to that, many of the HTTP conventions are still considered `mixed` stability. We've been waiting a bit for the specification to settle before committing to anything and have had a few experiments to support multiple versions of the schema but have not had much success. 

I think what that means for this PR is that it would have to change to support the envars before we are willing to provide initial feedback and consider including it in a release. 

If you have an immediate need to support semconv 1.20+ then I would recommend looking at performing schema migrations in the otel collector either using a Schema Processor or a Transform Processor:

- https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/processor/schemaprocessor
- https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/processor/transformprocessor

[^1]: https://opentelemetry.io/docs/specs/semconv/http/
[^2]: https://opentelemetry.io/docs/specs/otel/schemas/

tt:
@arielvalentin:

> Though I recognize this is extremely challenging, many of our users are relying on pre 1.0 schema name, so having the ability to support `http/dup` would be important for us to provide during a transition.

I agree this is important. I also want overlap for our current traces. That said, isn't this essentially the same as what I've done where both sets of attributes are being included?

I think the downside of not supporting these new conventions is that people starting on their observability journey today will have to go through a migration that could be avoided.

Anyway, it seems you could achieve this entirely through releases:

1. Version 1.0 contains old set ("default behavior")
2. Version 1.1 contains old and new set (`http/dup` equivalent)
3. Version 2.0 contains new set (`http` equivalent)

I can certainly add `OTEL_SEMCONV_STABILITY_OPT_IN` support if that's the preferred direction but I'm probably not going to do it for all HTTP clients. We fortunately don't use all of them. 😅

tt:
@arielvalentin, any appetite for converging the existing instrumentation towards the now deprecated attributes?

For instance, look at the instrumentation for Excon and Faraday:

https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/aedc42cbb4bf2f6aba5d4e4a69c9fa45aacd172a/instrumentation/excon/lib/opentelemetry/instrumentation/excon/middlewares/tracer_middleware.rb#L29-L36

https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/da831f0b622830364197d48fb03c9615f8231249/instrumentation/faraday/lib/opentelemetry/instrumentation/faraday/middlewares/tracer_middleware.rb#L49-L55

These are mostly different sets. I don't strictly need the newest conventions but having a common set of attributes would be really valuable.

arielvalentin:
> any appetite for converging the existing instrumentation towards the now deprecated attributes?

Yes. I love symmetry and very much dislike surprises. 

These instrumentations where likely written by different authors and we don't exactly have a "checklist" of what an instrumentation should emit but only that it should do the absolute minimum or order to add instrumentation.  

In some of those cases there is a bit of divergence in the data provided by the libraries themselves. 

In the example you referenced above, Faraday and Excon, one provides the full URL (`http.url`) while the other only provides the path (`http.target`)[^1]. 

I assert the authors were likely trying to avoid any additional object allocation and only provided the minimal data made available to the instrumentation by the library itself; so, it's a tradeoff here but depending on system load it may not be so bad. 

Faraday is another one of the odd cases, where those attributes really matter most to the underlying `adapter` `client` spans. Unlike the underlying adapters, Faraday provides a much nicer and symmetric abstraction since it has a URL object, where you may be able to extract all of the attributes and share them with the adapter spans by amending them to the `OpenTelemetry::Common::HTTP::ClientContext` shared attributes helper:

https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/2230403871a9ee5ed9bac4fc6a7205b7d0876cd8/instrumentation/faraday/lib/opentelemetry/instrumentation/faraday/middlewares/tracer_middleware.rb#L58

### What should we do next?

I am in favor of providing symmetric attributes in HTTP instrumentations, but if we could do so by avoiding allocating any additional objects in drivers that would be best. 

If `Faraday` were to include all of the attributes as part of `OpenTelemetry::Common::HTTP::ClientContext`, would that address your use case? 

If it does, then how would you feel about approaching it that way instead of updating individual drivers?
 
[^1]: https://opentelemetry.io/docs/specs/semconv/attributes-registry/http/#http-deprecated-attributes

tt:
@arielvalentin:

> If `Faraday` were to include all of the attributes as part of `OpenTelemetry::Common::HTTP::ClientContext`, would that address your use case?

I need a bit of help understanding this. Isn't `OpenTelemetry::Common::HTTP::ClientContext` for passing additional context in?

Issue #957, opened 2024-04-30
Changes Requested for ActiveSupport instrumentation


arielvalentin:
During the [SIG meeting on 2024-04-30](https://docs.google.com/document/d/1EaIbfDE1elWTWt3bhilggki_OCRJZoFCSYZJXgc9KsM/edit), I brought up a few challenges of working with the existing [ActiveSupport instrumentation](https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/06759156579cfa4b6a83bd5ff8dfdfdcbf3e9747/instrumentation/active_support/lib/opentelemetry/instrumentation/active_support/span_subscriber.rb#L18), that were brought to our attention while instrumenting ActionMailer: https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/887

There were issues with missing options, implementation and naming that makes it difficult to use and potentially error prone. The list below includes some of the changes I would like to see: 

- [ ] Support Setting Span Kind
- [ ] Default to using the AS event name instead of splitting and inverting event names
- [ ] Allow users to provide a proc that formats the span name
- [ ] Ingress/egress spans should default to the semantic conventions; while internal spans default to AS event name 
- [ ] Switch to allow lists instead of deny (`disallow...`) lists and default to excluding PII
- [ ] Prevent users from mutating the ActiveSupport payload when extracting span attributes in `notification_payload_transform`
- [ ] Map Rails key names to semantic convention key names

PR #1014, opened 2024-06-16
feat!: Custom ActiveSupport Span Names

arielvalentin:
The current implementation of ActiveSupport instrumentation sets the span name to the reverse tokenized name, e.g. `render_template.action_view` is converted to `action_view render_template`

This default behavior can sometimes seem counter intuitive for users who use ActiveSupport Notifications to instrument their own code or users who are familiar with Rails instrumentation names.

This change does a few things to address the issues listed above:

1. Uses the notification name by default as oppossed to the legacy span name
2. Allows users to provide a custom span name formatter lambda
3. Provides a proc with backward compatible span name formatter `OpenTelemetry::Instrumentation::ActiveSupport::LEGACY_NAME_FORMATTER`

See https://github.com/open-telemetry/opentelemetry-ruby-contrib/issues/957

### DIFF
--- a/instrumentation/active_support/lib/opentelemetry/instrumentation/active_support/span_subscriber.rb
+++ b/instrumentation/active_support/lib/opentelemetry/instrumentation/active_support/span_subscriber.rb
@@ -10,6 +10,7 @@ module OpenTelemetry
     module ActiveSupport
       LEGACY_NAME_FORMATTER = ->(name) { name.split('.')[0..1].reverse.join(' ') }
 
+      # rubocop:disable Metrics/ParameterLists
       # The SpanSubscriber is a special ActiveSupport::Notification subscription
       # handler which turns notifications into generic spans, taking care to handle
       # context appropriately.
@@ -58,9 +59,11 @@ module OpenTelemetry
         subscriber_object
       end
 
+      # rubocop:enable Metrics/ParameterLists
       class SpanSubscriber
         ALWAYS_VALID_PAYLOAD_TYPES = [TrueClass, FalseClass, String, Numeric, Symbol].freeze
 
+        # rubocop:disable Metrics/ParameterLists
         def initialize(name:, tracer:, notification_payload_transform: nil, disallowed_notification_payload_keys: nil, kind: nil, span_name_formatter: nil)
           @span_name = safe_span_name_for(span_name_formatter, name).dup.freeze
           @tracer = tracer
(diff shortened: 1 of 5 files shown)


instrumentation/active_support/lib/opentelemetry/instrumentation/active_support/span_subscriber.rb
```diff
@@ -19,11 +21,13 @@ def self.subscribe(
         tracer,
         pattern,
         notification_payload_transform = nil,
-        disallowed_notification_payload_keys = []
+        disallowed_notification_payload_keys = nil,
+        span_name_formatter = nil
```
stevenharman, line 25: I wonder if the intent might be more clear if we defaulted this to a passthrough rather than `nil`? Something like:

```ruby
PASSTHROUGH_NAME_FORMATTER = =>(name) { name.dup.freeze }

# …

span_name_formatter = PASSTHROUGH_NAME_FORMATTER
```

Maybe it doesn't matter. Six of one, half a dozen of the other, as they say?

instrumentation/active_support/lib/opentelemetry/instrumentation/active_support/span_subscriber.rb
```diff
@@ -125,6 +128,16 @@ def sanitized_value(value)
             value
           end
         end
+
+        # Helper method to try an shield the span name formatter from errors
+        #
+        # It wraps the user supplied formatter in a rescue block and returns the original name if a StandardError is raised by the formatter
+        def safe_span_name_for(span_name_formatter, name)
+          span_name_formatter&.call(name) || name
```
stevenharman, line 136: If we're going to go this route, should we `#dup#freeze` the `name` that we fall back to for consistency with how the legacy formatter now works?

instrumentation/active_support/lib/opentelemetry/instrumentation/active_support/span_subscriber.rb
arielvalentin: I think even though it is symmetric, I am going choose to minimize allocations.

instrumentation/active_support/lib/opentelemetry/instrumentation/active_support/span_subscriber.rb
arielvalentin: I added `dup.freeze` to the result of the method call.

stevenharman (approved):

arielvalentin:
gentle ping @open-telemetry/ruby-contrib-approvers

xuan-cao-swi (approved):

PR #1018, opened 2024-06-16
perf: Reduce Context Allocations in ActiveJob

arielvalentin:
After seeing changes to Rack::Events that reduced the number of active contexts required for ingress spans, I decided to apply the same changes to ActiveJob


instrumentation/active_job/lib/opentelemetry/instrumentation/active_job/handlers/default.rb
```diff
@@ -55,20 +55,20 @@ def start_span(name, _id, payload)
...
-            finish_span(span, tokens)
+            finish_span(span, token)
           end
 
           # Finishes the provided spans and also detaches the associated contexts
           #
           # @param span [OpenTelemetry::Trace::Span]
-          # @param tokens [Array] to unregister
-          def finish_span(span, tokens)
+          # @param tokens [Numeric] to unregister
```
stevenharman, line 70:
```suggestion
          # @param token [Numeric] to unregister
```

stevenharman (approved):
Looks great. Just one little rename that was missed in a comment.

instrumentation/active_job/lib/opentelemetry/instrumentation/active_job/handlers/default.rb
```diff
@@ -41,9 +41,9 @@ def start(name, id, payload)
           # @return [Hash] with the span and generated context tokens
           def start_span(name, _id, payload)
             span = tracer.start_span(name, attributes: @mapper.call(payload))
-            tokens = [OpenTelemetry::Context.attach(OpenTelemetry::Trace.context_with_span(span))]
+            token = OpenTelemetry::Context.attach(OpenTelemetry::Trace.context_with_span(span))
```
kaylareopelle, line 44: nit: since this bit of code is [repeated in the Enqueue class](https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/1018/files#diff-d693657995fb181905c5c85e82859e4812b1bfa4e4b3adcacba234b5a4842c00R33), does it make sense to move it to some sort of shared helper method?

kaylareopelle (approved):

instrumentation/active_job/lib/opentelemetry/instrumentation/active_job/handlers/default.rb
arielvalentin: Perhaps there is an opportunity to refactor out a template method here. 🤔 let me take a stab at that in a follow up PR.

arielvalentin: I took a look again. I think that adding a template method here is more complex than I would like since the Enqueue class is doing things like setting a span kind, attributes, and context propagation, which the Default handler does not do. 

If I add a helper method just for the return statement, it would effectively be calling a method with two values that return a hash and that would also result in structural duplication so I am not sure how much better off we would be.

PR #1026, opened 2024-06-18
feat: Extend Sidekiq instrumentation with common used attributes

pantuza:
## Contributions
This Pull Request adds two attributes to Sidekiq instrumentation: `retry` and `args`.

Neither of these tags are present on the [Semantic Convention for Messaging systems](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/messaging/messaging-spans.md).

## Use case

I am migrating one application that was using Open Tracing libraries. 
As can be [seen here](https://github.com/iaintshine/ruby-sidekiq-tracer/blob/f29955e129432f6e5e727d8f729920d9594a6c67/lib/sidekiq/tracer/commons.rb#L15C29-L15C51) the library adds these attributes. 

Now with the Otel Instrumentation I have noticed these missing span attributes. 
Thus, this PR adds it. It might be useful for other companies and engineers migrating their traces
infrastructure to Open Telemetry. 

I have tried to follow the same naming strategy for non Semantic Convention attributes as 
you are already using in this Sidekiq instrumentation ([example](https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/d80a9f36b51aeddfd9d01da6964846ed21bcdcaf/instrumentation/sidekiq/lib/opentelemetry/instrumentation/sidekiq/middlewares/client/tracer_middleware.rb#L20))

### DIFF
--- a/instrumentation/sidekiq/lib/opentelemetry/instrumentation/sidekiq/middlewares/client/tracer_middleware.rb
+++ b/instrumentation/sidekiq/lib/opentelemetry/instrumentation/sidekiq/middlewares/client/tracer_middleware.rb
@@ -18,7 +18,7 @@ module OpenTelemetry
               attributes = {
                 SemanticConventions::Trace::MESSAGING_SYSTEM => 'sidekiq',
                 'messaging.sidekiq.job_class' => job['wrapped']&.to_s || job['class'],
-                'messaging.sidekiq.retry' => job['retry'].to_s,
+                'messaging.sidekiq.retry' => job['retry'],
                 'messaging.sidekiq.args' => job['args'].join(', '),
                 SemanticConventions::Trace::MESSAGING_MESSAGE_ID => job['jid'],
                 SemanticConventions::Trace::MESSAGING_DESTINATION => job['queue'],
--- a/instrumentation/sidekiq/lib/opentelemetry/instrumentation/sidekiq/middlewares/client/tracer_middleware.rb
+++ b/instrumentation/sidekiq/lib/opentelemetry/instrumentation/sidekiq/middlewares/client/tracer_middleware.rb
@@ -18,6 +18,8 @@ module OpenTelemetry
               attributes = {
                 SemanticConventions::Trace::MESSAGING_SYSTEM => 'sidekiq',
                 'messaging.sidekiq.job_class' => job['wrapped']&.to_s || job['class'],
+                'messaging.sidekiq.retry' => job['retry'].to_s,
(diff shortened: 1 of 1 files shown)


instrumentation/sidekiq/lib/opentelemetry/instrumentation/sidekiq/middlewares/client/tracer_middleware.rb
```diff
@@ -18,6 +18,8 @@ def call(_worker_class, job, _queue, _redis_pool)
               attributes = {
                 SemanticConventions::Trace::MESSAGING_SYSTEM => 'sidekiq',
                 'messaging.sidekiq.job_class' => job['wrapped']&.to_s || job['class'],
+                'messaging.sidekiq.retry' => job['retry'].to_s,
+                'messaging.sidekiq.args' => job['args'].join(', '),
```
arielvalentin, line 22: Please consider amending your specific domain attributes to the job in custom code as part of your application. 

Including arguments in an unstructured value makes instrumentations susceptible to leaking PII and are incredibly difficult to redact in post processing.  

```suggestion
```

instrumentation/sidekiq/lib/opentelemetry/instrumentation/sidekiq/middlewares/client/tracer_middleware.rb
```diff
@@ -18,6 +18,8 @@ def call(_worker_class, job, _queue, _redis_pool)
               attributes = {
                 SemanticConventions::Trace::MESSAGING_SYSTEM => 'sidekiq',
                 'messaging.sidekiq.job_class' => job['wrapped']&.to_s || job['class'],
+                'messaging.sidekiq.retry' => job['retry'].to_s,
```
arielvalentin, line 21: Is this a Numeric value? Would this be more interesting to keep as an `Integer`?

arielvalentin (changes requested):

In addition to my comments, please add automated tests to avoid any regressions in the future.

instrumentation/sidekiq/lib/opentelemetry/instrumentation/sidekiq/middlewares/client/tracer_middleware.rb
pantuza: Yes, it is. I agree with you. I will update it.

pantuza:
Closing this PR since `args` would be definitely a problematic span attribute to be added due to PII constraints.

Issue #1031, opened 2024-06-20
Default position of Faraday tracer middleware changes when upgrading Faraday


composerinteralia:
The Faraday tracer middleware works by patching the `RackBuilder#adapter` method to add the middleware to the stack if it isn't already there:

https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/e2ba37dcdd220774b33035a2e2a2dd97977ace48/instrumentation/faraday/lib/opentelemetry/instrumentation/faraday/patches/rack_builder.rb#L14-L20

In Faraday 0.17, `#adapter` is generally called last (Faraday warns otherwise), causing the tracer middleware to get installed last. That puts the tracer middleware as the innermost middleware, running directly around the adapter.

In Faraday 1.x, `#adapter` is typically called quite early while initializing the `RackBuilder`, before any user middlewares have been added. That's happening now because of [this change](https://github.com/lostisland/faraday/commit/8503baea4bf2bc821cf2c63125929c9fc7ef9d31#diff-14e9d694fa7c552f422f14b43d88d203f14a7a155c3c29e31a64c6125b8b7464R71). That ends up putting the tracer middleware as the outermost middleware (regardless of whether the user later adds the tracer middleware explicitly), running around all of the other middlewares.

I'm not entirely sure which position is most desirable, but having it shift from innermost middleware to outermost middleware when upgrading Faraday might be surprising, and is a breaking change at least for us at GitHub.

Does anybody have thoughts about what is most correct here? If we want to preserve having this default to the innermost middleware, we might be able to patch `Faraday::Connection#initialize` instead, adding the tracer middleware last if it isn't already there.

arielvalentin:
Thanks for this report! 

The instrumentations in this repo are mostly ports of the DD Trace Instrumentations. My guess is that this use case was not considered at the time when the instrumentation was written. 

I believe that we want to update this instrumentation to ensure the tracer middleware is the first one in the list, which I believe is what you mean by outermost, like it does in the current version of DD Trace:

https://github.com/DataDog/dd-trace-rb/blob/4b1901101c0f4d1033558bd9c0ff2e1c2b963bc7/lib/datadog/tracing/contrib/faraday/patcher.rb#L33

Would you be amenable to submitting a fix for this?

composerinteralia:
[...]

PR #1032, opened 2024-06-20
fix: compatibility with Faraday v1

composerinteralia:
Fixes https://github.com/open-telemetry/opentelemetry-ruby-contrib/issues/1031

Faraday >= 1 changes where the #adapter method gets called, affecting the RackBuilder patch in this gem.

Without this change, the tracer middleware gets added to the beginning of the stack instead of the end, and explicitly adding the middleware would cause it to get added twice.

This change patches `Connection#initialize` for Faraday >= 1 to preserve the behavior we had with Faraday < 1. This is similar to the dd-trace-rb change made in https://github.com/DataDog/dd-trace-rb/pull/906.

There is some followup work here to explore moving the middleware to the beginning of the stack instead of the end, but we can work on that separately.


arielvalentin (approved):

I will ponder a way to write some installation tests to cover some of this.

PR #1039, opened 2024-06-24
feat: Use Default Span Name Format

arielvalentin:
It also allows users to rollback to legacy span name format.

Follow up to https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/1014


instrumentation/action_view/lib/opentelemetry/instrumentation/action_view/instrumentation.rb
```diff
@@ -7,7 +7,42 @@
...
+      #
+      # Specifies custom proc used to extract span attributes form the notification payload.
+      # Use this to rename keys, extract nested values, or perform any other custom logic.
+      #
+      # ### `:legacy_span_names`
+      #
+      # - `boolean`  **default** `false`
+      #
+      # Specifies whether spans names should use the legacy format where the subscription was reverse ordered and white space separated.
+      # If set to `true`, the span name will match the name of the notification itself.
```
kaylareopelle, line 35: What do you think about adding examples? It doesn't have to be like this, but as an example:
```suggestion
      # Specifies whether spans names should use the legacy format where the subscription was reverse ordered and white space separated. (Ex. `action_view render_template`)
      # If set to `true`, the span name will match the name of the notification itself. (Ex. `render_template.action_view`)
```

kaylareopelle (approved):
Looks good as-is! One small documentation suggestion.

kaylareopelle (approved):
Thank you!

PR #1040, opened 2024-06-24
fix: suppress header access with symbol key deprecation warning in Racecar Instrumentation

ewhorton:
Fix the Rdkafka deprecation warning for header access with symbol key in the Racecar instrumentation.

This copies the approach taken for the Rdkafka instrumentation [here](https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/884).

Both of these can be cleaned up if/when support for rdkafka 0.12.0 is removed.

@chrisholmes


kaylareopelle:
Hi @ewhorton! Thanks for your submission! Would you mind [signing the CLA?](https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/1040#issuecomment-2187102886) We can't accept your submission until you do.

ewhorton:
Hi again @kaylareopelle, sorry for the delay, we have just been able to get this signed now 🙂

kaylareopelle (approved):

instrumentation/racecar/lib/opentelemetry/instrumentation/racecar/process_message_subscriber.rb
```diff
@@ -4,14 +4,21 @@ module OpenTelemetry
   module Instrumentation
     # This class contains the ASN subsciber that instruments message processing
     class ProcessMessageSubscriber
+      GETTER = if Gem::Version.new(::Rdkafka::VERSION) >= Gem::Version.new('0.13.0')
```
arielvalentin, line 7: I would _prefer_ to see this evaluation occur in `install` process since we do other gem version compatibility checks but 🤷🏼‍♂️ it may make it more difficult to understand. WDYT @kaylareopelle ?

https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/b26b765b91b789bce25d47743326d3cc16b6feb7/instrumentation/racecar/lib/opentelemetry/instrumentation/racecar/instrumentation.rb#L36

arielvalentin (approved):

PR #1073, opened 2024-07-16
fix: register lambda span

xuan-cao-swi:
## Description

The aws lambda span was not registered with context after start_span, which cause current_span can't detect it. User couldn't append attributes outside call_wrapped span. This PR attach the span to current context and detach it before finish.

### DIFF
--- a/instrumentation/aws_lambda/lib/opentelemetry/instrumentation/aws_lambda/handler.rb
+++ b/instrumentation/aws_lambda/lib/opentelemetry/instrumentation/aws_lambda/handler.rb
@@ -47,6 +47,9 @@ module OpenTelemetry
               kind: span_kind
             )
 
+            trace_ctx = OpenTelemetry::Trace.context_with_span(span)
+            @trace_token = OpenTelemetry::Context.attach(trace_ctx)
+
             begin
               response = call_original_handler(event: event, context: context)
               status_code = response['statusCode'] || response[:statusCode] if response.is_a?(Hash)
@@ -63,6 +66,7 @@ module OpenTelemetry
               span&.record_exception(original_handler_error)
               span&.status = OpenTelemetry::Trace::Status.error(original_handler_error.message)
             end
+            OpenTelemetry::Context.detach(@trace_token) if @trace_token
             span&.finish
             OpenTelemetry.tracer_provider.force_flush(timeout: @flush_timeout)
           end
--- a/instrumentation/aws_lambda/lib/opentelemetry/instrumentation/aws_lambda/handler.rb
+++ b/instrumentation/aws_lambda/lib/opentelemetry/instrumentation/aws_lambda/handler.rb
@@ -39,6 +39,7 @@ module OpenTelemetry
 
           original_handler_error = nil
           original_response = nil
+          trace_token = nil
           OpenTelemetry::Context.with_current(parent_context) do
             span_attributes = otel_attributes(event, context)
             span = tracer.start_span(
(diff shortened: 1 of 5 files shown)


instrumentation/aws_lambda/lib/opentelemetry/instrumentation/aws_lambda/handler.rb
```diff
@@ -47,6 +47,9 @@ def call_wrapped(event:, context:)
               kind: span_kind
             )
 
+            trace_ctx = OpenTelemetry::Trace.context_with_span(span)
+            @trace_token = OpenTelemetry::Context.attach(trace_ctx)
```
kaylareopelle, line 51: Why did you decide to make this an instance variable?

xuan-cao-swi: In case the process fail, on the ensure part, `OpenTelemetry::Context.detach(@trace_token) if @trace_token` won't throw error (e.g. undefined local variable or method trace_token)

kaylareopelle: Got it! I understand wanting to prevent undefined errors, and I think we can achieve that with a local variable.

One option could be to update the existing guard clause to use `if defined?(trace_token)`.

Another option could be to add a line to define `trace_token` earlier on in the method, like we do for `span_kind`, and then use `if trace_token` in the guard clause:
```ruby
def method
  trace_token = nil
  # ... 
ensure
  OpenTelemetry::Context.detach(trace_token) if trace_token
end
```

Would one of these options work for you?

xuan-cao-swi: Yes, I will take the local variable, thanks!

kaylareopelle (approved):
Thank you!

instrumentation/aws_lambda/lib/opentelemetry/instrumentation/aws_lambda/handler.rb
```diff
@@ -47,6 +48,9 @@ def call_wrapped(event:, context:)
               kind: span_kind
             )
 
+            trace_ctx = OpenTelemetry::Trace.context_with_span(span)
+            trace_token = OpenTelemetry::Context.attach(trace_ctx)
```
cheempz, line 52: @xuan-cao-swi I'm curious whether you tried using `tracer.in_span` instead of `tracer.start_span` and having to explicitly manage the context. 

Since you're already using `OpenTelemetry::Context.with_current(parent_context)` in line 43 above to propagate any incoming context into the current one, `in_span` should be able to continue that and as a benefit record exception too. Some examples in exising "[server](https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/542fa91b796c4abb1b855b91c1af1d67894a2798/instrumentation/rack/lib/opentelemetry/instrumentation/rack/middlewares/tracer_middleware.rb)" or "[consumer](https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/542fa91b796c4abb1b855b91c1af1d67894a2798/instrumentation/sidekiq/lib/opentelemetry/instrumentation/sidekiq/middlewares/server/tracer_middleware.rb)" instrumentation that deal with the same concern of propagating incoming context from external process, setting it into current process's context, and starting a span.

xuan-cao-swi: Thanks, I agree in_span is a better choice for this instrumentation. Updated!

PR #1074, opened 2024-07-17
fix(que): Fix bulk_enqueue when enqueuing more than 5 jobs

laurglia:
Previously, with Que `bulk_enqueue` a new span was created for every job in the batch.

For each such span, a tag was added to _all_ jobs in the batch. This meant that if you enqueued 3 jobs, then 3 spans were created and each job had 3 tags pointing to 3 different spans.

This behavior became problematic when you tried to insert more than 5 jobs, as Que supports up to 5 tags per job. More specifically, a runtime error was raised by Que.

This commit fixes that bug by creating only a single span when using bulk_enqueue.

### DIFF
--- a/instrumentation/que/lib/opentelemetry/instrumentation/que/patches/que_job.rb
+++ b/instrumentation/que/lib/opentelemetry/instrumentation/que/patches/que_job.rb
@@ -19,6 +19,10 @@ module OpenTelemetry
           # Module to prepend to Que singleton class
           module ClassMethods
             def enqueue(*args, job_options: {}, **arg_opts)
+              # In Que version 2.1.0 `bulk_enqueue` was introduced.
+              # In that case, the span is created inside the `bulk_enqueue` method.
+              return super(*args, **arg_opts) if gem_version >= Gem::Version.new('2.1.0') && Thread.current[:que_jobs_to_bulk_insert]
+
               tracer = Que::Instrumentation.instance.tracer
               otel_config = Que::Instrumentation.instance.config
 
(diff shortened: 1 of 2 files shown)


instrumentation/que/lib/opentelemetry/instrumentation/que/patches/que_job.rb
```diff
@@ -67,6 +60,32 @@ def enqueue(*args, job_options: {}, **arg_opts)
...
+
+                tracer.in_span('publish', kind: :producer) do |span|
+                  super do
+                    yield
+
+                    job_attrs = Thread.current[:que_jobs_to_bulk_insert][:jobs_attrs]
+
+                    unless job_attrs.empty?
+                      span.name = "#{job_attrs.first[:job_class]} publish"
+                      span.add_attributes(QueJob.job_attributes(job_attrs.first))
```
indrekj, line 77: Does it add `messaging.message_id` to the attributes? It shouldn't in case of a bulk enqueue. The other properties seem correct because they should be the same according to the bulk enqueue [limitations](https://github.com/que-rb/que/blob/master/docs/README.md#enqueueing-jobs-in-bulk).

instrumentation/que/test/opentelemetry/instrumentation/que_test.rb
```diff
@@ -363,10 +416,10 @@ def self.run(first, second); end
 
           _(finished_spans.size).must_equal(2)
 
-          span1 = finished_spans.first
+          span1 = finished_spans.last
```
indrekj, line 419: did something change with ordering?

indrekj (approved):
(not an actual reviewer but) looks good to me

kaylareopelle:
Hi @laurglia! Thanks for your contribution. Would you mind [signing the CLA](https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/1074#issuecomment-2233328176)?

kaylareopelle:
Hi @laurglia! Any luck with the CLA?

kaylareopelle (approved):
Thank you, @laurglia! (and thank you @indrekj for your review too! 🙂)

PR #1075, opened 2024-07-19
fix: use AS::N subscriber for serialize events

### LINKED ISSUE #992: ActiveModelSerializer instrumentation does not make the span current
**Description of the bug**

Spans occuring during a render call using AM:S are not set as children span of the AM:S render span

image

Here, the spans below the serializer are made by the serializer code.

Operating system details: Linux, Ubuntu 20.04 LTS
RUBY_ENGINE: "ruby"
RUBY_VERSION: "3.2.2"
RUBY_DESCRIPTION: "ruby 3.2.2 (2023-03-30 revision e51014f9c0) [arm64-darwin22]"

No reproduction, but the issue is that the event handler is setting itself as the current span here: https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/main/instrumentation/active_model_serializers/lib/opentelemetry/instrumentation/active_model_serializers/event_handler.rb#L14-L20

Pinging @robbkidd as he helped me on this

robbkidd:
Resolves #992 

This mimicks a fair amount of the current ActionMailer and ActionView instrumentation.

The details for Context management (i.e. setting current span) are already handled by the OTel ActiveSupport instrumentation. Reuse the notifications subscriber here for ActiveModel serialization events.

Reworked the example app into two: one Rails which works with the usual SDK configuration and one standalone (no Rails) to demonstrate that the subscription needs to be made after the SDK configuration is complete. If the subscription is created during instrumentation install, the subscription's tracer will be a NO-OP API tracer and won't produce spans.

TODO:

- [ ] review/update README and doc comments for accurate usage steps


instrumentation/active_model_serializers/test/opentelemetry/instrumentation/active_model_serializers/instrumentation_test.rb
```diff
@@ -38,7 +39,7 @@
       _(exporter.finished_spans.size).must_equal 1
 
       _(span).must_be_kind_of OpenTelemetry::SDK::Trace::SpanData
-      _(span.name).must_equal 'ModelSerializer render'
+      _(span.name).must_equal 'render.active_model_serializers'
```
robbkidd, line 42: One disadvantage of using the current OTel AS::N SpanSubscriber is that the `span_name_formatter` callable only receives the notification event name as a parameter. It has no access to the event payload to generate a span name like previously.

instrumentation/active_model_serializers/example/standalone.rb
```diff
@@ -0,0 +1,55 @@
...
+end
+
+OpenTelemetry::SDK.configure do |c|
+  c.service_name = 'active_model_serializers_example'
+  c.use_all
+end
+
+# without Rails and the Railtie automation, must manually trigger
+# instrumentation subscription after SDK is configured
+OpenTelemetry::Instrumentation::ActiveModelSerializers.subscribe
```
robbkidd, line 27: I don't love this extra needed step outside the configure block to get the instrumentation to Actually Instrument when Rails wasn't around to detect and call Railtie hooks. I couldn't get the SpanScriber to have a real, configured Tracer when instantiating the subscriber within the instrumentation install step. I see that ActionMailer and ActionView are relying entirely on a Railtie to subscribe at the end of Rails initialization. I figured I would add this interface for folks who may be using ActiveModelSerializers outside of a full Rails app.

arielvalentin: I understand what you mean. 

Want to do some sort of hacky thing that checks if `Rails` is defined and use the Railtie, otherwise manually install?

instrumentation/active_model_serializers/lib/opentelemetry/instrumentation/active_model_serializers/instrumentation.rb
```diff
@@ -24,24 +32,39 @@ class Instrumentation < OpenTelemetry::Instrumentation::Base
...
+          self.class.instance.tracer
+        end
+
         def gem_version
           Gem::Version.new(::ActiveModel::Serializer::VERSION)
         end
 
-        def require_dependencies
-          require_relative 'event_handler'
+        def install_active_support_instrumenation
```
arielvalentin, line 52: You will want to pass the config object here otherwise it will never receive the options.

instrumentation/active_model_serializers/lib/opentelemetry/instrumentation/active_model_serializers/instrumentation.rb
```diff
@@ -24,24 +32,39 @@ class Instrumentation < OpenTelemetry::Instrumentation::Base
...
-            EventHandler.handle(start, finish, payload)
-          end
+        def require_dependencies
+          require_relative 'railtie'
         end
 
-        def event_name
-          'render.active_model_serializers'
+        def default_attribute_transformer
+          lambda { |payload|
```
arielvalentin, line 61: Bike shed time...

Is `serializer` a specific enough namespace not to collide with other instrumentations or sematic conventions?

Is `format` similar to a MIME type?

instrumentation/active_model_serializers/lib/opentelemetry/instrumentation/active_model_serializers/instrumentation.rb
```diff
@@ -24,24 +32,39 @@ class Instrumentation < OpenTelemetry::Instrumentation::Base
...
+        def install_active_support_instrumenation
+          OpenTelemetry::Instrumentation::ActiveSupport::Instrumentation.instance.install({})
         end
 
-        def register_event_handler
-          ::ActiveSupport::Notifications.subscribe(event_name) do |_name, start, finish, _id, payload|
-            EventHandler.handle(start, finish, payload)
-          end
+        def require_dependencies
+          require_relative 'railtie'
```
arielvalentin, line 57: Should this _only_ be required if `Rails` is defined?

instrumentation/active_model_serializers/lib/opentelemetry/instrumentation/active_model_serializers/railtie.rb
```diff
@@ -0,0 +1,24 @@
...
+# SPDX-License-Identifier: Apache-2.0
+
+module OpenTelemetry
+  module Instrumentation
+    module ActiveModelSerializers # :nodoc:
+      def self.subscribe
+        Instrumentation.instance.subscribe
+      end
+
+      if defined?(::Rails::Railtie)
```
arielvalentin, line 14: I see what you are doing. Instead of requiring this file only if `Rails::Railtie` is defined you are adding it to this file.

Is this how we do it in other instrumentations?

instrumentation/active_model_serializers/test/opentelemetry/instrumentation/active_model_serializers/instrumentation_test.rb
arielvalentin: Do you want to change that about the OTel ASN gem?

What about writing your own span subscriber?

instrumentation/active_model_serializers/lib/opentelemetry/instrumentation/active_model_serializers/instrumentation.rb
robbkidd: These are probably not the greatest attribute names. In this PR, I was attempting to only fix the means by which the instrumentation subscribed so that current span for activity occurring within the event would be accurate. If there are semantic conventions to address in the attribute names, I propose we figure out what better attribute names could be used on a separate issue.

instrumentation/active_model_serializers/lib/opentelemetry/instrumentation/active_model_serializers/instrumentation.rb
robbkidd: As you [noticed](https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/1075#pullrequestreview-2188591330), the check for Rails being defined is within railtie.rb.

instrumentation/active_model_serializers/lib/opentelemetry/instrumentation/active_model_serializers/railtie.rb
robbkidd: > Is this how we do it in other instrumentations?

No, the other instrumentation that have railties don't appear to have any guard around the Railtie class getting defined. I suspect that we ought to be doing something to guard that.

arielvalentin: Ok we should file an issue for this and fix that. This kind of inconsistency will come back to bite us.

instrumentation/active_model_serializers/test/opentelemetry/instrumentation/active_model_serializers/instrumentation_test.rb
robbkidd: I would very much like to change that about the subscriber in the AS::N instrumentation.

We chatted in CNCF Slack briefly around the idea of the OTel AS::N subscriber taking a block that receives a mutable span and the AS::N event payload. The instrumentation developer can perform whatever name and attribute hijinks they desire. (I propose we add to the documentation of this potential future interface a warning that the behavior, performance, and exception handling of that block is entirely the responsibility of the block's author.)

I can picture (vaguely) that any of the instrumentation we write that leverages this block-accepting AS::N subscriber could themselves accept a block to allow end-user customization.

instrumentation/active_model_serializers/lib/opentelemetry/instrumentation/active_model_serializers/instrumentation.rb
arielvalentin: NVM. I have no idea what in the world I am talking about.

arielvalentin:
@robbkidd mind taking a look at @kaylareopelle comments?

instrumentation/active_model_serializers/example/rails_app.rb
```diff
@@ -0,0 +1,72 @@
...
+gemfile(true) do
+  source 'https://rubygems.org'
+
+  gem 'rails'
+  gem 'active_model_serializers'
+  gem 'opentelemetry-api'
+  gem 'opentelemetry-common'
+  gem 'opentelemetry-instrumentation-active_model_serializers', path: '../'
+  gem 'opentelemetry-sdk'
+  gem 'opentelemetry-exporter-otlp'
```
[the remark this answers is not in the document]
robbkidd: Latest iteration [keeps this OTLP exporter dependency, but defaults to console output](https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/1075/commits/db4a04ab58ba5c4422a1767ebc88c93a30cb9114).

How's that hang together?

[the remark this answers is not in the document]
kaylareopelle: Those updates look great! Thanks!

kaylareopelle:
@robbkidd - Example code is cleared from me! 

It looks like there's a few other things you and @arielvalentin chatted about related to the subscriber and Railties. Is there more you'd like to do to address those questions? I'm not 100% following with a quick read, but happy to take a deeper look if that would help.

arielvalentin (approved):
A fast follow up to this would be adding some details in the README about what semantic attributes to expect being emitted by this instrumentation. 

Otherwise I think we are good to go.

arielvalentin:
@kaylareopelle I'm good merge if you are

robbkidd:
@kaylareopelle I think the ideas @arielvalentin and I had in conversations in this PR are not blockers to this PR and can/ought to be done in follow-ups.

kaylareopelle (approved):

PR #1090, opened 2024-07-30
fix: Add super calls to GraphqlTrace

rmosolgo:
Oops -- these trace modules are supposed to call `super`, where other trace modules may add instrumentation. (Calling `&block` hands control to GraphQL-Ruby, but skips other trace module code.) 

My guess is that these weren't converted to `super` when this module was migrated off of GraphQL-Ruby's own `PlatformTrace` module which worked differently (or used to have this same bug?). 

Fixes https://github.com/rmosolgo/graphql-ruby/issues/5044

TODO: 

-  ~~Along the way, I found an issue with how GraphQL::Ruby handles `GraphQL::Schema.trace_with(trace_mod)` when subclasses already exist and those subclasses have already received `.trace_with(...)` calls of their own. I'm trying to work that out in https://github.com/rmosolgo/graphql-ruby/pull/5045, but that issue is making this test fail (I think...)~~ Never mind, I think I misunderstood because of the ivar reset-related code in the test suite. When I modified my test to use a "fresh" schema class (and manually re-call the setup code), the test failed as expected, then passed when my patch was applied. So I think this is ready to go as-is.

### DIFF
--- a/instrumentation/graphql/lib/opentelemetry/instrumentation/graphql/tracers/graphql_trace.rb
+++ b/instrumentation/graphql/lib/opentelemetry/instrumentation/graphql/tracers/graphql_trace.rb
@@ -60,15 +60,15 @@ module OpenTelemetry
           end
 
           def execute_multiplex(multiplex:, &block)
-            tracer.in_span('graphql.execute_multiplex', &block)
+            tracer.in_span('graphql.execute_multiplex') { super }
           end
 
           def lex(query_string:, &block)
-            tracer.in_span('graphql.lex', &block)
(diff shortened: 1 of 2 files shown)


rmosolgo:
🤔 I force-pushed with with an updated commit message, hoping to pass the "conventional commits validation" test, but it's still failing. Is there a way to make it pass?

arielvalentin:
@rmosolgo it validates the title not the individual commits.

arielvalentin (approved):

rmosolgo:
Oops, thanks!

PR #1093, opened 2024-07-31
First draft of PR

AkhigbeEromo:


.github/workflows/ci-instrumentation.yml
```diff
@@ -51,6 +51,7 @@ jobs:
           - restclient
           - rspec
           - sinatra
+          - <instrumentation_name>
```
AkhigbeEromo, line 54: Would love to know what name should be here

Issue #978, opened 2024-05-13
Update instrumentation generator script output


kaylareopelle:
There's some output in the instrumentation generator script related to the CI that's inaccurate. Update the script to reflect the current files that need to be updated. 

See: https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/946/files#r1597226769

kaylareopelle:
@AkhigbeEromo I think this is a good one for you to start with! Let me know if you have any questions!

AkhigbeEromo:
Thank you @kaylareopelle, I will check it out

PR #1094, opened 2024-07-31
feat: Update instrumentation generator

AkhigbeEromo:
Updated the script to reflect the current files that need to be updated.
Updated the order of the methods generated in the instrumentation file
first draft of #978

### DIFF
--- a/.github/workflows/ci-instrumentation.yml
+++ b/.github/workflows/ci-instrumentation.yml
@@ -51,6 +51,7 @@ jobs:
           - restclient
           - rspec
           - sinatra
+          - <instrumentation_name>
         os:
           - ubuntu-latest
 
--- a/.github/workflows/ci-instrumentation.yml
+++ b/.github/workflows/ci-instrumentation.yml
@@ -51,7 +51,6 @@ jobs:
           - restclient
           - rspec
           - sinatra
-          - <instrumentation_name>
         os:
           - ubuntu-latest
 
--- a/.instrumentation_generator/test/instrumentation_generator_test.rb
+++ b/.instrumentation_generator/test/instrumentation_generator_test.rb
@@ -1,0 +1,84 @@
+# test/instrumentation_generator_test.rb
+
+require 'minitest/autorun'
+require 'mocha/minitest'
+require 'thor'
+require_relative '../instrumentation_generator'
+
+class InstrumentationGeneratorTest < Minitest::Test
+  def setup
+    @instrumentation_name = 'new_instrumentation'
+    @generator = InstrumentationGenerator.new([@instrumentation_name])
+
+    @generator.stubs(:template)
+    @generator.stubs(:insert_into_file)
+    @generator.stubs(:puts)
+  end
+
+  def test_root_files
+    @generator.root_files
+
+    assert @generator.template('templates/rubocop.yml.tt', 'instrumentation/new_instrumentation/.rubocop.yml')
+    assert @generator.template('templates/yardopts.tt', 'instrumentation/new_instrumentation/.yardopts')
+    assert @generator.template('templates/Appraisals', 'instrumentation/new_instrumentation/Appraisals')
+    assert @generator.template('templates/CHANGELOG.md.tt', 'instrumentation/new_instrumentation/CHANGELOG.md')
+    assert @generator.template('templates/Gemfile', 'instrumentation/new_instrumentation/Gemfile')
+    assert @generator.template('templates/LICENSE', 'instrumentation/new_instrumentation/LICENSE')
+    assert @generator.template('templates/gemspec.tt', 'instrumentation/new_instrumentation/opentelemetry-instrumentation-new_instrumentation.gemspec')
+    assert @generator.template('templates/Rakefile', 'instrumentation/new_instrumentation/Rakefile')
+    assert @generator.template('templates/Readme.md.tt', 'instrumentation/new_instrumentation/README.md')
+  end
+
+  def test_lib_files
+    @generator.lib_files
+
+    assert @generator.template('templates/lib/entrypoint.rb', 'instrumentation/new_instrumentation/lib/opentelemetry-instrumentation-new_instrumentation.rb')
+    assert @generator.template('templates/lib/instrumentation.rb.tt', 'instrumentation/new_instrumentation/lib/opentelemetry/instrumentation.rb')
+    assert @generator.template('templates/lib/instrumentation/instrumentation_name.rb.tt', 'instrumentation/new_instrumentation/lib/opentelemetry/instrumentation/new_instrumentation.rb')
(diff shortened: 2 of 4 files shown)


.github/workflows/ci-instrumentation.yml
```diff
@@ -51,6 +51,7 @@ jobs:
           - restclient
           - rspec
           - sinatra
+          - <instrumentation_name>
```
xuan-cao-swi, line 54: This is causing issue. <value> is not allowed in this kind workflow

xuan-cao-swi: Could you also check [use-conventional-commit-messages](https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/main/CONTRIBUTING.md#use-conventional-commit-messages) to update the title of PR

.instrumentation_generator/test/instrumentation_generator_test.rb
```diff
@@ -0,0 +1,84 @@
...
+
+    @generator.stubs(:template)
+    @generator.stubs(:insert_into_file)
+    @generator.stubs(:puts)
+  end
+
+  def test_root_files
+    @generator.root_files
+
+    assert @generator.template('templates/rubocop.yml.tt', 'instrumentation/new_instrumentation/.rubocop.yml')
```
kaylareopelle, line 21: First off, thank you for adding tests! It's great to have coverage on a tool that we use whenever we add new instrumentation. 

I like this example as the three options a Ruby unit test should evaluate[^1]: 
* It returns a value.
* It passes the work along to somewhere else (i.e., dispatches the work elsewhere).
* It causes a side effect.

In this test, it's almost as if we're testing whether the `@generator.template` method can be called. Since we don't define `@generator.template`, ([that's from `Thor`](https://www.rubydoc.info/gems/thor/Thor%2FActions:template)), I think we could take a different approach to make sure everything is working as expected.

Of the three options listed above, I think we want to make sure this method causes a side effect.

Since the `root_files` method is supposed to create these files from the ERB templates, what if we checked to see if the files were created when the method is run? 

That could look like: 
`assert File.exist?('instrumentation/new_instrumentation/.yardopts')`

Let me know if you want help brainstorming assertions related to side effects of the other methods.

[^1]: https://www.cloudbees.com/blog/unit-testing-in-ruby#what-should-be-tested

.instrumentation_generator/test/instrumentation_generator_test.rb
```diff
@@ -0,0 +1,84 @@
+# test/instrumentation_generator_test.rb
+
+require 'minitest/autorun'
+require 'mocha/minitest'
```
kaylareopelle, line 4: I wasn't able to run these tests locally. When I tried to run the file from the root directory using: 
```
bundle exec ruby .instrumentation_generator/test/instrumentation_generator_test.rb
```

I got the following error: 
```
/Users/kreopelle/.rbenv/versions/3.3.1/lib/ruby/gems/3.3.0/gems/mocha-1.16.1/lib/mocha/integration/mini_test/adapter.rb:26:in `included': uninitialized constant MiniTest (NameError)

          Mocha::ExpectationErrorFactory.exception_class = ::MiniTest::Assertion
                                                           ^^^^^^^^^^
Did you mean?  Minitest
        from /Users/kreopelle/.rbenv/versions/3.3.1/lib/ruby/gems/3.3.0/gems/mocha-1.16.1/lib/mocha/integration/mini_test.rb:50:in `include'
        from /Users/kreopelle/.rbenv/versions/3.3.1/lib/ruby/gems/3.3.0/gems/mocha-1.16.1/lib/mocha/integration/mini_test.rb:50:in `activate'
        from /Users/kreopelle/.rbenv/versions/3.3.1/lib/ruby/gems/3.3.0/gems/mocha-1.16.1/lib/mocha/minitest.rb:5:in `<top (required)>'
        from <internal:/Users/kreopelle/.rbenv/versions/3.3.1/lib/ruby/site_ruby/3.3.0/rubygems/core_ext/kernel_require.rb>:37:in `require'
        from <internal:/Users/kreopelle/.rbenv/versions/3.3.1/lib/ruby/site_ruby/3.3.0/rubygems/core_ext/kernel_require.rb>:37:in `require'
        from instrumentation_generator_test.rb:4:in `<main>'
```

It looks like `mocha` hasn't updated its call to `MiniTest::Assertion` to use the `Minitest` spelling. Minitest used to accept both spellings, but now it only accepts `Minitest`. 

Minitest has its own stubbing features that can be used, though you have to use them in a specific test and can't set them in a `setup` method. Here's one of my favorite articles on stubbing in Minitest: https://semaphoreci.com/community/tutorials/mocking-in-ruby-with-minitest 

When I removed `mocha/minitest` and took the `@generator.stubs` method calls out of the `setup` method, I was able to run the tests.

The only error I got was related to calling `puts`:
```
  1) Error:
InstrumentationGeneratorTest#test_update_ci_workflow:
NoMethodError: private method `puts' called for an instance of InstrumentationGenerator
    .instrumentation_generator/test/instrumentation_generator_test.rb:86:in `test_update_ci_workflow'

6 runs, 21 assertions, 0 failures, 1 errors, 0 skips
```

.instrumentation_generator/test/instrumentation_generator_test.rb
```diff
@@ -0,0 +1,84 @@
...
+
+class InstrumentationGeneratorTest < Minitest::Test
+  def setup
+    @instrumentation_name = 'new_instrumentation'
+    @generator = InstrumentationGenerator.new([@instrumentation_name])
+
+    @generator.stubs(:template)
+    @generator.stubs(:insert_into_file)
+    @generator.stubs(:puts)
+  end
```
kaylareopelle, line 16: Since running the generator creates files, I think the files should be deleted between the tests to make sure we have a clean slate for each test. 

Like the `setup` method you added, [minitest also has a `teardown` method.](https://www.rubydoc.info/stdlib/minitest/MiniTest%2FUnit%2FTestCase:teardown)

Maybe in `teardown` we could delete the directory we expect to get created if it exists?

kaylareopelle (changes requested):
@AkhigbeEromo, thank you for opening this PR! Amazing start! 😄 

I think this does a good job of dynamically updating the CI file and adding tests.

However, there's a little more to do to meet the changes in the PR description. 

For `Updated the script to reflect the current files that need to be updated.`, the `bin/instrumentation_generator` output should probably be removed. If we're going to add the gem to the CI file dynamically, then that warning isn't needed. 

For `Updated the order of the methods generated in the instrumentation file`, I think the [instrumentation.rb.tt](https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/0b58055b3ca23a4a4842d6e6ed38279f77991d90/.instrumentation_generator/templates/lib/instrumentation/instrumentation_name/instrumentation.rb.tt) template would need to get updated to add the calls to `compatible` and `option`. Here's a good example on the order of the methods and how they can be called: [pg/instrumentation.rb](https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/d70068ff1d6f33aa44937d8f2161229884105b14/instrumentation/pg/lib/opentelemetry/instrumentation/pg/instrumentation.rb)

Thanks again for your help!

PR #1107, opened 2024-08-08
revert: "feat: Use Default Span Name Format (#1039)"

arielvalentin:
This reverts commit f999e702430a55bc9f0fdf7b4748dc2f9b6341b8.

We can restore it if someone requests backward compatibility.


xuan-cao-swi (approved):

kaylareopelle (approved):

mrsimo:
Hi! This is a summary of my journey so far:

* Try to figure out why a recent update changed the naming of of spans.
* Find out that the new active_support changed the default. It's mentioned as a breaking change in the Changelog, but happened on a minor version bump, when it's quite hard to track all the changes when there's dozens of gems related to opentelemetry.
* Felt hopeful when I found the `legacy_span_names` option in https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/1039
* Saw that it was only implemented for action_view, thought about contributing and adding that option to the missing ones.
* Find it's being reverted here.

I really don't want to sound super dramatic, but it's been painful few hours. I'd like to provide some feedback, hope it's received with the good intentions that I'm writing it:

* It would have been useful to have a post install message on a breaking change that changes naming of spans. Or use a major version bump for breaking changes.
* When I found the culprit and saw [the Changelog](https://rubydoc.info/gems/opentelemetry-instrumentation-active_support/0.6.0/file/CHANGELOG.md), it says "BREAKING CHANGE: Custom ActiveSupport Span Names". Without looking at the code it seems like it means the span names can be configured, not that the format has changed. There is no link or mention of any PR, and the README doesn't say anything either. I have to look up the PR to see what it's about.
* I'd also like to explain why I find changing span names or attributes a big-ish deal. When looking at an individual trace the naming isn't important, but at least in our case, we have tail sampling rules or transformations that rely on the naming of spans and their attributes.

Having the option that was reverted here would have allowed me to upgrade the gems while I figure out and find time to update the rest of our otel-collector configuration. Without it I can't upgrade gems until I solve that. Again, it's not a big deal, but I find it preferable to be able to stay up to date on library upgrades automatically so we minimize the size of changes we deploy.

In any case, thanks a lot for the work you all do to provide and maintain these gems 🙏🏻

arielvalentin:
Thanks for the feedback. 

I will work on getting this change in and released on 2024-08-27

mrsimo:
Thank you! Much appreciated!

shioyama:
I'm really confused. Was this revert intended to be shipped in 0.7.3? Because AFAICT it's not there.

kaylareopelle:
Hi @shioyama, the revert was shipped in 0.7.2, but after @mrsimo's feedback, we released 0.7.3 with the default span name format feature. [Here's the diff.](https://github.com/open-telemetry/opentelemetry-ruby-contrib/compare/opentelemetry-instrumentation-action_view/v0.7.2...opentelemetry-instrumentation-action_view/v0.7.3) (though it displays changes for unrelated gems too)

Are you having trouble using the feature?

shioyama:
> Are you having trouble using the feature?

I'm trying to understand which versions have what, and it's very confusing. 

I think I understand now. The change in opentelemetry-instrumentation-active_support was what https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/1039 was meant to help handle for folks using opentelemetry-instrumetion-action_view. Reverting it was taking away the ability to customize behaviour, so the revert was reverted and the feature released in 0.7.3.

In our case it's fine, I've settled with just searching for both sets of tags for the foreseeable future, which works. Thanks for asking!

Issue #1097, opened 2024-07-31
Can't run tests in Docker Containers


matthewtusker:
I'm trying to follow [the instructions in the PG Instrumentation specs](https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/39056a3190561e87bfa4c5103508cffb0e133713/instrumentation/pg/test/opentelemetry/instrumentation/pg/instrumentation_test.rb#L14-L21) to run the tests, but the tests fail to load `pg` properly:

```
❯ docker-compose run ex-instrumentation-pg-test bundle exec rake test
[+] Creating 1/0
 ✔ Container opentelemetry-ruby-contrib-postgres-1  Running                                                                                         0.0s
/bundle/ruby/3.2.0/gems/activesupport-7.1.3.4/lib/active_support/core_ext/object/blank.rb:121: warning: method redefined; discarding old blank?
/bundle/ruby/3.2.0/gems/rubocop-1.64.1/lib/rubocop/core_ext/string.rb:19: warning: previous definition of blank? was here
<internal:/usr/local/lib/ruby/site_ruby/3.2.0/rubygems/core_ext/kernel_require.rb>:37:in `require': cannot load such file -- pg (LoadError)
        from <internal:/usr/local/lib/ruby/site_ruby/3.2.0/rubygems/core_ext/kernel_require.rb>:37:in `require'
        from /app/instrumentation/pg/test/opentelemetry/instrumentation/pg/instrumentation_test.rb:9:in `<top (required)>'
        from <internal:/usr/local/lib/ruby/site_ruby/3.2.0/rubygems/core_ext/kernel_require.rb>:37:in `require'
        from <internal:/usr/local/lib/ruby/site_ruby/3.2.0/rubygems/core_ext/kernel_require.rb>:37:in `require'
        from /bundle/ruby/3.2.0/gems/rake-13.2.1/lib/rake/rake_test_loader.rb:21:in `block in <main>'
        from /bundle/ruby/3.2.0/gems/rake-13.2.1/lib/rake/rake_test_loader.rb:6:in `select'
        from /bundle/ruby/3.2.0/gems/rake-13.2.1/lib/rake/rake_test_loader.rb:6:in `<main>'
rake aborted!
Command failed with status (1)
/bundle/ruby/3.2.0/gems/rake-13.2.1/exe/rake:27:in `<top (required)>'
/bundle/bin/bundle:108:in `load'
[...]

PR #1108, opened 2024-08-08
fix(pg): update versions to be tested (includes drop support for pg 1.2)

robbkidd:
* drop pg-1.2
    * This addresses the specific pg error found in #1097.
    * pg 1.2.x does not work under Ruby >= 3.0.
    * We currently support and test only Ruby >= 3.0.
* add pg-1.4
    * This used to be latest! Now latest in 1.5. 

With thanks to @matthewtusker for raising the issue and helping to resolve!


robbkidd:
Is this a `fix!:` change to bump the version number differently because of drop support?

arielvalentin (approved):

PR #1109, opened 2024-08-08
Feat: Add Mysql2 and Trilogy `db.collection.name` attribute

### LINKED ISSUE #1100: Add `db.collection.name` to mysql-based instrumentation libraries
The [db.collection.name](https://opentelemetry.io/docs/specs/semconv/database/database-spans/#:~:text=Required-,db.collection.name,-string) attribute is conditionally required for Database spans.

Update `mysql2` and `trilogy` instrumentations to emit this attribute.

hannahramadan:
The [db.collection.name](https://opentelemetry.io/docs/specs/semconv/database/database-spans/#:~:text=Required-,db.collection.name,-string) attribute is conditionally required for Database spans. This PR uses regex to detect the collection name and report it as the `db.collection.name` attribute for Mysql2 and Trilogy instrumentation.

Closes #1100


instrumentation/trilogy/lib/opentelemetry/instrumentation/trilogy/patches/client.rb
```diff
@@ -84,10 +89,16 @@ def client_attributes(sql = nil)
                 attributes[::OpenTelemetry::SemanticConventions::Trace::DB_STATEMENT] = sql
               end
             end
-
+            attributes.compact!
```
hannahramadan, line 92: I noticed other instrumentations were removing `nil` and thought Trilogy could benefit from this as well.

xuan-cao-swi: Just micro-optimization on `compact!` like this [comment](https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/693#discussion_r1360669417) and this [comment](https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/887#discussion_r1586410848) that you can consider it or leave it as it is.

xuan-cao-swi (approved):

instrumentation/trilogy/lib/opentelemetry/instrumentation/trilogy/patches/client.rb
hannahramadan: Thanks @xuan-cao-swi! It looks like we are doing `nil` checks in places, so using `compact!` maybe wasn't that helpful to begin with! I've removed this and added a `nil` check for `db.collection.name` before adding the attribute. https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/1109/commits/f9c244eaf83ee3e891748ad385363bc41644a48f

instrumentation/trilogy/test/opentelemetry/instrumentation/trilogy/instrumentation_test.rb
```diff
@@ -9,6 +9,16 @@
 require_relative '../../../../lib/opentelemetry/instrumentation/trilogy'
 require_relative '../../../../lib/opentelemetry/instrumentation/trilogy/patches/client'
 
+# This test suite requires a running mysql container and dedicated test container. We can use the same
```
kaylareopelle, line 12:
```suggestion
# This test suite requires a running mysql service within a dedicated test container. We can use the same
```

instrumentation/trilogy/test/opentelemetry/instrumentation/trilogy/instrumentation_test.rb
```diff
@@ -9,6 +9,16 @@
...
+# docker-compose file as the mysql2 instrumentation tests. The test container should have the mysql client.
```
kaylareopelle, line 13:
```suggestion
# docker-compose service as the mysql2 instrumentation tests. The test container should run the mysql client.
```

kaylareopelle (approved):
A few small wording things on the trilogy test instructions you can keep or leave. Other than that, looks great! Thank you!

instrumentation/trilogy/test/opentelemetry/instrumentation/trilogy/instrumentation_test.rb
hannahramadan: Thank you @kaylareopelle! Great suggestions.

instrumentation/mysql2/lib/opentelemetry/instrumentation/mysql2/patches/client.rb
```diff
@@ -83,9 +86,17 @@ def _otel_client_attributes
 
             attributes[SemanticConventions::Trace::DB_NAME] = _otel_database_name
             attributes[SemanticConventions::Trace::PEER_SERVICE] = config[:peer_service]
+            attributes['db.collection.name'] = collection_name(sql)
```
arielvalentin, line 89: I would like to avoid adding additional overhead to the instrumentation if possible. 

In cases where the SQL statement is omitted or not sanitized, there isn't any regexp scan that occurs. 

Is there a way to optimize it so that there isn't additional processing?

I would also like to consider that this `db.collection.name` matches newer versions of the OTel Schema and the instrumentations are still using pre-1.0 semantics. 

If I am not mistaken that would have been `db.table.name`. I think we should continue using pre-1.0 semantics until we have 
`OTEL_SEMCONV_STABILITY_OPT_IN` implemented

https://opentelemetry.io/docs/specs/semconv/database/database-spans/

arielvalentin (changes requested):

instrumentation/mysql2/lib/opentelemetry/instrumentation/mysql2/patches/client.rb
hannahramadan: Thanks @arielvalentin! It makes sense to keep consistent with pre-1.0 semantics. In this case, the attribute is `db.sql.table`. I've made this update! 

Re additional processing: I see your point about putting every SQL statement through a regexp scan. Because the table/collection name isn't available on the client, I'm not sure how else we'd be able to get this information. We could make table name an opt-in/opt-out attribute, but is that level of control something we want to provide? What do you think?

arielvalentin: I mean it's difficult to say. I feel like maybe this should only be done when the we also include the db.statement and I don't know if it would be possible to combine with the obfuscation code. 

Maybe I'm making too much of it? I'm not sure. 

Perhaps benchmarking will put me a bit at ease.

instrumentation/mysql2/lib/opentelemetry/instrumentation/mysql2/patches/client.rb
```diff
@@ -83,9 +91,16 @@ def _otel_client_attributes
 
             attributes[SemanticConventions::Trace::DB_NAME] = _otel_database_name
             attributes[SemanticConventions::Trace::PEER_SERVICE] = config[:peer_service]
+
             attributes
           end
 
+          def collection_name(sql)
+            Regexp.last_match[1] if sql =~ TABLE_NAME
```
arielvalentin, line 99: I could probably look this up myself but...

How does `Regexp.last_match` work? 

Is this thread safe? 

What happens if you are in a multi-threaded worker and an unrelated Regexp runs. Will it return the `last_match` from an unexpected regular expression?

hannahramadan: `Regexp.last_match(1)` is returning the first matched result and stopping. Since we just need the first match, this approach is faster than the previously used `.scan`, which was capturing every single match.

Re thread safety: I think so. Here's a [StackOverflow](https://stackoverflow.com/questions/9004245/is-regexp-last-match-thread-safe) answer that talks a bit about it. Because its being called inside an instance method, `Regexp.last_match()` should only be yielding the last capture that was evaluated inside the method, so no clobbering should be taking place.

instrumentation/mysql2/lib/opentelemetry/instrumentation/mysql2/patches/client.rb
hannahramadan: I don't know if only including the table name if we're recording the db.statement deviates from the [convention](https://opentelemetry.io/docs/specs/semconv/database/database-spans/#:~:text=Required-,db.collection.name,-string) (because as of now, the convention has it that recording the collection name isn't based on any condition, just availability).

The change to `Regexp.last_match(1)` and an updated regex string improved speed. In the following benchmark example, I used the MySQL obfuscation SQL code as a baseline, and found running the extra `db.collection.name` sql was 1.19x slower.

```
require 'benchmark/ipsa'

TABLE_NAME = /\b(?:(?:FROM|INTO|UPDATE)|(?:(?:CREATE|DROP|ALTER)\s+TABLE(?:\s+IF\s+(?:NOT\s+)?EXISTS)?))\s+["']?([\w.]+)["']?/i
MYSQL_OBFUSCATION = /(?-mix:'(?:[^']|'')*?(?:\\'.*|'(?!')))|(?-mix:"(?:[^"]|"")*?(?:\\".*|"(?!")))|(?-mix:(\$(?!\d)[^$]*?\$).*?(?:\1|$))|(?-mix:\{?(?:[0-9a-fA-F]\-*){32}\}?)|(?-mix:-?\b(?:[0-9]+\.)?[0-9]+([eE][+-]?[0-9]+)?\b)|(?i-mx:\b(?:true|false|null)\b)|(?-mix:0x[0-9a-fA-F]+)|(?i-mx:(?:#|--).*?(?=\r|\n|$))|(?m-ix:\/\*.*?\*\/)|(?-mix:q'\[.*?(?:\]'|$)|q'\{.*?(?:\}'|$)|q'\<.*?(?:\>'|$)|q'\(.*?(?:\)'|$))/
SQL = 'SELECT * FROM test_table'

def collection_name
  shared_operation
  Regexp.last_match(1) if SQL =~ TABLE_NAME
end

def no_collection_name
  shared_operation
  SQL
end

def shared_operation
  SQL.gsub(MYSQL_OBFUSCATION, '?')
end

Benchmark.ipsa do |x|
  x.report('collection_name') { collection_name }
  x.report('no_collection_name') { no_collection_name }

  x.compare!
end

Comment (hannahramadan) [reply instrumentation/mysql2/lib/opentelemetry/instrumentation/mysql2/patches/client.rb]: @arielvalentin We chatted about this in the SIG today and decided the reporting of table/collection names can be put behind a feature flag. What do you think? If that works, the remaining consideration is if the default on or off - do you have any preference on this?

Comment (hannahramadan) [reply instrumentation/mysql2/lib/opentelemetry/instrumentation/mysql2/patches/client.rb]: @arielvalentin made a suggestion for the config name `db_collection_name` and default value `include`, with `omit` as the other option: https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/1109/commits/83286689e8f288e5894bd3bcba08883606c30d6c

Comment (hannahramadan) [reply instrumentation/mysql2/lib/opentelemetry/instrumentation/mysql2/patches/client.rb]: Hi @arielvalentin! Wanted to check in and see if you had thoughts on the above.

Comment (arielvalentin) [reply instrumentation/mysql2/lib/opentelemetry/instrumentation/mysql2/patches/client.rb]: Thanks for your patience. I'm going to add this to my list to review by EoD tomorrow

**Comment by arielvalentin** [comment]
@hannahramadan Thank you for your patience waiting for my response. 

We discussed some of the details during the SIG on 2024-10-08 which I will share here:

## Default to `omit`

Adding additional regular expressions to extract data adds more overhead than I would like in high volume systems that are performance sensitive. I would like to make it so that users opt-in to this attribute. 

## Using Mixed Schema Versions

I think I mentioned this already in a separate comment, but I believe that our instrumentations are still emitting pre-1.0 schema attributes. I would have expected the attribute name to be [`db.sql.table`](https://github.com/open-telemetry/semantic-conventions/blob/96f8bda9bab363cb01e2441820bc83a5dad15801/docs/attributes-registry/db.md?plain=1#L231) and then once we had the DB Semconv Stability functionality it would emit `db.collection.name` and/or `db.sql.table` depending on the user selected option. 

I am of the opinion it may be a bit more confusing in this case to have the `db.statement` attribute along with the `db.collection.name`, when users I think would expect to see `db.query.text` instead; so, I think it is best to align on using deprecated attributes until we introduce the use of semconv stability. 

Is that not what we want?

## Config options are tied to a Schema Version

I think we set a bad precedent with setting individual attribute options because the names are now tied to a specific schema version. I am not certain what to do about that. 

Do we continue to have options when we move things along?  

## Remove Structural Duplication

Identical code is now going to appear in multiple instrumentations, which means that if we need to make changes it will require applying them to multiple locations. 

I think that it would be best to extract this functionality into a common library that could be shared across DB instrumentations. 

Though the gem is currently named `sql-obfuscation` I think it is the best place to include this functionality, where instead of it solely being responsible for sanitizing SQL, the gem could provide a helper that extracts or enriches a DB client span. It could use the instrumentation config to make decisions about what attributes to include and what logic to apply.  

Something like this:

```ruby
  SqlProcessor.append_db_attributes(span, sql, config)
```     

Given that this is more of a client side `processor` than it is a query sanitizer, then I think we should rename the gem to reflect that it is processing SQL and generating attributes for it.

An additional benefit to removing structural duplication is that it would reduce the number of places where we would have to add any DB attribute semconv stability compatibility.

## Are Span Processors out of the question?

This is something we did not discuss.

With the introduction of [`OnEnding`](https://opentelemetry.io/docs/specs/otel/trace/sdk/#onending), the specification leaves room for us to do some Span Processing _outside_ of the instrumentation code. 

This would mean that instrumentations could all emit the SQL and then the `SqlProcessor` could be run in `OnEnding` and potentially reduce some of the complexity in the DB instrumentations. 

The processor could extract attributes from the SQL and sanitize or omit the statements. This would change configurations so that these options could not be configured on the instrumentations anymore but rather in the span processor.  

I think that trade off may be acceptable since it is unlikely we will want to have different configurations for different SQL datastores. 

## All that being said 

We agreed that I would unblock this on the condition that this attribute be `omitted` by default and move forward with structural duplication and refactor the code in a future PR.

instrumentation/mysql2/lib/opentelemetry/instrumentation/mysql2/patches/client.rb
```diff
@@ -83,9 +91,16 @@ def _otel_client_attributes
...
             attributes[SemanticConventions::Trace::DB_NAME] = _otel_database_name
             attributes[SemanticConventions::Trace::PEER_SERVICE] = config[:peer_service]
+
             attributes
           end
 
+          def db_sql_table_name(sql)
+            Regexp.last_match(1) if sql =~ TABLE_NAME
+          rescue StandardError
+            nil
```
arielvalentin, line 101: If an error occurs and the attribute is set to `nil`, then the SDK will report an error. It would be best to avoid setting any invalid attributes. 

Please ensure that the attribute is not set in cases where the table name could not be extracted. 

https://github.com/open-telemetry/opentelemetry-ruby/blob/555b062ef9421784c132aa9b97b29ec637b13b0f/sdk/lib/opentelemetry/sdk/trace/span.rb#L277

https://github.com/open-telemetry/opentelemetry-ruby/blob/555b062ef9421784c132aa9b97b29ec637b13b0f/sdk/lib/opentelemetry/sdk/internal.rb#L57

arielvalentin (changes requested):

instrumentation/mysql2/lib/opentelemetry/instrumentation/mysql2/patches/client.rb
hannahramadan: Made some changes! Will log an error if trouble getting the table name and won't report. Is the error logging here okay?

instrumentation/mysql2/lib/opentelemetry/instrumentation/mysql2/patches/client.rb
```diff
@@ -98,6 +98,7 @@ def _otel_client_attributes(sql)
           def db_sql_table_name(sql)
             Regexp.last_match(1) if sql =~ TABLE_NAME
           rescue StandardError
```
arielvalentin, line 100:
```suggestion
          rescue StandardError => e
```

hannahramadan:
Thank you @arielvalentin and apologies for the delayed response here!

To your point, I think it may make sense to nail down how we want the design to look and then add this attribute.
> I think that it would be best to extract this functionality into a common library that could be shared across DB instrumentations.

Perhaps we:
1. Rename the `sql-obfuscation` to `sql-processor`
2. Move the logic to collect table/collection name there. Attributes that require sql parsing logic will also live there

For table/collection name attribute:

- **Default to omit** - I agree this should be and opt-in attribute.
- **Using Mixed Schema Versions** - Also a good call out. Agree. 
- **Config options are tied to a Schema Version** - We talked about potentially having levels, [like collector internal metrics](https://opentelemetry.io/docs/collector/internal-telemetry/#lists-of-internal-metrics). I'm not sure there are enough configs/options to to warrant grouping into levels, but what do we think about a config that accepts any array of values to collect? Old and new semantic versions could be accepted and aliased.

hannahramadan:
Since this PR was opened, guidelines were updated (https://github.com/open-telemetry/semantic-conventions/commit/3fa52acca3ec1530f9b88a38a29bfd876ec2205d) to **NOT**  extract the collection name from `db.query.text` when the database system supports query text with multiple collections in non-batch operations. This PR is doing that and should be closed.

PR #1120, opened 2024-08-15
fix(active-job): Prefix ::ActiveSupport when installing the instrumentation

lavoiesl:
Otherwise, `ActiveSupport` ends up resolving to `OpenTelemetry::Instrumentation::ActiveSupport`.

`OpenTelemetry::Instrumentation::Registry#install_instrumentation` rescues the exception, so the prepend ends up being (somewhat) silently dropped.

This results in the patch not being adding, which causes `OpenTelemetry.propagation.inject(job.__otel_headers)` to raise, dropping the span


arielvalentin (approved):

arielvalentin:
cc: @bensheldon

PR #1128, opened 2024-08-19
🔬 Metrics Prototype: Conditions everywhere edition

kaylareopelle:
A potential way to integrate metrics into the Instrumentation:
* If the OpenTelemetry::Metrics API is installed, go ahead and make things in base
* Check for a configuration option, `:send_metrics` to be `true` before generating telemetry
* Still very messy, do we like this general approach?


kaylareopelle:
I think this is too clunky. I'm going to put forth #1129 for an in-depth review instead.

PR #1129, opened 2024-08-19
feat: 🦠 Metrics prototype for Rack instrumentation

### LINKED ISSUE #1127: Determine the structure for Metrics instrumentation
Metrics are still veryyy experimental, but, for those willing to take the risk, I'd like to start building instrumentation that can emit metrics. 

Before we can do that, we need to decide how to keep the stable traces code safe from experimental metrics.

Current ideas:
* More than one condition needs to be true to turn on metrics (show that you really really want this)
  * First requirement: The presence of the Metrics API is detected
  * Second requirement: The configuration enables metrics 
* Only emit HTTP metrics that 
 
Potential approaches:
* Similar to the ConfigurationPatch that allows the Metrics SDK to use the Configurator, create a separate module that only gets loaded if the conditions to enable metrics are met 
* Put conditionals everywhere to check to see if metrics should be generated 
* Create separate libraries to emit metrics (not a huge fan on this one)

kaylareopelle:
Here's a prototype to add the `http.server.request.duration` metric to Rack instrumentation. 

The base gem looks for the presence of the `opentelemetry-metrics-api`, by checking to see if `OpenTelemetry::Metrics` is defined before it creates a meter. Then, the instrumentation checks for the `:send_metrics` configuration to be `true`. If both conditions are true, metrics will be generated. Guard clauses are used to determine whether the metrics-related code should run.

In order to test this with an application, you need to patch the `OpenTelemetry::SDK::Configurator#metrics_configuration_hook` to add a meter provider and metric reader before configuring the SDK. Here's an example that uses the periodic metric reader and the OTLP metrics exporter:
```ruby 
module MetricsPatch
  def metrics_configuration_hook
    OpenTelemetry.meter_provider = OpenTelemetry::SDK::Metrics::MeterProvider.new(resource: @resource)
    OpenTelemetry.meter_provider
      .add_metric_reader(
        OpenTelemetry::SDK::Metrics::Export::PeriodicMetricReader.new(
          exporter: OpenTelemetry::Exporter::OTLP::Metrics::MetricsExporter.new,
          export_interval_millis: 3000,
          export_timeout_millis: 10000))
  end
end

OpenTelemetry::SDK::Configurator.prepend(MetricsPatch)

    c.use 'OpenTelemetry::Instrumentation::Rack', { send_metrics: true }
end
``` 

The attributes assigned to the `http.server.request.duration` metric are drawn from the span attributes. When `OTEL_SEMCONV_STABILITY_OPT_IN` is ready, we should update the instrumentation to emit the stability-related attributes for both spans and metrics.

To use this with another instrumentation that installs Rack (ex. Sinatra), you'll need to update the gemspec for the instrumentation that installs Rack so that the `opentelemetry-instrumentation-base` and `opentelemetry-instrumentation-rack` dependencies aren't installed.

Closes #1127


instrumentation/base/lib/opentelemetry/instrumentation.rb
```diff
@@ -7,6 +7,7 @@
 require 'opentelemetry'
 require 'opentelemetry-registry'
 require 'opentelemetry/instrumentation/base'
+require 'opentelemetry/instrumentation/metrics_patch' if defined?(OpenTelemetry::Metrics) # maybe also add Env var check?
```
kaylareopelle, line 10: I'm thinking an environment variable check might be handy at this stage because the config gets registered later, so I wouldn't want to check for a `:send_metrics` config value here, whereas environment variables should be easily accessible.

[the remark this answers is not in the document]
xuan-cao-swi: I think these two safeguards are enough for users who don’t want metrics in instrumentation or wish to create custom metrics themselves (e.g., include the metrics SDK but not use it in instrumentation).

Is the patch (e.g., prepend) only because metrics are still experimental features? I was thinking of including them in the base but not installing if `OpenTelemetry::Metrics` is missing (similar to line 25 in `metrics_patch.rb`).

instrumentation/rack/lib/opentelemetry/instrumentation/rack/middlewares/metrics_patch.rb
```diff
@@ -0,0 +1,58 @@
...
+            OpenTelemetry::Instrumentation::Rack::Instrumentation.instance.meter
+          end
+
+          def http_server_request_duration_histogram
+            # TODO: Add Advice to set small explicit histogram bucket boundaries
+            # TODO: Does this need to be memoized?
+            @http_server_request_duration_histogram ||= meter.create_histogram('http.server.request.duration', unit: 's', description: 'Duration of HTTP server requests.')
+          end
+
+          def record_http_server_request_duration_metric(span)
```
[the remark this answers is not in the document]
xuan-cao-swi: If the metrics attribute only cares about HTTP-related attributes, can it emit metrics with the attributes that are based on the existing `span.attributes`? I'm thinking something like:
```ruby
# Assume only one of url.scheme and http.scheme will appear
span.attributes.select { |key, value| key.match(/http\..*/) || key.match(/url\..*/) }
```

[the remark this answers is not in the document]
kaylareopelle: I like where you're headed with this! There's a lot of duplication in the current code. However, there are some `http.*` and `url.*` attributes present on the span that should not be attached to metrics (ex. `url.full` is on spans, but not metrics). 

I wonder if we could use a key-based allowlist to select things instead? Maybe add a constant with all the HTTP metrics keys, and we search for those? 

Things might need to shift again when the `OTEL_SEMCONV_STABILITY_OPT_IN` variable is out in the world, but we can address that later.

instrumentation/base/lib/opentelemetry/instrumentation.rb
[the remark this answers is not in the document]
kaylareopelle: Thank you! I think that could work! Yes, the patch is just because the metrics are experimental features. I thought this approach might feel "safer" (since all our instrumentation runs through `base`), but I don't know if truly provides a benefit.

It also tries to follow the configurator patches for metrics in the SDK: 
* https://github.com/open-telemetry/opentelemetry-ruby/blob/main/metrics_sdk/lib/opentelemetry/sdk/metrics/configuration_patch.rb 
* https://github.com/open-telemetry/opentelemetry-ruby/blob/main/sdk/lib/opentelemetry/sdk/configurator.rb#L150 

I'm happy to cut out the patch and just put everything in base. [I sorta tried that with the other prototype](https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/1128/files#diff-7379de806670e7a913ebbe1d8e27945a6367cc83c94d79e55434de15cdd16732).

instrumentation/base/lib/opentelemetry/instrumentation/base.rb
```diff
@@ -4,6 +4,7 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 
+
```
kaylareopelle, line 7:
```suggestion
```

instrumentation/rack/lib/opentelemetry/instrumentation/rack/instrumentation.rb
```diff
@@ -29,8 +29,9 @@ class Instrumentation < OpenTelemetry::Instrumentation::Base
         option :untraced_requests,        default: nil,   validate: :callable
         option :response_propagators,     default: [],    validate: :array
         # This option is only valid for applications using Rack 2.0 or greater
-        option :use_rack_events,          default: true, validate: :boolean
-
+        option :use_rack_events,          default: true,  validate: :boolean
+        # TODO: This option currently exclusively uses the event handler, should we support old and new Rack?
+        option :send_metrics,             default: false, validate: :boolean
```
kaylareopelle, line 34:
```suggestion
        option :send_metrics,             default: false, validate: :boolean

```

instrumentation/rack/lib/opentelemetry/instrumentation/rack/instrumentation.rb
```diff
@@ -47,6 +48,7 @@ def middleware_args
           end
         end
 
+
```
kaylareopelle, line 51:
```suggestion
```

instrumentation/rack/lib/opentelemetry/instrumentation/rack/middlewares/event_handler.rb
```diff
@@ -158,6 +158,7 @@ def untraced_request?(env)
             false
           end
 
+
```
kaylareopelle, line 161:
```suggestion
```

instrumentation/rack/lib/opentelemetry/instrumentation/rack/middlewares/event_handler.rb
```diff
@@ -262,6 +264,38 @@ def create_span(parent_context, request)
...
+          def meter
+            return unless metrics_enabled?
+
+            OpenTelemetry::Instrumentation::Rack::Instrumentation.instance.meter
+          end
+
+          def http_server_request_duration_histogram
+            return unless metrics_enabled?
+
+            @http_server_request_duration_histogram ||= meter.create_histogram('http.server.request.duration', unit: 's', description: 'Duration of HTTP server requests.')
```
kaylareopelle, line 284:
```suggestion
            @http_server_request_duration_histogram ||= meter.create_histogram(
              'http.server.request.duration', 
              unit: 's', 
              description: 'Duration of HTTP server requests.'
            )
```

instrumentation/rack/lib/opentelemetry/instrumentation/rack/middlewares/event_handler.rb
```diff
@@ -262,6 +264,38 @@ def create_span(parent_context, request)
...
+          end
+
+          def record_http_server_request_duration_metric(span)
+            return unless metrics_enabled?
+            # find span duration
+            # end - start / a billion to convert nanoseconds to seconds
+            duration = (span.end_timestamp - span.start_timestamp) / Float(10**9)
+            # glean attributes
+            attrs = span.attributes.select {|k, _v| HTTP_SERVER_REQUEST_DURATION_ATTRS_FROM_SPAN.include?(k) }
```
kaylareopelle, line 293:
```suggestion
            attrs = span.attributes.select { |k, _v| HTTP_SERVER_REQUEST_DURATION_ATTRS_FROM_SPAN.include?(k) }
```

kaylareopelle:
@xuan-cao-swi  - Thank you for the feedback! I've updated the prototype to use conditions instead of patches and to add the attributes from the spans using `select` instead of adding each to a hash individually. Please take another look when you have a chance!

instrumentation/rack/lib/opentelemetry/instrumentation/rack/middlewares/event_handler.rb
```diff
@@ -262,6 +263,43 @@ def create_span(parent_context, request)
...
+
+            @http_server_request_duration_histogram ||= meter.create_histogram(
+              'http.server.request.duration',
+              unit: 's',
+              description: 'Duration of HTTP server requests.'
+            )
+          end
+
+          def record_http_server_request_duration_metric(span)
+            return unless metrics_enabled?
```
arielvalentin, line 291: I have a bunch of ideas and questions floating around in my head that could use your input being a metrics SDK expert. 

It may result in a separate dispatch call, but have you considered creating a separate handler for metric generation? That would remove the need for having conditionals in the tracer middleware.

What if our user has an o11y 1.0 mindset and only wanted to generate metrics and not traces? Would that not mean the span would be non_recording and the metric would always generate invalid histogram entries?

Is there anything in the spec that states this should be generated from span timestamps?

Is there an alternative implementation that generates metrics in the span processor pipeline?

E.g. there is a a processor that implements on_finish generated metrics for server and client spans regardless of the instrumentation?

What about one that decorates the BatchSpanProcessor export loop and generates metrics in the bsp thread to minimize the metrics from being generated in the critical path of the users request?

instrumentation/rack/lib/opentelemetry/instrumentation/rack/middlewares/metrics_patch.rb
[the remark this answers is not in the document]
xuan-cao-swi: Thanks for the updates. Yes, I like key-based allowlist `HTTP_SERVER_REQUEST_DURATION_ATTRS_FROM_SPAN`.
In the future, I think we can even have something like `HTTP_SERVER_REQUEST_DURATION_ATTRS_FROM_SPAN_V1_2` (`V1_2` = semantic version 1.2)

instrumentation/rack/lib/opentelemetry/instrumentation/rack/middlewares/event_handler.rb
xuan-cao-swi: > Is there anything in the spec that states this should be generated from span timestamps?

I don't think there is spec about where we should generate metrics. afaik, metrics spec only talks about what the metrics should looks like and how to export them.

> Is there an alternative implementation that generates metrics in the span processor pipeline?

To generate the metrics from span processor, then the processor need have meter and create the instrument in processor in on_start or on_finish, I am thinkhing something like this?
```ruby
def initialize
  ...
  @meter = ::OpenTelemetry.meter_provider.meter('sample_meter')
end

def on_start
  ...
  @histogram = @meter.create_histogram('histogram_name', unit: 'ms', description: 'measure sample data')
end

def on_finish
  ...
  span_time = span.end_timestamp - span.start_timestamp
  @histogram.record(span_time, attributes: {})
end
```

arielvalentin: Perhaps. 

What are other language sigs doing?

Is what I'm suggesting seemingly out of touch with what is the consistent pattern across language implementations?

instrumentation/rack/lib/opentelemetry/instrumentation/rack/middlewares/metrics_patch.rb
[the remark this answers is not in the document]
kaylareopelle: ooh, nice idea!

instrumentation/base/lib/opentelemetry/instrumentation.rb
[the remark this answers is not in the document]
kaylareopelle: Whoops, forgot to follow up here. I've changed the structure to use conditions instead of the patches.

instrumentation/rack/lib/opentelemetry/instrumentation/rack/middlewares/event_handler.rb
kaylareopelle: Love these questions! I have some research to do to follow-up, but here are my current thoughts.

The instrumentation is based on [Python's approach with the aiohttp-server](https://github.com/fidelity-contributions/open-telemetry-opentelemetry-python-contrib/blob/main/instrumentation/opentelemetry-instrumentation-aiohttp-server/src/opentelemetry/instrumentation/aiohttp_server/__init__.py). The [asgi instrumentation](https://github.com/fidelity-contributions/open-telemetry-opentelemetry-python-contrib/blob/main/instrumentation/opentelemetry-instrumentation-asgi/src/opentelemetry/instrumentation/asgi/__init__.py) has an example of what we could do when `OTEL_SEMCONV_STABILITY_OPT_IN` comes around. 

> It may result in a separate dispatch call, but have you considered creating a separate handler for metric generation? That would remove the need for having conditionals in the tracer middleware.

I think we could take that approach! The examples I've seen so far tend to record the metric as part of the span creation, but we don't have to and that might be safer until metrics are stable.

> What if our user has an o11y 1.0 mindset and only wanted to generate metrics and not traces? Would that not mean the span would be non_recording and the metric would always generate invalid histogram entries?

This is a great thing to consider. I think that'd be the case with this current implementation, and I'd like to make some tweaks for this scenario.

I don't see an option to create only metrics and not traces in the Python instrumentation, but perhaps this exists in other languages. I'll do a bit more digging. My read on the spec is that it should leverage the span if it's available, but if the span is not available, it should still record the metric. [spec](https://opentelemetry.io/docs/specs/semconv/http/http-metrics/#metric-httpserverrequestduration) 

> Is there anything in the spec that states this should be generated from span timestamps?

The spec says: "When this metric is reported alongside an HTTP server span, the metric value SHOULD be the same as the HTTP server span duration." 

AFAIK, we don't save duration on a span, just a start and end timestamp, so I figured this would be the way to create a duration that's the same as the HTTP server span duration. I might be interpreting the spec incorrectly. We could get the start and end timestamps in an independent fashion. That may be closer to what Python does.

> Is there an alternative implementation that generates metrics in the span processor pipeline?

I haven't seen this, and this isn't something I considered. I'll poke around other languages and see if I can find an example. I haven't looked too much at the BSP metrics, I'll take a look to get a better sense of how they work.

kaylareopelle: @arielvalentin - Thanks for your patience!

I spent some time looking at metrics instrumentation in other languages. The most common design I found was to create metrics within the same instrumentation file as spans, but to calculate the duration and collect the attributes in separate ways. It seems like metrics are stable in those languages, so this might not have always been the case. 
When it comes to creating a separate handler for metrics, I don't feel strongly one way or the other. Would you prefer to keep them in separate handlers? 

Here are some of the files I looked at. The links should be to the code that records the metric. In most cases, the same file also creates the histogram.
* [Python - aiohttp](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/instrumentation/opentelemetry-instrumentation-aiohttp-server/src/opentelemetry/instrumentation/aiohttp_server/__init__.py#L237)
* [Javascript - undici](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/23aae0132506fdb00fe30c8b143c24bd4f5a706a/plugins/node/instrumentation-undici/src/undici.ts#L462-L467) 
* [.NET - owin](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/177748ed2b5d216954b24d45a8511d9963288b1b/src/OpenTelemetry.Instrumentation.Owin/Implementation/DiagnosticsMiddleware.cs#L187-L192)
* [Go - net/http](https://github.com/Sovietaced/opentelemetry-go-contrib/blob/main/instrumentation/net/http/otelhttp/handler.go#L241)

For the span processor question, I didn't come across use of a span processor to collect metrics. Since span processors are part of the SDK and not the API, and instrumentation should only rely on the API this doesn't seem like quite the right fit. In addition, since spans and metrics should be independent of each other, I don't think we want a span processor to control the creation of another signal. What do you think?  

It looks like there are some features coming with the [meter configurator](https://opentelemetry.io/docs/specs/otel/metrics/sdk/#meterconfigurator) that allow users to control what data is collected by instrumentation scope. This could help someone control what telemetry they want from a given instrumentation scope, and support that Olly 1.0 mindset of sending only metrics, but not traces. I'll look into prototyping the experimental spec and use that to control whether metrics and traces are sent. We could probably do this with configs in the instrumentation too.

So as a follow-up, I'd like to:
* Separate metrics from traces in the instrumentation. The metrics data should not use the span directly for its attributes/duration. 
* See what it's like to make an event handler specifically for metrics
* Prototype the experimental metrics configuration features to control whether metrics/traces are sent

Let me know if there's anything I missed! Thanks again for your feedback!

kaylareopelle: @arielvalentin @xuan-cao-swi - I have a prototype ready that creates metrics in a separate event handler: https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/1213 

I decided to put it in a different PR to start to make it easier to compare the two approaches. 

Next on my list is to prototype the metrics configuration work, but that PR will likely live in the core repo.

Issue #1131, opened 2024-08-20
active_job: Context propagation is broken


lavoiesl:
**Description of the bug**

I believe https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/677#discussion_r1723998717 caused Context propagation to break.

This is a bit hard to detect since a link is created if the execution is within a span, but it links to the parent span rather than the span newly created

**Share details about your runtime**

Operating system details: Linux, Ubuntu 20.04 LTS
RUBY_ENGINE: "ruby"
RUBY_VERSION: "3.3.4"
RUBY_DESCRIPTION: "ruby 3.3.4 (2024-07-09 revision be1089c8ec) [arm64-darwin23]"

**Share a simplified reproduction if possible**

```rb
require "bundler/inline"

gemfile(true) do
  source "https://rubygems.org"

  gem "opentelemetry-api"
  gem "opentelemetry-sdk"
  gem "opentelemetry-instrumentation-active_job", "0.7.6"
  gem "activejob", "7.1.3.4"
  gem "activesupport", "7.1.3.4"
end

require "opentelemetry-api"
require "opentelemetry-sdk"
require "opentelemetry-instrumentation-active_job"
require "active_job"

EXPORTER = OpenTelemetry::SDK::Trace::Export::InMemorySpanExporter.new
span_processor = OpenTelemetry::SDK::Trace::Export::SimpleSpanProcessor.new(EXPORTER)

OpenTelemetry::SDK.configure do |c|
  c.add_span_processor(span_processor)
  c.use("OpenTelemetry::Instrumentation::ActiveJob", { propagation_style: :link, span_naming: :queue })
end

SimpleTracer = OpenTelemetry.tracer_provider.tracer("Bug Report")

class MyJob < ActiveJob::Base
  def perform
  end
end

class TestAdapterTest < ActiveSupport::TestCase
  setup do
    EXPORTER.reset
  end

  test "test adapter" do
    singleton_class.include(ActiveJob::TestHelper)
    ActiveJob::Base.queue_adapter = :test

    MyJob.perform_later
    perform_enqueued_jobs

    publish, process = EXPORTER.finished_spans.sort_by(&:start_timestamp)

    assert_nil(publish.links)

    # Fails, no links
    assert_not_nil(process.links)
    assert_equal(publish.hex_span_id, process.links.first.span_context.hex_span_id)
  end

  test "test adapter + in_span" do
[...]

PR #1132, opened 2024-08-21
fix(active-job): Propagate context between enqueue and perform

lavoiesl:
Fixes https://github.com/open-telemetry/opentelemetry-ruby-contrib/issues/1131

This fixes two issues:
1. Sets the current context before injecting for propagation
2. Specify `with_parent: parent_context` when performing

Those cases were not caught by tests because they were using the "async immediate" queue adapter, which executes within the same thread. It was working by accident. 

Added tests that match the report at https://github.com/open-telemetry/opentelemetry-ruby-contrib/issues/1131


arielvalentin (approved):

PR #1149, opened 2024-09-04
fix!: Return message when sql is over the obfuscation limit

### LINKED ISSUE #1146: `sql-obfuscation` does not sanitize SQL that exceed size limits
**Description of the bug**

A recent change to how SQL query comments are pre-prepended to the statement has resulted in triggering logic in the SQL obfuscation helper that bypasses executing the regular expression to sanitize the substring of query. 

https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/f817d6b941d0e43215c59e7d5c90f778ad824ee1/helpers/sql-obfuscation/lib/opentelemetry/helpers/sql_obfuscation.rb#L118

The regular expression does not seem to match on the comments index and ends up returning the raw contents of the SQL: 

```
/*service.name:foo,deployment.environtment:production,tracecontext:00-4bf92f3577b34da6a3ce929d0e0e4736-00f067aa0ba902b7-00,rails.route:examples/bars#index,host.name:baz-abc123.example.com*/ SELECT user.id FROM users where user.login = 'secretUserNameThatShouldBeObfuscated'... SQL truncated (> 2000 characters)
```
We must ensure that SQL is sanitized or omit the statement entirely. 

Operating system details: Linux, Ubuntu 20.04 LTS
RUBY_ENGINE: "ruby"
RUBY_VERSION: "3.3.4"
RAILS_VERSION: "8.0.0.alpha"

kaylareopelle:
When SQL queries had prepended comments and were over the configured obfuscation limit, the query would fail to obfuscate and sensitive data could be exposed. 

The previous approach was incompatible because the first index with a match would be the zero index, so the range would return the entire SQL string and append the truncation message to the end of it. 

Until we can find a way to make the first match approach work with prepended queries, let's return a message stating the query was over the limit.

Resolves #1146

### DIFF
--- a/helpers/sql-obfuscation/lib/opentelemetry/helpers/sql_obfuscation.rb
+++ b/helpers/sql-obfuscation/lib/opentelemetry/helpers/sql_obfuscation.rb
@@ -20,7 +20,7 @@ module OpenTelemetry
     #    `option :db_statement, default: :include, validate: %I[omit include obfuscate]`
     #  * `:obfuscation_limit`
     #    Example:
-    #    `option :obfuscation_limit, default: 2000, validate: :integer`
+    #    `option :obfuscation_limit, default: 16384, validate: :integer`
     #
     # If you want to add support for a new adapter, update the following
     # constants to include keys for your adapter:
--- a/helpers/sql-obfuscation/lib/opentelemetry/helpers/sql_obfuscation.rb
+++ b/helpers/sql-obfuscation/lib/opentelemetry/helpers/sql_obfuscation.rb
@@ -94,7 +94,7 @@ module OpenTelemetry
       # This is a SQL obfuscation utility intended for use in database adapter instrumentation.
       #
       # @param sql [String] The SQL to obfuscate.
-      # @param obfuscation_limit [optional Integer] The maximum length of an obfuscated sql statement.
+      # @param obfuscation_limit [optional Integer] the length at which the SQL string will not be obfuscated
       # @param adapter [optional Symbol] the type of database adapter calling the method. `:default`, `:mysql` and `:postgres` are supported.
       # @return [String] The SQL query string where the values are replaced with "?". When the sql statement exceeds the obufscation limit
       #  the first matched pair from the SQL statement will be returned, with an appended truncation message. If trunaction is unsuccessful,
@@ -102,7 +102,7 @@ module OpenTelemetry
       #
       # @api public
       def obfuscate_sql(sql, obfuscation_limit: 2000, adapter: :default)
-        return "SQL truncated (> #{obfuscation_limit} characters)" if sql.size > obfuscation_limit
+        return "SQL not obfuscated, query exceeds #{obfuscation_limit} characters" if sql.size > obfuscation_limit
 
         regex = case adapter
                 when :mysql
--- a/helpers/sql-obfuscation/lib/opentelemetry/helpers/sql_obfuscation.rb
+++ b/helpers/sql-obfuscation/lib/opentelemetry/helpers/sql_obfuscation.rb
@@ -62,8 +62,6 @@ module OpenTelemetry
                       hexadecimal_literals comments multi_line_comments]
       }.freeze
 
-      PREPENDED_COMMENT_REGEX = %r{^/\*.*\*/}
-
       PLACEHOLDER = '?'
 
       # We use these to check whether the query contains any quote characters
@@ -116,8 +114,8 @@ module OpenTelemetry
         # Original MySQL UTF-8 Encoding Fixes:
         # https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/160
         # https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/345
+        return "SQL truncated (> #{obfuscation_limit} characters)" if sql.size > obfuscation_limit
         sql = OpenTelemetry::Common::Utilities.utf8_encode(sql, binary: true)
-        return truncate_statement(sql, regex, obfuscation_limit) if sql.size > obfuscation_limit
 
         sql = sql.gsub(regex, PLACEHOLDER)
         return 'Failed to obfuscate SQL query - quote characters remained after obfuscation' if CLEANUP_REGEX[adapter].match(sql)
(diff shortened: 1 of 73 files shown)


helpers/sql-obfuscation/lib/opentelemetry/helpers/sql_obfuscation.rb
```diff
@@ -127,6 +129,8 @@ def obfuscate_sql(sql, obfuscation_limit: 2000, adapter: :default)
 
       # @api private
       def truncate_statement(sql, regex, limit)
+        sql = sql.gsub(PREPENDED_COMMENT_REGEX, PLACEHOLDER) if sql.match?(PREPENDED_COMMENT_REGEX)
```
arielvalentin, line 132: @kaylareopelle I think what we need here is to obfuscate the substring and amend the truncation message. 

Something like 

```ruby

if sql.length >= limit
  sql = sql[0..limit-1] + "SQL truncated (> #{limit} characters)"
end

sql.gsub(regex)
```

kaylareopelle, line 132: The problem seems to be that `sql.index(regex)` returns `0` when a prepended comment is present, so our trick to truncate the message at the first match doesn't work, and the un-obfuscated query is preserved. 

We could ditch the query altogether and just return the truncation message by altering the condition on line 136: 
```diff 
-        return truncation_message unless first_match_index
+       return truncation_message unless first_match_index && first_match_index > 0
```

But then the user would get something like "SQL truncated (> 2000 characters)" rather than a snippet of the query until the first match.

There might be more performant ways to do this, I'd love to hear other ideas.

My benchmark is a little wacky, so I'll just share the results for now. I can post it if anyone is interested. 

Here are the current results:
| case | i/s |
| ---- | ---- |
|**with comment, altered return on line 136** |  295981.8 i/s |
|**with comment, current code** (this has an unobfuscated result) |  262156.7 i/s - 1.13x slower |
|**without comment, current code** |    92768.5 i/s - 3.19x slower |
|**without comment, altered return on line 136** |  87121.0 i/s - 3.40x slower |
|**without comment, with PREPENDED_COMMENT_REGEX checks** |   79018.5 i/s - 3.75x slower |
|**with comment, with PREPENDED_COMMENT_REGEX** |   66864.0 i/s - 4.43x slower |

kaylareopelle: I'll try it out!

kaylareopelle: @arielvalentin - With this approach, if the limit is hit in the middle of a value that needs to be obfuscated, the obfuscation will fail for that value. 

The string will miss the closing part of the capture group, so the value won't be replaced with the placeholder.

This can be prevented by running the query through the CLEANUP_REGEX before sending it out, but every time the query was truncated within a sensitive value, `sql.statement` would equal `'Failed to obfuscate SQL query - quote characters remained after obfuscation'`.

I'll keep working on this tomorrow to see if there's another way around this problem, let me know if you have other ideas!

Just to clarify what we're working toward, it sounds like the ideal situation would be
* truncate the query to the limit
* obfuscate everything within that truncated string

Does that sound right?

arielvalentin: I suppose. 

I think if this gets too complicated, I think we should highly consider not processing the SQL at all and return the placeholder value.

I think we may also want to consider increasing the size limit. 

I don't know why we landed in 2K characters but it seems arbitrary. 

Maybe we publish benchmarks or something to give folks an idea of what to expect? 

IDK.

kaylareopelle: It looks like the 2000 limit [has been around since the beginning.](https://github.com/open-telemetry/opentelemetry-ruby-contrib/commit/ccfd3d19194bb8b55e93e96adfdd600e28b30637)

Here's the PR that introduced the current truncation style: https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/224 
[NR's limit is 16384](https://github.com/newrelic/newrelic-ruby-agent/blob/dev/lib/new_relic/agent/database.rb#L12). So we could consider changing the limit.

I like the idea of publishing benchmarks. When we decide what we want to do with the truncated queries, I'll add one that toggles the various configs.

> I think if this gets too complicated, I think we should highly consider not processing the SQL at all and return the placeholder value.

I agree. I'm still experimenting with different strategies, but have this timeboxed.

arielvalentin: I love how when we look at old PRs, I was the approver 🤦🏼 

I only have myself to blame.

arielvalentin: The original, original, original one 

https://github.com/aislinnnnn/opentelemetry-ruby/commit/1bd5e24fd4b7346b281dbb72d9b34dbab360f361

There isn't much context on the size choice. 

> [NR's limit is 16384](https://github.com/newrelic/newrelic-ruby-agent/blob/dev/lib/new_relic/agent/database.rb#L12). So we could consider changing the limit.

Agreed.

kaylareopelle: It looks like for OTel, [NR limits attribute values to 4095](https://github.com/newrelic/docs-website/blob/develop/src/content/docs/opentelemetry/best-practices/opentelemetry-otlp.mdx?plain=1#L209). So 4000 might be a better number here? 🤷

I can check in with other SIG members next week to see if they have suggestions on a reasonable limit.

arielvalentin:
Ready?

kaylareopelle:
Whoops, yes, @arielvalentin! Opened now.

arielvalentin:
cc: @reid-rigo Making you aware that we are reverting this functionality.

We will be looking to restore it in the near future, but we have identified an issue with sanitizing SQL queries that prepend comments, so we have to revert it now.

arielvalentin (approved):

mwear (approved):

xuan-cao-swi (approved):

PR #1176, opened 2024-09-20
chore: update env var in vitess docs

plantfansam:
Hopefully I can save someone some time 😄


kaylareopelle (approved):
Good find! Thank you!

PR #1178, opened 2024-09-20
fix: update references to logging exporter

codeboten:
This exporter has been replaced by the debug exporter and will be removed soon

Related to https://github.com/open-telemetry/opentelemetry-collector/pull/11037


kaylareopelle (changes requested):
Hi @codeboten! 

Thanks for your contribution! 

Unfortunately, this repo has similar problems to opentelemetry-ruby, in that it also uses a version of the collector that predates the debug exporter (though this repo uses 0.54.0, rather than 0.35.0).

See this comment on the other PR for more details: https://github.com/open-telemetry/opentelemetry-ruby/pull/1731#pullrequestreview-2319132038 

I'd appreciate your help getting things working if you have the time to assist! I imagine the fixes for one of the repos could easily translate to the other.

codeboten:
Updated the example in this repo as well and got a trace in jaeger

Screenshot 2024-09-20 at 12 45 39 PM

Note that I was getting some errors from instrumentations when i ran `docker compose up`, but i didn't have time to dig into those

kaylareopelle (approved):
This works for me too! Good to know about the errors. I didn't see anything that needed immediate attention. Thank you, @codeboten!

otelcol-config.yml
```diff
@@ -2,13 +2,15 @@ receivers:
...
     protocols:
       grpc:
+        endpoint: 0.0.0.0:4317
       http:
+        endpoint: 0.0.0.0:4318
 exporters:
-  jaeger:
-    endpoint: "jaeger:14250"
+  otlp:
+    endpoint: "jaeger:4317"
```
ericmustin, line 10: Should 'jaeger' also be modified here?

ericmustin: Oh hm, disregard i suppose? Do we need to expose this port on the jaeger container tho? Sorry been a minute since I've peaked at this setup

codeboten: Don't think it's necessary to expose the port on the jaeger container, as the data flows through the collector to jaeger.

ericmustin (approved):

Issue #1183, opened 2024-09-26
Change in behavior in activerecord from 6.1 to 7.0+


mrsimo:
**Description of the bug**

In Rails 6.1 (and 6.0) there used to be an ActiveRecord span for each read query in the form of `Model.find_by_sql`. We found a few uses for this span, even if the underlaying query was also reported by the lower level library instrumentation as well.

With Rails 7.0 this has changed and only a few queries have this. For instance, `Post.first` doesn't create a `Post.find_by_sql` span, but a `post.comments` does create a `Comment.find_by_sql`. In Rails 6.1 there'd both of them.

**Share details about your runtime**

My reproduction steps were run with `ruby 3.3.4 (2024-07-09 revision be1089c8ec) [arm64-darwin23]`, latest opentelemetry gems and various Rails versions (as seen in the script).

**Share a simplified reproduction if possible**

[Here's a gist](https://gist.github.com/mrsimo/11c45868554994a4b4920f764562965b) with a script to reproduce this. The output:

```
❯ for rails in 6.1 7.0 7.1 7.2; do echo "Rails version: $rails"; RAILS_VERSION=$rails ruby run.rb 1> /dev/null; done
Rails version: 6.1
OpenTelemetry::Instrumentation::Mysql2: select - {"db.statement"=>"SELECT `posts`.* FROM `posts` ORDER BY `posts`.`id` ASC LIMIT ?"}
OpenTelemetry::Instrumentation::ActiveRecord: Post.find_by_sql - {}
OpenTelemetry::Instrumentation::Mysql2: select - {"db.statement"=>"SELECT `comments`.* FROM `comments` WHERE `comments`.`post_id` = ?"}
OpenTelemetry::Instrumentation::ActiveRecord: Comment.find_by_sql - {}
Rails version: 7.0
OpenTelemetry::Instrumentation::Mysql2: select - {"db.statement"=>"SELECT `posts`.* FROM `posts` ORDER BY `posts`.`id` ASC LIMIT ?"}
OpenTelemetry::Instrumentation::Mysql2: select - {"db.statement"=>"SELECT `comments`.* FROM `comments` WHERE `comments`.`post_id` = ?"}
OpenTelemetry::Instrumentation::ActiveRecord: Comment.find_by_sql - {}
Rails version: 7.1
[...]

PR #1184, opened 2024-09-26
fix!: activerecord find_by_sql spans on Rails 7.0+

mrsimo:
For https://github.com/open-telemetry/opentelemetry-ruby-contrib/issues/1183

From my tests, using `_query_by_sql` to instrument is compatible with 7.0, 7.1 and 7.2, and is mostly equivalent to `find_by_sql`, with that one calling `_query_by_sql` itself. I believe this would be enough to bring back the same behavior we had in 6.1.

I am a bit conflicted with the name of these spans, since `find_by_sql` isn't entirely true anymore. Should they be called `_query_by_sql` in that case instead? Feels weird using the name of a private method, though. Should we do a more generic `read` or `query`? Help on this would be appreciated 😸 

Thank you!!

### DIFF
--- a/instrumentation/active_record/lib/opentelemetry/instrumentation/active_record/patches/querying.rb
+++ b/instrumentation/active_record/lib/opentelemetry/instrumentation/active_record/patches/querying.rb
@@ -18,17 +18,11 @@ module OpenTelemetry
 
           # Contains ActiveRecord::Querying to be patched
           module ClassMethods
-            if ::ActiveRecord.version >= Gem::Version.new('7.0.0')
-              def _query_by_sql(...)
-                tracer.in_span("#{self} query") do
-                  super
-                end
-              end
-            else
-              def find_by_sql(...)
-                tracer.in_span("#{self} query") do
-                  super
-                end
+            method_name = ::ActiveRecord.version >= Gem::Version.new('7.0.0') ? :_query_by_sql : :find_by_sql
+
+            define_method(method_name) do |*args, **kwargs|
+              tracer.in_span("#{self} query") do
+                super(*args, **kwargs)
               end
             end
 
--- a/instrumentation/active_record/lib/opentelemetry/instrumentation/active_record/patches/querying.rb
+++ b/instrumentation/active_record/lib/opentelemetry/instrumentation/active_record/patches/querying.rb
@@ -18,9 +18,17 @@ module OpenTelemetry
 
           # Contains ActiveRecord::Querying to be patched
           module ClassMethods
-            def find_by_sql(...)
-              tracer.in_span("#{self}.find_by_sql") do
-                super
+            if ::ActiveRecord.version >= Gem::Version.new('7.0.0')
+              def _query_by_sql(...)
+                tracer.in_span("#{self}.find_by_sql") do
+                  super
+                end
+              end
+            else
+              def find_by_sql(...)
+                tracer.in_span("#{self}.find_by_sql") do
+                  super
+                end
               end
             end
 
--- a/instrumentation/active_record/lib/opentelemetry/instrumentation/active_record/patches/querying.rb
+++ b/instrumentation/active_record/lib/opentelemetry/instrumentation/active_record/patches/querying.rb
@@ -20,13 +20,13 @@ module OpenTelemetry
           module ClassMethods
             if ::ActiveRecord.version >= Gem::Version.new('7.0.0')
               def _query_by_sql(...)
-                tracer.in_span("#{self}.find_by_sql") do
+                tracer.in_span("#{self} query") do
                   super
                 end
               end
             else
               def find_by_sql(...)
-                tracer.in_span("#{self}.find_by_sql") do
+                tracer.in_span("#{self} query") do
                   super
                 end
               end
--- a/instrumentation/active_record/test/instrumentation/active_record/patches/querying_test.rb
+++ b/instrumentation/active_record/test/instrumentation/active_record/patches/querying_test.rb
@@ -15,15 +15,15 @@ describe OpenTelemetry::Instrumentation::ActiveRecord::Patches::Querying do
 
   before { exporter.reset }
 
-  describe 'find_by_sql' do
+  describe 'query' do
     it 'traces' do
       Account.create!
 
       User.find_by_sql('SELECT * FROM users')
       Account.first.users.to_a
 
-      user_find_spans = spans.select { |s| s.name == 'User.find_by_sql' }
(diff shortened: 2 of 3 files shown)


mrsimo:
Sorry about those tests, somehow pushed a last minute thing that broke the build!

kaylareopelle:
> Sorry about those tests, somehow pushed a last minute thing that broke the build!

No problem, @mrsimo! Taking a look at this now. It looks like there's one Rubocop offense preventing the `active_record` suite from passing: https://github.com/open-telemetry/opentelemetry-ruby-contrib/actions/runs/11132211056/job/30938815177?pr=1184#step:6:418

instrumentation/active_record/lib/opentelemetry/instrumentation/active_record/patches/querying.rb
```diff
@@ -18,9 +18,17 @@ class << base
 
           # Contains ActiveRecord::Querying to be patched
           module ClassMethods
-            def find_by_sql(...)
-              tracer.in_span("#{self}.find_by_sql") do
-                super
+            if ::ActiveRecord.version >= Gem::Version.new('7.0.0')
+              def _query_by_sql(...)
+                tracer.in_span("#{self}.find_by_sql") do
```
kaylareopelle, line 23: I get what you're saying about the naming! If we change the name of the span, we should consider marking this update as a breaking change since all 7.0 spans would now have a new name.

I like your suggestion of `query` for 7.0+ spans since they don't roll up to `find_by_sql`. 

Unfortunately, instrumentation often has to resort to private methods to get the data users want, so `_query_by_sql` probably wouldn't be too much of a shock and perhaps easier for a user to connect the dots. 

I'd love to hear other opinions!

instrumentation/active_record/test/instrumentation/active_record/patches/querying_test.rb
```diff
@@ -17,10 +17,16 @@
 
   describe 'find_by_sql' do
     it 'traces' do
+      Account.create!
```
kaylareopelle, line 20: Would this updated test fail on 7.0 without the change to `patches/querying.rb`? If not, what do you think about adding a test to cover the scenario that drove this fix?

mrsimo: If I comment the changes to the patch, the test fails in 7.0, 7.1 and 7.2 with a:

```
  1) Failure:
OpenTelemetry::Instrumentation::ActiveRecord::Patches::Querying::find_by_sql#test_0001_traces [test/instrumentation/active_record/patches/querying_test.rb:29]:
Expected nil to not be nil.
```

Because the `Account.first` doesn't produce a span. I'd say the test is redgreening the problem and works across the supported versions. Would you prefer there's one test with the explicit `User.find_by_sql` and one that tests the other types of reads (the `Account.first.users`)?

instrumentation/active_record/lib/opentelemetry/instrumentation/active_record/patches/querying.rb
mrsimo: I feel like `query` (or similar) would be the most stable for the future. Since we're hooking onto internal AR methods, it's possible we might need to change methods, needing a breaking change every time.

The downside is that all other spans being generated have the name of the method we're hooking into, so it would be a bit inconsistent. We could make it configurable?

As a user of this instrumentation I'd lean towards something more generic and stable. I usually have the `mysql2` or `trilogy` spans with details about the exact query that ran, and as a developer I'd need to open the rails code to know what `_query_by_sql` even is. Even the `find_by_sql` was a bit confusing, I didn't know all reads (used to) go through that method.

kaylareopelle: Nice, I'm leaning toward `query` now too!

I hesitate to make it configurable. We have some configuration around span naming for ActiveSupport::Notifications in some of our Rails-related instrumentation gems, but in this scenario, since the span is not related to ActiveSupport::Notifications, I think having a single, consistent name would be best.

mrsimo: I pushed a change to use `query` instead of `find_by_sql`. Spans are now called `User query`, `Account query`, etc. I'm liking it!

xuan-cao-swi (approved):
Looks good to me!

You can ignore this suggestions since it's just a preference
```ruby
method_name = ::ActiveRecord.version >= Gem::Version.new('7.0.0') ? :_query_by_sql : :find_by_sql

define_method(method_name) do |*args, **kwargs|
  tracer.in_span("#{self} query") do
    super(*args, **kwargs)
  end
end
```
p.s. just checked active_record 3,4,5 and 6, they all use find_by_sql, so only 7+ they made big change, hopefully they won't do it again in 8 :grin:

mrsimo:
Hey @xuan-cao-swi! Thanks for the suggestion! An error happens with that version:

```
ArgumentError: wrong number of arguments (given 3, expected 1..2)
    /Users/albert/.rbenv/versions/3.3.5/lib/ruby/gems/3.3.0/gems/activerecord-6.1.7.8/lib/active_record/querying.rb:46:in `find_by_sql'
    lib/opentelemetry/instrumentation/active_record/patches/querying.rb:25:in `block (2 levels) in <module:ClassMethods>'
```

I'm not very familiar with `define_method`, but seems to have some limitations compared to defining the method explicitly, since we can use `(...)` and call `super` without any parameters.

xuan-cao-swi:
Sorry, I should test it before make a comment. It's missing the key-like arguments. I updated my comments (with `**kwargs`)

instrumentation/active_record/lib/opentelemetry/instrumentation/active_record/patches/querying.rb
kaylareopelle: Great to hear! Would you mind updating the PR title to mark this as a breaking change? I think this'll help anyone who uses `find_by_sql` in their observability backends know to update things accordingly. 

It can be marked as a breaking change by updating the title to start with `fix!:` instead of `fix:` 

If you don't think this should be a breaking change, let me know and we can discuss it further.

mrsimo: Done! Totally agree that it's a breaking change.

mrsimo:
@xuan-cao-swi cool! Pushed that change!

kaylareopelle (approved):
Thank you!

PR #1185, opened 2024-09-27
feat: subscribe to process.action_mailer notifications

mrsimo:
I mentioned this [last month](https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/887#pullrequestreview-2253869990).

I believe the `process.action_mailer` event is also very interesting since it will contain the template rendering. Here's an example of how it looks:

image

A couple things I'd like mention with regards to the implementation:

* I added two new options to set the disallowed keys and payload transform. I didn't know what to call them, the original settings had fairly generic names. I imagine it was assumed we'd only subscribe to one event, or that they could be reused for all of them. I don't feel like the `payload_transform` should be reused, since even the default is very specifically about renaming the other event's payload to follow the semantic conventions.
* Since they had different settings I split the Railtie code into two explicit calls to subscribe instead of iterating over an array of events.

I tried this branch with my own app, and I did find an initial issue with the Railtie code I'd written. I'm a bit nervous that the code from the Railtie isn't tested.

Also there's no test that actually sends an email with ActionMailer. I understand we're relying on `ActiveSupport::Notifications`, but I feel like it would be beneficial even if just to know if/when future Rails versions change the contents of the payload so we can decide if we want to handle that to maintain future compatibility. Would we be open to that?

Thank you!

### DIFF
--- a/instrumentation/action_mailer/README.md
+++ b/instrumentation/action_mailer/README.md
@@ -21,7 +21,7 @@ To use the instrumentation, call `use` with the name of the instrumentation:
 
 ```ruby
 OpenTelemetry::SDK.configure do |c|
-  # Use only the ActionMailer instrumentation 
+  # Use only the ActionMailer instrumentation
   c.use 'OpenTelemetry::Instrumentation::ActionMailer'
   # Use the ActionMailer instrumentation along with the rest of the Rails-related instrumentation
   c.use 'OpenTelemetry::Instrumentation::Rails'
@@ -44,8 +44,8 @@ See the table below for details of what [Rails Framework Hook Events](https://gu
 
 | Event Name | Creates Span? | Notes |
 | - | - | - |
-| `deliver.action_mailer` | :white_check_mark: | Creates an span with kind `internal` and email content and status|
-| `process.action_mailer` | :x: | Lack of useful info so ignored |
+| `deliver.action_mailer` | :white_check_mark: | Creates a span with kind `internal` and email content and status |
+| `process.action_mailer` | :white_check_mark: | Creates a span with kind `internal` that will include email rendering spans |
 
 ### Options
 
@@ -67,9 +67,9 @@ end
 
 ## Semantic Conventions
 
-Internal spans are named using the name of the `ActiveSupport` event that was provided (e.g. `action_mailer deliver`).
+Internal spans are named using the name of the `ActiveSupport` event that was provided (e.g. `deliver.action_mailer`).
 
-The following attributes from the notification payload for the `deliver.action_mailer` event are attached to `action_mailer deliver` spans:
+The following attributes from the notification payload for the `deliver.action_mailer` event are attached:
 
 | Attribute Name | Type | Notes |
 | - | - | - |
--- a/instrumentation/action_mailer/README.md
+++ b/instrumentation/action_mailer/README.md
@@ -69,7 +69,7 @@ end
 
 Internal spans are named using the name of the `ActiveSupport` event that was provided (e.g. `deliver.action_mailer`).
 
-The following attributes from the notification payload for the `deliver.action_mailer` event are attached:
+### Attributes attached to the `deliver.action_mailer` event payload
 
 | Attribute Name | Type | Notes |
 | - | - | - |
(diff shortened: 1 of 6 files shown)


instrumentation/action_mailer/README.md
```diff
@@ -67,9 +67,9 @@ end
 
 ## Semantic Conventions
 
-Internal spans are named using the name of the `ActiveSupport` event that was provided (e.g. `action_mailer deliver`).
+Internal spans are named using the name of the `ActiveSupport` event that was provided (e.g. `deliver.action_mailer`).
 
-The following attributes from the notification payload for the `deliver.action_mailer` event are attached to `action_mailer deliver` spans:
+The following attributes from the notification payload for the `deliver.action_mailer` event are attached:
```
kaylareopelle, line 72: Could you add a section describing the attributes captured by the `process.action_mailer` notification by default? 
https://edgeguides.rubyonrails.org/active_support_instrumentation.html#process-action-mailer

instrumentation/action_mailer/lib/opentelemetry/instrumentation/action_mailer/instrumentation.rb
```diff
@@ -27,7 +27,9 @@ class Instrumentation < OpenTelemetry::Instrumentation::Base
         end
 
         option :disallowed_notification_payload_keys, default: [], validate: :array
+        option :disallowed_process_payload_keys,      default: [], validate: :array
         option :notification_payload_transform,       default: nil, validate: :callable
+        option :process_payload_transform,            default: nil, validate: :callable
```
kaylareopelle, line 32: I get what you're saying in your PR description about the benefits of having additional configuration options for the process notification. 

I think we could use some documentation that goes into a little more detail about the scope of these different configs. 

The ActionView instrumentation has similar configuration options, and documents them like this: https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/main/instrumentation/action_view/lib/opentelemetry/instrumentation/action_view/instrumentation.rb#L10-L45 

Would you mind adding something similar for ActionMailer?

kaylareopelle (commented):
I think this is heading in the right direction! Thank you for your work so far. 

I have some documentation suggestions in the comments. 

As for the testing concern you raised in the PR description, I think it would be a great idea to find a way to simulate the process of sending mail within the test suite. Especially since you ran into failures using the tool that weren't revealed by running the tests. 

If you're looking for an example, I work for New Relic and our Ruby agent tests get a little closer to the mail sending process: https://github.com/newrelic/newrelic-ruby-agent/blob/dev/test/new_relic/agent/instrumentation/rails/action_mailer_subscriber.rb

instrumentation/action_mailer/Rakefile
```diff
@@ -15,6 +15,7 @@ Rake::TestTask.new :test do |t|
   t.libs << 'test'
   t.libs << 'lib'
   t.test_files = FileList['test/**/*_test.rb']
+  t.warning = false
```
mrsimo, line 18: I took the liberty of adding this, when adding the new tests that do ActionMailer deliveries the `mail` gems generates lots of warnings. I saw other instrumentation libraries already had it.

instrumentation/action_mailer/lib/opentelemetry/instrumentation/action_mailer/railtie.rb
```diff
@@ -7,24 +7,39 @@
...
         config.after_initialize do
           ::OpenTelemetry::Instrumentation::ActiveSupport::Instrumentation.instance.install({})
+          subscribe_to_deliver
+          subscribe_to_process
+        end
 
-          SUBSCRIPTIONS.each do |subscription_name|
-            config = ActionMailer::Instrumentation.instance.config
+        class << self
+          def subscribe_to_deliver
```
mrsimo, line 22: I moved these bits into class methods that the railtie callback calls. This way I can call them from the tests to subscribe (and unsubscribe) with different configurations.

mrsimo (commented):
Pushed some changes:

* The documentation changes requested, always appreciate if someone can double check them, english is not my first language. But thanks a lot for the suggestions 🧡, I love some good docs on the gems we use!
* An extra test that does a basic email delivery. Happy to iterate on it, pull it out to a different PR or whatever you prefer.
* I've been trying to figure out a way to test the railtie code, best I could come up with is pulling some bits out to class methods.

Thanks a lot for the review!!

kaylareopelle (approved):
I think this looks great! Thank you, @mrsimo!! 🚀

mrsimo:
@kaylareopelle thank you so much. Just so I'm not missing anything, is there anything left to do here? Wait patiently?

kaylareopelle:
@mrsimo, nothing else for you to do! I left the PR open to see if anyone else wanted look at it. If there aren't any new reviews by the end of the day tomorrow, I'll merge it in!

PR #1186, opened 2024-09-27
feat: Integration with V3 telemetry provider

jterapin:
**Description**: AWS SDK for Ruby released Observability support, which allows the SDK to emit telemetry data based on OpenTelemetry. See [OTelProvider](https://docs.aws.amazon.com/sdk-for-ruby/v3/api/Aws/Telemetry/OTelProvider.html) for more details.

This PR addresses integration with the support mentioned above.

**Other changes**:
* Refactor test files to separate files, includes specs for current AwsSdk plugin and patched plugin.
* Update README to include information about the instrumentation config options and integration with SDK's telemetry provider.
* Reduced number of appraisal definitions to the following sets: 
  * 1: Latest V3 release 
  * 2: V3 release prior to the SDK's Observability support
  * 3: Last updated V2 release
  * 4: Earliest V2 release

### DIFF
--- a/instrumentation/aws_sdk/Appraisals
+++ b/instrumentation/aws_sdk/Appraisals
@@ -4,56 +4,25 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 
-appraise 'aws-sdk-3.1' do
-  gem 'aws-sdk', '~> 3.1'
+appraise 'aws-sdk-3' do
+  gem 'aws-sdk-core', '~> 3'
+  gem 'aws-sdk-lambda', '~> 1'
+  gem 'aws-sdk-dynamodb', '~> 1'
+  gem 'aws-sdk-sns', '~> 1'
+  gem 'aws-sdk-sqs', '~> 1'
 end
 
-appraise 'aws-sdk-3.0' do
-  gem 'aws-sdk', '~> 3.0'
+# pre-Observability support in V3 SDK
+appraise 'aws-sdk-3.2' do
+  gem 'aws-sdk-core', '~> 3.202'
+  gem 'aws-sdk-lambda', '~> 1.127'
+  gem 'aws-sdk-dynamodb', '~> 1.118'
+  gem 'aws-sdk-sns', '~> 1.82'
+  gem 'aws-sdk-sqs', '~> 1.80'
 end
 
 appraise 'aws-sdk-2.11' do
   gem 'aws-sdk', '~> 2.11'
-end
-
-appraise 'aws-sdk-2.10' do
-  gem 'aws-sdk', '~> 2.10'
-end
-
-appraise 'aws-sdk-2.9' do
-  gem 'aws-sdk', '~> 2.9'
-end
-
-appraise 'aws-sdk-2.8' do
-  gem 'aws-sdk', '~> 2.8'
-end
-
-appraise 'aws-sdk-2.7' do
-  gem 'aws-sdk', '~> 2.7'
-end
-
-appraise 'aws-sdk-2.6' do
-  gem 'aws-sdk', '~> 2.6'
-end
-
-appraise 'aws-sdk-2.5' do
-  gem 'aws-sdk', '~> 2.5'
-end
-
-appraise 'aws-sdk-2.4' do
-  gem 'aws-sdk', '~> 2.4'
-end
-
-appraise 'aws-sdk-2.3' do
-  gem 'aws-sdk', '~> 2.3'
-end
-
-appraise 'aws-sdk-2.2' do
-  gem 'aws-sdk', '~> 2.2'
-end
-
-appraise 'aws-sdk-2.1' do
-  gem 'aws-sdk', '~> 2.1'
 end
 
 appraise 'aws-sdk-2.0' do
--- a/instrumentation/aws_sdk/Appraisals
+++ b/instrumentation/aws_sdk/Appraisals
@@ -13,7 +13,7 @@ appraise 'aws-sdk-3' do
 end
 
 # pre-Observability support in V3 SDK
-appraise 'aws-sdk-3.2' do
+appraise 'aws-sdk-3.202' do
   gem 'aws-sdk-core', '~> 3.202'
   gem 'aws-sdk-lambda', '~> 1.127'
   gem 'aws-sdk-dynamodb', '~> 1.118'
--- a/instrumentation/aws_sdk/README.md
+++ b/instrumentation/aws_sdk/README.md
@@ -32,6 +32,33 @@ OpenTelemetry::SDK.configure do |c|
   c.use_all
 end
 ```
+### Configuration options
+This instrumentation offers the following configuration options: 
+* `:inject_messaging_context` (default: `false`): When set to `true`, adds context key/value 
+ to Message Attributes for SQS/SNS messages.
+* `suppress_internal_instrumentation` (default: `false`): When set to `true`, any spans with 
+ span kind of `internal` are suppressed from traces.
+
+## Integration with SDK V3's Telemetry support
+AWS SDK for Ruby V3 added support for Observability which includes a configuration, 
+`telemetry_provider` and an OpenTelemetry-based telemetry provider. Only applies to
+AWS service gems released after 2024-09-03. 
+
+Using the OTel telemetry provider will give you insights about specific handlers 
+during the SDK request/response lifecycle.
+
+```ruby
+# configures the OpenTelemetry SDK with instrumentation defaults
+OpenTelemetry::SDK.configure do |c|
+  c.use 'OpenTelemetry::Instrumentation::AwsSdk'
+end
+
+# create otel provider and pass to client config
+otel_provider = Aws::Telemetry::OTelProvider.new
+client = Aws::S3::Client.new(telemetry_provider: otel_provider)
+```
+
+
 
 ## Example
 
--- a/instrumentation/aws_sdk/README.md
+++ b/instrumentation/aws_sdk/README.md
@@ -40,20 +40,19 @@ This instrumentation offers the following configuration options:
  span kind of `internal` are suppressed from traces.
 
 ## Integration with SDK V3's Telemetry support
-AWS SDK for Ruby V3 added support for Observability which includes a configuration, 
+AWS SDK for Ruby V3 added support for Observability which includes a new configuration, 
 `telemetry_provider` and an OpenTelemetry-based telemetry provider. Only applies to
 AWS service gems released after 2024-09-03. 
 
-Using the OTel telemetry provider will give you insights about specific handlers 
-during the SDK request/response lifecycle.
-
+Using later versions of these gems will give more details on the internal spans. 
+See below for example usage:
 ```ruby
 # configures the OpenTelemetry SDK with instrumentation defaults
 OpenTelemetry::SDK.configure do |c|
   c.use 'OpenTelemetry::Instrumentation::AwsSdk'
 end
 
-# create otel provider and pass to client config
+# create open-telemetry provider and pass to client config
 otel_provider = Aws::Telemetry::OTelProvider.new
 client = Aws::S3::Client.new(telemetry_provider: otel_provider)
 ```
--- a/instrumentation/aws_sdk/README.md
+++ b/instrumentation/aws_sdk/README.md
@@ -58,8 +58,6 @@ otel_provider = Aws::Telemetry::OTelProvider.new
 client = Aws::S3::Client.new(telemetry_provider: otel_provider)
 ```
 
-
-
 ## Example
 
 To run the example:
--- a/instrumentation/aws_sdk/lib/opentelemetry/instrumentation/aws_sdk/handler.rb
+++ b/instrumentation/aws_sdk/lib/opentelemetry/instrumentation/aws_sdk/handler.rb
@@ -12,21 +12,17 @@ module OpenTelemetry
         def call(context)
           return super unless context
 
-          service_id = service_name(context)
-          operation = context.operation&.name
-          client_method = "#{service_id}.#{operation}"
+          service_id = HandlerHelper.service_id(context, legacy: true)
+          client_method = HandlerHelper.client_method(service_id, context)
 
           tracer.in_span(
-            span_name(context, client_method, service_id),
-            attributes: attributes(context, client_method, service_id, operation),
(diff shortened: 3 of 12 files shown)


instrumentation/aws_sdk/Appraisals
```diff
@@ -4,58 +4,27 @@
...
 
-appraise 'aws-sdk-3.1' do
-  gem 'aws-sdk', '~> 3.1'
+appraise 'aws-sdk-3' do
+  gem 'aws-sdk-core', '~> 3'
+  gem 'aws-sdk-lambda', '~> 1'
+  gem 'aws-sdk-dynamodb', '~> 1'
+  gem 'aws-sdk-sns', '~> 1'
+  gem 'aws-sdk-sqs', '~> 1'
 end
```
jterapin, line 13: The recent version of `aws-sdk-3.x` does not contain the recent gem releases so I omitted these sets and just included the relevant gems for testing.

Also reduced the number of V2 appraisal definitions to just 2.

instrumentation/aws_sdk/Appraisals
```diff
@@ -4,58 +4,27 @@
...
 
-appraise 'aws-sdk-3.0' do
-  gem 'aws-sdk', '~> 3.0'
+# pre-Observability support in V3 SDK
+appraise 'aws-sdk-3.2' do
```
alextwoods, line 16: sdk versioning in v3 is a little weird with the aws-sdk, aws-sdk-core and aws-sdk-<servivce> gems versioned separately... 3.2 probably doesn't really make sense.  Maybe we use the core version, so 3.202?

instrumentation/aws_sdk/Appraisals
alextwoods: Additional benefit - the aws-sdk gem (which we dont really support anymore) also brings in tons of additional dependencies (almost 400 service gems at this point).

instrumentation/aws_sdk/README.md
```diff
@@ -32,6 +32,31 @@ OpenTelemetry::SDK.configure do |c|
...
+ to Message Attributes for SQS/SNS messages.
+* `suppress_internal_instrumentation` (default: `false`): When set to `true`, any spans with 
+ span kind of `internal` are suppressed from traces.
+
+## Integration with SDK V3's Telemetry support
+AWS SDK for Ruby V3 added support for Observability which includes a configuration, 
+`telemetry_provider` and an OpenTelemetry-based telemetry provider. Only applies to
+AWS service gems released after 2024-09-03. 
+
+Using the OTel telemetry provider will give you insights about specific handlers
```
alextwoods, line 47: This documentation is a little confusing here I think since I think the "Otel telemetry provider" is the SDK concept.   But maybe worth mentioning that using these later versions of the gem w/ this will give more detailed internal spans.

instrumentation/aws_sdk/lib/opentelemetry/instrumentation/aws_sdk/instrumentation.rb
```diff
@@ -43,10 +44,28 @@ def require_dependencies
...
+          end
+        end
+
+        def supports_telemetry_plugin?(klass)
+          telemetry_plugin? &&
+            klass.plugins.include?(Aws::Plugins::Telemetry)
+        end
+
+        def telemetry_plugin?
+          ::Aws.const_defined?('Plugins::Telemetry')
```
alextwoods, line 64: Is there a reason not to do `::Aws::Plugins.const_defined?(:Telemetry)`?

instrumentation/aws_sdk/lib/opentelemetry/instrumentation/aws_sdk/instrumentation.rb
```diff
@@ -43,10 +44,28 @@ def require_dependencies
...
+        end
+
+        def patch
```
alextwoods, line 67: Might be good to have a slightly more descriptive name like `patch_telemtry_plugin`.  And some documentation on what it does/why

instrumentation/aws_sdk/lib/opentelemetry/instrumentation/aws_sdk/patches/telemetry.rb
```diff
@@ -0,0 +1,89 @@
...
+          def attributes(context, client_method, service_id)
+            {
+              'aws.region' => context.config.region,
+              OpenTelemetry::SemanticConventions::Trace::RPC_SYSTEM => 'aws-api',
+              OpenTelemetry::SemanticConventions::Trace::RPC_SERVICE => service_id,
+              OpenTelemetry::SemanticConventions::Trace::RPC_METHOD => context.operation.name,
+              OpenTelemetry::SemanticConventions::Trace::CODE_FUNCTION => context.operation_name.to_s,
+              OpenTelemetry::SemanticConventions::Trace::CODE_NAMESPACE => 'Aws::Plugins::Telemetry'
+            }.tap do |attrs|
+              attrs[SemanticConventions::Trace::DB_SYSTEM] = 'dynamodb' if service_id == 'DynamoDB'
```
alextwoods, line 63: I'm not quite sure what the expectation on DB SYSTEM is - but potentially we should set this for RDS as well?

alextwoods (commented):
Nice! Looking good!

instrumentation/aws_sdk/lib/opentelemetry/instrumentation/aws_sdk/patches/telemetry.rb
jterapin: Per offline convo, we decided to leave this as it is since, `RDS` could be tied to many different database systems (e.g, aurora, postgresql, mysql, etc)

instrumentation/aws_sdk/lib/opentelemetry/instrumentation/aws_sdk/patches/telemetry.rb
```diff
@@ -0,0 +1,89 @@
...
+            context.config.api.metadata['serviceId'] ||
+              context.config.api.metadata['serviceAbbreviation'] ||
+              context.config.api.metadata['serviceFullName']
+          end
+
+          def client_method(service_id, context)
+            "#{service_id}.#{context.operation.name}".delete(' ')
+          end
+
+          def attributes(context, client_method, service_id)
```
alextwoods, line 54: Is there a way we can share this code between the legacy (ie, pre-sdk-observability support) and here?  Would be good to avoid any duplication of this and make it easier to add future attributes in only one spot.

jterapin: Good idea. I could create a module, like `HandlerHelper`(?) that contains shared methods. Bit of a broader name incase we want to include any methods relating to metrics or logging in the future.

instrumentation/aws_sdk/README.md
```diff
@@ -32,6 +32,30 @@ OpenTelemetry::SDK.configure do |c|
   c.use_all
 end
 ```
+### Configuration options
```
kaylareopelle, line 35: I love these updated configuration docs! We've also put similar content in the `instrumentation.rb` file for the library. Would you mind adding them there too? 

Here's an example for how we did it with Sidekiq: https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/main/instrumentation/sidekiq/lib/opentelemetry/instrumentation/sidekiq/instrumentation.rb

kaylareopelle (commented):
@jterapin - Thanks for this PR! I'll take another look once you've added [the `HandlerHelper` module](https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/1186/files#r1781760871). For now, the only thing that stands out to me is a comment about documentation.

instrumentation/aws_sdk/README.md
jterapin: Happy to!

alextwoods (approved):
Nice - Looking good!

instrumentation/aws_sdk/lib/opentelemetry/instrumentation/aws_sdk/patches/telemetry.rb
alextwoods: Yeah - I like the new HandlerHelper - I think that cleans it up well!

instrumentation/aws_sdk/test/opentelemetry/instrumentation_test.rb
```diff
@@ -55,216 +65,4 @@
...
-        _(last_span.attributes['aws.region']).must_include 'stubbed'
-        _(last_span.attributes['db.system']).must_be_nil
-
-        _(last_span.attributes[OpenTelemetry::SemanticConventions::Trace::HTTP_STATUS_CODE]).must_equal 200
-
-        _(last_span.status.code).must_equal OpenTelemetry::Trace::Status::UNSET
-      end
-    end
-
-    describe 'S3' do
```
kaylareopelle, line 79: I don't see a replacement for these S3 tests. Is that intentional?

instrumentation/aws_sdk/lib/opentelemetry/instrumentation/aws_sdk/handler_helper.rb
```diff
@@ -0,0 +1,81 @@
...
+
+# Copyright The OpenTelemetry Authors
+#
+# SPDX-License-Identifier: Apache-2.0
+
+module OpenTelemetry
+  module Instrumentation
+    module AwsSdk
+      # Utility module that contains shared methods between AwsSdk and Telemetry handlers
+      module HandlerHelper
```
kaylareopelle, line 11: Love this refactor! I noticed there isn't a test file for the HandlerHelper. Should it have tests of its own?

kaylareopelle (commented):
This is looking great! 😄 I have a few questions before we move forward

instrumentation/aws_sdk/lib/opentelemetry/instrumentation/aws_sdk/handler_helper.rb
```diff
@@ -0,0 +1,81 @@
...
+            when *MessagingHelper.supported_services
+              MessagingHelper.span_kind(client_method)
+            else
+              OpenTelemetry::Trace::SpanKind::CLIENT
+            end
+          end
+
+          def span_name(context, client_method, service_id, legacy: false)
+            case service_id
+            when *MessagingHelper.supported_services
```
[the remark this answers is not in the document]
jterapin: Hi! It [destructures an array](https://thoughtbot.com/blog/ruby-splat-operator) in this `when` condition, it turns into: `'SQS', 'SNS'`. I seen it mostly used in passing arguments.

instrumentation/aws_sdk/lib/opentelemetry/instrumentation/aws_sdk/messaging_helper.rb
```diff
@@ -10,12 +10,17 @@ module AwsSdk
...
+          SUPPORTED_SERVICES = %w[SQS SNS].freeze
           SQS_SEND_MESSAGE = 'SQS.SendMessage'
           SQS_SEND_MESSAGE_BATCH = 'SQS.SendMessageBatch'
           SQS_RECEIVE_MESSAGE = 'SQS.ReceiveMessage'
           SNS_PUBLISH = 'SNS.Publish'
           SEND_MESSAGE_CLIENT_METHODS = [SQS_SEND_MESSAGE, SQS_SEND_MESSAGE_BATCH, SNS_PUBLISH].freeze
 
+          def supported_services
+            SUPPORTED_SERVICES
+          end
```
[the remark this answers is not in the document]
jterapin: Good point, I think I actually meant to be a class constant here - so I'll revise and change up all the references. Thank you!

instrumentation/aws_sdk/lib/opentelemetry/instrumentation/aws_sdk/handler_helper.rb
jterapin: I initially thought it should but thinking a bit further... I felt that these methods are sufficiently tested via `handler_test` and `instrumentation_test` files so I didn't feel to explicitly have a test file for this.

instrumentation/aws_sdk/lib/opentelemetry/instrumentation/aws_sdk/handler.rb
```diff
@@ -12,21 +12,17 @@ class Handler < Seahorse::Client::Handler
         def call(context)
           return super unless context
 
-          service_id = service_name(context)
```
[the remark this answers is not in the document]
jterapin: Good catch, I'll add some comments. This legacy handler will be used in V2 and V3 pre-telemetry support.

instrumentation/aws_sdk/test/opentelemetry/instrumentation_test.rb
jterapin: Yes, this was intentional. The S3 gem is rather large (465 KB) so we decided to replace this with Lambda (140KB) to make the testing lightweight. Also - there's no S3-specific span attributes that are being generated so we felt Lambda was sufficient to test the same functionality.

instrumentation/aws_sdk/lib/opentelemetry/instrumentation/aws_sdk/handler_helper.rb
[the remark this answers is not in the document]
kaylareopelle: Ah! I didn't realize you could use that outside of when you were passing arguments. Neat!

instrumentation/aws_sdk/test/opentelemetry/instrumentation_test.rb
kaylareopelle: Got it, that makes a lot of sense. Thank you!

instrumentation/aws_sdk/lib/opentelemetry/instrumentation/aws_sdk/handler_helper.rb
kaylareopelle: Sounds good! In that case, I'm comfortable skipping a specific file for these tests.

kaylareopelle (approved):
Thanks, @jterapin! I'll leave this open for a few more days to see if other folks want to review it. I'll merge next week unless there's more feedback.

xuan-cao-swi (approved):

PR #1196, opened 2024-10-19
feat: Set span error only for 5xx response range

j15e:
👋 Hello!

I ran into a very specific problem with the Rack instrumentation. 

When a Rails Action Cable requests get traced by the Rack instrument, the [response code returned by Rails is -1](https://github.com/rails/rails/blob/094aeca1378fd901510959a956fc887df745f153/actioncable/lib/action_cable/connection/client_socket.rb#L71-L74) and the body is empty as defined in [Rack Hijacking spec](https://github.com/rack/rack/blob/main/SPEC.rdoc#hijacking-). 

In a nutshell, Rack hijacking exists to allow an app to take over the response streaming object to do special things like switching protocol to a web socket (just like Action Cable does).

This currently makes the Rack instrumentation add an error on the span because `-1` is (rightfully) outside the defined good http statues (100..499).

https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/c2ffafc85527a415bbe6d0fccece1130201634c1/instrumentation/rack/lib/opentelemetry/instrumentation/rack/middlewares/event_handler.rb#L46
https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/c2ffafc85527a415bbe6d0fccece1130201634c1/instrumentation/rack/lib/opentelemetry/instrumentation/rack/middlewares/event_handler.rb#L210-L215

This change will simply ignore the response object when the env marked the request as hijacked, which I think is was a Rack user would expect in this case.

Special thanks to @arielvalentin for pointing me to the Rack detailed spec when I initially asked the question about this problem in Slack [just here](https://cloud-native.slack.com/archives/C01NWKKMKMY/p1729277160681719).

### DIFF
--- a/instrumentation/rack/lib/opentelemetry/instrumentation/rack/middlewares/event_handler.rb
+++ b/instrumentation/rack/lib/opentelemetry/instrumentation/rack/middlewares/event_handler.rb
@@ -43,7 +43,6 @@ module OpenTelemetry
           include ::Rack::Events::Abstract
 
           OTEL_TOKEN_AND_SPAN = 'otel.rack.token_and_span'
-          GOOD_HTTP_STATUSES = (100..499)
 
           # Creates a server span for this current request using the incoming parent context
           # and registers them as the {current_span}
@@ -208,7 +207,7 @@ module OpenTelemetry
           end
 
           def add_response_attributes(span, response)
-            span.status = OpenTelemetry::Trace::Status.error unless GOOD_HTTP_STATUSES.include?(response.status.to_i)
+            span.status = OpenTelemetry::Trace::Status.error if response.server_error?
             attributes = extract_response_attributes(response)
             span.add_attributes(attributes)
           rescue StandardError => e
--- a/instrumentation/rack/lib/opentelemetry/instrumentation/rack/middlewares/tracer_middleware.rb
+++ b/instrumentation/rack/lib/opentelemetry/instrumentation/rack/middlewares/tracer_middleware.rb
@@ -152,7 +152,7 @@ module OpenTelemetry
           end
 
           def set_attributes_after_request(span, status, headers, _response)
-            span.status = OpenTelemetry::Trace::Status.error unless (100..499).cover?(status.to_i)
+            span.status = OpenTelemetry::Trace::Status.error if (500..599).cover?(status.to_i)
             span.set_attribute('http.status_code', status)
 
             # NOTE: if data is available, it would be good to do this:
--- a/instrumentation/rack/test/opentelemetry/instrumentation/rack/middlewares/event_handler_test.rb
+++ b/instrumentation/rack/test/opentelemetry/instrumentation/rack/middlewares/event_handler_test.rb
@@ -105,6 +105,19 @@ describe 'OpenTelemetry::Instrumentation::Rack::Middlewares::EventHandler' do
       _(proxy_event).must_be_nil
     end
 
+    describe 'with a hijacked response' do
+      let(:service) do
+        lambda do |env|
+          env['rack.hijack?'] = true
+          [-1, {}, []]
+        end
+      end
+
+      it 'sets the span status to "unset"' do
+        _(rack_span.status.code).must_equal OpenTelemetry::Trace::Status::UNSET
+      end
+    end
+
     describe 'when baggage is set' do
       let(:headers) do
         Hash(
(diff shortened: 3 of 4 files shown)


kaylareopelle (commented):
Thank you so much for this PR and for your great research on `rack.hijack`! 🚀

instrumentation/rack/lib/opentelemetry/instrumentation/rack/middlewares/event_handler.rb
```diff
@@ -109,7 +109,7 @@ def on_finish(request, response)
             span = OpenTelemetry::Instrumentation::Rack.current_span
             return unless span.recording?
 
-            add_response_attributes(span, response) if response
+            add_response_attributes(span, response) if response && !request.env.key?('rack.hijack?')
```
arielvalentin, line 112: @j15e The more I think about your comments in slack, the more I agree with this implementation:

> I see, so maybe changing the status check to > 500 would be a simpler and more pragmatic fix, is that what you think?[^1]

When we take a closer look at the spec:

> [Span Status](https://github.com/open-telemetry/opentelemetry-specification/tree/v1.37.0/specification/trace/api.md#set-status) MUST be left unset if HTTP status code was in the 1xx, 2xx or 3xx ranges, unless there was another error (e.g., network error receiving the response body; or 3xx codes with max redirects exceeded), in which case status MUST be set to Error.
> ...
> For HTTP status codes in the 5xx range, as well as any other code the client failed to interpret, span status SHOULD be set to Error.

I interpret that to mean that the code should only set the status to error in cases where the response is in the `5xx` ranges . There are other non-5xx cases that the code isn't currently accounting for but the spec isn't terribly explicit about it other than the case of the Max Redirects Exceeded, which is usually a client side constraint.  

Given the description the spec I think it is safe to change the code to look for Bad HTTP status codes in the 5xx range and leaving the span status unset in all other cases. 

This will remove our explicit checks for the Rack Hijack implementation and allow the response context propagators to execute. 

[^1]: https://cloud-native.slack.com/archives/C01NWKKMKMY/p1729384339814629?thread_ts=1729277160.681719&cid=C01NWKKMKMY
[^2]: https://opentelemetry.io/docs/specs/semconv/http/http-spans/#status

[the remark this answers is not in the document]
j15e: > Is it likely that if the request is a rack.hijack, users wouldn't want their allowed response headers to be added as span attributes?

Yes, but I think they won't be available within the rack response we have access here because when the response is "hijacked", everything including response headers are deal with by another IO stream

Interesting bits in Rails are around here, I think:

https://github.com/rails/rails/blob/fe0fb9fdf04f638996e785d3e291f5fe8bc138a0/actioncable/lib/action_cable/connection/client_socket.rb#L59-L74

As you can see there, it returns a dummy Rack call result `[ -1, {}, [] ]` where `-1` is the status code, `{}` are the headers and `[]` is the body.

instrumentation/rack/lib/opentelemetry/instrumentation/rack/middlewares/event_handler.rb
```diff
@@ -208,7 +207,7 @@ def detach_context(request)
           end
 
           def add_response_attributes(span, response)
-            span.status = OpenTelemetry::Trace::Status.error unless GOOD_HTTP_STATUSES.include?(response.status.to_i)
+            span.status = OpenTelemetry::Trace::Status.error if response.server_error?
```
j15e, line 210: I found out that here we have access to the final [Rack::Response](https://github.com/rack/rack/blob/8f1fe7d6b477d460b2365e2aadc2b348639f26f7/lib/rack/response.rb#L187) object so this existing `server_error?` method can be re-used (not the case in the middleware instrument version though)

j15e:
Thanks for your kind and detailed feedback! I adjusted the PR to change the way the status code is set (only for the 5xx response code range) rather than make a special case for the `rack.hijacked?` flag  (that felt unclean to me too!).

kaylareopelle (approved):
This looks good to me, thank you! I learned a lot during this review 😄

instrumentation/rack/lib/opentelemetry/instrumentation/rack/middlewares/tracer_middleware.rb
```diff
@@ -152,7 +152,7 @@ def create_request_span_name(request_uri_or_path_info, env)
           end
 
           def set_attributes_after_request(span, status, headers, _response)
-            span.status = OpenTelemetry::Trace::Status.error unless (100..499).cover?(status.to_i)
+            span.status = OpenTelemetry::Trace::Status.error if (500..599).cover?(status.to_i)
```
arielvalentin, line 155: based on your comment here https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/1196/files#r1811615817

Are you saying the `_response` object here is _not_ a `Rack::Response`?

j15e: Exactly, it is only the raw response array that is transmitted to middleware `[status, headers, body]`

arielvalentin (approved):

PR #1207, opened 2024-10-22
chore: Update rubocop from 1.66.0 to 1.67.0

kaylareopelle:
PR to update Rubocop across the board, which should auto-close the dependabot PRs.


arielvalentin (approved):
always a fun chore

PR #1214, opened 2024-10-24
fix: ActiveJob Propagate baggage information properly when performing

yoheyk:
After upgrading ActiveJob instrumentation to `0.7.2`, we have noticed that baggage information is not propagating properly via ActiveJob.

Looks like the deletion of this line is causing the issue (We are using the `child` propagation style):
[https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/1018/files#diff-bc60a254cc5ccc8843a524d0eb4c8[…]218817496946fa815a22a39a010b5ddL39](https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/1018/files#diff-bc60a254cc5ccc8843a524d0eb4c8cb47218817496946fa815a22a39a010b5ddL39)

Passing the extracted context as the parent for the consumer context seems to resolve this issue.

### DIFF
--- a/instrumentation/active_job/test/opentelemetry/instrumentation/active_job/handlers/perform_test.rb
+++ b/instrumentation/active_job/test/opentelemetry/instrumentation/active_job/handlers/perform_test.rb
@@ -188,21 +188,6 @@ describe OpenTelemetry::Instrumentation::ActiveJob::Handlers::Perform do
         _(process_span.links[0].span_context.span_id).must_equal(publish_span.span_id)
       end
 
-      it 'propagates baggage' do
-        ctx = OpenTelemetry::Baggage.set_value('testing_baggage', 'it_worked')
-        OpenTelemetry::Context.with_current(ctx) do
-          BaggageJob.perform_later
-        end
-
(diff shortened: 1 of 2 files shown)


instrumentation/active_job/test/opentelemetry/instrumentation/active_job/handlers/perform_test.rb
```diff
@@ -225,6 +210,21 @@
           _(process_span.links[0].span_context.trace_id).must_equal(publish_span.trace_id)
           _(process_span.links[0].span_context.span_id).must_equal(publish_span.span_id)
         end
+
+        it 'propagates baggage' do
```
yoheyk, line 214: this test case needs to be run with an async queue adapter

arielvalentin:
@yoheyk perhaps the better thing to do here is revert https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/1018

yoheyk:
we need to do more than just reverting #1018 
it makes the baggage propagation possible for the `child` propagation style but not for `link`

arielvalentin (approved):

PR #1221, opened 2024-11-01
feat: create span for mysql2 execute

xuan-cao-swi:
###  Description

Adding a span for the mysql2 execute function is similar to the pg `EXEC_PREPARED_ISH_METHODS`.

When using prepared statements with ActiveRecord, the first execution of a query will create a prepare span, which can log the query execution. However, re-execution will not trigger the prepare function; it will trigger the [execute](https://github.com/brianmario/mysql2/blob/0.5.6/lib/mysql2/statement.rb#L3) function directly. Since we currently don’t create a span for the execute function, I’d like to propose adding a span for the execute function.

See prepared statement execution log from mysql server:
```
# first time call the query
2024-11-01T19:35:33.716687Z	   12 Prepare	SELECT `customers`.* FROM `customers` WHERE `customers`.`contactLastName` = ? LIMIT ?
2024-11-01T19:35:33.716952Z	   12 Execute	SELECT `customers`.* FROM `customers` WHERE `customers`.`contactLastName` = 'Schmitt' LIMIT 1

# second time call the query
2024-11-01T19:35:38.256820Z	   12 Execute	SELECT `customers`.* FROM `customers` WHERE `customers`.`contactLastName` = 'Schmitt' LIMIT 1

# third time call the query with different binding value
2024-11-01T19:35:46.300242Z	   12 Execute	SELECT `customers`.* FROM `customers` WHERE `customers`.`contactLastName` = 'Snow' LIMIT 1
```

Even though there is no valuable information except the binding value for the prepared statement, it is useful for monitoring the MySQL query execution duration.

Sample span data:
```
otel-collector     | InstrumentationScope OpenTelemetry::Instrumentation::Mysql2 0.27.2
otel-collector     | Span #0
otel-collector     |     Trace ID       : e93599e8a955311184c69c52769adc26
otel-collector     |     Parent ID      : 949f6bc157ecbb9a
otel-collector     |     ID             : ef2baa0ac9121262
otel-collector     |     Name           : execute
otel-collector     |     Kind           : Client
otel-collector     |     TraceState     : sw=0000000000000000-01
otel-collector     |     Start time     : 2024-11-01 19:35:46.300091427 +0000 UTC
otel-collector     |     End time       : 2024-11-01 19:35:46.301052177 +0000 UTC
otel-collector     |     Status code    : Unset
otel-collector     |     Status message : 
otel-collector     | Attributes:
otel-collector     |      -> args: Str(["Snow", 1])
otel-collector     |      -> kwargs: Str({})
```


instrumentation/mysql2/lib/opentelemetry/instrumentation/mysql2/patches/statement.rb
```diff
@@ -0,0 +1,44 @@
...
+            ) do
+              super
+            end
+          end
+
+          private
+
+          def _otel_execute_attributes(args, kwargs)
+            if config[:db_statement] == :include
+              {'args' => args.to_s, 'kwargs' => kwargs.to_s}
```
arielvalentin, line 27: I do agree that we should amend all attributes without using explicit mappings for semantic conventions especially if it opens up a risk to exposing sensitive data. 

E.g. auth token or password would be exposed through telemetry collection, with a safeguards to remove them. 

I do realize we allow this to happen today with our existing `include` statement but I am leaning more and more each day to deleting that option all together and only allowing `sanitized`, `obfuscate`, or `omit` in the future.

xuan-cao-swi: Does `sanitized` mean specifically removing the sensitive data (e.g. token, password, ssn, health_id, etc.), but not other binding parameters such as address?

I believe most password stored in table should be hashed.

arielvalentin:
There are now experimental attributes [db.operation.parameter.<key>](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/attributes-registry/db.md)

https://github.com/open-telemetry/semantic-conventions/blob/main/docs/database/sql.md#attributes

kaylareopelle:
Hi @xuan-cao-swi, what are the next steps on this PR?

kaylareopelle:
On hold until we implement `SEMCONV_STABILITY_OPT_IN` for DB libraries

Issue #1172, opened 2024-09-17
Freeze all range objects


hannahramadan:
Freezing objects in Ruby provides several benefits including performance improvements and semantic clarity. 

Since ranges are not frozen in Ruby by default, it would be nice to freeze ranges that appear through the codebase.

Victorsesan:
@kaylareopelle Is this still up for grabs?, can you assign me to give it a try thanks

Victorsesan:
@kaylareopelle Cool thanks! Just wondering are the required ranges to be frozen that of Inclusive & Exclusive (100..399)? And can we create a different file in the lib directory to write a code to freeze the ranges in , also should we have the entire range as a single object frozen or have an implementation which creates individual frozen ranges?

kaylareopelle:
@Victorsesan  - I found a good example of this in the Rack instrumentation: 
https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/main/instrumentation/rack/lib/opentelemetry/instrumentation/rack/middlewares/event_handler.rb#L46

Victorsesan:
@kaylareopelle Just tried this out! I'm not sure how it was directed to a different section for review 😅hopefully i didn't do something wrong . Anyways please check it out let me know what you think, i used a constant instead of a .freeze method as requested to make the code more readable. I'm very much open for any corrections or improvement if needed

kaylareopelle:
Hi @Victorsesan, thank you for opening the PR! The changes look like they're on the right track! 

Our contributing guide has a few recommendations [about making modifications](https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/main/CONTRIBUTING.md#make-your-modifications) that I would like you to follow before we move forward.

1. Because we squash commits when we merge PRs in this repository, creating PRs off of your fork's `main` branch can cause problems. Would you mind closing your PR, creating a new branch, moving your changes to that branch, and opening a new PR from that branch? 

2. Also, we like to have our PR titles and commit messages follow the conventional commits format. Could you write the commit message on your new branch using a conventional commit style?

If you need more help with any of these steps, please let me know! 😄

kaylareopelle:
Hi @Victorsesan! Thanks for your work on this! I'm going to respond [to your latest comment on your old PR here.](https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/1197#issuecomment-2439010201)

[...]

PR #1222, opened 2024-11-02
 refactor: Freeze all range objects #1172

Victorsesan:
Ref #1172


arielvalentin (commented):
@open-telemetry/ruby-contrib-approvers now that I'm looking at this I'm wondering if there is an opportunity to extract status setter helper method rather than defining a range per http instrumentation and implementing the same logic in each instrumentation separately.

Something like this

```ruby
Semconv::Http::Client.set_status(span, http_status_code)
```

Which would then set the span status based on the http status code falling in the range of bad client status code. 

That would reduce the number of places we would need to update when satisfying the specification.

Maybe that could be a follow up PR and this perf refactoring is a first step?

kaylareopelle:
> @open-telemetry/ruby-contrib-approvers now that I'm looking at this I'm wondering if there is an opportunity to extract status setter helper method rather than defining a range per http instrumentation and implementing the same logic in each instrumentation separately.
> 
> Something like this
> 
> ```ruby
> Semconv::Http::Client.set_status(span, http_status_code)
> ```
> 
> Which would then set the span status based on the http status code falling in the range of bad client status code.
> 
> That would reduce the number of places we would need to update when satisfying the specification.
> 
> Maybe that could be a follow up PR and this perf refactoring is a first step?

@arielvalentin - I like that idea! My vote would be to keep this refactoring as a first step and open that up in a follow-up PR.

kaylareopelle:
Hi @Victorsesan! Thanks for re-opening this PR! 

It looks like there are a few [rubocop](https://github.com/rubocop/rubocop) (Ruby linter) failures on your changes. 

The general message is: 
```
Running RuboCop...
RuboCop failed!
Inspecting 17 files
.........C.......

Offenses:

lib/opentelemetry/instrumentation/http/patches/client.rb:14:39: C: [Correctable] Style/RedundantFreeze: Do not freeze immutable objects, as freezing them has no effect.
          HTTP_STATUS_SUCCESS_RANGE = (100..399).freeze
```

I think the solution is to remove `.freeze` since the constant makes the range immutable. 

If you'd like to run rubocop locally to test yourself, you can do so using:
```sh
docker compose run app
<container_sha>: /app$ cd instrumentation/http_client
<container_sha>: /app/instrumentation/http_client$ bundle install
<container_sha>: /app/instrumentation/http_client$ bundle exec rubocop
```

Victorsesan:
Thanks very much for reviewing @kaylareopelle . I have taken a look at the failures again from my end and have them sorted. Some end-of-line characters & carriage return test failures were also found which i converted with ```dos2unix``` and ran ```bundle exec rubocop --autocorrect``` to ensure all the offense in some files were fixed, so we can have some few modification/corrections shown in some files after the test was carried out.  Hopefully it was for the best , but if these additional fix aren't needed i will have them regress asap.

kaylareopelle:
Hi @Victorsesan! Thank you for your Rubocop updates! Great work so far! 

It looks like your latest commit resolved the Rubocop error only on HTTP Client instrumentation. I should've been more specific in my earlier instructions that they would only run the HTTP Client instrumentation.

The changes in your PR span multiple instrumentation libraries. This means you'll need to change directory into those libraries within the Docker container to run Rubocop on each of them. Alternatively, you could try a find-and-replace to remove the `.freeze` at the end of the new `HTTP_STATUS_SUCCESS_RANGE` constants.

If you'd like to try to run Rubocop locally on each of those libraries, you can do so by:
```sh
docker compose run app
<container_sha>: /app$ cd instrumentation/<gem name, ex. ethon>
<container_sha>: /app/instrumentation/<gem name>$ bundle install
<container_sha>: /app/instrumentation/<gem name>$ bundle exec rubocop --autocorrect
<container_sha>: /app/instrumentation/<gem name>$ cd ../<next gem name, ex. excon>
<container_sha>: /app/instrumentation/<next gem name>$ bundle install
<container_sha>: /app/instrumentation/<next gem name>$ bundle exec rubocop --autocorrect
# repeat for the remaining gems: faraday, http, httpx, net_http, restclient
```

Victorsesan:
Did i hit the mark this time, @kaylareopelle . I might have missed something tho🤔 , i had some offenses which i'm not sure i understand what they're.
Screenshot 2024-11-17 012736

kaylareopelle (approved):
Hi @Victorsesan, this looks great! 🎉 

Thank you so much for your patience and willingness to work with me through this. 

I think the extra offenses you saw are related to an issue with the way Windows terminates lines. [This Stack Overflow might help for the future.](https://stackoverflow.com/questions/73532361/rubocop-layout-endofline-carriage-return-character-detected-wsl2)

PR #1234, opened 2024-11-06
fix: get correct table name if table name is quoted

### LINKED ISSUE #1173: PG instrumentation does not parse quoted table names in queries into the 'db.collection.name' attribute
**Description of the bug**
The PG instrumentation library is using a regex [here](https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/a174d03213057f0a86ecf23251f2c9e072c0ec6e/instrumentation/pg/lib/opentelemetry/instrumentation/pg/patches/connection.rb#L18) to extract the table name from queries in order to set the `db.collection.name` trace attribute. This regex does not match quoted table names, for instance `SELECT * from "foo"` will not return a match. 

Operating system details: Linux, Ubuntu 24.04 
RUBY_ENGINE: "ruby"
RUBY_VERSION: "3.3.4"
RUBY_DESCRIPTION: "ruby 3.3.4 (2024-07-09 revision be1089c8ec) [x86_64-linux]"

This also happens on various production instances, I can get that information if needed.

I've been reproducing this by adding this to the tests for pg on a local fork
```diff
diff --git a/instrumentation/pg/test/opentelemetry/instrumentation/pg/instrumentation_test.rb b/instrumentation/pg/test/opentelemetry/instrumentation/pg
/instrumentation_test.rb
index cc50728e..7c1c2590 100644
--- a/instrumentation/pg/test/opentelemetry/instrumentation/pg/instrumentation_test.rb
+++ b/instrumentation/pg/test/opentelemetry/instrumentation/pg/instrumentation_test.rb
@@ -275,6 +275,13 @@ describe OpenTelemetry::Instrumentation::PG::Instrumentation do
       client.query('DROP TABLE test_table') # Drop table to avoid conflicts
     end

+    it 'extracts table name with quotes' do
+      client.query('CREATE TABLE "test_quotes_table" (personid int, name VARCHAR(50))')
+
+      _(span.attributes['db.collection.name']).must_equal '"test_quotes_table"'
+      client.query('DROP TABLE test_quotes_table') # Drop table to avoid conflicts
+    end
+
     describe 'when db_statement is obfuscate' do
       let(:config) { { db_statement: :obfuscate } }
```

This seems to fix it, though I'm pretty green at ruby and the inner workings of this library, so I don't know if there are knock-on effects here.
```diff
diff --git a/instrumentation/pg/lib/opentelemetry/instrumentation/pg/patches/connection.rb b/instrumentation/pg/lib/opentelemetry/instrumentation/pg/pat
ches/connection.rb
index 98814950..6330abee 100644
--- a/instrumentation/pg/lib/opentelemetry/instrumentation/pg/patches/connection.rb
+++ b/instrumentation/pg/lib/opentelemetry/instrumentation/pg/patches/connection.rb
@@ -15,7 +15,7 @@ module OpenTelemetry
         # Module to prepend to PG::Connection for instrumentation
         module Connection # rubocop:disable Metrics/ModuleLength
           # Capture the first word (including letters, digits, underscores, & '.', ) that follows common table commands
-          TABLE_NAME = /\b(?:FROM|INTO|UPDATE|CREATE\s+TABLE(?:\s+IF\s+NOT\s+EXISTS)?|DROP\s+TABLE(?:\s+IF\s+EXISTS)?|ALTER\s+TABLE(?:\s+IF\s+EXISTS)?)
\s+([\w\.]+)/i
+          TABLE_NAME = /\b(?:FROM|INTO|UPDATE|CREATE\s+TABLE(?:\s+IF\s+NOT\s+EXISTS)?|DROP\s+TABLE(?:\s+IF\s+EXISTS)?|ALTER\s+TABLE(?:\s+IF\s+EXISTS)?)
\s+(["\w\.]+)/i

           PG::Constants::EXEC_ISH_METHODS.each do |method|
             define_method method do |*args, &block|
```

HielkeJ:
TABLE_NAME is extracted if the table name is quoted. Should fix #1173


instrumentation/pg/test/fixtures/sql_table_name.json
```diff
@@ -50,5 +50,13 @@
...
+    },
+    {
+      "name": "table_name_with_single_quotes",
+      "sql": "SELECT columns FROM 'test_table'"
+    },
+    {
+      "name": "table_name_with_double_quotes",
+      "sql": "SELECT columns FROM \"test_table\""
     }
   ]
```
arielvalentin, line 62:
```suggestion
  ]

```

instrumentation/pg/test/fixtures/sql_table_name.json
```diff
@@ -50,5 +50,13 @@
...
       "name": "from_with_join",
       "sql": "SELECT columns FROM test_table JOIN table2 ON test_table.column = table2.column"
+    },
+    {
+      "name": "table_name_with_single_quotes",
+      "sql": "SELECT columns FROM 'test_table'"
+    },
+    {
+      "name": "table_name_with_double_quotes",
+      "sql": "SELECT columns FROM \"test_table\""
```
arielvalentin, line 60: Would you kindly edit this file and add a trailing new line? It was missing in the original comment and should have been there.

arielvalentin (approved):

@kaylareopelle @hannahramadan would you all mind comparing that to the new relic implementation to see if we are missing anything else?

scbjans (approved):

hannahramadan:
Thanks @HielkeJ!

I'm wondering if we should remove the single quotes from the regex. @mostfunkyduck [pointed out](https://github.com/open-telemetry/opentelemetry-ruby-contrib/issues/1173#issuecomment-2358618255) that single quotes are for string literals, not identifiers. What do you think @arielvalentin @scbjans

scbjans:
Thanks for pointing that out, I somehow read past that.
After reading those docs, I agree that the single quotes should be removed from the regex, since those aren't for identifiers.

HielkeJ:
I did a small update in the PR @hannahramadan. If I try in Postgres to to a select SELECT * FROM 'table', it returns a syntax error. I didn't test if it fails in opentelemetry-ruby-contrib as well with a syntax error, but at least in shouldn't run the query.

arielvalentin:
Given this new information I'm going to rescind my approval and scrutinize this a little more.

I'll leave the final approval to some other folks.

hannahramadan (approved):
This looks good to me with the recent updates.

kaylareopelle (approved):

PR #1242, opened 2024-11-15
test: Add Rails 7.2 to Appraisals

arielvalentin:
Now that Rails 8 has been released, we need to make sure our appraisals are up to date to explicitly include Rails 7.2


kaylareopelle (approved):
Thanks, Ariel! Do you think we should mention this now tests Rails 8.0 too? 

Also, I saw earlier test runs where Rails 2.3 was installed instead of the latest version on some gems. Did the `appraisal clean` addition fix that?

PR #1256, opened 2024-11-19
test: Update que Appraisals

### LINKED ISSUE #1110: que test suite failures
The nightly build for queue started failing with what seems to be incompatibilities with newer versions of Rails.

cc: @indrekj as the current maintainer of the que instrumentation. 

```console
1) Error:
  OpenTelemetry::Instrumentation::Que::span propagation::when propagation_style is set to link#test_0003_links spans together:
  NoMethodError: undefined method `clear_active_connections!' for class ActiveRecord::Base
      /opt/hostedtoolcache/Ruby/3.3.4/x64/lib/ruby/gems/3.3.0/gems/activerecord-7.2.0/lib/active_record/dynamic_matchers.rb:22:in `method_missing'
      /opt/hostedtoolcache/Ruby/3.3.4/x64/lib/ruby/gems/3.3.0/gems/que-1.4.0/lib/que/active_record/connection.rb:45:in `call'
      /opt/hostedtoolcache/Ruby/3.3.4/x64/lib/ruby/gems/3.3.0/gems/que-1.4.0/lib/que/utils/middleware.rb:36:in `invoke_middleware'
      (eval at /opt/hostedtoolcache/Ruby/3.3.4/x64/lib/ruby/gems/3.3.0/gems/que-1.4.0/lib/que/utils/middleware.rb:14):11:in `run_job_middleware'
      test/opentelemetry/instrumentation/que_test.rb:[170](https://github.com/open-telemetry/opentelemetry-ruby-contrib/actions/runs/10335505991/job/28610140236#step:4:181):in `block (4 levels) in <top (required)>'
```

https://github.com/open-telemetry/opentelemetry-ruby-contrib/actions/runs/10335505991/job/28610140236

kaylareopelle:
The latest version of the gem now supports Rails 7.2 and Rails 8.0

Resolves: https://github.com/open-telemetry/opentelemetry-ruby-contrib/issues/1110


instrumentation/que/Appraisals
```diff
@@ -6,12 +6,19 @@
...
 end
 
 appraise 'que-2.x' do
-  gem 'que', '~> 2.3'
+  gem 'que', '~> 2.4'
+  if Gem::Version.new(RUBY_VERSION) >= Gem::Version.new('3.1.0')
+    gem 'activerecord', '~> 7.2.0'
+  else
+    gem 'activerecord'
+  end
```
kaylareopelle, line 18: Without this, we'd skip testing on Rails 7.2 and jump straight to Rails 8.0. Que made a change to support Rails 7.2 in 2.4.1, but our instrumentation didn't need to change. 

Since these tests are for our instrumentation and not necessarily for activerecord compatibility with que, I'm not sure we need the condition. Thoughts?

arielvalentin: I think it really depends on how much "coverage" we want to verify. 

I am amenable to only testing against the latest version active record unless @indrekj  has any objections.

arielvalentin (approved):

PR #1259, opened 2024-11-20
fix: pass block argument in ActiveRecord `find_by_sql` patch

### LINKED ISSUE #1258: New ActiveRecord `find_by_sql` patch fails to pass `&block` argument
**Description of the bug**

The new patch for ActiveRecord's `find_by_sql`, added in #1184 and released in `opentelemetry-instrumentation-active_record` v0.8.0 alters application behaviour. The PR changed the patch from using argument forwarding (`def find_by_sql(...)`) to using `define_method` with explicitly passed arguments: https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/bf6394d2ef3918a1a67e03b773352598e40adcbc/instrumentation/active_record/lib/opentelemetry/instrumentation/active_record/patches/querying.rb#L20-L27

The problem with this approach is that the function signature of `find_by_sql` looks like this:

```ruby
# ActiveRecord v6 & v7
def find_by_sql(sql, binds = [], preparable: nil, &block)
# ActiveRecord v8
def find_by_sql(sql, binds = [], preparable: nil, allow_retry: false, &block)
```

Notice that `find_by_sql` takes `&block` as its final argument, which is neither a positional nor a keyword argument. The old argument forwarding approach automatically forwards `&block` as well, but the new patch's `|*args, **kwargs|` is not equivalent. As a result, the block is not passed to the `super()` call and hence not executed. I noticed this because `opentelemetry-instrumentation-active_record` v0.8.0 breaks some specs in a Rails application I work on; downgrading to v0.7.4 makes the specs pass again.

I believe just passing `&block` to the `do` block and specifying it in the `super()` call is all that is needed to resolve this:

```diff
diff --git a/instrumentation/active_record/lib/opentelemetry/instrumentation/active_record/patches/querying.rb b/instrumentation/active_record/lib/opentelemetry/instrumentation/active_record/patches/querying.rb
index 094575ae..0b9f3790 100644
--- a/instrumentation/active_record/lib/opentelemetry/instrumentation/active_record/patches/querying.rb
+++ b/instrumentation/active_record/lib/opentelemetry/instrumentation/active_record/patches/querying.rb
@@ -20,9 +20,9 @@ module OpenTelemetry
           module ClassMethods
             method_name = ::ActiveRecord.version >= Gem::Version.new('7.0.0') ? :_query_by_sql : :find_by_sql

-            define_method(method_name) do |*args, **kwargs|
+            define_method(method_name) do |*args, **kwargs, &block|
               tracer.in_span("#{self} query") do
-                super(*args, **kwargs)
+                super(*args, **kwargs, &block)
               end
             end
```

Monkey-patching the gem with the above diff makes my specs pass, at least.

Operating system details: Debian GNU/Linux trixie (testing)
RUBY_ENGINE: "ruby"
RUBY_VERSION: "3.0.6"
RUBY_DESCRIPTION: "ruby 3.0.6p216 (2023-03-30 revision 23a532679b) [x86_64-linux]"

I'm not familiar enough with OpenTelemetry or ActiveRecord's internals to provide a minimal repro, I'm sorry. I hope the above description and suggested fix help, at least.

All the best,
Ole

olepbr:
`find_by_sql` takes a block as its final argument, which is neither a positional nor a keyword argument, hence we need to pass it explicitly; before this PR, the block was not passed to the `super()` call and hence not executed, thus altering application behavior.

Fixes #1258.

### DIFF
--- a/instrumentation/active_record/lib/opentelemetry/instrumentation/active_record/patches/querying.rb
+++ b/instrumentation/active_record/lib/opentelemetry/instrumentation/active_record/patches/querying.rb
@@ -20,9 +20,9 @@ module OpenTelemetry
           module ClassMethods
             method_name = ::ActiveRecord.version >= Gem::Version.new('7.0.0') ? :_query_by_sql : :find_by_sql
 
-            define_method(method_name) do |*args, **kwargs|
+            define_method(method_name) do |*args, **kwargs, &block|
               tracer.in_span("#{self} query") do
-                super(*args, **kwargs)
+                super(*args, **kwargs, &block)
               end
             end
 
--- a/instrumentation/active_record/test/instrumentation/active_record/patches/querying_test.rb
+++ b/instrumentation/active_record/test/instrumentation/active_record/patches/querying_test.rb
@@ -14,6 +14,11 @@ describe OpenTelemetry::Instrumentation::ActiveRecord::Patches::Querying do
   let(:spans) { exporter.finished_spans }
 
   before { exporter.reset }
+  after do
+    ActiveRecord::Base.subclasses.each do |model|
+      model.connection.truncate(model.table_name)
+    end
+  end
 
   describe 'query' do
     it 'traces' do
@@ -27,6 +32,80 @@ describe OpenTelemetry::Instrumentation::ActiveRecord::Patches::Querying do
 
       _(user_find_spans.length).must_equal(2)
       _(account_find_span).wont_be_nil
+    end
+
+    describe 'find_by_sql' do
+      it 'creates a span' do
+        Account.create!
+
+        Account.find_by_sql('SELECT * FROM accounts')
+
+        account_find_span = spans.find { |s| s.name == 'Account query' }
+        _(account_find_span).wont_be_nil
+        _(account_find_span.attributes).must_be_empty
+      end
+
+      describe 'given a block' do
+        it 'creates a span' do
+          account = Account.create!
+
+          record_ids = []
+
+          Account.find_by_sql('SELECT * FROM accounts') do |record|
+            record_ids << record.id
+          end
+
+          account_find_span = spans.find { |s| s.name == 'Account query' }
+          _(account_find_span).wont_be_nil
+          _(account_find_span.attributes).must_be_empty
+
+          _(record_ids).must_equal([account.id])
+        end
+      end
+    end
+
+    describe 'find_by' do
+      it 'creates a span' do
+        account = Account.create!
+        User.create!(account: account)
+
+        Account.find_by(id: account.id)
+
+        account_find_span = spans.find { |s| s.name == 'Account query' }
+        _(account_find_span).wont_be_nil
+        _(account_find_span.attributes).must_be_empty
+      end
+    end
+
+    describe 'find' do
+      it 'creates a span' do
+        account = Account.create!
+        User.create!(account: account)
+
+        Account.find(account.id)
+
+        account_find_span = spans.find { |s| s.name == 'Account query' }
+        _(account_find_span).wont_be_nil
+        _(account_find_span.attributes).must_be_empty
+      end
+
+      describe 'given a block' do
+        it 'creates a span' do
+          account = Account.create!
+          User.create!(account: account)
+
+          record_ids = []
+
+          Account.find(account.id) do |record|
+            record_ids << record.id
+          end
+
+          account_find_span = spans.find { |s| s.name == 'Account query' }
+          _(account_find_span).wont_be_nil
+          _(account_find_span.attributes).must_be_empty
+          _(record_ids).must_equal([account.id])
+        end
+      end
     end
   end
 end


instrumentation/active_record/lib/opentelemetry/instrumentation/active_record/patches/querying.rb
```diff
@@ -20,9 +20,9 @@ class << base
           module ClassMethods
             method_name = ::ActiveRecord.version >= Gem::Version.new('7.0.0') ? :_query_by_sql : :find_by_sql
 
-            define_method(method_name) do |*args, **kwargs|
+            define_method(method_name) do |*args, **kwargs, &block|
               tracer.in_span("#{self} query") do
-                super(*args, **kwargs)
+                super(*args, **kwargs, &block)
```
arielvalentin, line 25: Since I am out of date with my meta-programming knowledge, do we have to explicitly pass the arguments here or is there implicit arguments that we can leverage here? 

Also, may I ask that you add an automated test here to mitigate future regressions?

arielvalentin:
@olepbr in addition to my comments, you are required to sign the EasyCLA before we are able to review and accept your contribution. 

Thanks.

instrumentation/active_record/lib/opentelemetry/instrumentation/active_record/patches/querying.rb
kaylareopelle:
```suggestion
                super
```
I'm pretty sure you can just call `super` without passing anything if you want to pass everything.

For example: 
```ruby 
# super_example.rb

class Parent
  def hello(a = 'hi', b: 'howdy', &blk)
    puts a
    puts b
    yield
  end
end

class Child < Parent
  def hello(*args, **kwargs, &blk)
    super
  end
end

c = Child.new
c.hello('hey') { |h| puts 41 }
```

When you run this Ruby file on Ruby 3.0.0 (our min supported version), the output is:
```
hey
howdy
41
```
In this example, we didn't explicitly pass anything to `super`, so everything was passed.

arielvalentin: Will this also work with define_method?

Can Ruby blocks use shorthand syntax to ignore arguments? E.g.

`define_method do |...|`

olepbr: @kaylareopelle While it's fine to do so in your example, it does not work in the context of `define_method()`. Specifically, your suggestion results in `RuntimeError: implicit argument passing of super from method defined by define_method() is not supported. Specify all arguments explicitly.` I don't know Ruby's internals well enough to understand _why_ that isn't allowed, but it hasn't been since [this commit](https://github.com/ruby/ruby/commit/62deb7dbbda5104c352abd723053bd1b18e50dbf) from 2007. (These days, that code lives in [vm_insnhelper.c](https://github.com/ruby/ruby/blob/bf718cef592d9c6c3229c2da0ee1169cfefa17c6/vm_insnhelper.c#L5026).) Given that, I conclude that the explicit arguments are indeed needed as long as we use `define_method()`.

(The PR that introduced this bug originally used argument forwarding, which does allow argument-less `super`, but a reviewer suggested `define_method` instead, leading to 1f30c8069a6ae7d3854287ae4cdebf959539ae8f. If there's a hard preference for argument-less `super`, I can rewrite the patch to use argument forwarding instead; I focused on making a minimal change.)

olepbr: > Also, may I ask that you add an automated test here to mitigate future regressions?

@arielvalentin I'm afraid I don't really have the skills to do so; I have written approximately 50 lines of Ruby in my entire life =) I'm happy to take guidance on how to write such a test, but I imagine it would be much quicker if someone more proficient did it. Given the severity of this bug, time is perhaps of the essence.

olepbr: @arielvalentin perhaps something like the below is sufficient? It fails on main, at least.

```diff
diff --git a/instrumentation/active_record/test/instrumentation/active_record/patches/querying_test.rb b/instrumentation/active_record/test/instrumentation/active_record/patches/querying_test.rb
index 879af50e..52b34095 100644
--- a/instrumentation/active_record/test/instrumentation/active_record/patches/querying_test.rb
+++ b/instrumentation/active_record/test/instrumentation/active_record/patches/querying_test.rb
@@ -22,11 +22,18 @@ describe OpenTelemetry::Instrumentation::ActiveRecord::Patches::Querying do
       User.find_by_sql('SELECT * FROM users')
       Account.first.users.to_a

+      # Check that blocks passed to `find_by_sql` are executed.
+      foo = nil
+      SuperUser.find_by_sql 'SELECT 1' do
+        foo = 'bar'
+      end
+
       user_find_spans = spans.select { |s| s.name == 'User query' }
       account_find_span = spans.find { |s| s.name == 'Account query' }

       _(user_find_spans.length).must_equal(2)
       _(account_find_span).wont_be_nil
+      _(foo).must_equal('bar')
```

arielvalentin: I think so. I have added some additional tests in a PR for reference for you:

https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/1260

You may choose to use them as a starting point if you like.

@open-telemetry/ruby-contrib-approvers TIL that [`ActiveRecord::QueryMethods.where`](https://api.rubyonrails.org/classes/ActiveRecord/QueryMethods.html#method-i-where) is not instrumented. See the test cases in the PR I linked above.

olepbr: @arielvalentin Many thanks! I'm going to sleep now (it's late in Europe), but can add a commit with some tests in the morning.

arielvalentin: I appreciate it thank you. From the examples I shared you may skip the `where` tests. I will open a separate feature request for that. 

cc: @simi if you are around tomorrow morning or afternoon (Central European Time) to help review the changes and get a release out once you are satisfied.

olepbr: @arielvalentin Changed my mind and committed it now instead =) I left the where tests out as they fail due to the lack of instrumentation. `find_by` apparently also does not accept a block, so I didn't specify a "given a block" test for it.

olepbr:
CI was failing due to a Rubocop offense triggered by the added tests; amended now.

arielvalentin (approved):

Issue #1194, opened 2024-10-14
Explore renaming/enhancing `opentelemetry-helpers-sql-obfuscation` gem


hannahramadan:
Currently, the `opentelemetry-helpers-sql-obfuscation` gem contains the logic for SQL sanitation; the name of this gem reflects this. However, there are additional instances where shared logic would be helpful for db instrumentations. 

This idea came up in context of adding a new attribute to db spans, where currently logic is duplicated across db instrumentations to parse SQL statements for the db collection/table name and add it to a span [[context](https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/1109#issuecomment-2401209163)].

We should consider what it takes to rename the obfuscation gem to something more broad, such as `sql-processor`, and refactoring shared logic in db instrumentations to make use of this gem.

scbjans:
hi @hannahramadan, 
previously we did something similar with the resource detectors:

1. create the new gem, add to CI etc.
2. update the old gem, setting a [post_install_message](https://guides.rubygems.org/specification-reference/#post_install_message) stating the old gem is deprecated and point users to the new gem (e.g. #639 )
3. update codebase to make sure new gem is used
4. clean up (e.g #659 )

PR #1263, opened 2024-11-23
feat: Initial Release SQL Helpers

arielvalentin:
SQL Helpers provides utilities that are intended for use with SQL database related instrumentations. 

The initial release adds support for a Shared DB Context Helper, whose goal is to replace the instrumentation specific helpers. 

Future releases will target replacing sanitization/obfuscation helpers.

cc: https://github.com/open-telemetry/opentelemetry-ruby-contrib/issues/1194


kaylareopelle:
Thanks for your work to reduce duplicate code! 

I'm getting a little stuck thinking through the future plans for the SQL helpers gem. They may not need to block this PR, and this might not be the right forum to talk this through. Please redirect me if there's a better place.

When we created the SQL obfuscation gem, IIRC we intentionally separated the MySQL-specific code into a separate gem, in the spirit of providing the minimum amount of additional code per library. It seemed like we wanted to keep helper libraries small, despite the potential annoyance of having extra libraries. 

I don't see this context management code in the PG instrumentation. However, I see it copied in the MySQL2, Trilogy and Redis libraries.

With this new gem, are we shifting our paradigm a bit? 

Even though it's a NoSQL library, Redis could still use the same code. However, I don't think it'd benefit from the SQL Obfuscation code.

At a high level, it seems like the context management code could be beneficial for non-SQL libraries as well, the only unique element is the name of the context key. Should the SQL scope for this helper be removed?

arielvalentin:
Great questions and observations. 

If I'm understanding you correctly, you're seeing  
an opportunity here for us to eliminate structural duplication between these context helpers since they are structurally all the same and only differ in the a context key.

I suspect that PG does not have a similar helper because Shopify does not use postgresql. They first introduced this helper for the MySQL2 gem but recently switched to Trilogy and needed the same functionality. I suspect copy/paste modify was the quickest way to do that.

I asked them if there would be any concern in consolidating those and there wasn't any objection. 

I did not have plans on going the same for PG but I think it's only right to be symmetric and follow up with some updates there.

My goal with this gem is to have shared helpers for all SQL backed Relational Databases so I don't think of sharing this gem with Redis. In the case of Redis, I think we'd want to look at some use cases to see where it may make sense to reuse the same SQL context helper. I'd be worried that SQL DB semantics would end up inadvertently added to Redis spans. But again that is speculative. 

We can chat more about this at the next SIG

.toys/.data/releases.yml
```diff
@@ -30,6 +30,10 @@ commit_lint:
 #  *  changelog_path: Path to CHANGLEOG.md relative to the gem directory.
 #     (Required only if it is not in the expected location.)
 gems:
+  - name: opentelemetry-helpers-sql
+    directory: helpers/sql
+    version_constant: [OpenTelemetry, helpers, Sql, VERSION]
```
kaylareopelle, line 35:
```suggestion
    version_constant: [OpenTelemetry, Helpers, Sql, VERSION]
```

helpers/sql/.yardopts
```diff
@@ -0,0 +1,9 @@
+--no-private
+--title=OpenTelemetry Sql Instrumentation
+--markup=markdown
+--main=README.md
+./lib/opentelemetry/instrumentation/**/*.rb
+./lib/opentelemetry/instrumentation.rb
```
kaylareopelle, line 6:
```suggestion
./lib/opentelemetry/helpers/**/*.rb
./lib/opentelemetry/helpers.rb
```

helpers/sql/README.md
```diff
@@ -0,0 +1,53 @@
+# OpenTelemetry Sql Helpers
+
+Todo: Add a description.
```
kaylareopelle, line 3: Do you want to add more of a description? I think things are explained well below.

helpers/sql/lib/opentelemetry/helpers/sql.rb
```diff
@@ -0,0 +1,56 @@
...
+
+# Copyright The OpenTelemetry Authors
+#
+# SPDX-License-Identifier: Apache-2.0
+
+require 'opentelemetry'
+
+module OpenTelemetry
+  module Helpers
+    # Contains the OpenTelemetry instrumentation for the Sql gem
```
kaylareopelle, line 11:
```suggestion
    # Contains helpers for OpenTelemetry instrumentation related to SQL
```

helpers/sql/README.md
```diff
@@ -0,0 +1,53 @@
...
+
+## How do I get started?
+
+Install the gem using:
+
+```console
+gem install opentelemetry-helpers-sql
+```
+
+Or, if you use [bundler][bundler-home], include `opentelemetry-helpers-sql` in your `Gemfile`.
```
kaylareopelle, line 13: Since this is helper gem, and most users are likely be instrumentation authors/gems, should we include gemspec instructions?

helpers/sql/opentelemetry-helpers-sql.gemspec
```diff
@@ -0,0 +1,46 @@
...
+require 'opentelemetry/helpers/sql/version'
+
+Gem::Specification.new do |spec|
+  spec.name        = 'opentelemetry-helpers-sql'
+  spec.version     = OpenTelemetry::Helpers::Sql::VERSION
+  spec.authors     = ['OpenTelemetry Authors']
+  spec.email       = ['cncf-opentelemetry-contributors@lists.cncf.io']
+
+  spec.summary     = 'Sql helpers for the OpenTelemetry framework'
+  spec.description = 'Sql helpers for the OpenTelemetry framework'
```
kaylareopelle, line 18:
```suggestion
  spec.summary     = 'SQL helpers for the OpenTelemetry framework'
  spec.description = 'SQL helpers for the OpenTelemetry framework'
```

helpers/sql/opentelemetry-helpers-sql.gemspec
```diff
@@ -0,0 +1,46 @@
...
+  spec.required_ruby_version = '>= 3.0'
+
+  spec.add_dependency 'opentelemetry-api', '~> 1.4.0'
+
+  spec.add_development_dependency 'appraisal', '~> 2.5'
+  spec.add_development_dependency 'bundler', '~> 2.4'
+  spec.add_development_dependency 'minitest', '~> 5.0'
+  spec.add_development_dependency 'opentelemetry-test-helpers', '~> 0.3'
+  spec.add_development_dependency 'rake', '~> 13.0'
+  spec.add_development_dependency 'rubocop', '~> 1.68.0'
```
kaylareopelle, line 35:
```suggestion
  spec.add_development_dependency 'rubocop', '~> 1.71.0'
```

helpers/sql/opentelemetry-helpers-sql.gemspec
```diff
@@ -0,0 +1,46 @@
...
+  spec.summary     = 'Sql helpers for the OpenTelemetry framework'
+  spec.description = 'Sql helpers for the OpenTelemetry framework'
+  spec.homepage    = 'https://github.com/open-telemetry/opentelemetry-ruby-contrib'
+  spec.license     = 'Apache-2.0'
+
+  spec.files = Dir.glob('lib/**/*.rb') +
+               Dir.glob('*.md') +
+               ['LICENSE', '.yardopts']
+  spec.require_paths = ['lib']
+  spec.required_ruby_version = '>= 3.0'
```
kaylareopelle, line 26:
```suggestion
  spec.required_ruby_version = '>= 3.1'
```

helpers/sql/opentelemetry-helpers-sql.gemspec
```diff
@@ -0,0 +1,46 @@
...
+
+  spec.add_dependency 'opentelemetry-api', '~> 1.4.0'
+
+  spec.add_development_dependency 'appraisal', '~> 2.5'
+  spec.add_development_dependency 'bundler', '~> 2.4'
+  spec.add_development_dependency 'minitest', '~> 5.0'
+  spec.add_development_dependency 'opentelemetry-test-helpers', '~> 0.3'
+  spec.add_development_dependency 'rake', '~> 13.0'
+  spec.add_development_dependency 'rubocop', '~> 1.68.0'
+  spec.add_development_dependency 'rubocop-performance', '~> 1.22.0'
```
kaylareopelle, line 36:
```suggestion
  spec.add_development_dependency 'rubocop-performance', '~> 1.23.0'
```

helpers/sql/lib/opentelemetry/helpers/sql/version.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+# Copyright The OpenTelemetry Authors
+#
+# SPDX-License-Identifier: Apache-2.0
+
+module OpenTelemetry
+  module Helpers
+    module Sql
+      VERSION = '0.0.0'
```
kaylareopelle, line 10: Will this bump itself when toys runs? I think we've been setting it to `'0.1.0'` on recent releases of brand new gems.

helpers/sql/README.md
```diff
@@ -0,0 +1,53 @@
...
+
+## Usage
+
+This gem is intended to be used by the instrumentation libraries to provide a common set of helpers for SQL-related spans. It is not intended to be used directly by applications.
+
+Some Database libraries do not have enough context to add sufficient details to client spans. In these cases, you can use the `OpenTelemetry::Helpers::Sql.with_attributes` to create a set of shared attributes to amend to a database span.
```
kaylareopelle, line 19:
```suggestion
Some database libraries do not have enough context to add sufficient details to client spans. In these cases, you can use the `OpenTelemetry::Helpers::Sql.with_attributes` to create a set of shared attributes to amend to a database span.
```

kaylareopelle (commented):
Thanks, Ariel! I noticed a few small wording things. Other than that, I think this is good to go. I appreciate your patience!

helpers/sql/lib/opentelemetry/helpers/sql/version.rb
arielvalentin: Yes. The toys gem will bump this.

helpers/sql/README.md
arielvalentin: Likely no. I will rework the README to target the most appropriate audience.

kaylareopelle (approved):

PR #1266, opened 2024-11-24
fix: Share Faraday Attrs with Adapter Spans

arielvalentin:
The Faraday gem has additional http attributes that are sometimes missing or difficult to derive in Adapter instrumentations.


kaylareopelle (approved):
Thank you!

PR #1271, opened 2024-11-27
refactor: Use SQL helpers

arielvalentin:


arielvalentin:
@kaylareopelle btw this was my first pass of consolidating the context helpers.

instrumentation/trilogy/opentelemetry-instrumentation-trilogy.gemspec
```diff
@@ -27,14 +27,14 @@ Gem::Specification.new do |spec|
 
   spec.add_dependency 'opentelemetry-api', '~> 1.0'
   spec.add_dependency 'opentelemetry-helpers-mysql'
+  spec.add_dependency 'opentelemetry-helpers-sql'
```
kaylareopelle, line 30: Could you add the new helpers gem to the gemspec for mysql2 and pg as well?

kaylareopelle (commented):
This is looking good! Thank you for the refactor! I'm glad we have existing tests that cover the context behavior 🙂

instrumentation/trilogy/opentelemetry-instrumentation-trilogy.gemspec
arielvalentin: OMG DID I NOT

kaylareopelle: I think adding the helpers gem to the Gemfile's `:test` group concealed the issue! The tests would load the gem through that path, so it didn't need to be in the main `gemspec`.

arielvalentin: ♻️

kaylareopelle (approved):

PR #1274, opened 2024-11-29
ci: 202 is Accepted for a non-dead link.

scbjans:
The Markdown Link Check job occasionally checks a link on rubydoc.info which needs to be rendered. In that case a HTTP `202` is returned.
However that leads to an error, e.g. [here](https://github.com/open-telemetry/opentelemetry-ruby-contrib/actions/runs/12072279789/job/33665893727?pr=1272#step:4:692):

```
ERROR: 1 dead links found!
[✖] https://www.rubydoc.info/gems/opentelemetry-propagator-ottrace → Status: 202
```

A HTTP `202` should be `Accepted`.

### DIFF
--- a/.markdown-link-check.json
+++ b/.markdown-link-check.json
@@ -6,5 +6,5 @@
   ],
   "timeout": "5s",
   "retryOn429": true,
-  "aliveStatusCodes": [200, 206, 429]
+  "aliveStatusCodes": [200, 202, 206, 429]
 }


xuan-cao-swi (approved):

arielvalentin:
Any idea why the action is failing?

xuan-cao-swi:
```
ERROR: 1 dead links found!
[✖] https://opensource.org/licenses/MIT → Status: 403
```

I curl the link returns `HTTP/2 301`. I guess the ci/action exceed the rate limit for the opensource.org?

scbjans:
> ```
> ERROR: 1 dead links found!
> [✖] https://opensource.org/licenses/MIT → Status: 403
> ```
> 
> I curl the link returns `HTTP/2 301`. I guess the ci/action exceed the rate limit for the opensource.org?

I would expect a `429` in that case..? Anyway, I see the author of the [action](https://github.com/gaurav-nelson/github-action-markdown-link-check) recommends switching to [Linkspector](https://github.com/UmbrellaDocs/action-linkspector). So maybe we should close this PR and put some effort into migrating to Linkspector?

arielvalentin:
Probably. 

Another thing we may want to try is only run the action if it detects changes to markdown files.

kaylareopelle:
@xuan-cao-swi, you're closer to this than I am. I think this issue was fixed with the changes in #1322. Is that right?

scbjans:
> @xuan-cao-swi, you're closer to this than I am. I think this issue was fixed with the changes in #1322. Is that right?

should be fixed indeed.
In any case, this PR is no longer valid -> closing this.

PR #1296, opened 2024-12-09
feat: Enable support for `Regexp` patterns when subscribing to Active Support's instrumentation Events

erichmachado:
This PR aims to introduce support for `Regexp` patterns to the `OpenTelemetry::Instrumentation::ActiveSupport.subscribe` method.

This should be in line with [a similar feature available on `ActiveSupport::Notifications.subscribe`](https://guides.rubyonrails.org/active_support_instrumentation.html#subscribing-to-an-event).

With this update, users will be able to:
1. Subscribe to multiple events at once, if desired.
2. Use the `Regexp` pattern directly without having to resort to a custom `span_name_formatter` to work around crashes when processing the event name.
3. Process the incoming event name if `span_name_formatter` is set, since the `name` parameter in the initializer includes a reference to the `Regexp` pattern in this case.

### DIFF
--- a/instrumentation/active_support/lib/opentelemetry/instrumentation/active_support/span_subscriber.rb
+++ b/instrumentation/active_support/lib/opentelemetry/instrumentation/active_support/span_subscriber.rb
@@ -27,7 +27,7 @@ module OpenTelemetry
         span_name_formatter: nil
       )
         subscriber = OpenTelemetry::Instrumentation::ActiveSupport::SpanSubscriber.new(
-          name: pattern,
+          pattern: pattern,
           tracer: tracer,
           notification_payload_transform: notification_payload_transform,
           disallowed_notification_payload_keys: disallowed_notification_payload_keys,
@@ -64,8 +64,8 @@ module OpenTelemetry
         ALWAYS_VALID_PAYLOAD_TYPES = [TrueClass, FalseClass, String, Numeric, Symbol].freeze
 
         # rubocop:disable Metrics/ParameterLists
-        def initialize(name:, tracer:, notification_payload_transform: nil, disallowed_notification_payload_keys: nil, kind: nil, span_name_formatter: nil)
-          @pattern = name
+        def initialize(pattern:, tracer:, notification_payload_transform: nil, disallowed_notification_payload_keys: nil, kind: nil, span_name_formatter: nil)
+          @pattern = pattern
           @tracer = tracer
           @notification_payload_transform = notification_payload_transform
           @disallowed_notification_payload_keys = Array(disallowed_notification_payload_keys)
--- a/instrumentation/active_support/lib/opentelemetry/instrumentation/active_support/span_subscriber.rb
+++ b/instrumentation/active_support/lib/opentelemetry/instrumentation/active_support/span_subscriber.rb
@@ -65,7 +65,7 @@ module OpenTelemetry
 
         # rubocop:disable Metrics/ParameterLists
         def initialize(name:, tracer:, notification_payload_transform: nil, disallowed_notification_payload_keys: nil, kind: nil, span_name_formatter: nil)
-          @name = name
+          @pattern = name
           @tracer = tracer
           @notification_payload_transform = notification_payload_transform
           @disallowed_notification_payload_keys = Array(disallowed_notification_payload_keys)
--- a/instrumentation/active_support/lib/opentelemetry/instrumentation/active_support/span_subscriber.rb
+++ b/instrumentation/active_support/lib/opentelemetry/instrumentation/active_support/span_subscriber.rb
@@ -75,7 +75,7 @@ module OpenTelemetry
         # rubocop:enable Metrics/ParameterLists
 
         def start(name, id, payload)
-          span = @tracer.start_span(span_name(name).dup.freeze, kind: @kind)
+          span = @tracer.start_span(safe_span_name_for(name), kind: @kind)
           token = OpenTelemetry::Context.attach(
             OpenTelemetry::Trace.context_with_span(span)
           )
@@ -135,15 +135,6 @@ module OpenTelemetry
             value.to_s
           else
             value
-          end
-        end
-
-        def span_name(name)
-          case @name
-          when Regexp
-            safe_span_name_for(name)
-          else
-            @span_name ||= safe_span_name_for(@name)
           end
         end
 
--- a/instrumentation/active_support/lib/opentelemetry/instrumentation/active_support/span_subscriber.rb
+++ b/instrumentation/active_support/lib/opentelemetry/instrumentation/active_support/span_subscriber.rb
@@ -65,16 +65,17 @@ module OpenTelemetry
 
         # rubocop:disable Metrics/ParameterLists
         def initialize(name:, tracer:, notification_payload_transform: nil, disallowed_notification_payload_keys: nil, kind: nil, span_name_formatter: nil)
-          @span_name = safe_span_name_for(span_name_formatter, name).dup.freeze
+          @name = name
           @tracer = tracer
           @notification_payload_transform = notification_payload_transform
           @disallowed_notification_payload_keys = Array(disallowed_notification_payload_keys)
           @kind = kind || :internal
+          @span_name_formatter = span_name_formatter
         end
         # rubocop:enable Metrics/ParameterLists
 
         def start(name, id, payload)
-          span = @tracer.start_span(@span_name, kind: @kind)
+          span = @tracer.start_span(span_name(name).dup.freeze, kind: @kind)
           token = OpenTelemetry::Context.attach(
             OpenTelemetry::Trace.context_with_span(span)
           )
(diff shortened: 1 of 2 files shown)


instrumentation/active_support/test/opentelemetry/instrumentation/active_support/span_subscriber_test.rb
```diff
@@ -205,123 +205,226 @@ def finish(name, id, payload)
...
+    after do
+      ActiveSupport::Notifications.notifier.all_listeners_for(notification_name).each do |listener|
+        ActiveSupport::Notifications.unsubscribe(listener)
+      end
     end
 
-    it 'does not trace an event by default' do
-      ActiveSupport::Notifications.subscribe(notification_name) do
-        # pass
+    describe 'when subscribing to events directly' do
```
erichmachado, line 214: The tests in this context are the same as before, I've just added one level of indentation.

instrumentation/active_support/test/opentelemetry/instrumentation/active_support/span_subscriber_test.rb
```diff
@@ -205,123 +205,226 @@ def finish(name, id, payload)
   end
 
   describe 'instrument' do
-    before do
-      ActiveSupport::Notifications.unsubscribe(notification_name)
+    after do
+      ActiveSupport::Notifications.notifier.all_listeners_for(notification_name).each do |listener|
+        ActiveSupport::Notifications.unsubscribe(listener)
+      end
```
erichmachado, line 211: This prevents subscriptions from leaking across tests. Just using `ActiveSupport::Notifications.unsubscribe` with a `Regexp` won't work as intended.

instrumentation/active_support/test/opentelemetry/instrumentation/active_support/span_subscriber_test.rb
```diff
@@ -205,123 +205,226 @@ def finish(name, id, payload)
...
+        _(last_span.name).must_equal('bar.foo')
+        _(last_span.attributes['extra']).must_equal('context')
+        _(last_span.kind).must_equal(:client)
+      end
     end
 
-    it 'finishes spans even when complex subscribers blow up' do
-      ActiveSupport::Notifications.subscribe(notification_name, CrashingEndSubscriber.new)
-      OpenTelemetry::Instrumentation::ActiveSupport.subscribe(tracer, notification_name)
+    describe 'when subscribing to events matching a regular expression' do
```
erichmachado, line 331: This context contains the same tests as above, but exercising the pattern subscription.

instrumentation/active_support/lib/opentelemetry/instrumentation/active_support/span_subscriber.rb
```diff
@@ -137,11 +138,20 @@ def sanitized_value(value)
           end
         end
 
+        def span_name(name)
```
arielvalentin, line 141: I see what you meant now. Duping the string will now result in a separate string allocation and an additional case statement per subscription call. The difference in use between the regexp matcher and string based subscriber is only for span names right?

I think maybe it would be best in this case then to use the span name found on the payload and then keep the name instance variable as more as a indexing value that we can use to unsubscribe registrations or inspect the object to know what it's subscribed to but not use it for span naming. Does that help?

erichmachado: Thank you for your review @arielvalentin!

> I think maybe it would be best in this case then to use the span name found on the payload and then keep the name instance variable as more as a indexing value that we can use to unsubscribe registrations or inspect the object to know what it's subscribed to but not use it for span naming. Does that help?

It does help. It's also a behavioural change, meaning this test will break:

https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/7bda424d9c1ab4c26be9cb9697558e7fcd1a3d3e/instrumentation/active_support/test/opentelemetry/instrumentation/active_support/span_subscriber_test.rb#L38-L41

Which is the behaviour I was trying to preserve there. In this case, would it make sense to just remove that test?

erichmachado: Thank you! I've updated that on b8c0fca1a6c44d1800b67639a975eb36a5d117c3. Can you please check if this is what you were looking for?

instrumentation/active_support/test/opentelemetry/instrumentation/active_support/span_subscriber_test.rb
```diff
@@ -205,123 +200,238 @@ def finish(name, id, payload)
...
+        OpenTelemetry::Instrumentation::ActiveSupport.subscribe(tracer, notification_pattern)
+
+        expect do
+          ActiveSupport::Notifications.instrument(notification_name, extra: 'context')
+        end.must_raise RuntimeError
+
+        _(last_span).wont_be_nil
+        _(last_span.name).must_equal(notification_name)
+        _(last_span.attributes['extra']).must_equal('context')
+      end
```
erichmachado, line 434: I had to skip these two tests on Active Support 6.1 since there's a bug in this version that prevents the subscriber from properly finishing the spans if another subscriber raises an error.

See https://github.com/rails/rails/pull/43282 for further details.

instrumentation/active_support/test/opentelemetry/instrumentation/active_support/span_subscriber_test.rb
```diff
@@ -205,123 +200,238 @@ def finish(name, id, payload)
...
```
erichmachado, line 434: I had to skip these two tests on Active Support 6.1 since there's a bug in this version that prevents the subscriber from properly finishing the spans if another subscriber raises an error.

See https://github.com/rails/rails/pull/43282 for further details.

instrumentation/active_support/lib/opentelemetry/instrumentation/active_support/span_subscriber.rb
```diff
@@ -65,16 +65,17 @@ class SpanSubscriber
 
         # rubocop:disable Metrics/ParameterLists
         def initialize(name:, tracer:, notification_payload_transform: nil, disallowed_notification_payload_keys: nil, kind: nil, span_name_formatter: nil)
-          @span_name = safe_span_name_for(span_name_formatter, name).dup.freeze
+          @name = name
```
arielvalentin, line 68: Please rename this attribute to `pattern` now that it will match the argument passed into AS notification subscribe. This will help us disambiguate the shadowed `name` parameter passed into the `start` and `finish` methods as well. 

https://api.rubyonrails.org/classes/ActiveSupport/Notifications.html#method-c-subscribe

instrumentation/active_support/test/opentelemetry/instrumentation/active_support/span_subscriber_test.rb
```diff
@@ -205,123 +200,238 @@ def finish(name, id, payload)
...
-    end
+        describe 'when using a custom formatter' do
+          let(:span_name_formatter) { ->(name) { "custom.#{name}" } }
 
-    it 'supports setting the span kind' do
-      OpenTelemetry::Instrumentation::ActiveSupport.subscribe(tracer, 'bar.foo', nil, [], kind: :client)
-      ActiveSupport::Notifications.instrument('bar.foo', extra: 'context')
+          it 'uses the user supplied formatter' do
+            OpenTelemetry::Instrumentation::ActiveSupport.subscribe(tracer, notification_pattern, nil, nil, span_name_formatter: span_name_formatter)
+            ActiveSupport::Notifications.instrument(notification_name, extra: 'context')
```
arielvalentin, line 365: Please confirm that the indentation changes are using spaces and not tabs. There are a few other sections where it looks like indentation changed and I want to make sure that you are using spaces there as well.

arielvalentin (changes requested):

instrumentation/active_support/lib/opentelemetry/instrumentation/active_support/span_subscriber.rb
erichmachado: I've updated the attribute name to `@pattern` on ae89d187f0f4260bc5527800754d0928a5c14851 @arielvalentin, as requested.

I didn't touch the `name` parameter to avoid disrupting the interface. Let me know if you want this to be renamed as well.

instrumentation/active_support/test/opentelemetry/instrumentation/active_support/span_subscriber_test.rb
erichmachado: I've double checked it here and it's all spaces. :+1:

instrumentation/active_support/lib/opentelemetry/instrumentation/active_support/span_subscriber.rb
arielvalentin: Yes please. This Subscriber is part of the internal API of this gem and i don't think  it is used externally at the moment.

erichmachado: Thank you for confirming. It should be fixed on ed978b2dd9109abd55ad5c437d95e84f2abe472f.

arielvalentin (approved):

Issue #1261, opened 2024-11-21
Prepare gRPC instrumentation donation for release


kaylareopelle:
Overall, the goal is to update the donated gRPC gem to match the other contrib projects' structure and tooling. This will allow us to use our existing CI that runs tests and releases gems with this new instrumentation.

TODOs:
- [x] Create a gRPC-specific example to replace the Gruf example
- [x] Move from Standard.rb to Rubocop and Rubocop-Performance to match our other gems
- [x] Migrate the tests from RSpec to Minitest
- [X] Add the gRPC gem to the relevant CI workflows (ci-instrumentation.yml) 
- [x] Add the gRPC gem to dependabot.yml
- [ ] Add the gRPC gem to the `opentelemetry-instrumentation-all` gem
- [x] Update the README to match our style
- [x] Update the Changelog for compatibility with toys
- [x] Update code of conduct to match OTel docs 
- [x] Remove RBS
- [x] Remove conflicting [Code of Conduct file](https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/92d59eb33cf7fd58242cde22a7346960078ff29f/instrumentation/grpc/CODE_OF_CONDUCT.md)

Relates to the code merged in #1164

michal-kazmierczak:
I believe we should now have completed the following tasks:
- Move from Standard.rb to Rubocop and Rubocop-Performance to match our other gems (https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/1341)
- Migrate the tests from RSpec to Minitest (https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/1330)
- Remove RBS (https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/1341)
- Remove conflicting Code of Conduct file (https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/1341)

By the way, I wanted to ask about `Update code of conduct to match OTel docs` - do instrumentation gems have a separate code of conduct? From what I can see, there is one global file in the root directory. I'm not sure if anything else than removing the conflicting file needs to be done.

kaylareopelle:
Hi @michal-kazmierczak! Thanks for your gRPC example PR! I updated the issue to reflect the remaining TODOs. 

[...]

PR #1303, opened 2024-12-13
feat: parts of the gRPC istrumentation

michal-kazmierczak:
This PR includes parts of the gRPC instrumentation adaptation outlined in: https://github.com/open-telemetry/opentelemetry-ruby-contrib/issues/1261

**This PR does not make the gem ready to release.**

The primary change is that the instrumentation leverages the use of interceptors (a concept parallel to middlewares). 

Also, there is a patch that instruments every stub (client) created after the OpenTelemetry installation. This aligns the instrumentation behavior with other libraries, so end-users do not need to manually add interceptors.

I decided not to include relatively easy changes that would make the diff quite more difficult to follow (i.e. migration to Rubocop). However, this PR also includes the migration from RSpec to minitest and other minor alignments (like installation of Appraisals or unified Rakefile).

It would be great if this PR could be reviewed and merged, allowing me to proceed with the remaining adaptations.


instrumentation/grpc/Appraisals
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+# Copyright The OpenTelemetry Authors
+#
+# SPDX-License-Identifier: Apache-2.0
+
+appraise 'grpc-1.68' do
+  gem 'grpc', '~> 1.68.0'
+end
```
kaylareopelle, line 9: Thanks for testing the latest version of gRPC! We aren't always consistent about updating the Appraisals when new versions of gems come out, so we often add an Appraisal for the gem without any version constraints so the CI can notify us if a new version breaks the instrumentation. 

Even though they'd be for the same version today, let's add an unversioned Appraisal too.

```suggestion
appraise 'grpc-1.68' do
  gem 'grpc', '~> 1.68.0'
end

appraise 'grpc-latest' do
  gem 'grpc'
end
```

instrumentation/grpc/Gemfile
```diff
@@ -9,9 +9,7 @@ source "https://rubygems.org"
 # Specify your gem's dependencies in opentelemetry-instrumentation-grpc.gemspec
 gemspec
 
-gem "rake", "~> 13.0"
-gem "rspec", "~> 3.0"
-gem "standard", "~> 1.3"
-gem "opentelemetry-sdk", "~> 1.0"
-gem "opentelemetry-test-helpers"
-gem "grpc"
```
kaylareopelle, line 17: It looks like `rake` and `opentelemetry-test-helpers` aren't in the gemspec file. I think we may want to explicitly declare `rake`. If you're not using the `opentelemetry-test-helpers`, that one probably doesn't need to carry over.

instrumentation/grpc/README.md
```diff
@@ -46,18 +46,13 @@ instrumentation.
...
-
-```ruby
-  otel = OpenTelemetry::Instrumentation::Grpc.client_interceptor
-  SomeService::Stub.new(host, credentials, *args, **kwargs, interceptors: [otel])
-```
-
 ## Development
 
-After checking out the repo, run `bin/setup` to install dependencies. Then, run `rake spec` to run the tests. You can also run `bin/console` for an interactive prompt that will allow you to experiment.
+Integration tests relay on a real gRPC server which is started by relevant tests. The proto definition is located in `test/support/proto/ping.proto`. Making changes to the proto definition requires re-creating grpc-generated code. To do this, run the following command:
```
kaylareopelle, line 51:
```suggestion
Integration tests rely on a real gRPC server that is started by relevant tests. The proto definition is located in `test/support/proto/ping.proto`. Making changes to the proto definition requires re-creating gRPC-generated code. To do this, run the following command:
```

instrumentation/grpc/lib/opentelemetry/instrumentation/grpc/patches/client_stub.rb
```diff
@@ -0,0 +1,27 @@
...
+
+# Copyright The OpenTelemetry Authors
+#
+# SPDX-License-Identifier: Apache-2.0
+
+module OpenTelemetry
+  module Instrumentation
+    module Grpc
+      module Patches
+        # Module to be prepended to force GRPC to use the client interceptor by
```
kaylareopelle, line 11:
```suggestion
        # Module to be prepended to force gRPC to use the client interceptor by
```

instrumentation/grpc/test/support/proto/ping_pb.rb
```diff
@@ -0,0 +1,18 @@
+# frozen_string_literal: true
```
kaylareopelle, line 1: Is there documentation to regenerate these files? When might someone need to regenerate them?

instrumentation/grpc/test/test_helper.rb
```diff
@@ -0,0 +1,23 @@
...
+
+# Copyright The OpenTelemetry Authors
+#
+# SPDX-License-Identifier: Apache-2.0
+
+require 'bundler/setup'
+Bundler.require(:default, :development, :test)
+
+require 'minitest/autorun'
+# require 'webmock/minitest'
```
kaylareopelle, line 11:
```suggestion
```

kaylareopelle (changes requested):
Thank you, @michal-kazmierczak! This looks great! I have a few comments on the files. 

Also, I noticed is this PR is from your `main` branch. We recommend contributors put their work on a branch other than `main` because the repo's strategy is to squash when we merge, which can cause problems when you update your fork after your `main` branch is merged.

instrumentation/grpc/Gemfile
michal-kazmierczak: Thank you for spotting this. You're correct that `rake` should be included in the gemspec file. I've also added `opentelemetry-test-helpers`, as it's likely only a matter of time before it becomes necessary.

instrumentation/grpc/test/support/proto/ping_pb.rb
michal-kazmierczak: I've added a brief note in the README under the Development section explaining how to regenerate these files  using `grpc_tools_ruby_protoc` https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/1303/files#diff-8ec47973e98f69236d5b7f178b429d4d18133f13c80f07a1f7d37042c1e7a11eR51

Do you think this provides enough detail? Is there any specific information you feel should be included?

kaylareopelle: Ah, my mistake! That seems sufficient. Thank you!

instrumentation/grpc/Gemfile
kaylareopelle: Thank you!

kaylareopelle (commented):
This is great, incremental progress. In my opinion, once the branch is updated, this is ready to merge, @michal-kazmierczak.

michal-kazmierczak:
> This is great, incremental progress. In my opinion, once the branch is updated, this is ready to merge, @michal-kazmierczak.

Thank you. I've rebased it now.

kaylareopelle:
Thanks for the updates, @michal-kazmierczak! I wanted to follow up on [this comment](https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/1303#pullrequestreview-2510115206):
> Also, I noticed is this PR is from your main branch. We recommend contributors put their work on a branch other than main because the repo's strategy is to squash when we merge, which can cause problems when you update your fork after your main branch is merged.

Here's [the section of our contributing guide](https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/main/CONTRIBUTING.md#always-work-in-a-branch-from-your-fork) that goes over this concept.

Would you copy your work onto a different branch within your fork and reopen the PR from that branch?

michal-kazmierczak:
@kaylareopelle thanks for pointing that out. Initially, I assumed that since the issue was only potentially problematic on my end, I could manage and leave it on the main branch. However, to avoid any confusion, I’ve now created a new PR (https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/1330), with the only difference being that it’s based on a feature branch from my fork. I'm closing this PR.

PR #1312, opened 2024-12-19
fix: Format gat commands

robertlaurin:
We don't know how to format [gat](https://github.com/memcached/memcached/blob/master/doc/protocol.txt#L425-L440), lets help the instrumentation do it properly.


plantfansam (approved):

ericmustin (approved):

fbogsany (approved):
Looks gat to me!

fbogsany (changes requested):
Looks like `gat` isn't supported in (one of?) the `dalli` version we test against. We should consider whether to bump the minimum version we support.

instrumentation/dalli/test/opentelemetry/instrumentation/dalli/instrumentation_test.rb
```diff
@@ -127,6 +127,14 @@
       _(span.name).must_equal 'set'
       _(span.attributes['db.statement']).must_equal 'set ?'
     end
+
+    it 'supports gat' do
+      dalli.gat('foo')
```
fbogsany, line 132: `gat` was [added to `dalli` in 3.0.0](https://github.com/petergoldstein/dalli/blob/f72450aeb23a1a553299157f2a070eb421994b2c/CHANGELOG.md?plain=1#L178). We could `skip` if the dalli version is < 3.0.0 or set 3.0.0 as the min supported version.

fbogsany: My preference is to `skip`, since we only need it in this test, and the test matrix _should_ cover 3.0.0.

kaylareopelle (approved):

fbogsany (approved):

PR #1321, opened 2025-01-06
Fix "unknown keyword: :label" ArgumentError in Rdkafka

venkatagiri:
Support for adding a label to a kafka payload(within rdkafka) was added in v0.15.0. https://github.com/karafka/rdkafka-ruby/pull/381
This PR updates the method signature to allow `label` keyword argument.


arielvalentin (changes requested):

I would like to request a few changes here:

1. Please add an automated test to avoid regressions
2. This change should ensure that the instrumentation requires a minimal version compatible for rdkafka v0.15.0 e.g. https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/main/instrumentation/rdkafka/lib/opentelemetry/instrumentation/rdkafka/instrumentation.rb#L12
3. We may want to consider releasing a version of the instrumentation that checks for a Max compatibility with v0.14.x and release that first then release this PR

PR #1324, opened 2025-01-07
feat: gradually add metrics capabilities to Instrumentation::Base

zvkemp:
Proposed implementation for adding **optional** metrics capabilities to existing instrumentation libraries.

See https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/1377, https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/1314, or https://github.com/zvkemp/opentelemetry-ruby-contrib/pull/1 (example of observables) for example implementations. (<- note that these PRs are based on top of this branch, so you may see some duplication between here and there)

Proposed api:

- Libraries with metrics must add a boolean option called `:metrics`.
- Each instrument is configured at compile/boot time — this prevents library authors from creating instruments which are identically named, but have different attributes, callbacks, or advisory parameters, which makes registering them quite difficult. The intention here is that they are declared once in the instrumentation, then subsequent lookups in the instrumentation _instance_ will spawn the instrument (if it has not been spawned already).
- All of the logic in base/metrics works in the absence of metrics-sdk or metrics-api, so library authors can add metrics to libraries without requiring users to use the metrics system at all. This means that any of the instrument-lookup instance methods may return nil (so this will need to be handled, e.g. via short-circuiting https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/1377/files#diff-6732dfa68439092c8ae9c61ca4dfd067a76e48f9cb6e37768b12aedb5cfb7f50R20)

TODO:
- [ ] requires https://github.com/open-telemetry/opentelemetry-ruby/pull/1804
- [ ] [FIXME](https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/1324/files#diff-7379de806670e7a913ebbe1d8e27945a6367cc83c94d79e55434de15cdd16732R365) should instruments not declared in advance return a valid instrument anyway?
- [x] ensure instrument opts are forwarded correctly, or validate them when they are declared
- [x] dependency load order is very important here; if the Metrics API isn't loaded before Instrumentation::Base, the conditional definitions are not run. We either need to hook into `Kernel.require` and detect when Metrics is loaded (probably too perilous to really consider, as any instrumentations with conditional metrics will need to be re-evaluated at this time), or eagerly attempt to require the metrics-api gem when Instrumentation::Base is loaded.

### DIFF
--- a/instrumentation/base/lib/opentelemetry/instrumentation/base.rb
+++ b/instrumentation/base/lib/opentelemetry/instrumentation/base.rb
@@ -69,8 +69,9 @@ module OpenTelemetry
           integer: ->(v) { v.is_a?(Integer) },
           string: ->(v) { v.is_a?(String) }
         }.freeze
+        SINGLETON_MUTEX = Thread::Mutex.new
 
-        private_constant :NAME_REGEX, :VALIDATORS
+        private_constant :NAME_REGEX, :VALIDATORS, :SINGLETON_MUTEX
 
         private :new
 
@@ -163,8 +164,10 @@ module OpenTelemetry
         end
 
         def instance
-          @instance ||= new(instrumentation_name, instrumentation_version, install_blk,
-                            present_blk, compatible_blk, options)
+          @instance || SINGLETON_MUTEX.synchronize do
+            @instance ||= new(instrumentation_name, instrumentation_version, install_blk,
+                              present_blk, compatible_blk, options)
+          end
         end
 
         private
@@ -189,13 +192,15 @@ module OpenTelemetry
         end
       end
 
-      attr_reader :name, :version, :config, :installed, :tracer
+      attr_reader :name, :version, :config, :installed, :tracer, :meter
 
       alias installed? installed
 
+      require_relative 'metrics'
+      prepend(OpenTelemetry::Instrumentation::Metrics)
+
       # rubocop:disable Metrics/ParameterLists
-      def initialize(name, version, install_blk, present_blk,
-                     compatible_blk, options)
+      def initialize(name, version, install_blk, present_blk, compatible_blk, options)
         @name = name
         @version = version
         @install_blk = install_blk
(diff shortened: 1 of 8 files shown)


instrumentation/base/lib/opentelemetry/instrumentation/base.rb
```diff
@@ -163,8 +164,10 @@ def option(name, default:, validate:)
         end
 
         def instance
-          @instance ||= new(instrumentation_name, instrumentation_version, install_blk,
-                            present_blk, compatible_blk, options)
+          @instance || SINGLETON_MUTEX.synchronize do
```
robertlaurin, line 167: Can you explain the rational for why this mutex lock is necessary?

instrumentation/base/lib/opentelemetry/instrumentation/metrics.rb
```diff
@@ -0,0 +1,201 @@
...
+  # any of the definition queries have been made may cause false negatives.
+  require 'opentelemetry-metrics-api'
+rescue LoadError # rubocop: disable Lint/SuppressedException
+end
+
+module OpenTelemetry
+  module Instrumentation
+    # Extensions to Instrumentation::Base that handle metrics instruments.
+    # The goal here is to allow metrics to be added gradually to instrumentation libraries,
+    # without requiring that the metrics-sdk or metrics-api gems are present in the bundle
```
robertlaurin, line 20: >without requiring that the metrics-sdk or metrics-api gems are present in the bundle

Instrumentation base has a [dependency on the opentelemetry-api gem](https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/main/instrumentation/base/opentelemetry-instrumentation-base.gemspec#L28), if we want to start adding metric support in the instrumentation libraries the way forward is to get the metrics api stable and merged in the api proper.

instrumentation/base/lib/opentelemetry/instrumentation/base.rb
```diff
@@ -317,13 +329,17 @@ def config_options(user_config)
       # will be OTEL_RUBY_INSTRUMENTATION_SINATRA_ENABLED. A value of 'false' will disable
       # the instrumentation, all other values will enable it.
       def enabled_by_env_var?
+        !disabled_by_env_var?
+      end
+
+      def disabled_by_env_var?
```
robertlaurin, line 335: I don't actually see where this is being used.

robertlaurin (changes requested):
Lets get the metrics api merged into opentelemetry-api instead of working around it in this fashion.

PR #1331, opened 2025-01-08
fix: Add Concurrent Ruby dependency to Rails

arielvalentin:
Rails relies on concurrent ruby for features like async queries.

This results in inconsistent traces since context is not properly propagated between threads.

This change adds Concurrent Ruby as a transitive dependency to mitigate confusion when uses the rails instrumentation independent of the all gem.


kaylareopelle (approved):
nice catch!

PR #1351, opened 2025-01-17
feat: track ActiveRecord async queries

### LINKED ISSUE #1219: [ActiveRecord] Add Attributes for Async Queries
`ActiveRecord` 7.1 introduced async query methods[^1], which leverages concurrency primitives via `Promise` API to defer executing queries. 

We would like to enrich `ActiveRecord` related spans to know whether or not they were executed from the async context. 

## Additional Notes 

The `sql.active_record` notification payload includes a `async` attribute:  
https://github.com/rails/rails/blob/616d3a7675c18ae2d38d3116de47e2e7308cfbd8/activerecord/lib/active_record/connection_adapters/abstract_adapter.rb#L1135

We may be able to enrich the `ActiveRecord` span and amend it as a Shared DB attribute for the driver itself however we do not currently use notifications in this instrumentation.

Worst case scenario, we monkey patch existing `async` methods, concurrency primitives. 

[^1]: https://github.com/rails/rails/pull/44446

zvkemp:
Fixes #1219.

- Patches `ActiveRecord::FutureResult` to accept a span name. Note that this is _only_ to make it internally consistent with the existing ActiveRecord instrumentation span names; I do not believe these follow the semconv recommendations for database operations, but that can be addressed for both sync and async queries in a future version. See notes below for why this ivar was necessary.
- Conditionally patches `ActiveRecord::ConnectionAdapters::ConnectionPool#schedule_query`, which is the method that pushes an async task into the executor. This patch is skipped if the ConcurrentRuby instrumentation is installed.
- The various async scenarios are documented in the tests, but see here and the accompanying issue for more detailed explanations.

todo list:
- [x] test with async aggregation queries (e.g. #async_count)
- [x] test with multiple concurrent queries (ended up not using any stacks or additional context, so there are unlikely to be issues here).
- [x] I don't love how the span name is derived in `FutureResult#execute_query` — the Rails authors seem to have just decided to forward arguments as a blob through all of the async stuff; these are re-expanded into named arguments in `raw_exec_query` back in the connection handler. It's never guaranteed that `name` will remain the second argument.
  - more on this: unfortunately we _also_ can't guarantee that the query is executed in an otel context that can be propagated from the scheduler method — i.e. we can set the desired span name into the context, which works _if_ the async executor enqueues the task or executes it inline, because we can copy that context into the task. However, if `to_a` is called before the task is picked up, the query is executed synchronously in the parent span/context. I think it will need to be copied in as an ivar on the future result instead.
- [x] Neither `User query` (existing otel instrumentation name) nor `User Load` (ActiveRecord name) follow the semconv guidelines for the span name ([the preferred name](https://opentelemetry.io/docs/specs/semconv/database/database-spans/#generating-a-summary-of-the-query-text) would be `SELECT users`, but changing this is probably not the job of this PR. I'll try to at least make this internally consistent for now (edit: fixed to use `User query`)
- [x] There is one test (the one that calls `skip`) that demonstrates a false-positive `async: true` attribute. This false positive should be fixed if possible (**edit**: calling this a 'wontfix' for now, as it's really a Rails bug).
- ~~[ ] Decide whether or not the context propagation here is necessary, or if the concurrent-ruby instrumentation will always be installed when the Rails instrumentation is installed (see #1219)~~ (punted on this for now by conditionally including it)
- [ ] Question for discussion — **is the `schedule [model] query` span useful at all**? Seems like there is a small amount of overhead in getting the async query into the executor and then subsequently running it, so perhaps? There is also the issue that if we include it, it becomes the parent of the async query spans, which may not start before the `schedule query` end_timestamp (so they are not enclosed by the parent at all).


arielvalentin:
@zvkemp thank you for taking the time to provide these changes. 

I have this on my list of changes to review and am currently trying to get some input/feedback from our Rails and Ruby core contributors at GitHub who requested this instrumentation #1219 (cc: @bensheldon) 

I appreciate your patience and understanding while we get as many core Rails folks a chance to review these changes as possible.

zvkemp:
keep

PR #1369, opened 2025-01-27
docs: required version in Rails README from 0.24 to 0.34

j15e:
I think this is a small typo, it should be 0.34 and not 0.24, correct? [As 0.35 dropped support for 6.1](https://github.com/open-telemetry/opentelemetry-ruby-contrib/releases/tag/opentelemetry-instrumentation-rails%2Fv0.35.0)


arielvalentin (approved):

arielvalentin:
✨ thanks for catching this @j15e

PR #1381, opened 2025-01-30
test: add metrics-test-helpers

zvkemp:
A new internal gem to facilitate testing instrumentation libraries with and without the metrics-api/sdk active. See the README.md for further information


arielvalentin:
👋🏼 I think this is better suited for the https://github.com/open-telemetry/opentelemetry-ruby repo or as a part of the existing https://github.com/open-telemetry/opentelemetry-ruby/tree/main/test_helpers gem

PR #1388, opened 2025-02-01
refactor: Define a single ruby required version

arielvalentin:
I _think_ this will make it easier to enforce the same min Ruby version, as opposed to using a _find and replace_ strategy or a script to bump all the versions.

IDK, wdyt?


kaylareopelle:
I like it! Defining the version in one place seems much easier to maintain. Down to experiment, especially since Ruby 3.1 will sunset at the end of March. Not too long before we could try again.

kaylareopelle (approved):
Thanks, @arielvalentin!

PR #1390, opened 2025-02-01
test: Make Net::HTTP Deterministic

arielvalentin:
We would occasionally get errors where the test would attempt to make actual HTTP calls to example.com because dedpending on the order the tests would run WebMock would be configured to allow network calls.

> OpenTelemetry::Instrumentation::Net::HTTP::Instrumentation::untraced?::untraced context#test_0002_no-ops on #connect:
> Net::OpenTimeout: Failed to open TCP connection to example.com:80 (execution expired)

https://github.com/open-telemetry/opentelemetry-ruby-contrib/actions/runs/13089176560/job/36523848687?pr=1388

This change ensures that WebMock is reset between test cases and raise an error when unexpected network requests occur.


arielvalentin:
The irony. 

image

PR #1412, opened 2025-02-17
fix: handle force_flush for rake task with arguments

TeamNautilus:
This PR fixes an issue where `force_flush` was not correctly triggered when executing a Rake task with arguments. Previously, `::Rake.application.top_level_tasks` stored task names including arguments (e.g., "task_name[arg]"), while the comparison in force_flush was done against name (which did not include arguments). This caused `force_flush` to be skipped for tasks with arguments.

# Changes in this PR
* Updated force_flush logic to strip arguments from top_level_tasks before comparison.
* Modified tests to cover cases where a Rake task has arguments.
* Ensured force_flush is correctly called when a task with arguments is invoked.


kaylareopelle:
👋 Hi @TeamNautilus! Thanks for opening this PR! Would you mind [signing the CLA?](https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/1412#issuecomment-2663120062)

arielvalentin (approved):

kaylareopelle (approved):

PR #1426, opened 2025-02-21
chore: Enable Gemfile Dev Deps

arielvalentin:
> Enforce that development dependencies for a gem are specified in Gemfile, rather than in the gemspec using add_development_dependency. Alternatively, using EnforcedStyle: gemspec, enforce that all dependencies are specified in gemspec, rather than in Gemfile.

https://docs.rubocop.org/rubocop/cops_gemspec.html#gemspecdevelopmentdependencies


bin/update-dependencies
```diff
@@ -8,7 +8,7 @@ if [ -z "$2" ] ; then
   echo "gem version is required!" && exit 2;
 fi
 
-for gemspec in $(git ls-files \*.gemspec)
+for gemspec in $(git ls-files \*Gemfile)
```
kaylareopelle, line 11: I know we usually use this for dev dependencies, but prior to this PR, the script could also be used with general dependencies too. if this is all that needs to be changed, what do you think about updating documentation (possibly in `CONTRIBUTING.md`) to note how to adjust the script to update dependencies in gemspecs?

kaylareopelle (approved):
Thank you! Approved as-is, one small docs suggestion.

P.S. I didn't realize how merging the Rubocop PR would impact this one! Apologies for the merge conflicts!

bin/update-dependencies
arielvalentin: True! I'll make sure this supports both.

kaylareopelle (approved):
Thank you!

kaylareopelle:
@arielvalentin - Thanks for the updates! Once the conflicts are resolved I'll merge this in. It doesn't look like I have the option to resolve them myself.

PR #1442, opened 2025-03-11
feat: update gem version requirement of rdkafka

joker1007:
The current latest version of rdkafka-ruby is v0.21.0, which is not available in the current instrumentation.
Therefore, update the version of gem requriement.

The changelog of rdkafka does not seem to show any breaking changes that affect tracing.


xuan-cao-swi:
I think it's ok to leave the `each_batch` function here or completely remove this function for new instrumentation-rdkafka version, rather than having the condition to check version each time.

rdkafka seems still keep the [each_batch](https://github.com/karafka/rdkafka-ruby/blob/main/lib/rdkafka/consumer.rb#L613) but with raise error. If user use > v0.20.0, and still use `each_batch` in their app, with or without instrumentation-rdkafka, they will get the error.

joker1007:
@xuan-cao-swi 
Thank you for your review comment!
I deleted the `each_batch` method completely.
Do you think this patch is OK?

instrumentation/rdkafka/lib/opentelemetry/instrumentation/rdkafka/instrumentation.rb
```diff
@@ -11,7 +11,7 @@ module Rdkafka
       class Instrumentation < OpenTelemetry::Instrumentation::Base
         compatible do
           gem_version = Gem::Version.new(::Rdkafka::VERSION)
-          Gem::Requirement.new('>= 0.10.0', '< 0.20.0').satisfied_by?(gem_version)
+          Gem::Requirement.new('>= 0.10.0', '< 0.22.0').satisfied_by?(gem_version)
```
xuan-cao-swi, line 14: I think @arielvalentin means to change to `Gem::Requirement.new('>= 0.20.0').satisfied_by?(gem_version)` for new instrumentation-rdkafka version

kaylareopelle:
Hi @joker1007, could you take a look at @xuan-cao-swi's comment?

joker1007:
@xuan-cao-swi @kaylareopelle 

I've been so busy for a couple of weeks that I haven't had time to check GitHub properly.
I am very sorry for the delay in responding to your comments.

kaylareopelle (commented):
Thanks @joker1007! I'm sorry for our subsequent delay. I really appreciate your patience. I was getting ready to merge this and noticed a few other things. 

I realized today that the supported version range of the rdkafka gem published on their repo is [0.18.0 - 0.22.0](https://github.com/karafka/rdkafka-ruby?tab=readme-ov-file#versions). We generally have instrumentation support for all the versions supported by the maintainers of the library. I think we need to increase the min version to 0.18.0. 

Would you mind making the following updates? 
* updating this PR to wrap a condition around `each_batch` based on the versions that support it, `0.19.0` and below? 
* update the `Appraisals` file to run the tests on the newly supported version range.

joker1007:
@kaylareopelle 
Thanks for your reply! I don't mind.

OK, I'm going to do the following tasks.

1. Revert `each_batch` method for v0.20.0 > V >= v0.18.0 (consider patch version)
2. Add version 0.20.0, 0.21.0, 0.22.0 to Appraisals

I can do more tasks if you discard versions that are not maintained from supported versions.

3. Increase Gem::Requirement version up to 0.18.0
4. Delete old versions earlier than 0.18.0 from Appraisals.

For now, I will implement 1 and 2.

joker1007:
@kaylareopelle 

It appears that approval is needed again to run the workflow.

kaylareopelle:
Hi @joker1007, thanks for your understanding. Please take care of 3 and 4 if you can as well!

joker1007:
@kaylareopelle 

I have done the tasks.

In summary, I have implemented the following

1. support the `each_batch` method under 0.20.0
2. add 0.20.0 and 0.21.0 to Appraisal and remove versions below 0.18.0
3. set Gem::Requirement to 0.18.0 or higher
4. remove if-statement considering codes under 0.13.0

kaylareopelle (approved):

kaylareopelle:
This looks great! Thank you for the changes, @joker1007.

PR #1446, opened 2025-03-12
test: Lock graphql 2.1 testing to 2.1.13

kaylareopelle:
GraphQl 2.1.14 is corrupted. Until that is resolved, lock the 2.1 tests to 2.1.13 and remove 2.1 from the appraisal loop.


kaylareopelle:
a new version has been released with a fix, so we don't need this anymore! 🎉

PR #1480, opened 2025-04-14
feat: support meta protocol instrumentation

nickamorim:
## Description

Dalli supports [Memcached's meta protocol](https://docs.memcached.org/protocols/meta/) as of [3.2.0](https://github.com/petergoldstein/dalli/blob/main/CHANGELOG.md#320). This PR adds support for instrumentation when clients are using the meta protocol.

### DIFF
--- a/instrumentation/dalli/lib/opentelemetry/instrumentation/dalli/instrumentation.rb
+++ b/instrumentation/dalli/lib/opentelemetry/instrumentation/dalli/instrumentation.rb
@@ -34,6 +34,7 @@ module OpenTelemetry
             ::Dalli::Server.prepend(Patches::Server)
           else
             ::Dalli::Protocol::Binary.prepend(Patches::Server)
+            ::Dalli::Protocol::Meta.prepend(Patches::Server) if defined?(::Dalli::Protocol::Meta)
           end
         end
       end
--- a/instrumentation/dalli/lib/opentelemetry/instrumentation/dalli/patches/server.rb
+++ b/instrumentation/dalli/lib/opentelemetry/instrumentation/dalli/patches/server.rb
@@ -8,7 +8,7 @@ module OpenTelemetry
   module Instrumentation
     module Dalli
       module Patches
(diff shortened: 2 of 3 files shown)


sam-obeid (approved):

instrumentation/dalli/lib/opentelemetry/instrumentation/dalli/instrumentation.rb
```diff
@@ -34,6 +34,7 @@ def add_patches
             ::Dalli::Server.prepend(Patches::Server)
           else
             ::Dalli::Protocol::Binary.prepend(Patches::Server)
+            ::Dalli::Protocol::Meta.prepend(Patches::Server) if defined?(::Dalli::Protocol::Meta)
```
kaylareopelle, line 37: The tests currently run on the default protocol. 
https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/06c8de064bec05c787f053967689bcfcd447c840/instrumentation/dalli/test/opentelemetry/instrumentation/dalli/instrumentation_test.rb#L18 

In the latest version, this should default to `binary`. Now that the protocol could be `:binary` or `:meta`, what do you think about updating the tests to run on a `:meta` protocol?

kaylareopelle (commented):
Hi @nickamorim! Thanks for your contribution! 

There's a comment in the server patch that should be updated too: https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/f631a26222b5cc3b20224c081b6d594568089044/instrumentation/dalli/lib/opentelemetry/instrumentation/dalli/patches/server.rb#L11

I wonder if we should rename the patch now that it applies to so many different classes?

nickamorim:
> I wonder if we should rename the patch now that it applies to so many different classes?

I'm not exactly sure what this would look like but it seems that `Dalli::Server` is deprecated in favour of specifying the protocol https://github.com/petergoldstein/dalli/blob/main/lib/dalli/server.rb#L4.

As an aside, did you want me to update the version too as part of this PR?

fbogsany (approved):

robertlaurin (approved):

Issue #1502, opened 2025-04-23
Additional low-level instrumentation spans created for some instrumentation libraries


ezhang6811:
Some instrumentation libraries, when enabled along with Net:Http instrumentation, generate multiple spans for a single request.

For example, both Faraday and AWS SDK instrumentation libraries generate multiple client spans (and thus multiple metrics) due to the underlying Net:Http instrumentation that is also invoked.

Would it be possible to suppress the Net:Http instrumentation in these higher level libraries such as Faraday and AWS SDK to prevent these extra spans when both libraries are enabled?

jterapin:
Hi there! AWS SDK for Ruby here. 

From what I observed, when using both opentelemetry-instrumentation's `aws_sdk` and `net_http` gems - you could configure `suppress_internal_instrumentation` to `true` on the `OpenTelemetry::Instrumentation::AwsSdk` to suppress internal spans within the AWS SDK. This suppression includes any spans generated by the `net_http` as well.

To demonstrate:
```ruby
require 'aws-sdk-s3'
require 'opentelemetry-instrumentation-net_http'
require 'opentelemetry-instrumentation-aws_sdk'
require 'opentelemetry-exporter-otlp'

ENV['OTEL_TRACES_EXPORTER'] ||= 'console'

OpenTelemetry::SDK.configure do |c|
  c.use 'OpenTelemetry::Instrumentation::Net::HTTP'
  c.use 'OpenTelemetry::Instrumentation::AwsSdk', {
    suppress_internal_instrumentation: true
  }
end

otel_provider = Aws::Telemetry::OTelProvider.new
client = Aws::S3::Client.new(telemetry_provider: otel_provider)
client.list_buckets
```

And its output:
```
I, [2025-04-24T07:25:32.189012 #70251]  INFO -- : Instrumentation: 
  OpenTelemetry::Instrumentation::AwsSdk was successfully installed with the following options 
  {inject_messaging_context: false, suppress_internal_instrumentation: true}
I, [2025-04-24T07:25:32.189452 #70251]  INFO -- : Instrumentation: 
  OpenTelemetry::Instrumentation::Net::HTTP was successfully installed with the following options
   {untraced_hosts: []}
#<struct OpenTelemetry::SDK::Trace::SpanData
 name="S3.ListBuckets",
 kind=:client,
 status=#<OpenTelemetry::Trace::Status:0x0000000140c200b0 @code=1, @description="">,
 ...
>
```

[...]

PR #1506, opened 2025-04-26
feat: suppress internal spans with Faraday instrumentation

ezhang6811:
Addresses #1502 for Faraday library instrumentation. Allow user to set `suppress_internal_instrumentation` flag when configuring Faraday instrumentation to suppress additional client spans generated from underlying Net::Http instrumentation.

Before merging, waiting for update on Github issue to see if internal span suppression can be made default behavior for both Faraday and AWS SDK.

### DIFF
--- a/instrumentation/faraday/README.md
+++ b/instrumentation/faraday/README.md
@@ -19,7 +19,7 @@ To install the instrumentation, call `use` with the name of the instrumentation.
 ```ruby
 OpenTelemetry::SDK.configure do |c|
   c.use 'OpenTelemetry::Instrumentation::Faraday'
-  suppress_internal_instrumentation: true
+  enable_internal_instrumentation: false
 end
 ```
 
@@ -33,8 +33,8 @@ end
 
 ### Configuration options
 This instrumentation offers the following configuration options: 
-* `suppress_internal_instrumentation` (default: `false`): When set to `true`, any spans with 
- span kind of `internal` are suppressed from traces.
+* `enable_internal_instrumentation` (default: `false`): When set to `true`, any spans with 
+ span kind of `internal` are included in traces.
 
 ## Examples
 
--- a/instrumentation/faraday/README.md
+++ b/instrumentation/faraday/README.md
@@ -18,8 +18,9 @@ To install the instrumentation, call `use` with the name of the instrumentation.
 
 ```ruby
 OpenTelemetry::SDK.configure do |c|
-  c.use 'OpenTelemetry::Instrumentation::Faraday'
-  enable_internal_instrumentation: false
+  c.use 'OpenTelemetry::Instrumentation::Faraday', {
+    enable_internal_instrumentation: false
+  }
 end
 ```
 
--- a/instrumentation/faraday/README.md
+++ b/instrumentation/faraday/README.md
@@ -19,6 +19,7 @@ To install the instrumentation, call `use` with the name of the instrumentation.
 ```ruby
 OpenTelemetry::SDK.configure do |c|
   c.use 'OpenTelemetry::Instrumentation::Faraday'
+  suppress_internal_instrumentation: true
 end
 ```
 
@@ -29,6 +30,11 @@ OpenTelemetry::SDK.configure do |c|
   c.use_all
 end
 ```
+
+### Configuration options
+This instrumentation offers the following configuration options: 
+* `suppress_internal_instrumentation` (default: `false`): When set to `true`, any spans with 
+ span kind of `internal` are suppressed from traces.
 
 ## Examples
 
--- a/instrumentation/faraday/lib/opentelemetry/instrumentation/faraday/instrumentation.rb
+++ b/instrumentation/faraday/lib/opentelemetry/instrumentation/faraday/instrumentation.rb
@@ -28,7 +28,7 @@ module OpenTelemetry
 
         option :span_kind, default: :client, validate: %i[client internal]
         option :peer_service, default: nil, validate: :string
-        option :suppress_internal_instrumentation, default: false, validate: :boolean
+        option :enable_internal_instrumentation, default: false, validate: :boolean
 
         private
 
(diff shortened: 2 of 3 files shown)


instrumentation/faraday/lib/opentelemetry/instrumentation/faraday/middlewares/tracer_middleware.rb
```diff
@@ -40,7 +40,13 @@ def call(env)
               ) do |span|
                 OpenTelemetry.propagation.inject(env.request_headers)
 
-                app.call(env).on_complete { |resp| trace_response(span, resp.status) }
+                if config[:suppress_internal_instrumentation]
+                  OpenTelemetry::Common::Utilities.untraced do
```
arielvalentin, line 44: I believe that using untraced will suppress instrumentation for all child spans. 

Is this what you want?

instrumentation/faraday/lib/opentelemetry/instrumentation/faraday/instrumentation.rb
```diff
@@ -28,6 +28,7 @@ class Instrumentation < OpenTelemetry::Instrumentation::Base
 
         option :span_kind, default: :client, validate: %i[client internal]
```
arielvalentin, line 29: What about adding an option here like `:none`?

instrumentation/faraday/lib/opentelemetry/instrumentation/faraday/middlewares/tracer_middleware.rb
ezhang6811: Yes, currently Faraday will generate child spans for its under-the-hood Net::Http instrumentation which creates an additional dependency metric for each request. We want to give the customer the option (preferably default) to suppress this.

instrumentation/faraday/lib/opentelemetry/instrumentation/faraday/instrumentation.rb
ezhang6811: Could you elaborate on this? The current behavior of setting the flag to false will ensure no spans are suppressed.

instrumentation/faraday/lib/opentelemetry/instrumentation/faraday/middlewares/tracer_middleware.rb
ezhang6811: Are you referring to the configuration below?
```
'OpenTelemetry::Instrumentation::Faraday' => {
    suppress_internal_instrumentation: true,
    span_kind: :internal
}
```
This may not make sense, but I think flexibility to configure the options separately is better. Some users may also want client spans where the internal instrumentation isn't suppressed for better transparency, which discourages the coupling of the two settings.

Also, why do you prefer the polymorphism approach when the conditional is simple enough?

ezhang6811: Are you referring to the configuration below?
```
'OpenTelemetry::Instrumentation::Faraday' => {
    suppress_internal_instrumentation: true,
    span_kind: :internal
}
```
This may not make sense, but I think flexibility to configure the options separately is better. Some users may also want client spans where the internal instrumentation isn't suppressed for better transparency, which discourages the coupling of the two settings.

Also, why do you prefer the polymorphism approach when the conditional is simple enough?

arielvalentin: > Are you referring to the configuration below?
> 
> ```
> 
> 'OpenTelemetry::Instrumentation::Faraday' => {
> 
>     suppress_internal_instrumentation: true,
> 
>     span_kind: :internal
> 
> }
> 
> ```
> 
> This may not make sense, but I think flexibility to configure the options separately is better. Some users may also want client spans where the internal instrumentation isn't suppressed for better transparency, which discourages the coupling of the two settings.
> 

The more configuration options we offer the more complex the instrumentations are to understand for our users.

The combination of having an `internal` span and also suppressing the child span would be an invalid combination. I don't think it's one we want to support. 

> 
> Also, why do you prefer the polymorphism approach when the conditional is simple enough?

Yes, I agree that this conditional check is a minimal implementation but I also believe that the instrumentation should do the minimal possible to achieve its goal. The config option cannot change for the life of the instrumentation so we are checking it with each request.

Yes, it will likely be optimized away by a JIT at some point and it's just a quick check, but for now anything we can remove from the users request path we should consider doing. _Keep in mind that my comment was more of a nitpick and not a blocker._

instrumentation/faraday/README.md
```diff
@@ -19,6 +19,7 @@ To install the instrumentation, call `use` with the name of the instrumentation.
 ```ruby
   c.use 'OpenTelemetry::Instrumentation::Faraday'
+  suppress_internal_instrumentation: true
```
arielvalentin, line 22: I see a related PR from @jterapin renaming the option as `enable_internal_instrumentation`, which is language I think I prefer when naming the options.  

Ideally these options would have symmetric naming and functionality.

https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/1533

Would you be amenable to changing the name?

ezhang6811: Updated.

arielvalentin (approved):

arielvalentin (approved):

instrumentation/faraday/README.md
```diff
@@ -19,6 +19,7 @@ To install the instrumentation, call `use` with the name of the instrumentation.
 ```ruby
   c.use 'OpenTelemetry::Instrumentation::Faraday'
+  enable_internal_instrumentation: false
```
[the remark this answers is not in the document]
ezhang6811: Fixed, thanks for catching that

arielvalentin:
@ezhang6811 if the tests all pass I will merge and you should expect a Release this Tuesday 2025-06-03

Issue #1515, opened 2025-05-01
EC2 resource detector missing constant


xuan-cao-swi:
**Description of the bug**

Missing constant
```
#<NameError: uninitialized constant OpenTelemetry::SemanticConventions

          RESOURCE = OpenTelemetry::SemanticConventions::Resource
                                    ^^^^^^^^^^^^^^^^^^^^^>
/root/.rbenv/versions/3.1.0/lib/ruby/gems/3.1.0/gems/opentelemetry-resource-detector-aws-0.2.0/lib/opentelemetry/resource/detector/aws/ec2.rb:34:in `<module:EC2>'
```

I have to add `require 'opentelemetry/semantic_conventions/resource'` in `opentelemetry/resource/detector/aws/ec2.rb ` to resolve the issue.

**Share details about your runtime**

Operating system details: Linux, Ubuntu 22.04.5 LTS
RUBY_ENGINE: "ruby"
RUBY_VERSION: "3.1.0"
RUBY_DESCRIPTION: "ruby 3.1.0p0 (2021-12-25 revision fb4df44d16) [aarch64-linux]"

xuan-cao-swi:
cc @yiyuan-he

yiyuan-he:
Thanks for finding this Xuan - I'll take a look and PR a fix for this today

PR #1520, opened 2025-05-02
fix: add missing semantic conventions require in AWS resource detectors

yiyuan-he:
## What does this pull request do?
Fixes an issue where AWS resource detectors (EC2, ECS, Lambda) were using `OpenTelemetry::SemanticConventions::Resource` without requiring the module that defines it. This caused "uninitialized constant" errors when using these detectors in production environments where the semantic conventions module wasn't already loaded.

This PR implements a dual strategy for dependency management:

1. Added explicit `require 'opentelemetry/semantic_conventions/resource'` statements to each AWS detector file to ensure proper dependency loading regardless of the runtime environment.
2. Maintained these individual requires while also adding the require statement to the top-level detector file.

This approach ensures that we are explicit about the following behavior:
- When the gem is required normally via `require 'opentelemetry-resource-detector-aws'`, all dependencies are loaded properly.
- When individual detectors are required directly (e.g., `require 'opentelemetry/resource/detector/aws/ec2'`), they still have all their necessary dependencies.

## Issues
https://github.com/open-telemetry/opentelemetry-ruby-contrib/issues/1515

### DIFF
--- a/resources/aws/lib/opentelemetry/resource/detector/aws/lambda.rb
+++ b/resources/aws/lib/opentelemetry/resource/detector/aws/lambda.rb
@@ -4,6 +4,8 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 
+require 'opentelemetry/semantic_conventions/resource'
+
 module OpenTelemetry
   module Resource
     module Detector
--- a/resources/aws/lib/opentelemetry-resource-detector-aws.rb
+++ b/resources/aws/lib/opentelemetry-resource-detector-aws.rb
@@ -4,4 +4,5 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 
+require 'opentelemetry/semantic_conventions/resource'
 require_relative 'opentelemetry/resource/detector'
--- a/resources/aws/lib/opentelemetry/resource/detector/aws/ec2.rb
+++ b/resources/aws/lib/opentelemetry/resource/detector/aws/ec2.rb
@@ -7,6 +7,7 @@
 require 'net/http'
 require 'json'
 require 'opentelemetry/common'
+require 'opentelemetry/semantic_conventions/resource'
 
 module OpenTelemetry
   module Resource
--- a/resources/aws/lib/opentelemetry/resource/detector/aws/ecs.rb
+++ b/resources/aws/lib/opentelemetry/resource/detector/aws/ecs.rb
@@ -8,6 +8,7 @@ require 'net/http'
 require 'json'
 require 'socket'
 require 'opentelemetry/common'
+require 'opentelemetry/semantic_conventions/resource'
 
 module OpenTelemetry
   module Resource


resources/aws/lib/opentelemetry/resource/detector/aws/lambda.rb
```diff
@@ -4,6 +4,8 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 
+require 'opentelemetry/semantic_conventions/resource'
```
arielvalentin, line 7: Is there any disadvantage to requiring this dependency once at the top level?

`resources/aws/lib/opentelemetry-resource-detector-aws.rb`

xuan-cao-swi: Yes, it should be fine to include in the top level.

I was testing through `require 'opentelemetry/resource/detector/aws/ec2'`, and I got the issue.

Actually, if I do `require 'opentelemetry-resource-detector-aws'`, then it doesn't need require semantic_conventions/resource (e.g. no uninitialized constant error).

I guess I should require the detector through the gem name, not gem file.

arielvalentin: Yeah I think our instructions on `opentelemetry.io` are also a bit misleading. 

In most cases bundler will require the root of the gem and that should be enough to start the require process of all of the files down the tree so we should encourage requiring top levels files as much as possible.

yiyuan-he: Thanks Ariel and Xuan for the additional context and feedback. I'll push a commit to require the dependency at `resources/aws/lib/opentelemetry-resource-detector-aws.rb` instead.

However, i'm curious if we should move forward with these changes and implicitly support this unrecommended dependency import method.

arielvalentin: The top level directory requires the `opentelemetry/sdk` which requires the transitive dependencies: 

https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/main/resources/aws/lib/opentelemetry/resource/detector.rb#L7

https://github.com/open-telemetry/opentelemetry-ruby/blob/main/sdk/lib/opentelemetry/sdk.rb#L10

Is your preference that all of these dependencies be explicitly required? 

Is that because each sub detector may be used in isolation without loading the other detectors?

If so, then for consistency I think you all should treat each as their own nested library and explicitly require all dependencies.

yiyuan-he: Thanks Ariel - that's helpful context. Yes, the intention is to allow sub-detectors to be used in isolation. I'll update the PR to explicitly require all dependencies for consistency across detectors. This should make it clearer for users which dependencies are needed when using specific detectors independently.

yiyuan-he: Updated the PR to add the semantic conventions require to the top-level file while maintaining the individual requires in each detector file.

This should ensure:
  1. When the gem is required normally via `require 'opentelemetry-resource-detector-aws'`, all dependencies are loaded properly.
  2. When individual detectors are required directly (e.g., `require'opentelemetry/resource/detector/aws/ec2'`), they still have all their necessary dependencies.

Added some notes to the PR description as well for documentation purposes.

kaylareopelle (approved):
Thanks, @yiyuan-he!

arielvalentin (approved):

Issue #1470, opened 2025-04-03
AWS SDK instrumentation does not work for V3


srprash:
### Description of the bug
When trying to instrument AWS SDK for Ruby V3 clients, there are no spans produced. However, I do see client spans produced when I downgrade to V2 of the AWS SDK for Ruby.

### Share details about your runtime

Operating system details: MacOS
RUBY_ENGINE: "ruby"
RUBY_VERSION: "3.4.2"
RUBY_DESCRIPTION: "ruby 3.4.2 (2025-02-15 revision d2930f8e7a) +PRISM [arm64-darwin24]"

### Share a simplified reproduction if possible

*I am using the [example present in the aws sdk instrumentation](https://github.com/open-telemetry/opentelemetry-ruby-contrib/tree/main/instrumentation/aws_sdk/example) to demonstrate the issue.*

```
$ git clone https://github.com/open-telemetry/opentelemetry-ruby-contrib.git
$ cd opentelemetry-ruby-contrib/instrumentation/aws_sdk/example

$ bundle install
$ ruby trace_demonstration.rb
```

#### Output
```
/opt/homebrew/lib/ruby/gems/3.4.0/gems/aws-sdk-datapipeline-1.65.0/lib/aws-sdk-datapipeline/types.rb:283: warning: redefining 'object_id' may cause serious problems
I, [2025-04-03T14:10:20.401075 #39887]  INFO -- : Instrumentation: OpenTelemetry::Instrumentation::AwsSdk was successfully installed with the following options {inject_messaging_context: false, suppress_internal_instrumentation: false}
/opt/homebrew/lib/ruby/gems/3.4.0/gems/aws-sdk-core-3.222.1/lib/seahorse/client/plugins/raise_response_errors.rb:17:in 'Seahorse::Client::Plugins::RaiseResponseErrors::Handler#call': Invalid parameter: TopicArn or TargetArn Reason: no value for required parameter (Aws::SNS::Errors::InvalidParameter)
	from /opt/homebrew/lib/ruby/gems/3.4.0/gems/aws-sdk-core-3.222.1/lib/aws-sdk-core/plugins/checksum_algorithm.rb:169:in 'Aws::Plugins::ChecksumAlgorithm::OptionHandler#call'
[...]

PR #1522, opened 2025-05-04
docs: Use AWS SDK v3 in example

j15e:
This PR updates the AWS SDK documentation example to use the SDK v3 as discussed in #1470

It pretty much now implement the suggested example in the README itself:

https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/51f6328426db28829f6ca1cddb916d5d94cddb6b/instrumentation/aws_sdk/README.md?plain=1#L47-L58

And I added the version restriction in the example Gemfile to avoid any confusion on the SDK version used. 

I also added `rexml`, because starting from Ruby 3.0 it isn't shipped with Ruby core, but [aws-sdk-core needs an XML parser](https://github.com/aws/aws-sdk-ruby/blob/9bca9ab73addfd600e833b5c2b6ef57df517952f/gems/aws-sdk-core/lib/aws-sdk-core/xml/parser.rb#L70-L73). It works on the AWS SDK CI without specifying rexml, because [webmock](https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/51f6328426db28829f6ca1cddb916d5d94cddb6b/instrumentation/aws_sdk/Gemfile#L20) requires rexml, but we don't have this gem in the example.

If you run for example `AWS_ACCESS_KEY_ID=123 AWS_SECRET_ACCESS_KEY=qwer AWS_REGION=us-east-1 ruby trace_demonstration.rb` you will now get the example trace working :

image

### DIFF
--- a/instrumentation/aws_sdk/example/Gemfile
+++ b/instrumentation/aws_sdk/example/Gemfile
@@ -2,10 +2,10 @@
 
 source 'https://rubygems.org'
 
-gem 'aws-sdk-core', '~> 3'
-gem 'aws-sdk-sns', '~> 1'
+gem 'aws-sdk-core', '~> 3.203.0'
+gem 'aws-sdk-sns', '~> 1.83.0'
 gem 'opentelemetry-api'
-gem 'opentelemetry-instrumentation-aws_sdk'
+gem 'opentelemetry-instrumentation-aws_sdk', '~> 0.7.0'
 gem 'opentelemetry-sdk'
 # AWS SDK core needs an XML library
 gem 'rexml'
--- a/instrumentation/aws_sdk/example/Gemfile
+++ b/instrumentation/aws_sdk/example/Gemfile
@@ -2,7 +2,10 @@
 
 source 'https://rubygems.org'
 
-gem 'aws-sdk'
+gem 'aws-sdk-core', '~> 3'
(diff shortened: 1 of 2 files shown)


kaylareopelle:
cc @jterapin @alextwoods

instrumentation/aws_sdk/example/Gemfile
```diff
@@ -2,7 +2,10 @@
 
 source 'https://rubygems.org'
 
-gem 'aws-sdk'
+gem 'aws-sdk-core', '~> 3'
+gem 'aws-sdk-sns', '~> 1'
 gem 'opentelemetry-api'
 gem 'opentelemetry-instrumentation-aws_sdk'
```
jterapin, line 8: We will have to set min gem versions to when the V3 Observability support/refactoring was released:

```suggestion
gem 'aws-sdk-core', '~> 3.203.0'
gem 'aws-sdk-sns', '~> 1.83.0'
gem 'opentelemetry-api'
gem 'opentelemetry-instrumentation-aws_sdk', '~> 0.7.0'
```

jterapin (approved):
This is great! Thank you for the contribution! Much appreciated.

kaylareopelle (approved):

PR #1533, opened 2025-05-13
feat!: Suppress internal spans by default

jterapin:
#1502

**Description**: The `aws_sdk` instrumentation now suppresses internal spans by default. This ensures that the additional low-level instrumentation does not generate additional costs for customers unknowingly. 

The following work has been done to accommodate:
* Deprecated `:suppress_internal_instrumentation` config 
* Introduced a new config, `:enable_internal_instrumentation` which is set false by default

To enable internal spans, users will have explicitly configure as follows:
```ruby
  c.use 'OpenTelemetry::Instrumentation::AwsSdk', { enable_internal_instrumentation: true }
end
```


ericmustin:
Was this or the related Issue discussed at a SIG? seems like i'm missing some context, at the very least I have some ambivalence about multiple config options that do the same thing.

jterapin:
> Was this or the related Issue discussed at a SIG? seems like i'm missing some context, at the very least I have some ambivalence about multiple config options that do the same thing.

I have not attended SIG so most likely not. 

This PR drops the `:suppress_internal_instrumentation` config entirely in a favor of `:enable_internal_instrumentation` since the config name is quite explanatory of its default behavior.  That being said - I'm fine with just reusing the old config but flipping the config default.

instrumentation/aws_sdk/README.md
```diff
@@ -36,8 +36,10 @@ end
 This instrumentation offers the following configuration options: 
 * `:inject_messaging_context` (default: `false`): When set to `true`, adds context key/value 
  to Message Attributes for SQS/SNS messages.
-* `suppress_internal_instrumentation` (default: `false`): When set to `true`, any spans with 
- span kind of `internal` are suppressed from traces.
+* `:enable_internal_instrumentation` (default: `false`): When set to `true`, any spans with 
+ span kind of `internal` are traced.
+* `:suppress_internal_instrumentation`: **Deprecated**. This configuration has been 
+ deprecated in a favor of `:enable_internal_instrumentation`
```
arielvalentin, line 42: It looks like you are removing this option altogether resulting in a breaking change. 

If you want to treat it as deprecated, as opposed to removed, I would recommend leaving the option present and giving precedence to the new option.

kaylareopelle: Hi @jterapin, thanks for submitting this PR! I'm with @arielvalentin -- I think we should keep the old config around for a while as a documented deprecated option. Perhaps we could remove it entirely as part of a 1.x release? 

What do you think?

jterapin: Hello there! Sorry for the late response. Totally agreed! 

I'll try to make the suggested change sometime before this week ends.

jterapin: Updated pr, keeping deprecated config and added warn if it is set true (since it indicates to me that's its actively being used).

instrumentation/aws_sdk/lib/opentelemetry/instrumentation/aws_sdk/handler.rb
```diff
@@ -23,7 +23,7 @@ def call(context)
           ) do |span|
             MessagingHelper.inject_context_if_supported(context, client_method, service_id)
 
-            if HandlerHelper.instrumentation_config[:suppress_internal_instrumentation]
+            if HandlerHelper.skip_internal_instrumentation?
```
arielvalentin, line 26: @kaylareopelle I realize this is a common pattern we have in the code base already but I would recommend we look into refactoring cases like this where we create separate mixins for "untraced" instead of evaluating this logic with every request.

instrumentation/aws_sdk/lib/opentelemetry/instrumentation/aws_sdk/instrumentation.rb
```diff
@@ -65,6 +69,15 @@ def gem_version
 
         private
 
+        def resolve_config(config)
```
arielvalentin, line 72: How are we providing backward compatability for the deprecated option?

Is it not supposed to work at all?

jterapin: Well - this configuration handling was more trickier than i thought.....

The main challenge is that configuration hashes always contain default values, which makes it difficult to identify which settings were explicitly set by users. 

I tried to solve this by removing the `:suppress_internal_instrumentation` option completely to detect user config, but discovered that unrecognized settings aren't passed through to the install block.

Then, I realized that users would only want to touch this setting to suppress internal traces- which happens to be the default behavior anyway. Any thoughts on mitigating this?

arielvalentin: Then I suppose it can be a breaking change there. We would need to call out in the release notes that the option is no longer available.

arielvalentin (approved):

Issue #1434, opened 2025-03-05
Support for AWS X-Ray Remote Sampling


jj22ee:
### What problem do you want to solve?

OTel Ruby current does not support [centralized/remote sampling from AWS X-Ray](https://docs.aws.amazon.com/xray/latest/devguide/xray-concepts.html#xray-concepts-sampling).

### Describe the solution you'd like

OTel Ruby to support X-Ray Remote Sampling, similarly to the [.NET](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/tree/main/src/OpenTelemetry.Sampler.AWS) and [Java](https://github.com/open-telemetry/opentelemetry-java-contrib/tree/main/aws-xray/src/main/java/io/opentelemetry/contrib/awsxray) support for X-Ray Sampling. Also see "xray" sampler in the [specification here](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/configuration/sdk-environment-variables.md).

### Additional Context

I can look into implementing this X-Ray Sampler for Ruby.

kaylareopelle:
Hi @jj22ee, thanks for reaching out! We'd love your help implementing this. Would you like me to assign this issue to you?

cc @jterapin @alextwoods

jj22ee:
You can assign the issue to me, thanks!

PR #1536, opened 2025-05-15
feat: AWS X-Ray Remote Sampler Part 3 - Add Rate Limiter and Sampling Targets Poller Logic

jj22ee:
## Description
- Followup PR of https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/1498

This is the 3rd and final PR to address https://github.com/open-telemetry/opentelemetry-ruby-contrib/issues/1434 in order for OTel Ruby to support X-Ray Remote Sampling.

Changes:
- Added logic to fetch sampling targets for each `sampling_rule_applier`.
  - The sampling targets are periodically fetched every 10 seconds by making the [GetSamplingTargets API call to X-Ray](https://docs.aws.amazon.com/xray/latest/api/API_GetSamplingTargets.html).
  - The targets determine the reservoir quota and the rate at which a `sampling_rule_applier` will sample the requests.
  - Each rule applier keeps and updates a sampling statistics document which is required in `GetSamplingTargets` call to determine the next target
- Added a `rate_limiting_sampler` (applied before the fixed rate sampler) to be used in each rule applier.
  - Together these samplers determine how many requests to sample every second and what percentage of additional requests to sample in that second.
  - The `fallback_sampler` is updated to be a combination of above samplers to sample `1 req/sec` and `5%` of additional requests in that second.
- Added `example/xray_sampling_on_rails_demonstration.ru` that runs an example rails application that uses the OpenTelemetry X-Ray Sampler, with example instructions in the comments at the bottom of the file.

## Testing

- Unit Tests
- The current implementation was tested against [X-Ray's Centralized Sampling Integration tests](https://github.com/aws-observability/aws-otel-community/tree/master/centralized-sampling-tests#start-integration-tests).


jj22ee:
Thanks @kaylareopelle! This new gem will be ready for releasing after this PR.

jj22ee:
Also requesting @lukeina2z for review.

sampler/xray/example/xray_sampling_on_rails_demonstration.ru
```diff
@@ -0,0 +1,101 @@
...
+    render inline: 'Successfully called "/" endpoint'
+  end
+
+  def test
+    render inline: 'Successfully called "/test" endpoint'
+  end
+end
+
+ENV['OTEL_TRACES_EXPORTER'] = 'console'
+ENV['OTEL_SERVICE_NAME'] = 'xray-sampler-on-rails-service'
```
lukeina2z: enable override.
```suggestion
ENV['OTEL_TRACES_EXPORTER'] ||= 'console'
ENV['OTEL_SERVICE_NAME'] ||= 'xray-sampler-on-rails-service'
```

sampler/xray/example/xray_sampling_on_rails_demonstration.ru
```diff
@@ -0,0 +1,101 @@
...
+  "service.name"=>"xray-sampler-on-rails-service"
+}))
+
+App.initialize!
+
+run App
+
+#### Running and using the Sample App
+# To run this example run the `rackup` command with this file
+# Example: rackup trace_request_demonstration.ru
```
lukeina2z: "trace_request_demonstration", typo?

sampler/xray/lib/opentelemetry/sampler/xray/rate_limiter.rb
```diff
@@ -0,0 +1,48 @@
...
+        def initialize(quota, max_balance_in_seconds = 1)
+          @max_balance_millis = max_balance_in_seconds * 1000.0
+          @quota = quota
+          @wallet_floor_millis = Time.now.to_f * 1000
+          # current "balance" would be `ceiling - floor`
+          @lock = Mutex.new
+        end
+
+        def take(cost = 1)
+          return false if @quota.zero?
```
lukeina2z: <= 0?

lukeina2z (approved):

sampler/xray/lib/opentelemetry/sampler/xray/rate_limiter.rb
jj22ee: Done. This is usually set by X-Ray and cannot be set to <=0, but still safer none-the-less.

sampler/xray/example/xray_sampling_on_rails_demonstration.ru
jj22ee: Yep, should be `xray_sampling_on_rails_demonstration.ru`.

jj22ee:
@kaylareopelle Final XRay Sampler PR is ready for your review!

sampler/xray/example/xray_sampling_on_rails_demonstration.ru
```diff
@@ -0,0 +1,102 @@
...
+  def test
+    render inline: 'Successfully called "/test" endpoint'
+  end
+end
+
+ENV['OTEL_TRACES_EXPORTER'] ||= 'console'
+ENV['OTEL_SERVICE_NAME'] ||= 'xray-sampler-on-rails-service'
+
+OpenTelemetry::SDK.configure do |c|
+  c.use_all({ 'OpenTelemetry::Instrumentation::ActiveRecord' => { enabled: false } })
```
kaylareopelle: Does ActiveRecord instrumentation have to be off for the sampler to work?

kaylareopelle (commented):
Hi @jj22ee, I took a look last week, but got pulled onto other things before I could finish. I have one question for now. I'll try to finish up today.

sampler/xray/lib/opentelemetry/sampler/xray/aws_xray_remote_sampler.rb
```diff
@@ -4,14 +4,14 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 
-require 'net/http'
 require 'json'
+require 'net/http'
```
kaylareopelle: I don't see Net::HTTP getting used in this file. Why is it required?

kaylareopelle (commented):
Finished taking a look. I think this is ready to go. I would like to understand what's going on with ActiveRecord in the example before approving.

sampler/xray/lib/opentelemetry/sampler/xray/aws_xray_remote_sampler.rb
jj22ee: Nice catch, not required here. I had this here before I moved all the usage of `Net::HTTP` into the `aws_xray_sampling_client.rb` file, so I'll remove this.

sampler/xray/example/xray_sampling_on_rails_demonstration.ru
jj22ee: Oops. The Sampler should work with/without this instrumentation. I pretty much blindly copied this config from the [rails instrumentation README](https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/main/instrumentation/rails/README.md#configuration-options) when I was using the rails example to create the sampler example. I changed it back to just use `c.use_all`, and confirmed the sampling still works.

kaylareopelle: Thank you!

kaylareopelle (approved):

PR #1537, opened 2025-05-15
test: Add less restrictive gem spec for mongo gem

donoghuc:
It looks like latest features (including jruby 10 enablement) will require the latest minor release of the mongo gem.

Specifically:

The latest mongo gem relaxes the range for the ruby-bson gem. This would allow the uptake of https://github.com/mongodb/bson-ruby/pull/353 if it is accepted.

### DIFF
--- a/instrumentation/mongo/Appraisals
+++ b/instrumentation/mongo/Appraisals
@@ -4,6 +4,6 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 
-appraise 'mongo-2.13' do
-  gem 'mongo', '~> 2.13.0'
+appraise 'mongo-2' do
+  gem 'mongo', '~> 2.13'
 end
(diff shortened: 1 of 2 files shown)


instrumentation/mongo/Appraisals
```diff
@@ -5,5 +5,5 @@
 # SPDX-License-Identifier: Apache-2.0
 
 appraise 'mongo-2.13' do
```
kaylareopelle, line 7: Some of the mongo tests are failing on 3.4 with this update.

Seeing that the latest version is 2.21.0, what do you think about adding a new appraisal for the most recent version? 

Many instrumentation test suites have an appraisal without a version specified to make sure it's compatible with the latest release.

donoghuc: Followed the pattern in other appraisal files (just doing major version when constrained to minor version). Pushed a commit to fix the tests, still digging in to impacts of those changes to generate a good changelog.

kaylareopelle: Thanks!

arielvalentin (approved):

PR #1547, opened 2025-05-22
feat: http semconv opt in files

hannahramadan:

While this approach introduces multiple new and similar files and tests (potentially making debugging and maintenance more difficult), the tradeoff ensures we're only patching the HTTP class once and makes the eventual cleanup of the old conventions mostly a matter of deleting files and small refactors. [Another approach](https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/1166) (also taken by Python) introduces a `StabilityMode` class that checks calls to determine which attributes to send, however this approach adds complexity of calling a method every time an attribute is added. 

Hopefully with the temporary nature of this work, this solution can be seen as a way to start the transition. The approach can also be easily duplicated to other HTTP libraries and support [the similar DB migration](https://opentelemetry.io/docs/specs/semconv/non-normative/db-migration/).


kaylareopelle (approved):
I love the simplicity of this approach, especially since we're only required to keep it for six months. Migrating to the stable semantic conventions also makes metrics integration much easier. Thank you!

xuan-cao-swi (approved):

kaylareopelle:
We're planning to merge this on Monday, June 16 if there are no blocking comments.

ericmustin (approved):

PR #1548, opened 2025-05-23
fix: compact Dalli attributes

nickamorim:
## Description

Dalli supports connecting to Memcached over a [UNIX domain socket](https://github.com/petergoldstein/dalli/wiki/Server-Configuration#individual-server-configuration-strings) in which case the [`port` is nil](https://github.com/Shopify/dalli/blob/main/lib/dalli/protocol/server_config_parser.rb#L60-L65). This will raise `OpenTelemetry error: invalid span attribute value type NilClass for key 'net.peer.port' on span 'set' (OpenTelemetry::Error)` ([source](https://github.com/Shopify/opentelemetry-ruby-shopify/blob/main/opentelemetry-shopify/lib/opentelemetry/shopify.rb#L291-L301)).

This PR attempts to address this by compacting the `attributes`. I'm curious if folks think the test is digging too deep into Dalli internals. I just mainly wanted to avoid spinning up a separate Memcached server that is listening over a UNIX domain socket to test this. I'm all ears for other approaches too.

### DIFF
--- a/instrumentation/dalli/lib/opentelemetry/instrumentation/dalli/patches/server.rb
+++ b/instrumentation/dalli/lib/opentelemetry/instrumentation/dalli/patches/server.rb
@@ -15,9 +15,10 @@ module OpenTelemetry
             attributes = {
               'db.system' => 'memcached',
               'db.operation' => operation,
-              'net.peer.name' => hostname,
-              'net.peer.port' => port
+              'net.peer.name' => hostname
             }
+            attributes['net.peer.port'] = port if port
+
             if config[:db_statement] == :include
               attributes['db.statement'] = Utils.format_command(operation, args)
             elsif config[:db_statement] == :obfuscate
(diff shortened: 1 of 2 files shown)


oosidat (approved):

instrumentation/dalli/lib/opentelemetry/instrumentation/dalli/patches/server.rb
```diff
@@ -17,7 +17,7 @@ def request(op, *args) # rubocop:disable Naming/MethodParameterName
               'db.operation' => operation,
               'net.peer.name' => hostname,
               'net.peer.port' => port
-            }
+            }.compact
```
robertlaurin, line 20: `.compact!` means one less hash allocation, it can return `nil` so don't call it on the closing `}`. 
```suggestion
            }
            attributes.compact!
```

robertlaurin (approved):
Will merge on CI completion

instrumentation/dalli/lib/opentelemetry/instrumentation/dalli/patches/server.rb
```diff
@@ -25,6 +25,7 @@ def request(op, *args) # rubocop:disable Naming/MethodParameterName
             end
 
             attributes['peer.service'] = config[:peer_service] if config[:peer_service]
+            attributes.compact!
```
arielvalentin, line 28: We typically want to avoid using `compact!` whenever possilbe since it is a bit slower than conditional assigments. 

You can see a benchmark here someone shared:

https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/693#discussion_r1360669417

arielvalentin (approved):

PR #1555, opened 2025-06-03
fix: rack span class naming

robertlaurin:
### What

We don't want to override the span name when an exceptions app is configured, so we skip the span naming step `action_dispatch.exception` is set in the request env.

### Why

Consider a simple rails app with the an issues controller and create action. The app is configured with an exceptions app we'll just call it the exceptions controller with a show action.

The app receives a request to `IssuesController#create`, it raises, and it subsequently handled by `ExceptionsController#show`.

When preserving the original name you can do simple queries like:

`span_name = IssuesController#create where status = 500` or `span_name = IssuesController#create group by status`

With the exceptions app renaming those queries look something like the following:

`span_name in (IssuesController#create, ExceptionsController#show) and http.route = whatever_the_route_was where status = 500`

### Added context

I don't think this change was intentional but rather incidental when we changed the default from class naming to semantic convention naming.

### DIFF
--- a/instrumentation/action_pack/lib/opentelemetry/instrumentation/action_pack/handlers/action_controller.rb
+++ b/instrumentation/action_pack/lib/opentelemetry/instrumentation/action_pack/handlers/action_controller.rb
@@ -23,9 +23,34 @@ module OpenTelemetry
           # @param payload [Hash] the payload passed as a method argument
           # @return [Hash] the payload passed as a method argument
           def start(_name, _id, payload)
-            span_name, attributes = to_span_name_and_attributes(payload)
+            request = payload[:request]
+            # It seems that there are cases in Rails functional tests where it bypasses the routing system and the `action_dispatch.route_uri_pattern` header not being set.
+            # Our Test suite executes the routing system so we are unable to recreate this error case.
+            # https://github.com/rails/rails/blob/747f85f200e7bb2c1a31b4e26e5a5655e2dc0cdc/actionpack/lib/action_dispatch/http/request.rb#L160
+            http_route = request.route_uri_pattern&.chomp('(.:format)') if request.respond_to?(:route_uri_pattern)
+
+            attributes = {
+              OpenTelemetry::SemanticConventions::Trace::CODE_NAMESPACE => String(payload[:controller]),
+              OpenTelemetry::SemanticConventions::Trace::CODE_FUNCTION => String(payload[:action])
+            }
+            attributes[OpenTelemetry::SemanticConventions::Trace::HTTP_ROUTE] = http_route if http_route
+            attributes[OpenTelemetry::SemanticConventions::Trace::HTTP_TARGET] = request.filtered_path if request.filtered_path != request.fullpath
 
             span = OpenTelemetry::Instrumentation::Rack.current_span
+            span_name = if @span_naming == :semconv
+                          if http_route
+                            "#{request.method} #{http_route}"
+                          else
+                            "#{request.method} /#{payload.dig(:params, :controller)}/#{payload.dig(:params, :action)}"
+                          end
+                        # If there is an exception we want to keep the original span name
+                        # so it is easier to see where the request was routed to.
+                        elsif request.env['action_dispatch.exception']
+                          span.name
+                        else
+                          "#{payload[:controller]}##{payload[:action]}"
+                        end
+
             span.name = span_name
             span.add_attributes(attributes)
           rescue StandardError => e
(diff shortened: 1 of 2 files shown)


instrumentation/action_pack/test/opentelemetry/instrumentation/action_pack/handlers/action_controller_test.rb
```diff
@@ -215,11 +215,12 @@
 
   describe 'when the application has exceptions_app configured' do
     let(:rails_app) { AppConfig.initialize_app(use_exceptions_app: true) }
+    let(:config) { { span_naming: :class } }
 
     it 'does not overwrite the span name from the controller that raised' do
       get 'internal_server_error'
 
-      _(span.name).must_match(/^GET/)
```
robertlaurin, line 222: 😢 https://github.com/open-telemetry/opentelemetry-ruby-contrib/commit/acac4fd6b835b4ad8023e3bc6d8b454bc922c726#diff-2260f0004922c41669690c903aad041cf1df67590243c52eae659d67e7ce5574R208-R214

instrumentation/action_pack/lib/opentelemetry/instrumentation/action_pack/handlers/action_controller.rb
```diff
@@ -44,35 +69,6 @@ def finish(_name, _id, payload)
...
-            attributes[OpenTelemetry::SemanticConventions::Trace::HTTP_ROUTE] = http_route if http_route
-            attributes[OpenTelemetry::SemanticConventions::Trace::HTTP_TARGET] = request.filtered_path if request.filtered_path != request.fullpath
-
-            if @span_naming == :semconv
-              return ["#{request.method} #{http_route}", attributes] if http_route
-
-              return ["#{request.method} /#{payload.dig(:params, :controller)}/#{payload.dig(:params, :action)}", attributes]
-            end
-
-            ["#{payload[:controller]}##{payload[:action]}", attributes]
```
robertlaurin, line 74: Inlined this method as it's not used elsewhere, and by doing so we can skip the array allocation used to return the result.

instrumentation/action_pack/lib/opentelemetry/instrumentation/action_pack/handlers/action_controller.rb
```diff
@@ -23,9 +23,34 @@ def initialize(config)
...
-            span_name, attributes = to_span_name_and_attributes(payload)
+            request = payload[:request]
+            # It seems that there are cases in Rails functional tests where it bypasses the routing system and the `action_dispatch.route_uri_pattern` header not being set.
+            # Our Test suite executes the routing system so we are unable to recreate this error case.
+            # https://github.com/rails/rails/blob/747f85f200e7bb2c1a31b4e26e5a5655e2dc0cdc/actionpack/lib/action_dispatch/http/request.rb#L160
+            http_route = request.route_uri_pattern&.chomp('(.:format)') if request.respond_to?(:route_uri_pattern)
+
+            attributes = {
+              OpenTelemetry::SemanticConventions::Trace::CODE_NAMESPACE => String(payload[:controller]),
+              OpenTelemetry::SemanticConventions::Trace::CODE_FUNCTION => String(payload[:action])
```
robertlaurin, line 34: The old behaviour would allow this to be overwritten with the exceptions app class and action, so what is currently here keeps that. Do we want to revisit having this also preserve the original controller/action?

I'm a bit split, because it gives a clear bread crumb that an exceptions app handled it, but that's not really obvious...

ericmustin (approved):

instrumentation/action_pack/lib/opentelemetry/instrumentation/action_pack/handlers/action_controller.rb
arielvalentin: I don't have a strong opinion about how class naming should work. 

If someone configures an exception app they are intentionally trying to perform an internal redirect to another controller. 

We don't have a way a to model what the user intended on doing and what the server actually executed if it is configured to handle requests to the same routes differently.

For me that is where the semconv naming is most helpful because as long as the route is resolved, then we could add additional attributes that describe the controller action or middleware that eventually handled the request.

instrumentation/action_pack/test/opentelemetry/instrumentation/action_pack/handlers/action_controller_test.rb
arielvalentin: Are we losing test coverage here for the semconv naming case?

fbogsany (approved):

instrumentation/action_pack/test/opentelemetry/instrumentation/action_pack/handlers/action_controller_test.rb
robertlaurin: Technically no because it wasn't actually covering what that test was intended for, this only really applies to the class naming config option. The path remains the path regardless of it ending up with an exceptions controller or not.

instrumentation/action_pack/lib/opentelemetry/instrumentation/action_pack/handlers/action_controller.rb
robertlaurin: Yeah I think I'll leave it as is. 

>For me that is where the semconv naming is most helpful because as long as the route is resolved, then we could add additional attributes that describe the controller action or middleware that eventually handled the request.

I agree, but there's just such a strong convention of thinking about rails apps in the form of controllers and actions, that we want to continue to support that ease of association. 

If I really wanted to make an argument against my position, the rack and controller span should be separate spans as they are quite different.

PR #1561, opened 2025-06-16
feat: Ethon HTTP semantic convention stability migration

hannahramadan:


instrumentation/ethon/lib/opentelemetry/instrumentation/ethon/patches/stable/easy.rb
```diff
@@ -0,0 +1,130 @@
...
+              }
+
+              uri = _otel_cleanse_uri(url)
+              if uri
+                instrumentation_attrs['url.full'] = uri.to_s
+                instrumentation_attrs['server.address'] = uri.host if uri.host
+              end
+
+              config = Ethon::Instrumentation.instance.config
+              instrumentation_attrs['peer.service'] = config[:peer_service] if config[:peer_service]
```
kaylareopelle, line 105: I'm not quite sure what the right call is for this attribute. I don't see it in the modern semantic conventions. The old description was: 
```      
# The [`service.name`](../../resource/semantic_conventions/README.md#service) of the remote service. SHOULD be equal to the actual `service.name` resource attribute of the remote service if any
```

But since we have it as a config option, maybe it should stick around?

instrumentation/ethon/lib/opentelemetry/instrumentation/ethon/patches/stable/easy.rb
```diff
@@ -0,0 +1,130 @@
...
+      module Patches
+        # Module using stable HTTP semantic conventions
+        module Stable
+          # Ethon::Easy patch for instrumentation
+          module Easy
+            ACTION_NAMES_TO_HTTP_METHODS = Hash.new do |h, k|
+              # #to_s is required because user input could be symbol or string
+              h[k] = k.to_s.upcase
+            end
+            HTTP_METHODS_TO_SPAN_NAMES = Hash.new { |h, k| h[k] = "HTTP #{k}" }
```
kaylareopelle: I know we're primarily focused on attributes, but span names also fall into semantic conventions. I think the `HTTP` may need to be removed from the name. https://opentelemetry.io/docs/specs/semconv/non-normative/http-migration/#http-client-span-name

I'm not sure if we need to emit multiple spans in the `dup` category... one with each name? 🤔

instrumentation/ethon/lib/opentelemetry/instrumentation/ethon/patches/stable/easy.rb
```diff
@@ -0,0 +1,130 @@
...
+            def reset
+              super
+            ensure
+              @otel_span = nil
+              @otel_method = nil
+              @otel_original_headers = nil
+            end
+
+            def otel_before_request
+              method = 'N/A' # Could be GET or not HTTP at all
```
kaylareopelle: I wonder if this should be `_OTHER`? https://github.com/open-telemetry/semantic-conventions/blob/main/docs/http/http-spans.md?plain=1#L177

But if it's getting used for the span name, maybe it just needs to be `HTTP`? 
https://opentelemetry.io/docs/specs/semconv/non-normative/http-migration/#http-client-span-name

instrumentation/ethon/lib/opentelemetry/instrumentation/ethon/patches/stable/easy.rb
hannahramadan: I had kept `peer.service` around in all instrumentation attributes since I didn't see it change and its a config. This is a good question though. I will see if this was addressed anywhere, especially for the config option piece.

instrumentation/ethon/lib/opentelemetry/instrumentation/ethon/patches/stable/easy.rb
hannahramadan: Yes - great call out! Now, when the method isn't detected, the span name is `HTTP` and the method attribute is `_OTHER`. For users who opt in to receive both attributes, the will still get the old semantic convention 'N/A' reported


https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/1561/commits/1dd68106a6da9196b538769b8ab08d2e56d7cf8a

instrumentation/ethon/lib/opentelemetry/instrumentation/ethon/patches/stable/easy.rb
hannahramadan: `HTTP` was removed here, but we will not be emitting multiple spans and just use the newer span name if `dup` is enabled. This is following the Python decision to name the single span based on the newer conventions. Here is an [example instrumentation PR](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/2631) from Python - where we can see the span name in the PR description ◡̈

instrumentation/ethon/lib/opentelemetry/instrumentation/ethon/patches/stable/easy.rb
hannahramadan: `peer.service` appears to still be in development. I feel we should leave this as-is for now, unless there are other thoughts.

https://opentelemetry.io/docs/specs/semconv/registry/attributes/peer/

kaylareopelle (approved):
Thanks, @hannahramadan!

ericmustin (approved):

instrumentation/ethon/lib/opentelemetry/instrumentation/ethon/patches/stable/easy.rb
kaylareopelle: Sounds good. Thanks for looking into it!

PR #1562, opened 2025-06-16
fix: action_pack always assuming sdk spans

robertlaurin:
`span.name` is not a method on api/non recording spans. This small refactor changes the logic around renaming to stop relying on the presence of `span.name` for when the naming convention is `class_name`. 

Adds a small regression test to prevent this mistake from being re-introduced.

### DIFF
--- a/instrumentation/action_pack/lib/opentelemetry/instrumentation/action_pack/handlers/action_controller.rb
+++ b/instrumentation/action_pack/lib/opentelemetry/instrumentation/action_pack/handlers/action_controller.rb
@@ -23,6 +23,9 @@ module OpenTelemetry
           # @param payload [Hash] the payload passed as a method argument
           # @return [Hash] the payload passed as a method argument
           def start(_name, _id, payload)
+            span = OpenTelemetry::Instrumentation::Rack.current_span
+            return unless span.recording?
+
             request = payload[:request]
             # It seems that there are cases in Rails functional tests where it bypasses the routing system and the `action_dispatch.route_uri_pattern` header not being set.
             # Our Test suite executes the routing system so we are unable to recreate this error case.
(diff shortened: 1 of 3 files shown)


instrumentation/action_pack/lib/opentelemetry/instrumentation/action_pack/handlers/action_controller.rb
```diff
@@ -37,21 +37,18 @@ def start(_name, _id, payload)
...
+            if @span_naming == :semconv
+              span.name = if http_route
                             "#{request.method} #{http_route}"
                           else
                             "#{request.method} /#{payload.dig(:params, :controller)}/#{payload.dig(:params, :action)}"
                           end
-                        # If there is an exception we want to keep the original span name
-                        # so it is easier to see where the request was routed to.
-                        elsif request.env['action_dispatch.exception']
-                          span.name
```
robertlaurin, line 49: API spans have a setter, but not a getter for `name`.

arielvalentin (approved):

instrumentation/action_pack/lib/opentelemetry/instrumentation/action_pack/handlers/action_controller.rb
```diff
@@ -37,21 +37,18 @@ def start(_name, _id, payload)
             attributes[OpenTelemetry::SemanticConventions::Trace::HTTP_TARGET] = request.filtered_path if request.filtered_path != request.fullpath
 
             span = OpenTelemetry::Instrumentation::Rack.current_span
-            span_name = if @span_naming == :semconv
-                          if http_route
+            if @span_naming == :semconv
```
arielvalentin, line 42: Should we skip this entire logic unless `span.recording?`?

fbogsany (approved):

instrumentation/action_pack/lib/opentelemetry/instrumentation/action_pack/handlers/action_controller.rb
robertlaurin: Yeah, that's a good improvement to make. Pushed up that change.

PR #1568, opened 2025-06-18
feat: rest-client HTTP semantic convention stability migration

hannahramadan:


instrumentation/restclient/README.md
```diff
@@ -48,3 +48,19 @@ Apache 2.0 license. See [LICENSE][license-github] for more information.
...
+
+To facilitate the migration to stable semantic conventions, you can use the `OTEL_SEMCONV_STABILITY_OPT_IN` environment variable. This variable allows you to opt-in to the new stable conventions, ensuring compatibility and future-proofing your instrumentation.
+
+When setting the value for `OTEL_SEMCONV_STABILITY_OPT_IN`, you can specify which conventions you wish to adopt:
+
+- `http` - Emits the stable HTTP and networking conventions and ceases emitting the old conventions previously emitted by the instrumentation.
+- `http/dup` - Emits both the old and stable HTTP and networking conventions, enabling a phased rollout of the stable semantic conventions.
+- Default behavior (in the absence of either value) is to continue emitting the old HTTP and networking conventions the instrumentation previously emitted.
+
+During the transition from old to stable conventions, Faraday instrumentation code comes in three patch versions: `dup`, `old`, and `stable`. These versions are identical except for the attributes they send. Any changes to Faraday instrumentation should consider all three patches.
```
kaylareopelle:
```suggestion
During the transition from old to stable conventions, RestClient instrumentation code comes in three patch versions: `dup`, `old`, and `stable`. These versions are identical except for the attributes they send. Any changes to RestClient instrumentation should consider all three patches.
```

kaylareopelle (changes requested):
One typo in the README, other than that this is good to go!

instrumentation/restclient/README.md
ericmustin: couple copy/paste typos nothing major

ericmustin (approved):
some small readme typos but nothing blocking

kaylareopelle (approved):

PR #1569, opened 2025-06-18
feat: excon HTTP semantic convention stability migration

hannahramadan:


instrumentation/excon/lib/opentelemetry/instrumentation/excon/middlewares/stable/tracer_middleware.rb
```diff
@@ -0,0 +1,112 @@
...
+
+            # Constant for the HTTP status range
+            HTTP_STATUS_SUCCESS_RANGE = (100..399)
+
+            def request_call(datum)
+              return @stack.request_call(datum) if untraced?(datum)
+
+              http_method = HTTP_METHODS_TO_UPPERCASE[datum[:method]]
+              attributes = {
+                OpenTelemetry::SemanticConventions::Trace::HTTP_HOST => datum[:host],
```
kaylareopelle: This one feels a little tricky. 

So I see the constant points to `http.host` in the SemConv gem. 
It has this description above it: 

> The value of the [HTTP host header](https://tools.ietf.org/html/rfc7230#section-5.4). An empty  Host header should also be reported, see note
>       @note When the header is present but empty the attribute SHOULD be set to the empty string. Note that this is a valid situation that is expected in certain cases, according the aforementioned [section of RFC 7230](https://tools.ietf.org/html/rfc7230#section-5.4). When the header is not set the attribute MUST NOT be set

This sounds like it could be `server.address` or `http.response.headers.host`. However, I also see `server.address` below that uses the `hostname` key, where this attr uses the `host` key, so they could be different values.

Since I don't see `http.host` in the stable conventions, we should probably take this out, but I don't know if it should be melded with `server.address` or get its own attribute name. 

What do you think? Am I missing something?

instrumentation/excon/lib/opentelemetry/instrumentation/excon/patches/dup/socket.rb
```diff
@@ -0,0 +1,65 @@
...
+                'server.port' => conn_port
+              }.merge!(OpenTelemetry::Common::HTTP::ClientContext.attributes)
+
+              if is_a?(::Excon::SSLSocket) && @data[:proxy]
+                span_name = 'CONNECT'
+                span_kind = :client
+              else
+                span_name = 'connect'
+                span_kind = :internal
+              end
```
kaylareopelle, line 40: What's the difference between the caps and lowercase here?

instrumentation/excon/test/opentelemetry/instrumentation/excon/stable/instrumentation_test.rb
```diff
@@ -0,0 +1,344 @@
...
+      Excon.get('http://example.com/success?hello=there')
+
+      _(exporter.finished_spans.size).must_equal 1
+      _(span.attributes['url.query']).must_equal 'hello=there'
+      assert_requested(
+        :get,
+        'http://example.com/success?hello=there',
+        headers: { 'Traceparent' => "00-#{span.hex_trace_id}-#{span.hex_span_id}-01" }
+      )
+    end
```
kaylareopelle, line 153: ❤️

instrumentation/excon/lib/opentelemetry/instrumentation/excon/middlewares/stable/tracer_middleware.rb
kaylareopelle: We discussed this -- we're going to remove `http.host` since it seems like this value is in `server.address` as well.

instrumentation/excon/lib/opentelemetry/instrumentation/excon/patches/dup/socket.rb
kaylareopelle: @ericmustin - Do you have any background on why the client span name is capitalized, but the internal span name is not?

instrumentation/excon/lib/opentelemetry/instrumentation/excon/middlewares/stable/tracer_middleware.rb
hannahramadan: Removed in https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/1569/commits/44397308b1e8429ce45c335a74660d5b3f0708f1

instrumentation/excon/lib/opentelemetry/instrumentation/excon/patches/dup/socket.rb
ericmustin: No clue I can dig around dd-trace-rb and see if It is merely a carryover from something unintentional there

kaylareopelle: That's okay, thank you. I did a little digging through git blame. Excon based it off Net::HTTP. This is the PR that introduced it to Net::HTTP: https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/129/ 

I think it's to help distinguish internal from client spans. I don't think we need to hold this PR back until we figure it out.

kaylareopelle (approved):

PR #1570, opened 2025-06-20
fix: update span name when semconv stability is enabled

hannahramadan:
[Semantic conventions 1.34.0 updates span names](https://opentelemetry.io/docs/specs/semconv/non-normative/http-migration/#http-server-span-name) to drop the leading `HTTP`, for example `HTTP GET` becomes just `GET`. To facilitate the transition from old to new HTTP semantic conventions, a new stability mode was introduced in #1547. However, the PR did not account for the span name update.

If `OTEL_SEMCONV_STABILITY_OPT_IN` is set to either `http` or `http/dup`, the span name should be inline with the new conventions. This PR makes this update and drops the leading `HTTP` from span names.


kaylareopelle (approved):
Thank you!

PR #1572, opened 2025-06-24
feat: NET::HTTP semantic convention stability migration

hannahramadan:

**Special note about NET::HTTP changes**: `http.target` is split into two separate attributes, `url.path` and `url.query`, under stable conventions. Unlike some other HTTP libraries, NET::HTTP reports these attributes as one path. Because of this, [some additional work](https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/1572/files#diff-d9e62ed8c73a2cab2378a1e488dd7261f33583c6e54e3ab0b15e9fc138d244baR38-R40) was done to to separate the path from the query in order to report them as separate attributes.


instrumentation/net_http/lib/opentelemetry/instrumentation/net/http/patches/dup/instrumentation.rb
```diff
@@ -0,0 +1,129 @@
...
+                  OpenTelemetry::SemanticConventions::Trace::NET_PEER_NAME => @address,
+                  OpenTelemetry::SemanticConventions::Trace::NET_PEER_PORT => @port,
+                  'http.request.method' => req.method,
+                  'url.scheme' => USE_SSL_TO_SCHEME[use_ssl?],
+                  'server.address' => @address,
+                  'server.port' => @port
+                }.merge!(OpenTelemetry::Common::HTTP::ClientContext.attributes)
+
+                path, query = split_path_and_query(req.path)
+                attributes['url.path'] = path unless attributes.key?('url.path')
```
hannahramadan: We check if `url.path` already exists to account for possible attributes previously set by a user. For example, `url.path' => 'REDACTED`

instrumentation/net_http/Appraisals
```diff
@@ -0,0 +1,21 @@
...
+# https://opentelemetry.io/docs/specs/semconv/non-normative/http-migration/
+
+semconv_stability = %w[dup stable old]
+
+semconv_stability.each do |mode|
+  appraise "net-http-#{mode}" do
+    # NOOP - net-http is part of the Ruby standard library.
+    # We are only using Appraisals to allow testing of the
+    # different stability modes. The file and the Appraisal
+    # gem can be removed once semconv migration is complete.
```
kaylareopelle, line 19: Nice!

instrumentation/net_http/lib/opentelemetry/instrumentation/net/http/patches/dup/instrumentation.rb
```diff
@@ -0,0 +1,130 @@
...
+                  'server.port' => conn_port
+                }.merge!(OpenTelemetry::Common::HTTP::ClientContext.attributes)
+
+                if use_ssl? && proxy?
+                  span_name = 'CONNECT'
+                  span_kind = :client
+                else
+                  span_name = 'connect'
+                  span_kind = :internal
+                end
```
kaylareopelle, line 82: Same question here as on Excon: why is connect capitalized in one case and lowercase in the other?

hannahramadan: I'm not sure, but it seems to be intentional. I'll see if I can dig up anything on this. Also discussed in https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/1569#discussion_r2237535360

kaylareopelle: Resolved: https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/1569#discussion_r2261122328

kaylareopelle (approved):

instrumentation/net_http/test/opentelemetry/instrumentation/net/http/dup/instrumentation_test.rb
```diff
@@ -0,0 +1,366 @@
...
+
+        instrumentation.install(config)
+      end
+
+      it 'does not create a span when request ignored using a string' do
+        Net::HTTP.get('foobar.com', '/body')
+        _(exporter.finished_spans.size).must_equal 0
+      end
+
+      it 'does not create a span when request ignored using a regexp' do
```
kaylareopelle, line 195: @hannahramadan - The Net::HTTP failing test looks related to changes from a recent PR. Those fixes may need to be incorporated into the dup/old/stable tests on this one: 
https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/1604/files#diff-99a3ac2b3a9bb5c2d057a09c5101ddb9d5b472344b41269b8a9f1269bceabaaa

PR #1587, opened 2025-07-09
feat: add net_ldap instrumentation

scbjans:
Largely based on prior work in https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/1432 by @Hareramrai and @jdehaan

Since that PR seems stale, I've updated it with some small edits to comply with release management using toys.

Should fix: github.com/open-telemetry/opentelemetry-ruby-contrib/issues/669


.toys/.data/releases.yml
```diff
@@ -231,7 +237,8 @@ gems:
 
   - name: opentelemetry-propagator-google_cloud_trace_context
     directory: propagator/google_cloud_trace_context
-    version_constant: [OpenTelemetry, Propagator, GoogleCloudTraceContext, VERSION]
+    version_constant:
```
arielvalentin: Why did you add new lines to the GCP related definitions?

instrumentation/net_ldap/lib/opentelemetry/instrumentation/net/ldap/instrumentation.rb
```diff
@@ -0,0 +1,39 @@
...
+#
+# SPDX-License-Identifier: Apache-2.0
+
+module OpenTelemetry
+  module Instrumentation
+    module Net
+      module LDAP
+        # The Instrumentation class contains logic to detect and install the Net::LDAP
+        # instrumentation
+        class Instrumentation < OpenTelemetry::Instrumentation::Base
```
arielvalentin, line 13: Is there a min version requirement?

instrumentation/net_ldap/lib/opentelemetry/instrumentation/net/ldap/instrumentation_service.rb
```diff
@@ -0,0 +1,84 @@
...
+
+            tracer.in_span(
+              event,
+              attributes: attributes.compact,
+              kind: :client
+            ) do |span|
+              yield(payload).tap do |response|
+                annotate_span_with_response!(span, response)
+              end
+            rescue StandardError => e
```
arielvalentin: Doesn't `in_span` already handle Error Recording and setting the status?

instrumentation/net_ldap/lib/opentelemetry/instrumentation/net/ldap/instrumentation_service.rb
```diff
@@ -0,0 +1,84 @@
...
+
+          def tracer
+            LDAP::Instrumentation.instance.tracer
+          end
+
+          def instrumentation_config
+            LDAP::Instrumentation.instance.config
+          end
+
+          def annotate_span_with_response!(span, response)
```
arielvalentin: Only add a `!` if you have an existing method that does not have a `!`

instrumentation/net_ldap/lib/opentelemetry/instrumentation/net/ldap/instrumentation_service.rb
```diff
@@ -0,0 +1,84 @@
...
+              status_code ||= response.result_code
+              error_message = response.error_message.to_s
+              message = ::Net::LDAP.result2string(status_code)
+            end
+            span.set_attribute('ldap.status_code', status_code)
+            span.set_attribute('ldap.message', message) unless message.empty?
+
+            return if ::Net::LDAP::ResultCodesNonError.include?(status_code)
+
+            span.record_exception(error_message) unless error_message.empty?
```
arielvalentin: I believe record_exception is an explicit helper for _exception_ events. Consider using the Span Status Code description instead.

instrumentation/net_ldap/lib/opentelemetry/instrumentation/net/ldap/instrumentation_service.rb
```diff
@@ -0,0 +1,84 @@
...
+            message = ''
+            error_message = ''
+
+            if response.is_a?(::Net::LDAP::PDU)
+              status_code ||= response.result_code
+              error_message = response.error_message.to_s
+              message = ::Net::LDAP.result2string(status_code)
+            end
+            span.set_attribute('ldap.status_code', status_code)
+            span.set_attribute('ldap.message', message) unless message.empty?
```
arielvalentin: I think you should consider building up a local hash and then setting as many attributes as possible using `add_attributes` to avoid having to lock the mutex multiple times.

instrumentation/net_ldap/lib/opentelemetry/instrumentation/net/ldap/instrumentation_service.rb
```diff
@@ -0,0 +1,84 @@
...
+          def tracer
+            LDAP::Instrumentation.instance.tracer
+          end
+
+          def instrumentation_config
+            LDAP::Instrumentation.instance.config
+          end
+
+          def annotate_span_with_response!(span, response)
+            return unless response
```
arielvalentin: you should avoid a method dispatch if possible and add this check to L39 instead

release-please-config.json
```diff
@@ -227,4 +231,4 @@
       "version-file": "lib/opentelemetry/resource/detector/google_cloud_platform/version.rb"
     }
   }
-}
+}
```
arielvalentin:
```suggestion
}

```

instrumentation/net_ldap/lib/opentelemetry/instrumentation/net/ldap/instrumentation_service.rb
```diff
@@ -0,0 +1,84 @@
...
+          def instrument(event, payload)
+            attributes = {
+              'ldap.auth' => auth.except(:password).to_json,
+              'ldap.base' => base,
+              'ldap.encryption' => encryption.to_json,
+              'ldap.payload' => payload.to_json,
+              OpenTelemetry::SemanticConventions::Trace::NET_PEER_NAME => host || hosts,
+              OpenTelemetry::SemanticConventions::Trace::NET_PEER_PORT => port,
+              OpenTelemetry::SemanticConventions::Trace::PEER_SERVICE => instrumentation_config[:peer_service]
+            }.merge!(OpenTelemetry::Common::HTTP::ClientContext.attributes)
```
arielvalentin: why is this amending HTTP Client attributes? How is that relevant to LDAP lookups?

instrumentation/net_ldap/lib/opentelemetry/instrumentation/net/ldap/instrumentation_service.rb
```diff
@@ -0,0 +1,84 @@
...
+          def initialize(args = {})
+            @host = args[:host]
+            @port = args[:port]
+            @hosts = args[:hosts]
+            @auth = args[:auth]
+            @base = args[:base]
+            @encryption = args[:encryption]
+          end
+
+          def instrument(event, payload)
```
arielvalentin, line 27: Do we want more specific attribute mappers here for each of these events? 

https://github.com/search?q=repo%3Aruby-ldap%2Fruby-net-ldap+%2Finstrument+%22%2F&type=code

I do not particularly like the practice of dumping json strings where we _do not know if the contents of the provided payload have sensitive information in them_.

instrumentation/net_ldap/lib/opentelemetry/instrumentation/net/ldap/instrumentation_service.rb
```diff
@@ -0,0 +1,84 @@
...
+              'ldap.encryption' => encryption.to_json,
+              'ldap.payload' => payload.to_json,
+              OpenTelemetry::SemanticConventions::Trace::NET_PEER_NAME => host || hosts,
+              OpenTelemetry::SemanticConventions::Trace::NET_PEER_PORT => port,
+              OpenTelemetry::SemanticConventions::Trace::PEER_SERVICE => instrumentation_config[:peer_service]
+            }.merge!(OpenTelemetry::Common::HTTP::ClientContext.attributes)
+
+            tracer.in_span(
+              event,
+              attributes: attributes.compact,
```
arielvalentin: This allocates and extra hash object. 

lets avoid that by checking if the values are present before adding them to the hash.

instrumentation/net_ldap/lib/opentelemetry/instrumentation/net/ldap/instrumentation_service.rb
```diff
@@ -0,0 +1,84 @@
...
+              kind: :client
```
arielvalentin: Do you think LDAP qualifies as RPC? Would using RPC client semantics be right for this?
Would using ECS style events make more sense for these?

.toys/.data/releases.yml
scbjans: Thanks for your quick review @arielvalentin !

I guess this is from the `format on save` option in VSCode which I accidentally committed. I'll revert it.

I'll respond to the other comments this Friday.

instrumentation/net_ldap/lib/opentelemetry/instrumentation/net/ldap/instrumentation.rb
jdehaan: FYI: We tested the original implementation using gem 'net-ldap' 0.17.0 if that info helps in finding out some requirements. As it's a 0.x version we assumed all bets are off in this case....

instrumentation/net_ldap/lib/opentelemetry/instrumentation/net/ldap/patches/instrumentation.rb
```diff
@@ -0,0 +1,31 @@
...
+  module Instrumentation
+    module Net
+      module LDAP
+        module Patches
+          # Module to prepend to Net::LDAP for instrumentation
+          module Instrumentation
+            def initialize(args = {})
+              super
+
+              @instrumentation_service = args[:instrumentation_service] || OpenTelemetry::Instrumentation::Net::LDAP::InstrumentationService.new({
```
arielvalentin, line 17: Another thought I have here, what if we added this otel instrumentation as a decorator to the injected service?

instrumentation/net_ldap/lib/opentelemetry/instrumentation/net/ldap/instrumentation_service.rb
scbjans: TBH I just took the already existing stale PR that originally was opened to add this instrumentation. That did the job for me, and I thought it might be beneficial for others to have this instrumentation more easily available.
I rushed this a bit and skipped the due diligence (sorry!), so I'm glad you ask.

It doesn't: I removed it.

instrumentation/net_ldap/lib/opentelemetry/instrumentation/net/ldap/instrumentation_service.rb
scbjans: https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/1587/commits/0583992b7cf2d4262a48672efdc8b8885ab7b1e3

instrumentation/net_ldap/lib/opentelemetry/instrumentation/net/ldap/instrumentation_service.rb
scbjans: https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/1587/commits/0583992b7cf2d4262a48672efdc8b8885ab7b1e3

instrumentation/net_ldap/lib/opentelemetry/instrumentation/net/ldap/instrumentation_service.rb
scbjans: https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/1587/commits/0583992b7cf2d4262a48672efdc8b8885ab7b1e3

instrumentation/net_ldap/lib/opentelemetry/instrumentation/net/ldap/instrumentation_service.rb
scbjans: https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/1587/commits/0583992b7cf2d4262a48672efdc8b8885ab7b1e3

instrumentation/net_ldap/lib/opentelemetry/instrumentation/net/ldap/instrumentation_service.rb
scbjans: Yes it does, but it doesn't add the `ldap.error_message` attribute. I've updated this a bit, and in https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/1587/commits/0583992b7cf2d4262a48672efdc8b8885ab7b1e3 we'll rescue specifically on a `::Net::LDAP::Error`

instrumentation/net_ldap/lib/opentelemetry/instrumentation/net/ldap/instrumentation.rb
scbjans: Thanks, I only tested this with `net-ldap` 0.19.0. I'm inclined to say we should set a minimal version to `0.17.1`, since in that version Ruby3.x support was added (as in added to CI), since our minimum Ruby version is 3.1 (which is also EOL, shall I put up a PR for that as well?)
https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/1587/commits/cb5f1c4376bb1b54bdcdc82ce351d98af75ecd7a

instrumentation/net_ldap/lib/opentelemetry/instrumentation/net/ldap/instrumentation_service.rb
scbjans: I was inclined to say LDAP qualifies as RPC, but my AI friend does not agree with me. Apparently LDAP itself is not RPC, but it is often combined with RPC (e.g. when joining computers to Windows domains), but we're only instrumenting LDAP in this case. Semconv doesn't specify anything for `ldap`, and other languages don't seem to have ldap instrumentation.

I'm not sure what to do here.

instrumentation/net_ldap/lib/opentelemetry/instrumentation/net/ldap/instrumentation.rb
jdehaan: That sounds very reasonable to me, we are also planning to upgrade anyways soon.

instrumentation/net_ldap/lib/opentelemetry/instrumentation/net/ldap/instrumentation_service.rb
scbjans: I agree.

My naive thought was that it would be handled in `net-ldap` itself, but upon closer inspection I see that this instrumentation can leak e.g. passwords in cleartext.
I'll need some time to figure out which specific attributes to include.

scbjans: Fixed and added test in 3d2ff26717b7fc74b4ac379f48a075a28afde089

instrumentation/net_ldap/example/Gemfile
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+source 'https://rubygems.org'
+
+gem 'net-ldap'
+gem 'opentelemetry-api'
+gem 'opentelemetry-common'
+gem 'opentelemetry-instrumentation-base'
```
kaylareopelle: should be able to leave these out, since they're dependencies of the other gems.
```suggestion
```

instrumentation/net_ldap/lib/opentelemetry/instrumentation/net/ldap/instrumentation_service.rb
kaylareopelle: Good points. I was also wondering about SemConv as I'm reading through this. I asked the [#opentelemetry-semantic-conventions](https://cloud-native.slack.com/archives/C041APFBYQP/p1758067382218469) channel on CNCF Slack for advice.

kaylareopelle (commented):
Hi @scbjans, thanks for opening this PR. 

I'm not very familiar with LDAP, so I'll take another pass once I learn a bit more. For now, I left a few comments.

In addition to those comments, would you mind updating the CODEOWNERS file to list you as the owner of this gem? PR assignment for CODEOWNERS is currently broken, but I'm hoping to get it fixed soon.

instrumentation/net_ldap/lib/opentelemetry/instrumentation/net/ldap/instrumentation_service.rb
```diff
@@ -0,0 +1,88 @@
...
+          end
+
+          def instrument(event, payload)
+            attributes = {
+              'ldap.auth' => auth.except(:password).to_json,
+              'ldap.base' => base,
+              'ldap.encryption' => encryption.to_json,
+              'ldap.payload' => payload.to_json,
+              OpenTelemetry::SemanticConventions::Trace::NET_PEER_NAME => host || hosts,
+              OpenTelemetry::SemanticConventions::Trace::NET_PEER_PORT => port,
```
thompson-tomo: These should be in network namespace rather than net.

instrumentation/net_ldap/lib/opentelemetry/instrumentation/net/ldap/instrumentation_service.rb
```diff
@@ -0,0 +1,88 @@
...
+            tracer.in_span(
+              event,
+              attributes: AttributeMapper.map(attributes),
+              kind: :client
+            ) do |span|
+              yield(payload).tap do |response|
+                annotate_span_with_response(span, response) if response
+              end
+            rescue ::Net::LDAP::Error => e
+              span.set_attribute('ldap.error_message', "#{e.class}: #{e.message}")
```
thompson-tomo: Should we not be using error.type & error.message?

instrumentation/net_ldap/lib/opentelemetry/instrumentation/net/ldap/instrumentation_service.rb
```diff
@@ -0,0 +1,88 @@
...
+
+            if response.is_a?(::Net::LDAP::PDU)
+              status_code ||= response.result_code
+              error_message = response.error_message.to_s
+              message = ::Net::LDAP.result2string(status_code)
+            end
+            attributes = {
+              'ldap.status_code' => status_code
+            }
+            attributes['ldap.message'] = message unless message.empty?
```
thompson-tomo: Perhaps ldap.response.message?

instrumentation/net_ldap/lib/opentelemetry/instrumentation/net/ldap/instrumentation_service.rb
```diff
@@ -0,0 +1,88 @@
...
+            @base = args[:base]
+            @encryption = args[:encryption]
+          end
+
+          def instrument(event, payload)
+            attributes = {
+              'ldap.auth' => auth.except(:password).to_json,
+              'ldap.base' => base,
+              'ldap.encryption' => encryption.to_json,
+              'ldap.payload' => payload.to_json,
```
thompson-tomo: Perhaps ldap.request.message?

instrumentation/net_ldap/lib/opentelemetry/instrumentation/net/ldap/instrumentation_service.rb
```diff
@@ -0,0 +1,88 @@
...
+            @port = args[:port]
+            @hosts = args[:hosts]
+            @auth = args[:auth]
+            @base = args[:base]
+            @encryption = args[:encryption]
+          end
+
+          def instrument(event, payload)
+            attributes = {
+              'ldap.auth' => auth.except(:password).to_json,
```
thompson-tomo: Do we need auth given we removed it from db etc?

instrumentation/net_ldap/lib/opentelemetry/instrumentation/net/ldap/instrumentation_service.rb
```diff
@@ -0,0 +1,88 @@
...
+            message = ''
+            error_message = ''
+
+            if response.is_a?(::Net::LDAP::PDU)
+              status_code ||= response.result_code
+              error_message = response.error_message.to_s
+              message = ::Net::LDAP.result2string(status_code)
+            end
+            attributes = {
+              'ldap.status_code' => status_code
```
thompson-tomo: Perhaps ldap.response.status_code?

instrumentation/net_ldap/lib/opentelemetry/instrumentation/net/ldap/instrumentation_service.rb
```diff
@@ -0,0 +1,88 @@
...
+            }
+            attributes['ldap.message'] = message unless message.empty?
+            attributes['ldap.error_message'] = error_message unless error_message.empty?
```
thompson-tomo: Perhaps error.message?

instrumentation/net_ldap/lib/opentelemetry/instrumentation/net/ldap/instrumentation_service.rb
```diff
@@ -0,0 +1,88 @@
...
+          def instrument(event, payload)
+            attributes = {
+              'ldap.auth' => auth.except(:password).to_json,
+              'ldap.base' => base,
+              'ldap.encryption' => encryption.to_json,
+              'ldap.payload' => payload.to_json,
+              OpenTelemetry::SemanticConventions::Trace::NET_PEER_NAME => host || hosts,
+              OpenTelemetry::SemanticConventions::Trace::NET_PEER_PORT => port,
+              OpenTelemetry::SemanticConventions::Trace::PEER_SERVICE => instrumentation_config[:peer_service]
+            }
```
thompson-tomo, line 46: Should we not be adding network.protocol.name with a  value of ldap & potentially network.type?

instrumentation/net_ldap/lib/opentelemetry/instrumentation/net/ldap/instrumentation_service.rb
scbjans: Thanks! I see the new opentelemetry-semantic-conventions gem has been released yesterday, I'll update this PR accordingly

instrumentation/net_ldap/lib/opentelemetry/instrumentation/net/ldap/instrumentation_service.rb
scbjans: I like that, brings it more in line with the HTTP semconv

instrumentation/net_ldap/lib/opentelemetry/instrumentation/net/ldap/instrumentation_service.rb
scbjans: Do they need to be in `network` namespace, or in the `server` namespace? (see [deprecation notice](https://opentelemetry.io/docs/specs/semconv/registry/attributes/network/#net-peer-name) )

thompson-tomo: Possibly both as per https://opentelemetry.io/docs/specs/semconv/general/attributes/#clientserver-examples-using-networkpeer

instrumentation/net_ldap/lib/opentelemetry/instrumentation/net/ldap/instrumentation_service.rb
thompson-tomo: I think a good starting point would be to capture the name of the operation being performed ie open, read, parse_pdu.

scbjans:
@kaylareopelle thanks for your feedback; I updated it in https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/1587/commits/7264acb6e8dbfccb0c2e11c04d56b3984068a18c

@thompson-tomo thank you as well for your feedback, I believe I addressed it in https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/1587/commits/44a1d5413a831ae8f3c325e57fd00ce4d592c569
However, tests are failing, because the new `opentelemetry-semantic_conventions` is missing some `require`s of the `SemConv` module. I'll have to tidy up my fixes for that and submit a PR

instrumentation/net_ldap/lib/opentelemetry/instrumentation/net/ldap/instrumentation_service.rb
thompson-tomo: Resolved.

instrumentation/net_ldap/lib/opentelemetry/instrumentation/net/ldap/instrumentation_service.rb
thompson-tomo: Resolved.

instrumentation/net_ldap/lib/opentelemetry/instrumentation/net/ldap/instrumentation_service.rb
thompson-tomo: Resolved.

instrumentation/net_ldap/lib/opentelemetry/instrumentation/net/ldap/instrumentation_service.rb
thompson-tomo: Resolved.

instrumentation/net_ldap/lib/opentelemetry/instrumentation/net/ldap/instrumentation_service.rb
thompson-tomo: Resolved.

instrumentation/net_ldap/lib/opentelemetry/instrumentation/net/ldap/instrumentation_service.rb
```diff
@@ -0,0 +1,95 @@
+# frozen_string_literal: true
+
+# Copyright The OpenTelemetry Authors
+#
+# SPDX-License-Identifier: Apache-2.0
+
+require 'opentelemetry/semconv/incubating/peer'
```
scbjans, line 7: I do not like this line; but I'll add a test to make sure we catch it when the constant is removed from `Incubating`

scbjans: https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/1587/commits/111f18c81dd12f26cc0faa0d4599205148686638

instrumentation/net_ldap/lib/opentelemetry/instrumentation/net/ldap/instrumentation_service.rb
scbjans: I'm not sure why it was removed from db, but I think it's valuable to see which user connects to ldap; e.g. to troubleshoot permissions

instrumentation/net_ldap/lib/opentelemetry/instrumentation/net/ldap/instrumentation_service.rb
kaylareopelle: Heard back from the thread. There aren't any specific LDAP conventions we should adhere to. Instead, they suggested we use attributes from other relevant categories like [network](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/registry/attributes/network.md) and [server](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/registry/attributes/server.md).

instrumentation/net_ldap/lib/opentelemetry/instrumentation/net/ldap/instrumentation_service.rb
thompson-tomo: I believe it was in relation to the range of different auth methods/mechanism that could be used ie certificates vs user credentials.

The name Auth doesn't feel scoped enough as it could mean many things. In ldap v3 SASL is supported which opens up additional techniques.

instrumentation/net_ldap/lib/opentelemetry/instrumentation/net/ldap/instrumentation_service.rb
```diff
@@ -0,0 +1,95 @@
...
+            @hosts = args[:hosts]
+            @auth = args[:auth]
+            @base = args[:base]
+            @encryption = args[:encryption]
+          end
+
+          def instrument(event, payload)
+            attributes = {
+              'ldap.auth' => auth.except(:password).to_json,
+              'ldap.base' => base,
```
thompson-tomo: Should this become `ldap.tree.base` or similar that way we have potential of adding other properties for the tree.

instrumentation/net_ldap/lib/opentelemetry/instrumentation/net/ldap/instrumentation_service.rb
```diff
@@ -0,0 +1,95 @@
...
+              'ldap.encryption' => encryption.to_json,
```
thompson-tomo: Not sure of the benefits of this given it could be a hash of properties.

instrumentation/net_ldap/lib/opentelemetry/instrumentation/net/ldap/instrumentation_service.rb
scbjans: I'm sorry for the slow turnaround.

In my testsetup, this currently produces: `ldap.auth: { method: "simple", username: "<username-of-ldap-bind-user>" }`.
This is directly based on the [auth hash of the net-ldap gem](https://github.com/ruby-ldap/ruby-net-ldap/blob/master/lib/net/ldap.rb#L469-L472)
Would it be better to have `ldap.auth.method` and `ldap.auth.username`? Or shouldn't they be declared under `ldap.auth` at all?

instrumentation/net_ldap/lib/opentelemetry/instrumentation/net/ldap/instrumentation_service.rb
scbjans: I don't foresee that yet, but it's probably best to be futureproof. I'll make that change.

instrumentation/net_ldap/lib/opentelemetry/instrumentation/net/ldap/instrumentation_service.rb
scbjans: This is directly based on the [encryption hash of the net-ldap gem](https://github.com/ruby-ldap/ruby-net-ldap/blob/master/lib/net/ldap.rb#L486-L543)
In my testsetup, this currently produces: `ldap.encryption: { method: "simple_tls", tls_options: { verify_mode: 1 } }`. I can imagine this should better fit under
the [TLS](https://opentelemetry.io/docs/specs/semconv/registry/attributes/tls/) namespace, however, I don't know how we should define attributes like `verify_mode`, or `ca_file`. Might be best to leave it out (at least for now)?

instrumentation/net_ldap/lib/opentelemetry/instrumentation/net/ldap/instrumentation_service.rb
thompson-tomo: If we retain it, then we should go with ldap.auth.method etc

instrumentation/net_ldap/lib/opentelemetry/instrumentation/net/ldap/instrumentation_service.rb
thompson-tomo: Agree best to leave it out for now and can be added if need arises but use the tls namespace where appropriate.

instrumentation/net_ldap/lib/opentelemetry/instrumentation/net/ldap/instrumentation_service.rb
```diff
@@ -0,0 +1,95 @@
...
+            attributes = {
+              'ldap.auth' => auth.except(:password).to_json,
+              'ldap.tree.base' => base,
+              'ldap.encryption' => encryption.to_json,
+              'ldap.request.message' => payload.to_json,
+              OpenTelemetry::SemConv::SERVER::SERVER_ADDRESS => host || hosts,
+              OpenTelemetry::SemConv::SERVER::SERVER_PORT => port,
+              OpenTelemetry::SemConv::Incubating::PEER::PEER_SERVICE => instrumentation_config[:peer_service],
+              OpenTelemetry::SemConv::NETWORK::NETWORK_PROTOCOL_NAME => 'ldap'
+            }
```
thompson-tomo: Looking at the semconv docs, is there potential for us to set network.transport as well as potentially the network.peer.* atttibutes?

instrumentation/net_ldap/lib/opentelemetry/instrumentation/net/ldap/instrumentation_service.rb
```diff
@@ -0,0 +1,95 @@
...
+          def instrument(event, payload)
+            attributes = {
+              'ldap.auth' => auth.except(:password).to_json,
+              'ldap.tree.base' => base,
+              'ldap.encryption' => encryption.to_json,
+              'ldap.request.message' => payload.to_json,
+              OpenTelemetry::SemConv::SERVER::SERVER_ADDRESS => host || hosts,
+              OpenTelemetry::SemConv::SERVER::SERVER_PORT => port,
+              OpenTelemetry::SemConv::Incubating::PEER::PEER_SERVICE => instrumentation_config[:peer_service],
+              OpenTelemetry::SemConv::NETWORK::NETWORK_PROTOCOL_NAME => 'ldap'
```
thompson-tomo: Is there a way to also include the protocol version?

scbjans: Done in https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/1587/commits/97837bc17ca55cc5a7cd5c4e7bcc44f2e10501b3

instrumentation/net_ldap/lib/opentelemetry/instrumentation/net/ldap/instrumentation_service.rb
scbjans: `net-ldap` only supports `tcp`, so I've added that in https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/1587/commits/97837bc17ca55cc5a7cd5c4e7bcc44f2e10501b3. Since the `net-ldap` gem abstracts away some connection details, I don't see an easy way to add the `network.peer.*` attributes

instrumentation/net_ldap/lib/opentelemetry/instrumentation/net/ldap/instrumentation_service.rb
scbjans: Removed it in https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/1587/commits/97837bc17ca55cc5a7cd5c4e7bcc44f2e10501b3

instrumentation/net_ldap/lib/opentelemetry/instrumentation/net/ldap/instrumentation_service.rb
scbjans: Done so in https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/1587/commits/dc069a9769e46b33198681b2a25df93ab3a0eb08

instrumentation/net_ldap/lib/opentelemetry/instrumentation/net/ldap/instrumentation_service.rb
scbjans: Renamed in https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/1587/commits/97837bc17ca55cc5a7cd5c4e7bcc44f2e10501b3

instrumentation/net_ldap/lib/opentelemetry/instrumentation/net/ldap/instrumentation_service.rb
scbjans: Spans name is set to the event name, e.g. `search.net_ldap` (see [here](https://github.com/search?q=repo%3Aruby-ldap%2Fruby-net-ldap+%2Finstrument+%22%2F&type=code) for a list of possible span names)

thompson-tomo: Sure, but if you look at https://opentelemetry.io/docs/specs/semconv/http/http-spans/ as well as others the span name is usually defined as a templated string which uses the attributes to build the name.

scbjans: Thank you, I really appreciate your time and feedback!

I've updated the span name to be `LDAP <operation>` (modeled after the default [Rack instrumentation span naming](https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/main/instrumentation/rack/README.md?plain=1#L64) ). I also found that before I always set span_kind to client; updated this in https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/1587/commits/f0338fcd493dca9d16a66de2477f03bd66085048 to correctly set it to `internal` for internal spans as well.

thompson-tomo: No worries at all. 🙂 I feel that we have gotten the signal definition into a great shape which is consistent with already defined signals.

My last suggestion based on my last comment would be to add the ldap operation as an attribute `ldap.operation.type` which would have the same value as used in the span name template.

scbjans: Thanks, added the attribute in https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/1587/commits/f39b2dee7e0010cd087fd2573daafbe20feefd11

thompson-tomo (approved):
Span definition & attribute usage looks good especially as it is consistent with existing sem conv definitions. 👍 

Will allow others to validate the coding patterns/structure.

kaylareopelle (approved):

instrumentation/net_ldap/lib/opentelemetry/instrumentation/net/ldap/patches/instrumentation.rb
scbjans: Sorry, I've seem to have missed this earlier. Right now we use the method as prescribed in [ruby-net-ldap](https://github.com/ruby-ldap/ruby-net-ldap/blob/master/lib/net/ldap.rb#L482)

instrumentation/net_ldap/lib/opentelemetry/instrumentation/net/ldap/instrumentation_service.rb
```diff
@@ -0,0 +1,112 @@
...
+            operation_type, span_kind = event.split('.').then { |s| [s.first, s.last == 'net_ldap_connection' ? :internal : :client] }
+
+            attributes = {
+              'ldap.auth.method' => auth[:method].to_s,
+              'ldap.auth.username' => auth[:username].to_s,
+              'ldap.operation.type' => operation_type,
+              'ldap.request.message' => begin
+                payload.to_json
+              rescue JSON::GeneratorError
+                'Could not generate JSON'
```
scbjans: Previously this broke when someone used binary data (e.g. an `objectSid` in Active Directory) in the filter. Would it be better if this is set to `nil` in that case? (that would result in the `ldap.request.message` attribute not available on the span, which might be unexpected?)

scbjans: After thinking about this, I think it's better to set it to `nil`, instead of changing the value type as well.
Done so in https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/1587/commits/fe142ea760501550c8ac75c6d663ddb42a9c691b

instrumentation/net_ldap/lib/opentelemetry/instrumentation/net/ldap/attribute_mapper.rb
```diff
@@ -0,0 +1,55 @@
...
+#
+# SPDX-License-Identifier: Apache-2.0
+
+module OpenTelemetry
+  module Instrumentation
+    module Net
+      module LDAP
+        # attribute mapper to redact sensitive keys
+        class AttributeMapper
+          SENSITIVE_KEYS = %w[userPassword unicodePwd lmPassword msDS-ManagedPassword ntPassword authPassword krbPrincipalKey sambaNTPassword sambaLMPassword].freeze
```
arielvalentin: I think it is safer to have an allow list as opposed to a deny list. 

This will prevent any new sensitive keys from sneaking in.

instrumentation/net_ldap/lib/opentelemetry/instrumentation/net/ldap/instrumentation_service.rb
```diff
@@ -0,0 +1,112 @@
...
+              end,
+              'ldap.tree.base' => base,
+              OpenTelemetry::SemConv::SERVER::SERVER_ADDRESS => host || hosts,
+              OpenTelemetry::SemConv::SERVER::SERVER_PORT => port,
+              OpenTelemetry::SemConv::Incubating::PEER::PEER_SERVICE => instrumentation_config[:peer_service],
+              OpenTelemetry::SemConv::NETWORK::NETWORK_TRANSPORT => 'tcp',
+              OpenTelemetry::SemConv::NETWORK::NETWORK_PROTOCOL_NAME => 'ldap',
+              OpenTelemetry::SemConv::NETWORK::NETWORK_PROTOCOL_VERSION => ::Net::LDAP::Connection::LdapVersion
+            }
+            attributes.delete_if { |_key, value| value.nil? }
```
arielvalentin: We have moved away from using `delete_if` or `compact` since it is significantly slower than conditionally adding new entries in the hash. 

If its your preference to clear out items instead use then `compact!`

instrumentation/net_ldap/lib/opentelemetry/instrumentation/net/ldap/instrumentation_service.rb
```diff
@@ -0,0 +1,112 @@
...
+                    annotate_span_with_response(span, response) if response
+                  end
+                end
+              else
+                yield(payload).tap do |response|
+                  annotate_span_with_response(span, response) if response
+                end
+              end
+            rescue ::Net::LDAP::Error => e
+              span.add_attributes({
```
arielvalentin: Why would this not use `record_exception`? 

Do you not need the backtrace?

Is the message always encoded as Unicode with UTF-8?

instrumentation/net_ldap/lib/opentelemetry/instrumentation/net/ldap/instrumentation_service.rb
```diff
@@ -0,0 +1,112 @@
...
+            if response.is_a?(::Net::LDAP::PDU)
+              status_code ||= response.result_code
+              error_message = response.error_message.to_s
+              message = ::Net::LDAP.result2string(status_code)
+            end
+            attributes = {
+              'ldap.response.status_code' => status_code
+            }
+            attributes['ldap.response.message'] = message unless message.empty?
+            attributes['error.message'] = error_message unless error_message.empty?
```
arielvalentin: similarly are these always unicode utf-8?

instrumentation/net_ldap/lib/opentelemetry/instrumentation/net/ldap/instrumentation_service.rb
scbjans: Thanks again for your review.

There is no good reason `record_exception` isn't used, fixed it in https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/1587/commits/bce45ffd543dc723d72cf136556f8378378e39e3

According to [the RFC (4.1.2 String Types)](https://www.rfc-editor.org/rfc/rfc4511) this is always ISO10646 UTF-8 encoded, so it should be fine.

instrumentation/net_ldap/lib/opentelemetry/instrumentation/net/ldap/attribute_mapper.rb
scbjans: agreed, changed to an allowlist in https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/1587/commits/bce45ffd543dc723d72cf136556f8378378e39e3

@jdehaan @Hareramrai could you please have a look if there are any attributes you are missing in [attribute_mapper.rb](https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/1587/changes#diff-df71c7a42759e119f8d3922f904865489f2047fa86bd356301ffce80450326ca)

instrumentation/net_ldap/lib/opentelemetry/instrumentation/net/ldap/attribute_mapper.rb
```diff
@@ -0,0 +1,62 @@
...
+        # attribute mapper to redact keys which are not allowed
+        class AttributeMapper
+          LDAP_GENERAL_ATTRIBUTES = %w[attributes base filter ignore_server_caps left op operations paged_searches_supported replace right scope].freeze
+          LDAP_OBJECT_ATTRIBUTES  = %w[accountExpires codePage countryCode cn description displayName distinguishedName dn
+                                       givenName instanceType mail memberOf name objectCategory objectClass pwdChangedTime pwdLastSet
+                                       sAMAccountName userAccountControl userPrincipalName].freeze
+          SPAN_ATTRIBUTES =         %w[exception.message exception.stacktrace exception.type ldap.auth.method ldap.auth.username ldap.error.message
+                                       ldap.operation.type ldap.request.message ldap.response.status_code ldap.tree.base network.protocol.name
+                                       network.protocol.version network.transport peer.service server.address server.port].freeze
+          ALLOWED_KEYS = (LDAP_GENERAL_ATTRIBUTES + LDAP_OBJECT_ATTRIBUTES + SPAN_ATTRIBUTES).freeze
```
arielvalentin, line 24: Have you considered using a `Set` to avoid a linear search?

instrumentation/net_ldap/lib/opentelemetry/instrumentation/net/ldap/attribute_mapper.rb
```diff
@@ -0,0 +1,62 @@
...
+# SPDX-License-Identifier: Apache-2.0
+
+module OpenTelemetry
+  module Instrumentation
+    module Net
+      module LDAP
+        # attribute mapper to redact keys which are not allowed
+        class AttributeMapper
+          LDAP_GENERAL_ATTRIBUTES = %w[attributes base filter ignore_server_caps left op operations paged_searches_supported replace right scope].freeze
+          LDAP_OBJECT_ATTRIBUTES  = %w[accountExpires codePage countryCode cn description displayName distinguishedName dn
```
arielvalentin: For those fields that are object types, will it always be a JSON string?

instrumentation/net_ldap/lib/opentelemetry/instrumentation/net/ldap/attribute_mapper.rb
```diff
@@ -0,0 +1,62 @@
...
+              end
+            when Array
+              # Special case: LDAP operation tuple like ["replace", "unicodePwd", ["value"]]
+              if obj.size == 3 && obj[1].is_a?(String) && !ALLOWED_KEYS.include?(obj[1])
+                [obj[0], obj[1], ['[REDACTED]']]
+              else
+                obj.map { |item| deep_map(item) }
+              end
+            when String
+              map_json(obj)
```
arielvalentin, line 59: This looks like it will attempt to parse both JSON and String values but falling back to the raw string value. 

Are there specific fields that are JSON vs string literals? Can we avoid raising exceptions by checking that the string is a object type that needs to be parsed??

instrumentation/net_ldap/lib/opentelemetry/instrumentation/net/ldap/attribute_mapper.rb
scbjans: I had not considered that, but now that I did, that makes perfect sense: https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/1587/commits/409b612812ced9519449d11a9103c793c216dcbc

instrumentation/net_ldap/lib/opentelemetry/instrumentation/net/ldap/attribute_mapper.rb
scbjans: yes, for these included fields, it will always be a JSON string.

Note: we've encountered an edge-case when someone searched with binary data in the filter. That was fixed in https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/1587/changes/621d92566d49d6595a8bc89288a6b340c034f272 but shouldn't even occur anymore, now we've switched to an allowlist for the attributes.

instrumentation/net_ldap/lib/opentelemetry/instrumentation/net/ldap/attribute_mapper.rb
scbjans: I'm not aware of a check where it could be validated if the String needs to be parsed.

The best I could think of, is checking if the String starts with `{` or `[`. That should avoid raising exceptions.

instrumentation/net_ldap/Gemfile
```diff
@@ -0,0 +1,32 @@
...
+  gem 'opentelemetry-instrumentation-base', path: '../base'
+  if RUBY_VERSION >= '3.4'
+    gem 'base64'
+    gem 'logger'
+    gem 'mutex_m'
+  end
+
+  # ostruct is not part of the standard library from Ruby4+
+  # from net-ldap 0.20+ it is defined as a gem dependency
+  gem 'ostruct' if Gem::Version.new(RUBY_VERSION) >= Gem::Version.new('4')
```
scbjans: For net-ldap 0.20+ `ostruct` is defined as a gem dependency.
I'm also fine with updating the `compatible` block, so that only net-ldap 0.20+ is supported, if that is preferred

instrumentation/net_ldap/lib/opentelemetry/instrumentation/net/ldap/instrumentation_service.rb
```diff
@@ -0,0 +1,110 @@
...
+                  yield(payload).tap do |response|
+                    annotate_span_with_response(span, response) if response
+                  end
+                end
+              else
+                yield(payload).tap do |response|
+                  annotate_span_with_response(span, response) if response
+                end
+              end
+            rescue ::Net::LDAP::Error => e
```
arielvalentin: Doesn't `in_span` already record an exception? 
Won't this record it twice?

instrumentation/net_ldap/test/opentelemetry/instrumentation/net/ldap/instrumentation_test.rb
```diff
@@ -0,0 +1,199 @@
...
+        ldap.connection = FakeConnection.new
+
+        assert_raises Net::LDAP::Error do
+          ldap.add({})
+        end
+
+        _(exporter.finished_spans.size).must_equal 1
+        _(span.name).must_equal 'LDAP add'
+        _(span.kind).must_equal :client
+        _(span.events[0].attributes['exception.message']).must_equal('Connection timed out - user specified timeout')
```
arielvalentin, line 155: The test does not check for this, but I think this will end up with 2 exception events as I mentioned above

arielvalentin (commented):
One last question regarding exception handling. After that I think the 6 month long journey comes to an end.

instrumentation/net_ldap/Gemfile
scbjans: According to [this comment](https://github.com/open-telemetry/opentelemetry-ruby/pull/2016#issuecomment-3740492166) the problem was with an old version of `rake` requiring `openstruct`. I'll update the `Appraisals` to keep the compatibility, and move `openstruct` there

instrumentation/net_ldap/lib/opentelemetry/instrumentation/net/ldap/instrumentation_service.rb
scbjans: indeed it does! Fixed in https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/1587/commits/23069fd67b469e0b75e791cc3e3f951062d6a9e9

scbjans:
> One last question regarding exception handling. After that I think the 6 month long journey comes to an end.

Thank you (and @kaylareopelle and @thompson-tomo) so much for helping me get this in a better shape!

instrumentation/net_ldap/Gemfile
```diff
@@ -0,0 +1,28 @@
...
+source 'https://rubygems.org'
+
+gemspec
+
+group :test do
+  gem 'appraisal', '~> 2.5'
+  gem 'minitest', '~> 5.0'
+  gem 'opentelemetry-sdk', '~> 1.1'
+  gem 'opentelemetry-test-helpers', '~> 0.3'
+  gem 'rake', '~> 13.3'
```
arielvalentin, line 16:
```suggestion
  gem 'rake', '>= 13.3'
```

arielvalentin (approved):

kaylareopelle:
Hi @scbjans, thank you for your patience with us on this PR. I'd love to get it merged in and released. There's a merge conflict with the release-please-config. Could you take a look at it?

scbjans:
> Hi @scbjans, thank you for your patience with us on this PR. I'd love to get it merged in and released. There's a merge conflict with the release-please-config. Could you take a look at it?

@kaylareopelle no worries, merge conflict is fixed. Failing test seem unrelated.

PR #1588, opened 2025-07-10
feat: Begin HTTPClient semantic convention stability migration

hannahramadan:


kaylareopelle (approved):
🦖

ericmustin (approved):

kaylareopelle:
Same Net::HTTP error persists as in #1599

PR #1589, opened 2025-07-10
feat: HTTPX semantic convention stability opt in

hannahramadan:


instrumentation/httpx/lib/opentelemetry/instrumentation/httpx/dup/plugin.rb
```diff
@@ -0,0 +1,136 @@
...
+                'url.full' => "#{uri.scheme}://#{uri.host}",
+                'server.address' => uri.host,
+                'server.port' => uri.port
+              }
+
+              attributes['url.query'] = uri.query unless uri.query.nil?
+              attributes[OpenTelemetry::SemanticConventions::Trace::PEER_SERVICE] = config[:peer_service] if config[:peer_service]
+              attributes.merge!(OpenTelemetry::Common::HTTP::ClientContext.attributes)
+
+              span = tracer.start_span(verb.to_s, attributes: attributes, kind: :client, start_timestamp: start_time)
```
kaylareopelle: I don't know if we need the `to_s` here. `verb` is called directly for the old attributes on L79. Symbols aren't valid attribute values, so I believe the SDK would throw an error if it wasn't a String.

instrumentation/httpx/lib/opentelemetry/instrumentation/httpx/stable/plugin.rb
```diff
@@ -0,0 +1,128 @@
...
+
+            # return a span initialized with the +@request+ state.
+            def initialize_span(request, start_time = ::Time.now)
+              verb = request.verb
+              uri = request.uri
+
+              config = HTTPX::Instrumentation.instance.config
+
+              attributes = {
+                OpenTelemetry::SemanticConventions::Trace::HTTP_HOST => uri.host,
```
kaylareopelle: I asked a question about the `http.host` attribute in the Excon PR: https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/1569#discussion_r2214438922 

Same question applies here.

instrumentation/httpx/lib/opentelemetry/instrumentation/httpx/stable/plugin.rb
```diff
@@ -0,0 +1,128 @@
...
+                'url.path' => uri.path,
+                'url.full' => "#{uri.scheme}://#{uri.host}",
+                'server.address' => uri.host,
+                'server.port' => uri.port
+              }
+              attributes['url.query'] = uri.query unless uri.query.nil?
+              attributes[OpenTelemetry::SemanticConventions::Trace::PEER_SERVICE] = config[:peer_service] if config[:peer_service]
+              attributes.merge!(OpenTelemetry::Common::HTTP::ClientContext.attributes)
+
+              span = tracer.start_span(verb.to_s, attributes: attributes, kind: :client, start_timestamp: start_time)
```
kaylareopelle: If you remove the `.to_s` in `dup`, it should be removed here too.

ericmustin (approved):

instrumentation/httpx/lib/opentelemetry/instrumentation/httpx/dup/plugin.rb
hannahramadan: Removed in https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/1589/commits/f9cbcc89a36e8c33dc2189a24294e8dea167a944

instrumentation/httpx/lib/opentelemetry/instrumentation/httpx/stable/plugin.rb
hannahramadan: removed in https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/1589/commits/f9cbcc89a36e8c33dc2189a24294e8dea167a944

instrumentation/httpx/lib/opentelemetry/instrumentation/httpx/stable/plugin.rb
hannahramadan: We decided to remove `http.host` from stable instrumentations, since it is not a part of stable conventions. https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/1589/commits/f9cbcc89a36e8c33dc2189a24294e8dea167a944

kaylareopelle (approved):
🎉

PR #1592, opened 2025-07-11
feat: Faraday semantic convention stability opt in

hannahramadan:


instrumentation/faraday/lib/opentelemetry/instrumentation/faraday/middlewares/dup/tracer_middleware.rb
```diff
@@ -0,0 +1,95 @@
...
+              http_method = HTTP_METHODS_SYMBOL_TO_STRING[env.method]
+              config = Faraday::Instrumentation.instance.config
+
+              attributes = span_creation_attributes(
+                http_method: http_method, url: env.url, config: config
+              )
+
+              OpenTelemetry::Common::HTTP::ClientContext.with_attributes(attributes) do |attrs, _|
+                tracer.in_span(
+                  http_method.to_s, attributes: attrs, kind: config.fetch(:span_kind)
```
kaylareopelle: I don't think `to_s` is needed here. The value gets converted to a string on L31 when `http_method` is assigned.

instrumentation/faraday/lib/opentelemetry/instrumentation/faraday/middlewares/dup/tracer_middleware.rb
```diff
@@ -0,0 +1,95 @@
...
+            end
+
+            private
+
+            def span_creation_attributes(http_method:, url:, config:)
+              attrs = {
+                'http.method' => http_method,
+                'http.request.method' => http_method,
+                'http.url' => OpenTelemetry::Common::Utilities.cleanse_url(url.to_s),
+                'url.full' => OpenTelemetry::Common::Utilities.cleanse_url(url.to_s),
```
kaylareopelle: When reviewing this PR it stood out to me that we're calling the `cleanse_url` method twice. Since this is `dup`, I think it's reasonable for people to expect a little extra overhead. 

Do you have a preference for keeping the double method call?

(this method is also called in Excon and RestClient, so if a change is made here it might be helpful to change the others for consistency)

instrumentation/faraday/lib/opentelemetry/instrumentation/faraday/middlewares/dup/tracer_middleware.rb
```diff
@@ -0,0 +1,95 @@
...
+                'faraday.adapter.name' => app.class.name
+              }
+              attrs['net.peer.name'] = url.host if url.host
+              attrs['server.address'] = url.host if url.host
```
kaylareopelle: nit: 
```suggestion
              if url.host
                attrs['net.peer.name'] = url.host
                attrs['server.address'] = url.host
              end
```

I didn't notice/look for this in the other dups so we might not need to change anything.

instrumentation/faraday/lib/opentelemetry/instrumentation/faraday/middlewares/dup/tracer_middleware.rb
hannahramadan: Cleaned up ◡̈ https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/1592/commits/14dc0896231d06c68b65ce42b182da3c3df2a280#diff-9427e9a90e18e668f165a84c6d0879679b9e3cd5d8572130e51674cefad7e7e2R62

I'll check out Excon and RestClient

instrumentation/faraday/lib/opentelemetry/instrumentation/faraday/middlewares/dup/tracer_middleware.rb
hannahramadan: Thank you - Fixed https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/1592/commits/14dc0896231d06c68b65ce42b182da3c3df2a280#diff-9427e9a90e18e668f165a84c6d0879679b9e3cd5d8572130e51674cefad7e7e2R40

instrumentation/faraday/lib/opentelemetry/instrumentation/faraday/middlewares/dup/tracer_middleware.rb
hannahramadan: Excon https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/1569/commits/76bc2bcf314108ef52248fabbf84759d7fb97884
RestClient - this PR has been already merged, but I can open up a new PR for this change if we'd like. What do you think @kaylareopelle

kaylareopelle: @hannahramadan - If it's not too much trouble, let's open up a new PR for RestClient to keep things consistent. Thank you!

kaylareopelle (approved):

PR #1594, opened 2025-07-14
feat: Rack semantic stability opt in

hannahramadan:


instrumentation/grape/example/trace_demonstration.rb
```diff
@@ -43,8 +43,11 @@ class ExampleAPI < Grape::API
 # Set up fake Rack application
 builder = Rack::Builder.app do
   # Integration is automatic in web frameworks but plain Rack applications require this line.
+  # - middleware_args_old to emit old HTTP semantic conventions
+  # - middleware_args_stable to emit stable HTTP semantic conventions
+  # - middleware_args_dup to emit both old and stable HTTP semantic conventions
   # Enable it in your config.ru.
-  use *OpenTelemetry::Instrumentation::Rack::Instrumentation.instance.middleware_args
+  use *OpenTelemetry::Instrumentation::Rack::Instrumentation.instance.middleware_args_old
```
hannahramadan: Updating this Grape example to patch the old module.

instrumentation/sinatra/lib/opentelemetry/instrumentation/sinatra/instrumentation.rb
```diff
@@ -51,7 +51,19 @@ class Instrumentation < OpenTelemetry::Instrumentation::Base
...
+          if config[:install_rack]
+            if values.include?('http/dup')
+              app.use(*OpenTelemetry::Instrumentation::Rack::Instrumentation.instance.middleware_args_dup)
+            elsif values.include?('http')
+              app.use(*OpenTelemetry::Instrumentation::Rack::Instrumentation.instance.middleware_args_stable)
+            else
+              app.use(*OpenTelemetry::Instrumentation::Rack::Instrumentation.instance.middleware_args_old)
+            end
+          end
+
```
hannahramadan, line 66: Sinatra instrumentation relies on installing Rack middleware.

instrumentation/action_pack/lib/opentelemetry/instrumentation/action_pack/railtie.rb
```diff
@@ -12,10 +12,25 @@ class Railtie < ::Rails::Railtie
...
+            app.middleware.insert_before(
+              0,
+              *OpenTelemetry::Instrumentation::Rack::Instrumentation.instance.middleware_args_stable
+            )
+          else
+            app.middleware.insert_before(
+              0,
+              *OpenTelemetry::Instrumentation::Rack::Instrumentation.instance.middleware_args_old
+            )
+          end
```
kaylareopelle: I get that this is just temporary, so feel free to ignore me, but this is a little hard for me to read. 

Does it work to pick the instance in an if/else statement and then call  the `insert_before` only once? 

```suggestion
          middleware_args = if values.include?('http/dup')
             OpenTelemetry::Instrumentation::Rack::Instrumentation.instance.middleware_args_dup
          elsif values.include?('http')
            OpenTelemetry::Instrumentation::Rack::Instrumentation.instance.middleware_args_stable
          else
            OpenTelemetry::Instrumentation::Rack::Instrumentation.instance.middleware_args_old
          end

           app.middleware.insert_before(
             0,
             *middleware_args
           )
```

kaylareopelle: (You might want to check this outside of GH too, I have no idea if my spacing, etc. is correct in the suggestion 😅)

instrumentation/grape/example/trace_demonstration.rb
kaylareopelle: Could someone use the env var with the example if they wanted to?

instrumentation/grape/test/test_helper.rb
```diff
@@ -41,7 +41,17 @@ def unsubscribe
...
+    values = stability_opt_in.split(',').map(&:strip)
+
+    if values.include?('http/dup')
+      use(*OpenTelemetry::Instrumentation::Rack::Instrumentation.instance.middleware_args_dup)
+    elsif values.include?('http')
+      use(*OpenTelemetry::Instrumentation::Rack::Instrumentation.instance.middleware_args_stable)
+    else
+      use(*OpenTelemetry::Instrumentation::Rack::Instrumentation.instance.middleware_args_old)
+    end
+
```
kaylareopelle: Since this is a test_helper, does it make sense to update the grape appraisal to use the different env vars?

instrumentation/sinatra/example/config.ru
```diff
@@ -47,5 +47,5 @@ class App < Sinatra::Base
   end
 end
 
-use(*OpenTelemetry::Instrumentation::Rack::Instrumentation.instance.middleware_args)
+use(*OpenTelemetry::Instrumentation::Rack::Instrumentation.instance.middleware_args_old)
```
kaylareopelle: What do you think about adding a comment here to explain what's going on?

instrumentation/sinatra/lib/opentelemetry/instrumentation/sinatra/instrumentation.rb
```diff
@@ -51,7 +51,19 @@ class Instrumentation < OpenTelemetry::Instrumentation::Base
         end
 
         def install_middleware(app)
-          app.use(*OpenTelemetry::Instrumentation::Rack::Instrumentation.instance.middleware_args) if config[:install_rack]
+          stability_opt_in = ENV.fetch('OTEL_SEMCONV_STABILITY_OPT_IN', '')
+          values = stability_opt_in.split(',').map(&:strip)
```
kaylareopelle: Since we only need these variables `if config[:install_rack]` is true, what do you think about moving them inside the `if` block?

instrumentation/rack/README.md
```diff
@@ -101,3 +101,19 @@ The `opentelemetry-instrumentation-rack` gem is distributed under the Apache 2.0
 [community-meetings]: https://github.com/open-telemetry/community#community-meetings
 [slack-channel]: https://cloud-native.slack.com/archives/C01NWKKMKMY
 [discussions-url]: https://github.com/open-telemetry/opentelemetry-ruby/discussions
+
+## HTTP semantic convention stability
```
kaylareopelle, line 105: What do you think about adding something about SemConv stability to the READMEs for the libraries that depend on Rack?

instrumentation/rack/lib/opentelemetry/instrumentation/rack/middlewares/dup/tracer_middleware.rb
```diff
@@ -0,0 +1,209 @@
...
+
+            def finish_span(context)
+              OpenTelemetry::Trace.current_span(context).finish if context
+            end
+
+            def tracer
+              OpenTelemetry::Instrumentation::Rack::Instrumentation.instance.tracer
+            end
+
+            def request_span_attributes(env:)
```
kaylareopelle, line 126: It looks like the user agent attribute was accidentally removed. In the stable version, the key is `user_agent.original`

```ruby
attributes['http.user_agent'] = env['HTTP_USER_AGENT'] if env['HTTP_USER_AGENT']
```

instrumentation/rack/lib/opentelemetry/instrumentation/rack/middlewares/dup/tracer_middleware.rb
```diff
@@ -0,0 +1,209 @@
...
+            #
+            # see http://github.com/open-telemetry/opentelemetry-specification/pull/416/files
+            def create_request_span_name(request_uri_or_path_info, env)
+              # NOTE: dd-trace-rb has implemented 'quantization' (which lowers url cardinality)
+              #       see Datadog::Quantization::HTTP.url
+
+              if (implementation = config[:url_quantization])
+                implementation.call(request_uri_or_path_info, env)
+              else
+                env['REQUEST_METHOD'].to_s
```
kaylareopelle: I don't think we need to call `to_s` on this, but I could be wrong. It might be worth looking at Rack's code to see what `'REQUEST_METHOD'` is allowed to be.

instrumentation/rack/lib/opentelemetry/instrumentation/rack/middlewares/dup/event_handler.rb
```diff
@@ -0,0 +1,277 @@
...
+                'http.host' => env['HTTP_HOST'] || 'unknown',
+                'http.scheme' => env['rack.url_scheme'],
+                'http.target' => env['QUERY_STRING'].empty? ? env['PATH_INFO'] : "#{env['PATH_INFO']}?#{env['QUERY_STRING']}",
+                'http.request.method' => env['REQUEST_METHOD'],
+                'url.scheme' => env['rack.url_scheme'],
+                'url.path' => env['PATH_INFO']
+              }
+
+              attributes['url.query'] = env['QUERY_STRING'] unless env['QUERY_STRING'].empty?
+              attributes['http.user_agent'] = env['HTTP_USER_AGENT'] if env['HTTP_USER_AGENT']
```
kaylareopelle: In the stable conventions, I believe this is `user_agent.original`

instrumentation/rack/lib/opentelemetry/instrumentation/rack/middlewares/dup/event_handler.rb
```diff
@@ -0,0 +1,277 @@
...
+                request.env,
+                context: context,
+                getter: OpenTelemetry::Common::Propagation.rack_env_getter
+              )
+            end
+
+            def request_span_attributes(env)
+              attributes = {
+                'http.method' => env['REQUEST_METHOD'],
+                'http.host' => env['HTTP_HOST'] || 'unknown',
```
kaylareopelle, line 193: Do you think this is 1:1 with `server.address`? 

https://opentelemetry.io/docs/specs/semconv/non-normative/http-migration/#http-server-span-attributes

instrumentation/rack/lib/opentelemetry/instrumentation/rack/middlewares/stable/event_handler.rb
```diff
@@ -0,0 +1,271 @@
...
+            end
+
+            def untraced_request?(env)
+              return true if untraced_endpoints.include?(env['PATH_INFO'])
+              return true if untraced_requests&.call(env)
+
+              false
+            end
+
+            # https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/data-http.md#name
```
kaylareopelle:
```suggestion
            # https://opentelemetry.io/docs/specs/semconv/http/http-spans/#name
```

instrumentation/rack/lib/opentelemetry/instrumentation/rack/middlewares/stable/event_handler.rb
```diff
@@ -0,0 +1,271 @@
...
+            #
+            # recommendation: span.name(s) should be low-cardinality (e.g.,
+            # strip off query param value, keep param name)
+            #
+            # see http://github.com/open-telemetry/opentelemetry-specification/pull/416/files
```
kaylareopelle:
```suggestion
```

instrumentation/rack/lib/opentelemetry/instrumentation/rack/middlewares/stable/event_handler.rb
```diff
@@ -0,0 +1,271 @@
...
+              return true if untraced_endpoints.include?(env['PATH_INFO'])
+              return true if untraced_requests&.call(env)
+
+              false
+            end
+
+            # https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/data-http.md#name
+            #
+            # recommendation: span.name(s) should be low-cardinality (e.g.,
+            # strip off query param value, keep param name)
```
kaylareopelle, line 164: I think this summary is pretty close to what we have today; though the param convo is in the PR linked below, but not the stable spec

instrumentation/rack/lib/opentelemetry/instrumentation/rack/middlewares/stable/event_handler.rb
```diff
@@ -0,0 +1,271 @@
...
+            def request_span_attributes(env)
+              attributes = {
+                'http.request.method' => env['REQUEST_METHOD'],
+                'http.host' => env['HTTP_HOST'] || 'unknown',
+                'url.scheme' => env['rack.url_scheme'],
+                'url.path' => env['PATH_INFO']
+              }
+
+              attributes['url.query'] = env['QUERY_STRING'] unless env['QUERY_STRING'].empty?
+              attributes['http.user_agent'] = env['HTTP_USER_AGENT'] if env['HTTP_USER_AGENT']
```
kaylareopelle: https://opentelemetry.io/docs/specs/semconv/non-normative/http-migration/#migrating-from--v1180 
```suggestion
              attributes['user_agent.original'] = env['HTTP_USER_AGENT'] if env['HTTP_USER_AGENT']
```

instrumentation/rack/lib/opentelemetry/instrumentation/rack/middlewares/stable/tracer_middleware.rb
```diff
@@ -0,0 +1,206 @@
...
+            end
+
+            def tracer
+              OpenTelemetry::Instrumentation::Rack::Instrumentation.instance.tracer
+            end
+
+            def request_span_attributes(env:)
+              attributes = {
+                'http.request.method' => env['REQUEST_METHOD'],
+                'http.host' => env['HTTP_HOST'] || 'unknown',
```
kaylareopelle: What do you think about making this `server.address`? https://opentelemetry.io/docs/specs/semconv/non-normative/http-migration/#http-server-span-attributes

instrumentation/rack/lib/opentelemetry/instrumentation/rack/middlewares/stable/tracer_middleware.rb
```diff
@@ -0,0 +1,206 @@
...
+                'url.scheme' => env['rack.url_scheme'],
+                'url.path' => env['PATH_INFO']
+              }
+
+              attributes['url.query'] = env['QUERY_STRING'] unless env['QUERY_STRING'].empty?
+              attributes['http.user_agent'] = env['HTTP_USER_AGENT'] if env['HTTP_USER_AGENT']
```
kaylareopelle:
```suggestion
              attributes['user_agent.original'] = env['HTTP_USER_AGENT'] if env['HTTP_USER_AGENT']
```

instrumentation/rack/lib/opentelemetry/instrumentation/rack/middlewares/stable/tracer_middleware.rb
```diff
@@ -0,0 +1,206 @@
...
+            #
+            # see http://github.com/open-telemetry/opentelemetry-specification/pull/416/files
+            def create_request_span_name(request_uri_or_path_info, env)
+              # NOTE: dd-trace-rb has implemented 'quantization' (which lowers url cardinality)
+              #       see Datadog::Quantization::HTTP.url
+
+              if (implementation = config[:url_quantization])
+                implementation.call(request_uri_or_path_info, env)
+              else
+                env['REQUEST_METHOD'].to_s
```
kaylareopelle: If this gets changed in dup, should change it here too

instrumentation/rack/test/opentelemetry/instrumentation/rack/instrumentation_test.rb
```diff
@@ -12,6 +12,8 @@
   let(:config) { {} }
 
   before do
+    skip unless ENV['BUNDLE_GEMFILE'].include?('old')
```
kaylareopelle: Is this a situation where we could test the dup and stable without getting into problems with patching? Thinking about what kind of testing would be helpful for the `middleware_args_*` methods

kaylareopelle (changes requested):
Whew! This one covers a lot of ground. Great work catching the changes needed in the other instrumentations. Thanks for taking this on. 

I think the PR description may need a little tweaking for the HTTP server spans that Rack produces. 
> HTTP server span names SHOULD be {method} {http.route} if there is a (low-cardinality) http.route available (see below for the exact definition of the [{method}](https://github.com/open-telemetry/semantic-conventions/blob/v1.23.1/docs/http/http-spans.md#method-placeholder) placeholder).
>
>If there is no (low-cardinality) http.route available, HTTP server span names SHOULD be [{method}](https://github.com/open-telemetry/semantic-conventions/blob/v1.23.1/docs/http/http-spans.md#method-placeholder).

Happy to go over any of the comments if you have questions. Mostly, I was consulting the server span parts of this doc: https://opentelemetry.io/docs/specs/semconv/non-normative/http-migration/#http-server-span-attributes

instrumentation/grape/example/trace_demonstration.rb
hannahramadan: I don't think so. The original example looks like this is a config line that users must add either way!

instrumentation/sinatra/example/config.ru
hannahramadan: Done ◡̈ https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/1594/commits/fe46f3d4407518b6f6f0f8f4d6bf8b182066f51a

instrumentation/sinatra/lib/opentelemetry/instrumentation/sinatra/instrumentation.rb
hannahramadan: great idea! done https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/1594/commits/fe46f3d4407518b6f6f0f8f4d6bf8b182066f51a

instrumentation/rack/lib/opentelemetry/instrumentation/rack/middlewares/dup/tracer_middleware.rb
hannahramadan: Great idea! Looks like Rack calls .to_s, so we should always be getting a string https://github.com/rack/rack/blob/6828a1767e4f3bd535de506d95ff5686b254fa90/lib/rack/method_override.rb#L31

instrumentation/rack/lib/opentelemetry/instrumentation/rack/middlewares/dup/event_handler.rb
hannahramadan: Yes! Updated https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/1594/commits/516883111a79f5cfb74649babd79590f578352c8

instrumentation/rack/lib/opentelemetry/instrumentation/rack/middlewares/dup/tracer_middleware.rb
hannahramadan: Great catch. Added back https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/1594/commits/516883111a79f5cfb74649babd79590f578352c8

instrumentation/rack/lib/opentelemetry/instrumentation/rack/middlewares/stable/tracer_middleware.rb
hannahramadan: changed https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/1594/commits/516883111a79f5cfb74649babd79590f578352c8

instrumentation/rack/lib/opentelemetry/instrumentation/rack/middlewares/stable/tracer_middleware.rb
hannahramadan: made this change https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/1594/commits/516883111a79f5cfb74649babd79590f578352c8

instrumentation/rack/test/opentelemetry/instrumentation/rack/instrumentation_test.rb
hannahramadan: Looks like there weren't any issues with patching and testing! Added files for this test https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/1594/commits/74aa619581785d3913335115d3c8db2f781af0fc

instrumentation/rack/README.md
hannahramadan: Done https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/1594/commits/6dca891bf0b820b58a493d966e9fefb3c80fedce

instrumentation/sinatra/lib/opentelemetry/instrumentation/sinatra/instrumentation.rb
hannahramadan: Added tests for this too https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/1594/commits/6dca891bf0b820b58a493d966e9fefb3c80fedce

instrumentation/action_pack/lib/opentelemetry/instrumentation/action_pack/railtie.rb
ericmustin:
```
F, [2025-07-30T22:25:29.574298 #5831] FATAL -- :   
  RuntimeError (a little hell):
```

Here's a fun Rubocop error message, if that's what you're into. Anyway I think this just needs rubocop run on it, as it's complaining in the test suite.

instrumentation/rack/lib/opentelemetry/instrumentation/rack/instrumentation.rb
```diff
@@ -35,23 +36,62 @@ class Instrumentation < OpenTelemetry::Instrumentation::Base
...
         #   Rack::Builder.new do
-        #     use *OpenTelemetry::Instrumentation::Rack::Instrumenation.instance.middleware_args
+        #     use *OpenTelemetry::Instrumentation::Rack::Instrumenation.instance.middleware_args_old
         #     run lambda { |_arg| [200, { 'Content-Type' => 'text/plain' }, body] }
         #   end
         # @return [Array] consisting of a middleware and arguments used in rack builders
-        def middleware_args
-          if config.fetch(:use_rack_events, false) == true && defined?(OpenTelemetry::Instrumentation::Rack::Middlewares::EventHandler)
-            [::Rack::Events, [OpenTelemetry::Instrumentation::Rack::Middlewares::EventHandler.new]]
+        def middleware_args_old
```
ericmustin: Would this break existing monkey patches of the rack instrumentation, as this is not private? Is that a reasonable concern? I understand we are not 1.x here yet so it's not technically a semver question, but just trying to be mindful around touching rack instrumentation as it's probably got a rather large install base at this point.

ericmustin (changes requested):
I had one open question otherwise lgtm pending rubocop fixes

instrumentation/rack/lib/opentelemetry/instrumentation/rack/instrumentation.rb
hannahramadan: Fantastic call out - it probably would break some stuff! I've aliased the new and old method definitions, as well as added a test https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/1594/commits/72908226f8a8a71135f98e9017900409309ad0d5

instrumentation/action_pack/lib/opentelemetry/instrumentation/action_pack/railtie.rb
hannahramadan: Ran rubocop through and fixed an alignment issue. seems all good now https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/1594/commits/07529050d5e1819ffa7d12de893c242c7c58d93f

instrumentation/grape/test/test_helper.rb
hannahramadan: @kaylareopelle I decided it might be best to revert changes to this test suite, and instead rely on the original method call, `middleware_args`, which is now aliased to `middleware_args_old`. It seemed like a lot of testing of the same idea: that the correct middleware is installed and correct attribute names are used. We test both in Rack and Sinatra ◡̈ This PR is so heavy I'm nervous to keep adding!

kaylareopelle (approved):

ericmustin (approved):

PR #1616, opened 2025-08-06
feat: allow for propagation_style in Sidekiq instrumentation to have dynamic option

### LINKED ISSUE #991: Make Sidekiq `propagation_style` able to be changed on a job-by-job basis
Right now, you can choose to have _all_ jobs connected to their "producer" (enqueuing) spans via a `link` or via a standard parent-child span relationship (`child`) via the global `propagation_style` option. However, that level of granularity is not always sufficient for accurately representing a user-journey as a trace.

For example, let's say we have a web request where a user is uploading an document to convert it from one file type to another. This request handler
1. Enqueues a background job to perform that file conversion
2. Enqueues a separate background job for analytics processing

In terms of a user journey, we only actually care about (1) and not about (2). Ideally, I'd be able to connect the enqueue span to the file processing job performance span with a parent-child relationship and use a link for the other enqueue.

I'm open to implementing this functionality if such a feature would be welcome.

hibachrach:
This still permits static values (and environment variable to configure them) but allows users to opt-in to having a callable/dynamic value for the given option.

Resolves open-telemetry/opentelemetry-ruby-contrib#991

### DIFF
--- a/instrumentation/base/lib/opentelemetry/instrumentation/base.rb
+++ b/instrumentation/base/lib/opentelemetry/instrumentation/base.rb
@@ -4,6 +4,8 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 
+require 'opentelemetry/instrumentation/dynamic_validator'
+
 module OpenTelemetry
   module Instrumentation
     # The Base class holds all metadata and configuration for an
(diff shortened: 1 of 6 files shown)


hibachrach:
FYI, I opted to change the config handling code to allow for more complex config value types, but happy to add a new callable config option that would take precedent over the current enum value instead and revert all of that. Not sure which would be preferred.

PR #1628, opened 2025-08-12
feat: Add HTTPX instrumentation to all

kaylareopelle:
Now, HTTPX instrumentation will be automatically applied if the HTTPX gem is installed in a user's environment.


ericmustin (approved):
lgtm

PR #1634, opened 2025-08-14
fix: improve Ethon exception handling

mspletz:
## What

This pull request improves error handling for Ethon instrumentation by ensuring that spans are properly closed and marked as errors when an exception occurs during the HTTP request execution, before the `complete` method is called.  

## Why

Today, we've observed that some `ethon.client` http POST spans from this otel library continue unexpectedly for a long time (up to 13 hours). This has led us to believe that there are occasions where the `perform` method ran into an unhandled exception and never properly made it to the `complete` method to close the span.


kaylareopelle (approved):
Thanks, @mspletz!

PR #1637, opened 2025-08-16
feat: Puma plugin to gracefully shut down providers

bdewater-thatch:
Following discussion at https://github.com/open-telemetry/opentelemetry-ruby/discussions/1357 and [on Slack](https://cloud-native.slack.com/archives/C01NWKKMKMY/p1754426751777279) I decided to take a moment and write a Puma plugin to shutdown tracer/meter/logger providers and make sure nothing is lost before the process exits. Given that puma is the default for newly generated Rails apps, this should make things "just work" a little better.

First time contributor, I'm sure I'm missing things :) any feedback welcome!

### DIFF
--- a/instrumentation/puma/Appraisals
+++ b/instrumentation/puma/Appraisals
@@ -1,0 +1,15 @@
+# frozen_string_literal: true
+
+# Copyright The OpenTelemetry Authors
+#
+# SPDX-License-Identifier: Apache-2.0
+
+## TODO: Include the supported version to be tested here.
+## Example:
+# appraise 'rack-2.1' do
+#   gem 'rack', '~> 2.1.2'
+# end
+
+# appraise 'rack-2.0' do
+#   gem 'rack', '2.0.8'
+# end
--- a/instrumentation/puma/Appraisals
+++ b/instrumentation/puma/Appraisals
@@ -4,12 +4,10 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 
-## TODO: Include the supported version to be tested here.
-## Example:
-# appraise 'rack-2.1' do
-#   gem 'rack', '~> 2.1.2'
-# end
+appraise 'puma-7.0' do
+  gem 'puma', '~> 7.0'
+end
 
-# appraise 'rack-2.0' do
-#   gem 'rack', '2.0.8'
-# end
+appraise 'puma-6.0' do
+  gem 'puma', '~> 6.0'
+end
--- a/instrumentation/puma/Gemfile
+++ b/instrumentation/puma/Gemfile
@@ -1,0 +1,27 @@
+# frozen_string_literal: true
+
+# Copyright The OpenTelemetry Authors
+#
+# SPDX-License-Identifier: Apache-2.0
+
+source 'https://rubygems.org'
+
+# DO NOT ADD DEPENDENCIES HERE!
+# Please declare a minimum development dependency in the gemspec,
+# then target specific versions in the Appraisals file.
+
+gemspec
+
+group :test do
+  gem 'opentelemetry-instrumentation-base', path: '../base'
+  gem 'appraisal', '~> 2.5'
+  gem 'bundler', '~> 2.4'
+  gem 'minitest', '~> 5.0'
+  gem 'opentelemetry-sdk', '~> 1.0'
+  gem 'opentelemetry-test-helpers', '~> 0.3'
+  gem 'rake', '~> 13.0'
+  gem 'rubocop', '~> 1.79.1'
+  gem 'rubocop-performance', '~> 1.25.0'
+  gem 'simplecov', '~> 0.17.1'
+  gem 'yard', '~> 0.9'
+end
--- a/instrumentation/puma/Gemfile
+++ b/instrumentation/puma/Gemfile
@@ -6,10 +6,6 @@
 
 source 'https://rubygems.org'
 
-# DO NOT ADD DEPENDENCIES HERE!
-# Please declare a minimum development dependency in the gemspec,
-# then target specific versions in the Appraisals file.
-
 gemspec
 
 group :test do
--- a/instrumentation/puma/lib/opentelemetry/instrumentation/puma/plugin.rb
+++ b/instrumentation/puma/lib/opentelemetry/instrumentation/puma/plugin.rb
@@ -1,0 +1,36 @@
+# frozen_string_literal: true
+
+# Copyright The OpenTelemetry Authors
+#
+# SPDX-License-Identifier: Apache-2.0
+
+require 'puma/plugin'
+
+module OpenTelemetry
+  module Instrumentation
+    module Puma
+      # The Puma plugin for OpenTelemetry
+      class Plugin < ::Puma::Plugin
+        ::Puma::Plugins.register('opentelemetry', self)
+
+        def start(launcher)
+          launcher.events.on_stopped do
+            shutdown_providers
+          end
+
+          launcher.events.on_restart do
+            shutdown_providers
+          end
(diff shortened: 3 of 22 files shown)


instrumentation/puma/Appraisals
```diff
@@ -0,0 +1,15 @@
...
+
+## TODO: Include the supported version to be tested here.
+## Example:
+# appraise 'rack-2.1' do
+#   gem 'rack', '~> 2.1.2'
+# end
+
+# appraise 'rack-2.0' do
+#   gem 'rack', '2.0.8'
+# end
```
kaylareopelle: Calling out this TODO! 🙂 We usually like to test the latest version of the gem and at least one major version pinned. If those end up being the same version, feel free to just use the latest.

instrumentation/puma/Gemfile
```diff
@@ -0,0 +1,27 @@
...
+# Copyright The OpenTelemetry Authors
+#
+# SPDX-License-Identifier: Apache-2.0
+
+source 'https://rubygems.org'
+
+# DO NOT ADD DEPENDENCIES HERE!
+# Please declare a minimum development dependency in the gemspec,
+# then target specific versions in the Appraisals file.
+
```
kaylareopelle: Whoops, this isn't accurate anymore. I updated the generator in #1641. Let's remove it.
```suggestion
```

instrumentation/puma/opentelemetry-instrumentation-puma.gemspec
```diff
@@ -0,0 +1,38 @@
...
+
+  spec.files = Dir.glob('lib/**/*.rb') +
+               Dir.glob('*.md') +
+               ['LICENSE', '.yardopts']
+  spec.require_paths = ['lib']
+  spec.required_ruby_version = ">= #{File.read(File.expand_path('../../gemspecs/RUBY_REQUIREMENT', __dir__))}"
+
+  spec.add_dependency 'opentelemetry-api', '~> 1.4.0'
+  spec.add_dependency 'opentelemetry-instrumentation-base', '~> 0.23.0'
+  spec.add_dependency 'puma', '~> 6.0'
```
kaylareopelle: We don't like to require the gem we're instrumenting in the Gemspec. This interferes with dependency detection when using `opentelemetry-instrumentation-all`. If you'd like to set a minimum version for Puma, the `puma/instrumentation.rb` file is a better place! 

We use the `compatible` method to check versions. [This file is a good example of how to use it](https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/717f1c39bf231d8901c7b9d3d62e0ce20a955819/instrumentation/action_pack/lib/opentelemetry/instrumentation/action_pack/instrumentation.rb).

```suggestion
```

instrumentation/puma/lib/opentelemetry/instrumentation/puma/plugin.rb
```diff
@@ -0,0 +1,36 @@
...
+            shutdown_providers
+          end
+        end
+
+        private
+
+        def shutdown_providers
+          OpenTelemetry.tracer_provider.shutdown
+          OpenTelemetry.meter_provider.shutdown if OpenTelemetry.respond_to?(:meter_provider)
+          OpenTelemetry.logger_provider.shutdown if OpenTelemetry.respond_to?(:logger_provider)
```
kaylareopelle, line 41: Love the inclusion of metrics and logs! 🎉

kaylareopelle (commented):
Thank you for your contribution, @bdewater-thatch! 😄 I took a first pass on this PR and left a few comments related to the repo's best practices. Overall, this looks good! I still want to do more testing and thinking before I approve. 

The main thing that stands out to me, is that this instrumentation isn't emitting any telemetry. (I apologize, [we started talking Puma stats in Slack](https://cloud-native.slack.com/archives/C01NWKKMKMY/p1750441808338309) ages ago and I lost track of things 😅). 

I'm not sure if that's a problem for the initial release. Perhaps calling out what the gem does a little more in the README is enough for now, since the shutdown problem already plagues users. What do you think?

cc'ing the folks in the [on_shutdown slack thread](https://cloud-native.slack.com/archives/C01NWKKMKMY/p1754426751777279): @arielvalentin, @hibachrach, @wsmoak

instrumentation/puma/lib/opentelemetry/instrumentation/puma/plugin.rb
bdewater-thatch: It's a bit forward looking but figured it wouldn't hurt :) 

I forgot to ask but was wondering if you think this should include a configurable timeout option, and if so what the default should be (`nil`?). We host our apps on Render, which has a [30s timeout by default](https://render.com/docs/deploys#graceful-shutdown) for graceful shutdown but is configurable via API/IaC to be up to 300s.

kaylareopelle: Ooh, that's a good question. I'm not sure. Thinking out loud here: 

Timeouts could technically be set through [`OTEL_BSP_EXPORT_TIMEOUT`](https://opentelemetry.io/docs/specs/otel/configuration/sdk-environment-variables/#batch-span-processor) since the tracer provider will iterate through all the span processors on shutdown and they'll use that setting. However, the tracer provider's `shutdown` method has its own timeout that wraps the processors, which would have a different value. 

The API doesn't have a `shutdown` method for the TracerProvider, and one isn't mentioned in the spec. The SDK defines the `shutdown` method. Instrumentation should be implementation-agnostic, so we don't include the SDK as a dependency in instrumentation. I wonder if we need to add a no-op shutdown API to allow this to work? And if that's even spec compliant? 🤔

scbjans: I'd like to add that there is indeed no `shutdown` method in the API; therefore we need to guard for when someone uses [the OTEL_SDK_DISABLED envvar](https://opentelemetry.io/docs/specs/otel/configuration/sdk-environment-variables/#general-sdk-configuration).

robbkidd: 🤔 Could `try()` help here maybe with a warning along the lines of "do you have an OTel SDK loaded?" logged if there isn't a `shutdown()` method implemented on the configured tracer_provider?

bdewater-thatch: I'm not following the "shutdown method in the API" discussion. I've looked at what Sidekiq does (which is seemingly the only example of instrumentation calling shutdown) and it doesn't do anything additional checks? https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/2edff11f55f89cc760695f8bb6422950c492437d/instrumentation/sidekiq/lib/opentelemetry/instrumentation/sidekiq/instrumentation.rb#L134-L136

bdewater-thatch: I found [this article](https://signoz.io/comparisons/opentelemetry-api-vs-sdk/) to help me understand the difference between API and SDK. I've added an early return based on [this example](https://github.com/open-telemetry/opentelemetry-ruby/blob/a6ce85bc55e06437814c5c13c477b6273d8b9566/sdk/lib/opentelemetry/sdk.rb#L64) - since that already logs a warning I don't think we need another log here?

xuan-cao-swi:
> The main thing that stands out to me, is that this instrumentation isn't emitting any telemetry.

I have the same concern. The plugin could be in a different folder (e.g. `/plugins`, similar to resource detector), and then user can install them independently as they needed.

PR #1645, opened 2025-08-20
fix!: check if span has the attributes method to avoid internal error

xuan-cao-swi:
### Description

In rare cases, the span may be a no-op span because it is a [non-recording span](https://github.com/open-telemetry/opentelemetry-ruby/blob/opentelemetry-sdk/v1.8.1/sdk/lib/opentelemetry/sdk/trace/tracer_provider.rb#L166). If the span attempts to access its attributes, it will raise an error. A check has been added to prevent this internal error.


instrumentation/aws_lambda/lib/opentelemetry/instrumentation/aws_lambda/wrap.rb
```diff
@@ -43,7 +43,7 @@ def wrap_lambda(event:, context:, handler:, flush_timeout: DEFAULT_FLUSH_TIMEOUT
               begin
                 response = yield
 
-                unless span.attributes.key?(OpenTelemetry::SemanticConventions::Trace::HTTP_STATUS_CODE)
+                if span.respond_to?(:attributes) && !span.attributes.key?(OpenTelemetry::SemanticConventions::Trace::HTTP_STATUS_CODE)
```
robertlaurin: Probably can just check if the span is recording, right?
```suggestion
                if span.recording? && !span.attributes.key?(OpenTelemetry::SemanticConventions::Trace::HTTP_STATUS_CODE)
```

robertlaurin: How is this span attribute getting set anyways? Instrumentation should really not be making any decisions based of the contents of the spans attributes.

xuan-cao-swi: aws lambda instrumentation attribute is set through `otel_attributes(event, context)` from [in_span](https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/opentelemetry-instrumentation-aws_lambda/v0.3.0/instrumentation/aws_lambda/lib/opentelemetry/instrumentation/aws_lambda/wrap.rb#L42); and then it will add the attribute `HTTP_STATUS_CODE`.

> Instrumentation should really not be making any decisions based of the contents of the spans attributes.

This part only sets attributes like HTTP_STATUS_CODE as part of span metadata.

The issue is that if the span is not recording, it won't have the attributes method, which can raise an error. This error could overshadow the user's original Lambda function error, causing confusion.

kaylareopelle:
Hi @spongenee, thanks for reaching out! Have you given this branch a try? Does it resolve the problem for you?

xuan-cao-swi:
Hi @spongenee , one way (very hacky way) is that you can edit this file directly, and then pack everything into zip file. With the zip file, you can upload to layer.

After download the [otel-lambda](https://github.com/open-telemetry/opentelemetry-lambda), you can navigate to the path `cd ruby/src`, and then run `sam build -u -t template.yml` (change the ruby version in template.yaml and `otel/layer/Makefile` if needed). You will have the `.aws-sam` folder. That folder is basically your layer content, and you can modify code there (e.g. instrumentation-aws_lambda). Finally, you can run `./zip_ruby_layer.sh` to get the zip file `opentelemetry-ruby-layer.zip`.

kaylareopelle (approved):

kaylareopelle:
@spongenee - That makes sense! Apologies for missing that earlier!

kaylareopelle (commented):
@xuan-cao-swi I think this looks good and seems like it will solve the problem. Could you add a regression test? Since this fix is really breaking things for users, I'm fine to get it out into the world without the test, so you can add it later on.

PR #1653, opened 2025-08-29
feat: Trilogy: introduce record_exception setting

kirs:
In apps that use advanced exception tracking, having raw Trilogy error on the span can be misleading because it's likely handled by ActiveRecord. Surfacing it as an actual exception introduces some misdirection here.

I'd like to suggest that we make this behaviour configurable via `record_exception` option. It would be set to `true` by default, preserving existing behaviour.

@robertlaurin @fbogsany

### DIFF
--- a/instrumentation/trilogy/lib/opentelemetry/instrumentation/trilogy/patches/client.rb
+++ b/instrumentation/trilogy/lib/opentelemetry/instrumentation/trilogy/patches/client.rb
@@ -19,7 +19,8 @@ module OpenTelemetry
             tracer.in_span(
               'connect',
               attributes: client_attributes.merge!(OpenTelemetry::Instrumentation::Trilogy.attributes),
-              kind: :client
+              kind: :client,
+              record_exception: config[:record_exception]
             ) do
               super
             end
@@ -29,7 +30,8 @@ module OpenTelemetry
             tracer.in_span(
               'ping',
               attributes: client_attributes.merge!(OpenTelemetry::Instrumentation::Trilogy.attributes),
-              kind: :client
+              kind: :client,
+              record_exception: config[:record_exception]
             ) do
               super
             end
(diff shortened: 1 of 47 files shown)


ericmustin (commented):
Looks fine to me, happy to approve , We have some prior patterns like this in other instrumentation(ie rack/rails), and this seems like a useful knob to make the instrumentation more flexible. I am not an active trilogy user at this time so I defer to the folks working with that gem daily.

instrumentation/trilogy/lib/opentelemetry/instrumentation/trilogy/patches/client.rb
```diff
@@ -16,7 +16,7 @@ module Client
           def initialize(options = {})
             @connection_options = options # This is normally done by Trilogy#initialize
 
-            tracer.in_span(
+            in_span(
```
robertlaurin: @fbogsany suggested we improve the `in_span` convenience method to allow for the desired behaviour so all that is required in the PR is to update the calls to in span.

Related: https://github.com/open-telemetry/opentelemetry-ruby/pull/1911

kaylareopelle: @kirs - what do you think about updating this PR to use the changes to the `in_span` method from open-telemetry/opentelemetry-ruby#1911?

robertlaurin:
@kirs sorry for the slow turnaround but if you update the gemspec dependency of the `opentelemetry-api` to `1.7.0` you can make use of the kwarg added in https://github.com/open-telemetry/opentelemetry-ruby/pull/1911

robertlaurin (approved):

instrumentation/trilogy/opentelemetry-instrumentation-trilogy.gemspec
```diff
@@ -25,7 +25,7 @@ Gem::Specification.new do |spec|
   spec.require_paths = ['lib']
   spec.required_ruby_version = ">= #{File.read(File.expand_path('../../gemspecs/RUBY_REQUIREMENT', __dir__))}"
 
-  spec.add_dependency 'opentelemetry-api', '~> 1.0'
+  spec.add_dependency 'opentelemetry-api', '~> 1.7.0'
```
robertlaurin:
```suggestion
  spec.add_dependency 'opentelemetry-api', '~> 1.7'
```

fbogsany (approved):

PR #1660, opened 2025-09-05
fix: unify rack middleware_args

sergiofbsilva:
Standalone rack applications need to initialize middlewares explicitly:

```ruby
Rack::Builder.new do
	use *OpenTelemetry::Instrumentation::Rack::Instrumentation.instance.middleware_args
	run lambda { |_arg| [200, { 'Content-Type' => 'text/plain' }, body] }
 end
```
However this doesn't take into account the new `OTEL_SEMCONV_STABILITY_OPT_IN` environment variable.

This PR ensures rack middleware_args are selected consistently based on `OTEL_SEMCONV_STABILITY_OPT_IN` environment variable.


ericmustin:
Thanks for opening this @sergiofbsilva . We chatted on this briefly in SIG, just wanted to clarify

- To confirm, the issue here that you're using rack as a standalone app, and the env var `OTEL_SEMCONV_STABILITY_OPT_IN`  isn't being respected because it isn't being by the evaluated by the `middleware_args` helper method?

cc @hannahramadan

sergiofbsilva:
> Thanks for opening this @sergiofbsilva . We chatted on this briefly in SIG, just wanted to clarify
> 
> * To confirm, the issue here that you're using rack as a standalone app, and the env var `OTEL_SEMCONV_STABILITY_OPT_IN`  isn't being respected because it isn't being by the evaluated by the `middleware_args` helper method?
> 
> cc @hannahramadan

Yes, that's right. We want to use `middleware_args` without all the logic around `OTEL_SEMCONV_STABILITY_OPT_IN` being on the API client.

instrumentation/sinatra/lib/opentelemetry/instrumentation/sinatra/instrumentation.rb
```diff
@@ -51,19 +51,7 @@ class Instrumentation < OpenTelemetry::Instrumentation::Base
...
-            if values.include?('http/dup')
-              app.use(*OpenTelemetry::Instrumentation::Rack::Instrumentation.instance.middleware_args_dup)
-            elsif values.include?('http')
-              app.use(*OpenTelemetry::Instrumentation::Rack::Instrumentation.instance.middleware_args_stable)
-            else
-              app.use(*OpenTelemetry::Instrumentation::Rack::Instrumentation.instance.middleware_args)
-            end
-          end
-
+          app.use(*OpenTelemetry::Instrumentation::Rack::Instrumentation.instance.middleware_args) if config[:install_rack]
```
hannahramadan, line 54: This all looks so great. Thank you for working on this. 

Two Sinatra tests are using the direct stable/dup methods. It would be great to update those tests to better match the changes you've made by replacing `middleware_args_stable` and `middleware_args_dup` with `middleware_args`.

The files are:
- `sinatra/test/opentelemetry/instrumentation/sinatra_dup_http_test`
- `sinatra/test/opentelemetry/instrumentation/sinatra_stable_http_test`

sergiofbsilva:
@hannahramadan updated the tests, CI was failing, updated branch.

hannahramadan (approved):
🚀

kaylareopelle (approved):
Thanks, @sergiofbsilva! This is much cleaner.

kaylareopelle:
@sergiofbsilva - I'm changing the conventional commit prefix to `fix` so that this triggers a new release. Thanks for your contribution!

PR #1667, opened 2025-09-11
feat: add Render resource detector

bdewater-thatch:
This implements a resource detector for the Render platform. I used the AWS EC2 implementation and [Heroku](https://opentelemetry.io/docs/specs/semconv/resource/cloud-provider/heroku/) conventions as inspiration.

Docs:
- https://render.com/docs/environment-variables
- https://render.com/docs/metrics-streams#universal-properties to compare with the metrics the Render platform can emit

### DIFF
--- a/resources/render/lib/opentelemetry/resource/detector/render.rb
+++ b/resources/render/lib/opentelemetry/resource/detector/render.rb
@@ -1,0 +1,37 @@
+# frozen_string_literal: true
+
+# Copyright The OpenTelemetry Authors
+#
+# SPDX-License-Identifier: Apache-2.0
+
+module OpenTelemetry
+  module Resource
+    module Detector
+      # Render contains detect class method for determining Render environment resource attributes
+      module Render
+        extend self
+
+        RESOURCE = OpenTelemetry::SemanticConventions::Resource
+
+        def detect
+          resource_attributes = {}
+          if ENV.fetch('RENDER', false) == 'true'
+            resource_attributes[RESOURCE::CLOUD_PROVIDER] = 'render'
+            resource_attributes['render.is_pull_request'] = ENV.fetch('IS_PULL_REQUEST', 'false')
+            resource_attributes['render.git.branch'] = ENV.fetch('RENDER_GIT_BRANCH', nil)
+            resource_attributes['render.git.repo_slug'] = ENV.fetch('RENDER_GIT_REPO_SLUG', nil)
+            resource_attributes[RESOURCE::SERVICE_INSTANCE_ID] = ENV.fetch('RENDER_INSTANCE_ID', nil)
+            resource_attributes[RESOURCE::SERVICE_NAME] = ENV.fetch('RENDER_SERVICE_NAME', 'unknown_service')
+            resource_attributes[RESOURCE::SERVICE_VERSION] = ENV.fetch('RENDER_GIT_COMMIT', nil)
+
+            resource_attributes.delete_if { |_key, value| value.nil? || value.empty? }
+          else
+            OpenTelemetry.logger.warn('Render resource detector did not detect Render environment')
+          end
+
+          OpenTelemetry::SDK::Resources::Resource.create(resource_attributes)
+        end
+      end
+    end
+  end
+end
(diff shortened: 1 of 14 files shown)


resources/render/lib/opentelemetry/resource/detector/render.rb
```diff
@@ -0,0 +1,37 @@
...
+        extend self
+
+        RESOURCE = OpenTelemetry::SemanticConventions::Resource
+
+        def detect
+          resource_attributes = {}
+          if ENV.fetch('RENDER', false) == 'true'
+            resource_attributes[RESOURCE::CLOUD_PROVIDER] = 'render'
+            resource_attributes['render.is_pull_request'] = ENV.fetch('IS_PULL_REQUEST', 'false')
+            resource_attributes['render.git.branch'] = ENV.fetch('RENDER_GIT_BRANCH', nil)
```
thompson-tomo, line 21: Should be `vcs.ref.head.name`

resources/render/lib/opentelemetry/resource/detector/render.rb
```diff
@@ -0,0 +1,37 @@
...
+            resource_attributes['render.git.repo_slug'] = ENV.fetch('RENDER_GIT_REPO_SLUG', nil)
```
thompson-tomo, line 22: Could we construct a `vcs.repository.url.full` using the slug?

resources/render/lib/opentelemetry/resource/detector/render.rb
```diff
@@ -0,0 +1,37 @@
...
+            resource_attributes[RESOURCE::SERVICE_INSTANCE_ID] = ENV.fetch('RENDER_INSTANCE_ID', nil)
+            resource_attributes[RESOURCE::SERVICE_NAME] = ENV.fetch('RENDER_SERVICE_NAME', 'unknown_service')
+            resource_attributes[RESOURCE::SERVICE_VERSION] = ENV.fetch('RENDER_GIT_COMMIT', nil)
```
thompson-tomo, line 25: Should this not be using a vcs.ref.head.revision attribute instead?

resources/render/lib/opentelemetry/resource/detector/render.rb
bdewater-thatch: Thanks, will change.

resources/render/lib/opentelemetry/resource/detector/render.rb
bdewater-thatch: I think both can apply, eg we deploy from `main` and do not have any other kind of version identifier for our app. Software like GitLab does have a version number and a revision, so I'll change this to allow for that kind of thing.

resources/render/lib/opentelemetry/resource/detector/render.rb
bdewater-thatch: No. Render supports multiple git backend but does not expose this information via an environment variable.

thompson-tomo: In that case I would suggest vcs.repository.slug and get the attribute defined.

resources/render/lib/opentelemetry/resource/detector/render.rb
thompson-tomo: For me usually the service attributes should be driven by the application and not by the build/deployment system. Maybe you could use artifact.name/version to capture these variables.

bdewater-thatch: I see https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/44337aae5f228ac1d938d7a75b67b940fdf066fc/instrumentation/rails/lib/opentelemetry/instrumentation/rails/railtie.rb#L20 does that for service name, makes sense.

thompson-tomo: So looking at what other solutions ie heroku, gcp, cloud foundary do which is capture this info in their own namespace. Hence i would not set the service.name/version attributes via this resource detector but rather introduce with render.app.name & render.app.version to store this.

PR #1672, opened 2025-09-15
Report isolation level on ActiveRecord transaction

kirs:
This is a very useful field if your app is in progress of switching in between isolation levels.


instrumentation/active_record/lib/opentelemetry/instrumentation/active_record/patches/transactions_class_methods.rb
```diff
@@ -18,9 +18,13 @@ class << base
 
           # Contains ActiveRecord::Transactions::ClassMethods to be patched
           module ClassMethods
-            def transaction(...)
-              tracer.in_span('ActiveRecord.transaction', attributes: { 'code.namespace' => name }) do
-                super
+            def transaction(*args, **kwargs, &block)
+              attributes = { 'code.namespace' => name }
```
xuan-cao-swi, line 22: nit: 
```ruby
attributes = {
  'code.namespace' => name,
  'db.transaction_isolation' => kwargs[:isolation]
}.compact!
```

instrumentation/active_record/lib/opentelemetry/instrumentation/active_record/patches/transactions_class_methods.rb
```diff
@@ -18,9 +18,13 @@ class << base
...
+              if kwargs[:isolation]
+                attributes['db.transaction_isolation'] = kwargs[:isolation].to_s
+              end
+              tracer.in_span('ActiveRecord.transaction', attributes: attributes) do
+                super(*args, **kwargs, &block)
```
xuan-cao-swi, line 27: `super` is enough

instrumentation/active_record/lib/opentelemetry/instrumentation/active_record/patches/transactions_class_methods.rb
robertlaurin: @xuan-cao-swi this would potentially set the attributes hash to nil. 


https://ruby-doc.org/3.4.1/Hash.html#method-i-compact-21
>compact! → self or nil
>Returns self with all its nil-valued entries removed (in place):
>Returns nil if no entries were removed.

robertlaurin (approved):

instrumentation/active_record/lib/opentelemetry/instrumentation/active_record/patches/transactions_class_methods.rb
```diff
@@ -18,9 +18,13 @@ class << base
 
           # Contains ActiveRecord::Transactions::ClassMethods to be patched
           module ClassMethods
-            def transaction(...)
-              tracer.in_span('ActiveRecord.transaction', attributes: { 'code.namespace' => name }) do
-                super
+            def transaction(*args, **kwargs, &block)
+              attributes = { 'code.namespace' => name }
+              if kwargs[:isolation]
+                attributes['db.transaction_isolation'] = kwargs[:isolation].to_s
```
arielvalentin, line 24: Nit: There is no semantic convention for this. 

In practice only 1 other instrumentation implements it and namespaces it under the specific DB vendor since I believe Isolation Levels are rather specific to ACID SQL style databases. 

https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/c8000c3a13bc75c5033ab5c0ccbc1dd87f2c8069/receiver/sqlserverreceiver/documentation.md#attributes-11

Having said all that, I think that its fine to be under the `db` namespace but name it `db.tansaction_isolation_level` for symmetry with the existing usage.

🙏🏼 if you would kindly submit a PR to add this semconv attribute upstream. 

Do you all use db client spans as well? Would you like this attribute to appear on those spans as well or is this span good enough?

arielvalentin (changes requested):
Thanks for this submission. Before we can accept it though, the tests must pass. 

The tests are currently failing https://github.com/open-telemetry/opentelemetry-ruby-contrib/actions/runs/17834374868/job/51063006548#step:3:191

instrumentation/active_record/lib/opentelemetry/instrumentation/active_record/patches/transactions_class_methods.rb
thompson-tomo: I would go with the attribute naming `db.tansaction.isolation_level` so that we could also have `db.tansaction.id`. I also suggest raising a pr to add it to semconv.

PR #1678, opened 2025-09-16
feat: add anthropic instrumentation to instrumentation_all

hannahramadan:
This PR adds otel instrumentation for the `anthropic` gem to `instrumentation_all`


kaylareopelle (approved):
🚀

PR #1684, opened 2025-09-17
fix!: aws_lambda - check if span has the attributes method to avoid internal error

xuan-cao-swi:


kaylareopelle:
@xuan-cao-swi - thank you, I also updated the title so that it will make sense in the changelog for instrumentation-all.
the failure on `all` here is known, because we're going to bump to 0.4 in the release PR.

kaylareopelle (approved):

kaylareopelle:
We got the permissions issue fixed and were able to release without this change, so I'm going to close this PR. Thanks for your help, @xuan-cao-swi.

Issue #1431, opened 2025-02-26
Grape - Improve readability when multiple versions are defined


katpadi:
In Grape, it's possible to define multiple versions e.g.:
```ruby
class V1 < Grape::API
  version 'v1', 'v2'

  get '/foos' do
    # your code for GET /foo
  end
end

```

This produces a value for path and `http.route` that looks like this:
```ruby
/["v1", "v2"]/foos/

```
Reference:
https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/79fa5876d0ed95d928d4beb589b04be16b9b59f7/instrumentation/grape/lib/opentelemetry/instrumentation/grape/event_handler.rb#L96

Perhaps this is to promote low cardinality URL as per specs but would it hurt if we add an ability or configuration to maybe produce something like:

```ruby
/{v1|v2}/foos/
/(v1|v2)/foos/
/v1|v2/foos/
/{v1|v2}/foos/
/:version/foos/ # Static value?
```
I'm not sure if a pipe or any of my recommended symbols violate any spec but it's mainly for readability as `["v1", "v2"]` looks like a typo or a misconfiguration.

Thoughts?

arielvalentin:
Cc: @muripic

PR #1687, opened 2025-09-18
feat: add option to allow user format the version from path in grape

xuan-cao-swi:
### Description

Based on issue https://github.com/open-telemetry/opentelemetry-ruby-contrib/issues/1431: when a Grape API has multiple versions (for example:

```ruby
class V5 < Grape::API
  version 'v5', 'v6', 'v7', using: :path

  get '/other' do
    # ...
  end
end
```

—although using multiple versions in this way is not best practice and may be rare), the route string can become `/api/[\"v5\", \"v6\", \"v7\"]//other`, which, as the user points out, is not well formatted (see attributes).

```console
attributes=
  {"http.request.method"=>"GET",
   "server.address"=>"localhost:3000",
   "url.scheme"=>"http",
   "url.path"=>"/api/v5/other",
   "user_agent.original"=>"curl/7.74.0",
   "sw.is_entry_span"=>true,
   "code.namespace"=>"V5",
   "http.route"=>"/api/[\"v5\", \"v6\", \"v7\"]//other",
   "http.response.status_code"=>200}
```

This PR adds a configuration option that allows users to format the version string as they prefer. For example, to produce `/api/{v5|v6|v7}/other`, a user can provide a lambda like:

```ruby
  c.use 'OpenTelemetry::Instrumentation::Grape',
    {
      version_format: ->(version) {
        version.is_a?(Array) ? "{#{version.join('|')}}" : version.to_s
      }
    }
end
```

If the user does not provide a config option, the instrumentation keeps the original formatting (for example: `"/api/[\"v5\", \"v6\", \"v7\"]//other"`).


instrumentation/grape/lib/opentelemetry/instrumentation/grape/event_handler.rb
```diff
@@ -93,7 +93,8 @@ def path(endpoint)
             return '' unless endpoint.routes
 
             namespace = endpoint.routes.first.namespace
-            version = endpoint.routes.first.options[:version]&.to_s
+            version = endpoint.routes.first.options[:version]
+            version = config[:version_format].call(version)
```
thompson-tomo, line 97: Rather than making it configurable could we not add the `url.template` attribute with a default of `/:version/foos/` etc. That way http.route is mirroring what grape/server is providing as per semconv.

instrumentation/grape/Gemfile
```diff
@@ -15,6 +15,9 @@ group :test do
   gem 'rake', '>= 13'
   gem 'rubocop', '~> 1.87.0'
   gem 'rubocop-performance', '~> 1.26.0'
+  gem 'rubocop-minitest', '~> 0.39.1'
+  gem 'rubocop-rspec', '~> 3.10', '>= 3.10.2'
```
thompson-tomo, line 19: Can we not simplify it to
```suggestion
  gem 'rubocop-rspec', '~> 3.10.2'
```

Given we have renovate to bump it?

PR #1689, opened 2025-09-19
chore: replace bitnami redis images with official images

### LINKED ISSUE #1683: Remove bitnami images
Bitnami is deleting their public docker catalog.

See https://github.com/bitnami/containers/issues/83267 for details. 

Currently, our Redis images are from Bitnami. We may be able to use [Redis official](https://hub.docker.com/_/redis) images instead.

There's a brownout September 17 - the morning of September 19, so tests will fail today. The images will be removed September 29.

ericmustin:
Closes #1683 

_note: we may want to bump the image to a more recent major version of redis (ie 8.x), but that can be done in a separate PR. At a glance I wasn't sure which major versions the gem supported._

### DIFF
--- a/.github/workflows/ci-instrumentation-with-services.yml
+++ b/.github/workflows/ci-instrumentation-with-services.yml
@@ -212,7 +212,7 @@ jobs:
           build: true
     services:
       redis:
-        image: bitnami/redis:6.2
+        image: redis:6.2
         ports:
           - 6379:6379
         options: >-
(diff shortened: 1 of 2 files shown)


arielvalentin (approved):

arielvalentin (changes requested):
I approved prematurely. The connections fail because the service requires authentication.

ericmustin:
@arielvalentin yup I'm trying to work thru the differences. looks like the vanilla images don't support some of the magic env var defaults that the bitnami ones did.

arielvalentin (approved):

PR #1691, opened 2025-09-24
fix: add Active Support as runtime dependency in Racecar instrumentation

razumau:
Installation of Racecar instrumentation silently fails without Active Support. Since this instrumentation doesn’t work at all without Active Support, it should be listed as a runtime dependency.

It was already being installed in test and example Gemfiles.


arielvalentin (changes requested):
Please see my comments inline.

PR #1695, opened 2025-09-25
chore: relax inst all version constraints

robertlaurin:
Allow anyone using instrumentation all to take the latest minor version without having to wait for a release of instrumentation all.


instrumentation/all/opentelemetry-instrumentation-all.gemspec
```diff
@@ -26,43 +26,43 @@ Gem::Specification.new do |spec|
...
-  spec.add_dependency 'opentelemetry-instrumentation-rake', '~> 0.3.1'
-  spec.add_dependency 'opentelemetry-instrumentation-rdkafka', '~> 0.7.0'
-  spec.add_dependency 'opentelemetry-instrumentation-redis', '~> 0.26.1'
-  spec.add_dependency 'opentelemetry-instrumentation-resque', '~> 0.6.0'
-  spec.add_dependency 'opentelemetry-instrumentation-restclient', '~> 0.24.0'
-  spec.add_dependency 'opentelemetry-instrumentation-ruby_kafka', '~> 0.22.0'
-  spec.add_dependency 'opentelemetry-instrumentation-sidekiq', '~> 0.26.0'
-  spec.add_dependency 'opentelemetry-instrumentation-sinatra', '~> 0.26.0'
-  spec.add_dependency 'opentelemetry-instrumentation-trilogy', '~> 0.61.0'
+  spec.add_dependency 'opentelemetry-instrumentation-active_model_serializers', '~> 0.22'
```
arielvalentin, line 29: Please update the README explaining the implications of this change. 

Warn users that they may receive unexpected updates of breaking changes from transitive dependencies and they should _pin_ gem versions to avoid that.

Alternatively we can explore a "all-in-one" gem that includes all of the source code in a single gem instead of relying on gem dependencies or add scripts/actions to bump the all gem when a minor version bump of the instrumentations are released. 

That would make it possible to release individual gems and simplify the release process a bit.

kaylareopelle (changes requested):
I agree with Ariel's comment, I think we need to update our documentation if we take this approach. 

Also, how does this change updates to the all gem? Should we still update the gemspec to reflect the latest version, or leave it all as-is? I lean toward updating the gemspec anyway to help communicate changes applied to the all gem through the changelog, but I'm not set on the idea.

PR #1698, opened 2025-09-27
feat: Bump minimum API Version to 1.7

arielvalentin:
The instrumentation base gem already has a dependency on the API, so declaring the dependency again isn't necessary.

Removing the extra declaration in individual instrumentation gemspecs removes the need to keep all instrumentation gems in sync with each other.

## What am I trying to fix?

```console
Installing latest version of published gems and running tests
Fetching gem metadata from https://rubygems.org/........
Resolving dependencies...
Could not find compatible versions

Because opentelemetry-instrumentation-trilogy >= 0.62.0 depends on opentelemetry-api ~> 1.7
  and every version of opentelemetry-instrumentation-anthropic depends on opentelemetry-api ~> 1.6.0,
  opentelemetry-instrumentation-trilogy >= 0.62.0 is incompatible with opentelemetry-instrumentation-anthropic >= 0.
So, because Gemfile depends on opentelemetry-instrumentation-anthropic = 0.1.0
  and Gemfile depends on opentelemetry-instrumentation-trilogy = 0.62.0,
  version solving has failed.
```

### DIFF
--- a/.instrumentation_generator/templates/gemspec.tt
+++ b/.instrumentation_generator/templates/gemspec.tt
@@ -25,7 +25,6 @@ Gem::Specification.new do |spec|
   spec.require_paths = ['lib']
   spec.required_ruby_version = ">= #{File.read(File.expand_path('../../gemspecs/RUBY_REQUIREMENT', __dir__))}"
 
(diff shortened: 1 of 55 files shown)


ericmustin (approved):

kaylareopelle (approved):
love this solution. thank you!

PR #1704, opened 2025-09-29
fix: fallback to legacy bitnami for redis images

### LINKED ISSUE #1703: Installation Errors 2025-09-29
Canary installation tests are failing:

https://github.com/open-telemetry/opentelemetry-ruby-contrib/actions/runs/18099693122/job/51499167553

```
Installing latest version of published gems and running tests
Fetching gem metadata from https://rubygems.org/........
Resolving dependencies...
Bundler found conflicting requirements for the Ruby version:
  In Gemfile:
    sidekiq was resolved to 8.0.7, which depends on
      Ruby (>= 3.2.0)

  Current Ruby version:
    Ruby (= 3.1.7)

Bundler could not find compatible versions for gem "opentelemetry-api":
  In Gemfile:
opentelemetry-instrumentation-anthropic (= 0.1.0) was resolved to 0.1.0,
which depends on
      opentelemetry-api (~> 1.6.0)

opentelemetry-instrumentation-trilogy (= 0.62.0) was resolved to 0.62.0,
which depends on
      opentelemetry-api (~> 1.7)
Bundler could not find compatible versions for gem "rdkafka":
  In Gemfile:
    rdkafka

    racecar was resolved to 2.12.0, which depends on
      rdkafka (>= 0.15.0)
```

ericmustin:
Closes https://github.com/open-telemetry/opentelemetry-ruby-contrib/issues/1703

Workflow run: https://github.com/open-telemetry/opentelemetry-ruby-contrib/actions/runs/18099693122/job/51499167553

The new Redis images I introduced in https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/1689 would require setting a password via cli, which is more complicated to configure than I had expected, and requires different configurations for both docker compose and gh workflows. It would appear for gh workflows, where we can't just modify the entrypoint script, we may have to find a different (or build/publish our own) image for redis. It requires further investigation.

I am limited on time but we can fallback to bitnamilegacy for now to unblock, which has a tradeoff of being unmaintained.

### DIFF
--- a/.github/workflows/ci-instrumentation-with-services.yml
+++ b/.github/workflows/ci-instrumentation-with-services.yml
@@ -212,12 +212,11 @@ jobs:
           build: true
     services:
       redis:
-        image: redis:6.2
+        image: bitnamilegacy/redis:6.2
         ports:
           - 6379:6379
-        command: ["redis-server", "--appendonly", "yes", "--requirepass", "${REDIS_PASSWORD}"]
         options: >-
-          --health-cmd "sh -c 'redis-cli -a \"$REDIS_PASSWORD\" ping | grep PONG'"
+          --health-cmd "redis-cli ping"
           --health-interval 10s
           --health-timeout 5s
           --health-retries 5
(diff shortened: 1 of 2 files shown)


arielvalentin (approved):

kaylareopelle (approved):
Thank you!

PR #1707, opened 2025-09-30
fix: Min OTel Ruby API 1.7

arielvalentin:
A bad changelog commit lead to base not being published to add min requirement for transitive api gem to v1.7

This change should bring the gems up to date.


kaylareopelle (approved):
thank you!

ericmustin (approved):

robbkidd (approved):
Let's do this again.

PR #1711, opened 2025-10-01
docs: Enhance README

dan-corneanu:
Enhance README with additional configuration options for ActiveJob instrumentation.

### DIFF
--- a/instrumentation/active_job/README.md
+++ b/instrumentation/active_job/README.md
@@ -30,6 +30,27 @@ OpenTelemetry::SDK.configure do |c|
 end
 ```
 
+## Configuration Options
+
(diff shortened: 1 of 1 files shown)


kaylareopelle (changes requested):
Thank you for this PR! The text looks good. I have some feedback on the placement. 

Lines 44-50 should be above the configuration section. They go along with the earlier code snippet related to installing a single instrumentation.

Also, what do you think about adding a `Configuration`/`Configuration Options` heading, or something to that effect? It would keep the document's structure more consistent.

kaylareopelle (approved):
Thank you!

arielvalentin (approved):

PR #1712, opened 2025-10-02
fix: Capture correct descriptions for RSpec one-liner examples

americodls:
RSpec one-liner examples (e.g., `it { is_expected.not_to be_nil }`) generate descriptions after execution.

Updated formatter to capture the final description in `example_finished` instead of relying on the incomplete description at `example_started`.


kaylareopelle (approved):
Hi @americodls, thanks for catching this and providing a fix!

chrisholmes (approved):

instrumentation/rspec/lib/opentelemetry/instrumentation/rspec/formatter.rb
```diff
@@ -68,8 +68,12 @@ def example_started(notification)
 
         def example_finished(notification)
           pop_and_finalize_span do |span|
-            result = notification.example.execution_result
+            example = notification.example
+            result = example.execution_result
 
+            # Update name and full_description for one-liner examples where description is generated after execution
+            span.name = example.description
+            span.set_attribute('rspec.example.full_description', example.full_description.to_s)
```
arielvalentin: 💁🏼 Use `add_attributes` as opposed to setting each attribute separately because each call requires a lock when changing a span

kaylareopelle: Great catch. I agree. @americodls, can you update the method call to combine the attributes on line 76 and 77 into one `add_attributes` call?

americodls: Thank you for the feedback. I just fixed that.

PR #1713, opened 2025-10-02
feat: add example.id attribute to RSpec instrumentation

americodls:
This PR adds RSpec's `example.id` as a span attribute to provide unique identification for each test example.

The existing `rspec.example.location` attribute cannot uniquely identify dynamically generated examples (e.g., examples created within loops or using data tables) because multiple examples share the same source file location. RSpec's `example.id` provides a stable, unique identifier across test runs even for these dynamic cases.

Example format: `./spec/foo_spec.rb[1:2:3]`

This enables better correlation and tracking of individual test executions in observability platforms.


kaylareopelle (approved):
Thanks for your contribution, @americodls. This makes sense to me. I'll leave it open for a few more days to see if any of the other approvers/maintainers have feedback.

chrisholmes (approved):
Nice, I think this will solve a problem we have with Turnip

PR #1721, opened 2025-10-07
feat: Add FactoryBot instrumentation

americodls:
Adds OpenTelemetry instrumentation for the FactoryBot gem to provide visibility into test data creation patterns during testing.

Subscribes to the factory_bot.run_factory ActiveSupport::Notifications event to capture all factory operations without monkey-patching. Supports FactoryBot 4.0+ and captures all factory strategies (create, build, build_stubbed, attributes_for) including batch operations (create_list, build_list, etc.).

Span naming: FactoryBot.{strategy}({factory_name})
Example: FactoryBot.build(user), FactoryBot.create_list(post)

Span attributes:
- factory_bot.strategy: Internal strategy name (build, create, stub, attributes_for)
- factory_bot.factory_name: Name of the factory being used
- factory_bot.traits: Comma-separated list of traits applied (if any)


instrumentation/factory_bot/lib/opentelemetry/instrumentation/factory_bot/version.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+# Copyright The OpenTelemetry Authors
+#
+# SPDX-License-Identifier: Apache-2.0
+
+module OpenTelemetry
+  module Instrumentation
+    module FactoryBot
+      VERSION = '0.1.0'
```
arielvalentin:
```suggestion
      VERSION = '0.0.0'
```
Versions are managed by our release tooling. This initial release number is expected to be all 0 https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/main/.instrumentation_generator/templates/lib/instrumentation/instrumentation_name/version.rb.tt#L10

instrumentation/factory_bot/opentelemetry-instrumentation-factory_bot.gemspec
```diff
@@ -0,0 +1,37 @@
...
+  spec.homepage    = 'https://github.com/open-telemetry/opentelemetry-ruby-contrib'
+  spec.license     = 'Apache-2.0'
+
+  spec.files = Dir.glob('lib/**/*.rb') +
+               Dir.glob('*.md') +
+               ['LICENSE', '.yardopts']
+  spec.require_paths = ['lib']
+  spec.required_ruby_version = ">= #{File.read(File.expand_path('../../gemspecs/RUBY_REQUIREMENT', __dir__))}"
+
+  spec.add_dependency 'activesupport', '>= 4.0'
```
arielvalentin: Do not add any transitive runtime dependencies to your gemspec. We do not want for users to inadvertently install 3rd party dependencies that may conflict with their application production code.  

It looks like this is already a transitive dependency of `FactoryBot` so as long as they have installed this gem, then active support should be made available to your instrumentation https://github.com/thoughtbot/factory_bot/blob/main/factory_bot.gemspec#L29

```suggestion
```

kaylareopelle:
Hi @americodls, thanks for opening this PR. Would you be interested in being a codeowner for this gem? That would involve reviewing PRs and issues related to the factory bot instrumentation. We don't have the codeowner behavior automated in GitHub, but we're currently documenting this in the CODEOWNERS file in the repo.

simi (approved):

simi:
@americodls would you mind to share more how do you instrument testing also?

instrumentation/factory_bot/lib/opentelemetry/instrumentation/factory_bot/run_factory_subscriber.rb
```diff
@@ -0,0 +1,61 @@
...
+          # Map user-facing strategy names to internal strategy names
+          internal_strategy = case strategy_symbol
+                              when :build_stubbed then 'stub'
+                              else strategy_symbol.to_s
+                              end
+
+          span_name = "FactoryBot.#{strategy_symbol}(#{factory_name})"
+
+          attrs = {
+            'factory_bot.strategy' => internal_strategy,
```
thompson-tomo, line 30: Suggestion name this test.data.strategy given it could be used elsewhere and we already have the test namespace?

instrumentation/factory_bot/lib/opentelemetry/instrumentation/factory_bot/run_factory_subscriber.rb
```diff
@@ -0,0 +1,61 @@
...
+            'factory_bot.factory_name' => factory_name.to_s
+          }
+
+          attrs['factory_bot.traits'] = traits.join(',') if traits.any?
```
thompson-tomo: Suggestion name this test.data.traits given it could be used elsewhere and we already have the test namespace?

thompson-tomo: Shouldn't value be an array?

thompson-tomo: If test.data.* is adopted, I would add a test.data.provider.name attribute.

instrumentation/factory_bot/lib/opentelemetry/instrumentation/factory_bot/run_factory_subscriber.rb
americodls: Thanks for the feedback!

I checked the [OpenTelemetry test semantic conventions](https://opentelemetry.io/docs/specs/semconv/attributes-registry/test/) and found that the test.* namespace currently only defines attributes for test execution metadata (test cases, suites, and their statuses).

Since test.data isn’t part of the spec yet, using it now could risk conflicts if it’s added later. The current factory_bot.* approach keeps things scoped to this library’s specific domain.

I’d suggest keeping the dedicated namespace for now — we can always generalize to something like test.data.* if similar fixture/factory patterns start appearing across other testing libraries. That way, we’re being intentional rather than speculative about the abstraction.

Of course, I’m open to other perspectives — input from the maintainers would be especially valuable given their broader experience in this space.

thompson-tomo: So perhaps the following helps [defining attributes](https://opentelemetry.io/docs/specs/semconv/how-to-write-conventions/#defining-attributes)

> When defining an attribute for a narrow use case, think about potential broader use cases. For example, if creating a system-specific attribute, evaluate whether other systems in the same domain might need a similar attribute in the future.

I see the attributes identified as falling into that category as anything generating data needs a strategy etc.

I would even suggest raising a pr to get the attributes/spans added.

instrumentation/factory_bot/CHANGELOG.md
```diff
@@ -0,0 +1,8 @@
+# Release History: opentelemetry-instrumentation-factory_bot
+
+## v0.1.0 / 2025-01-03
+
+* Initial release
+* Instruments FactoryBot operations via ActiveSupport::Notifications
+* Captures create, build, build_stubbed, and attributes_for strategies
+* Includes factory name, strategy, and traits in span attributes
```
arielvalentin:
```suggestion
```

You will need to remove these notes because it conflicts with what is auto generated by our release tooling. 

https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/main/.instrumentation_generator/templates/CHANGELOG.md.tt

arielvalentin (changes requested):
Thank you for your submission! 

There are a few required changes you will have to make to avoid issues/conflicts with our release tooling. 

I will have to reach out for help to other users who are FactoryBot users to help with a review before we merge.

instrumentation/factory_bot/lib/opentelemetry/instrumentation/factory_bot/run_factory_subscriber.rb
americodls: Addressed here: [811782a](https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/1721/commits/811782a25c7dbfb6e3dc49ffe21d62981fcef188)

instrumentation/factory_bot/lib/opentelemetry/instrumentation/factory_bot/run_factory_subscriber.rb
arielvalentin: Although I agree with @thompson-tomo that we need to do our best to adhere to the spec for attribute naming; I am amenable to releasing custom attributes for our instrumentations since they are all pre-1.0 at this point and we are still working through figuring out how the version semantic conventions used in instrumentation libraries. 

The factory bot use case seems very narrow since the spans are implementation specific. Other testing object mother or factory libraries will likely use different domain language based on DSLs and helper methods. We may find that higher level span names may have some overlap but library specific attributes like `traits` are likely not going to be a good fit for a "broader" use case. 

@americodls Are there any other testing libraries that are currently instrumented in other languages? 

Is there enough traction there to start up a conversation between SIGs?

americodls: > @americodls Are there any other testing libraries that are currently instrumented in other languages?
> 
> Is there enough traction there to start up a conversation between SIGs?

I haven't found anything that's why I suggested to "go as is".

arielvalentin (approved):

arielvalentin (changes requested):
@americodls thank you again for your submission. 

Although I approved this PR, I am going to have request some additional changes. 

Our CI workflows have changed and this gems test suite is not longer running by default.

_You will need to add entries in the following actions workflow:_
- https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/1e018082a19f42a5f159838145b614d13537a5a4/.github/workflows/ci-instrumentation-full.yml#L8
- https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/1e018082a19f42a5f159838145b614d13537a5a4/.github/workflows/ci-instrumentation-full.yml#L67 

In addition to that please add `factorybot` as a gem dependency in our canary install test:

https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/1e018082a19f42a5f159838145b614d13537a5a4/releases/Gemfile#L14

Thank you for your patience.

americodls:
@arielvalentin

Thanks for your feedback!
I've gone ahead and fixed the issues you pointed out 🙂

Let me know if there’s anything else you’d like me to tweak!

arielvalentin (approved):

PR #1725, opened 2025-10-08
fix: Fixing missing OpenTelemetry::Context detach on Excon instrumentation

pafuent:
When a span is not recorded, for example when you are sampling traces, the OpenTelemetry::Context attached on `request_call` is never detached. This happens because `handle_response` is doing nothing if `span.recording?` is false.

Here is example of the error that the missing detach generated:
```
ERROR -- : OpenTelemetry error: calls to detach should match corresponding calls to attach.
```

### DIFF
--- a/instrumentation/excon/lib/opentelemetry/instrumentation/excon/middlewares/dup/tracer_middleware.rb
+++ b/instrumentation/excon/lib/opentelemetry/instrumentation/excon/middlewares/dup/tracer_middleware.rb
@@ -84,7 +84,8 @@ module OpenTelemetry
 
             def handle_response(datum)
               datum.delete(:otel_span)&.tap do |span|
-                OpenTelemetry::Context.detach(datum.delete(:otel_token)) if datum.include?(:otel_token)
+                token = datum.delete(:otel_token)
+                OpenTelemetry::Context.detach(token) if token
                 return unless span.recording?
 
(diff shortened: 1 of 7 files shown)


instrumentation/excon/lib/opentelemetry/instrumentation/excon/middlewares/dup/tracer_middleware.rb
```diff
@@ -84,6 +84,7 @@ def self.around_default_stack
 
             def handle_response(datum)
               datum.delete(:otel_span)&.tap do |span|
+                OpenTelemetry::Context.detach(datum.delete(:otel_token)) if datum.include?(:otel_token)
```
arielvalentin: Nit: will delete raise an error?

If not, I think it would be best to minimize the number of method calls.

```suggestion
token = datum.include?(:otel_token)

OpenTelemetry::Context.detach(token) if token 
```

pafuent: No, delete returns nil

arielvalentin (approved):
Thank you for this submission and fix!

kaylareopelle (approved):
Thank you! I really appreciate you adding the fix to all three conventions versions too! 🏆

PR #1738, opened 2025-10-11
feat: ActiveRecord instrumentation improvements

arielvalentin:


instrumentation/active_record/lib/opentelemetry/instrumentation/active_record/handlers/sql_handler.rb
```diff
@@ -0,0 +1,69 @@
...
+
+          # Invoked by ActiveSupport::Notifications at the end of the instrumentation block
+          #
+          # @param name [String] of the Event
+          # @param id [String] of the event
+          # @param payload [Hash] containing SQL execution information
+          def finish(name, id, payload)
+            attributes = {
+              'db.active_record.async' => payload[:async] == true,
+              'db.active_record.cached' => payload[:cached] == true
```
thompson-tomo: Thoughts about making this `rails.active_record.*` for consistency with the semconv pattern of not putting product specific attributes in the domain ie db namespace.

arielvalentin: Thanks for the feedback.

I wasn't sure what direction to head since there isn't any semconv guidance for ORM of database abstraction libraries. 

Namespacing under rails works for me

PR #1746, opened 2025-10-18
feat: use formatted_message in Ruby logger instrumentation

### LINKED ISSUE #1745: Consider using formatted_message in Ruby Logger instrumentation
In https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/983 there was a question about whether to use `msg` or `formatted_message` in the Ruby logger instrumentation.

Now that it has been released using `msg`, this issue is open to decide whether to change to `formatted_message`.

Image

wsmoak:
Use `formatted_message` in the Ruby logger instrumentation so that the fully decorated log message is sent.

Fixes #1745


ericmustin (approved):
Is there a test we want here to prevent a regression ?

kaylareopelle (approved):
Thanks, @wsmoak! As to your question about timestamps and process IDs, I think checking to see that the severity and body/message are both included in the body attribute should be sufficient for now for regression protection.

robbkidd (approved):

xuan-cao-swi (approved):

PR #1758, opened 2025-10-24
feat: Semconv Span Helpers

arielvalentin:
Introduces a helper gem for use in instrumentations that will standardize semantic convention naming.

The initial implementation adds an implementation for HTTP Client span names: https://opentelemetry.io/docs/specs/semconv/http/http-spans/#name 

The usage of this helper is intended to replace uses in all client gems (excon, http, net-http, etc...). This will allow wrapper libraries to enrich span names without the need for custom code and reuse the existing HTTP ClientContext helper.

### DIFF
--- a/.github/workflows/ci-contrib.yml
+++ b/.github/workflows/ci-contrib.yml
@@ -24,6 +24,7 @@ jobs:
       fail-fast: false
       matrix:
         gem:
+          - semconv
           - sql
           - mysql
           - sql-obfuscation
--- a/helpers/semconv/CHANGELOG.md
+++ b/helpers/semconv/CHANGELOG.md
@@ -1,0 +1,1 @@
+# Release History: opentelemetry-helpers-http
--- a/helpers/semconv/README.md
+++ b/helpers/semconv/README.md
@@ -1,0 +1,93 @@
+# OpenTelemetry Semantic Convention Helpers
+
+This gem provides semantic convention helpers for the OpenTelemetry framework, including HTTP span naming utilities. It is intended to be used by the instrumentation libraries and is not intended to be used directly by applications.
+
+## Installation
+
+Add a line similar to this in your `gemspec`:
+
+```ruby
+spec.add_dependency 'opentelemetry-helpers-semconv', '~> 0.1' # Use the appropriate version
+```
+
+Update your `Gemfile` to use the latest version of the gem in the contrib, if you are working locally in this repository e.g.
+
+```ruby
+group :test do
+  gem 'opentelemetry-helpers-semconv', path: '../../helpers/semconv' # Use the appropriate path
+end
+```
+
+## Usage
+
+### HTTP Span Naming
+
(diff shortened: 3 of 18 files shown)


.github/workflows/ci-contrib.yml
```diff
@@ -24,6 +24,7 @@ jobs:
       fail-fast: false
       matrix:
         gem:
+          - semconv
```
kaylareopelle: I'm a little worried about the name, but am not quite sure what to call it instead. Mostly leaving this as a note to think about later.

helpers/semconv/CHANGELOG.md
```diff
@@ -0,0 +1 @@
+# Release History: opentelemetry-helpers-http
```
kaylareopelle: This is inconsistent with the other places.

helpers/semconv/README.md
```diff
@@ -0,0 +1,107 @@
...
+
+The HTTP Client helper supports both current and legacy OpenTelemetry semantic conventions:
+
+- **Current**: `http.request.method`, `url.template` (client spans only)
+- **Legacy**: `http.method` (deprecated but supported for compatibility)
+- **Behavior**: Prefers current conventions over legacy ones
+- **Method normalization**: Automatically converts standard HTTP methods to uppercase (e.g., 'get' → 'GET')
+- **Fallbacks**: Returns 'HTTP' when no recognizable attributes are present
+
+**Important**: The `url.template` attribute is only applicable to HTTP client spans according to OpenTelemetry semantic conventions. Server spans should use `http.route` or other appropriate attributes.
```
kaylareopelle: How does it control these things?

helpers/semconv/opentelemetry-helpers-semconv.gemspec
```diff
@@ -0,0 +1,36 @@
...
+  spec.summary     = 'Semconv helpers for the OpenTelemetry framework'
+  spec.description = 'Semconv helpers for the OpenTelemetry framework'
+  spec.homepage    = 'https://github.com/open-telemetry/opentelemetry-ruby-contrib'
+  spec.license     = 'Apache-2.0'
+
+  spec.files = Dir.glob('lib/**/*.rb') +
+               Dir.glob('*.md') +
+               ['LICENSE', '.yardopts']
+  spec.require_paths = ['lib']
+  spec.required_ruby_version = ">= #{File.read(File.expand_path('../../gemspecs/RUBY_REQUIREMENT', __dir__))}"
```
kaylareopelle: If the other PR gets merged first, we'll want to update this.

helpers/semconv/test/test_helper.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+# Copyright The OpenTelemetry Authors
+#
+# SPDX-License-Identifier: Apache-2.0
+
+require 'bundler/setup'
+Bundler.require(:default, :development, :test)
+
```
kaylareopelle: Should add simplecov here too...maybe? Does it run without it?

PR #1759, opened 2025-10-24
fix: does not trace resolve type unless enable

xuanduc987:
Currently, a span with "empty" name is generated


arielvalentin:
cc: @rmosolgo Request for review

arielvalentin:
@xuanduc987 looks like you have a failing Rubocop checks. Please address those and we can look to merge.

arielvalentin (approved):

PR #1760, opened 2025-10-24
fix: patch dataloader to propagate context to new fiber

### LINKED ISSUE #127: GraphQL dataloaders not compatible with opentelemetry tracing plugin
Ruby 2.7.3p183 (2021-04-05 revision 6847ee089d) [x86_64-darwin20]
GraphQL Gem Version: 2.0.13

Options `enable_platform_field` and `enable_platform_authorized` are not compatible with GraphQL gem's own fiber based dataloaders. Turning on these options while using the built-in dataloaders results in missing data and logs like:
~~~
ERROR -- : OpenTelemetry error: calls to detach should match corresponding calls to attach.
ERROR -- : OpenTelemetry error: calls to detach should match corresponding calls to attach.
ERROR -- : OpenTelemetry error: calls to detach should match corresponding calls to attach.
ERROR -- : OpenTelemetry error: calls to detach should match corresponding calls to attach.
ERROR -- : OpenTelemetry error: calls to detach should match corresponding calls to attach.
ERROR -- : OpenTelemetry error: calls to detach should match corresponding calls to attach.
~~~

I'm assuming this has to do with [the way these data loaders work](https://graphql-ruby.org/dataloader/adopting.html). Somehow, this appeared to be working fine while using [NewRelic tracing](https://github.com/rmosolgo/graphql-ruby/blob/master/lib/graphql/tracing/new_relic_tracing.rb) and it was able to capture traces with details on field hops and authorizations. Removing the dataloaders fixes this problem, but we need them to prevent N+1 query issues.

~~~ruby
# field method
def team
  dataloader.with(Sources::Record, Team).load(object.team_id)
end

# authorized? method
def self.authorized?(object, context)
  can_access_team?(context.dataloader.with(Sources::Record, Team).load(object.team_id))
end
~~~

xuanduc987:
I think this should fix #127

### DIFF
--- a/instrumentation/graphql/lib/opentelemetry/instrumentation/graphql/instrumentation.rb
+++ b/instrumentation/graphql/lib/opentelemetry/instrumentation/graphql/instrumentation.rb
@@ -27,6 +27,8 @@ module OpenTelemetry
             require_relative 'tracers/graphql_trace'
             install_new_tracer(config)
           end
+
+          patch
         end
 
         present do
@@ -95,6 +97,13 @@ module OpenTelemetry
               OpenTelemetry.logger.error("Unable to patch schema #{schema}: #{e.message}")
             end
           end
+        end
+
+        def patch
+          return if gem_version < Gem::Version.new('2.1.8')
+
+          require_relative 'patches/dataloader'
+          ::GraphQL::Dataloader.prepend(Patches::Dataloader)
         end
       end
     end
--- a/instrumentation/graphql/lib/opentelemetry/instrumentation/graphql/instrumentation.rb
+++ b/instrumentation/graphql/lib/opentelemetry/instrumentation/graphql/instrumentation.rb
@@ -43,6 +43,10 @@ module OpenTelemetry
           Gem::Requirement.new('>= 2.0.19').satisfied_by?(gem_version)
         end
 
+        def dataloader_has_spawn_fiber?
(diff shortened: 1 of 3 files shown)


instrumentation/graphql/lib/opentelemetry/instrumentation/graphql/instrumentation.rb
```diff
@@ -96,6 +98,13 @@ def install_new_tracer(config = {})
             end
           end
         end
+
+        def patch
+          return if gem_version < Gem::Version.new('2.1.8')
```
xuanduc987: The target of the patch, `Dataloader#spawn_fiber` method existed from version 2.1.8

rmosolgo:
LGTM :+1:

instrumentation/graphql/lib/opentelemetry/instrumentation/graphql/instrumentation.rb
arielvalentin: Ideally we would have some automated tests that test this functionality to avoid regressions. 

@rmosolgo do you have any recommendations on how to set up a test to cover this?

xuanduc987: I had added a test for the expected behavior of `#spawn_fiber`.
I also confirmed that it failed without the patch.

arielvalentin (approved):

PR #1763, opened 2025-10-27
feat: Instrument PG connect

kirs:
It's surprising that we don't do this for PG while we do this for Trilogy. Without this, it's impossible to attribute any time spent establishing the connection.


instrumentation/pg/lib/opentelemetry/instrumentation/pg/patches/connection.rb
```diff
@@ -12,6 +12,56 @@ module OpenTelemetry
...
+                conn = super(*args)
+                set_connection_attributes(span, conn, config)
+                conn
+              end
+            end
+          end
+
+          private
+
+          def set_connection_attributes(span, conn, config)
```
arielvalentin, line 19: ℹ️ Avoid adding methods to mixins. 

Add these methods to instrumentation helper modules instead.

instrumentation/pg/lib/opentelemetry/instrumentation/pg/patches/connection.rb
```diff
@@ -12,6 +12,56 @@ module OpenTelemetry
...
+                end
+                result
+              else
+                # No block given, just set attributes after connection
+                conn = super(*args)
+                set_connection_attributes(span, conn, config)
+                conn
+              end
+            end
+          end
```
arielvalentin: Since we do not use the arguments, would it not be enough to forward them?

```suggestion
          def new(...)
            tracer = PG::Instrumentation.instance.tracer
            config = PG::Instrumentation.instance.config

            tracer.in_span('connect', kind: :client) do |span|
                conn = super
                set_connection_attributes(span, conn, config)
                return conn
              end
            end
          end
```

Would you also kindly reference to choice to override `new` as opposed to `initialize` here as well?

https://github.com/ged/ruby-pg/blob/master/lib/pg/connection.rb#L870

arielvalentin (commented):
Thank you for your contribution. I have left some comments for you to review.

kaylareopelle (commented):
Thank you for this contribution, @kirs! There's one Rubocop failure on the PG suite to address: 
```
lib/opentelemetry/instrumentation/pg/patches/connection.rb:15:9: 
  C: Style/Documentation: Missing top-level documentation comment for module 
    OpenTelemetry::Instrumentation::PG::Patches::ConnectionHelper.
        module ConnectionHelper
        ^^^^^^^^^^^^^^^^^^^^^^^ 
```

kirs:
fixed, thank you!

instrumentation/pg/lib/opentelemetry/instrumentation/pg/patches/connection.rb
```diff
@@ -12,6 +12,65 @@ module OpenTelemetry
...
+        # We also need to override the aliases (open, connect, async_connect) because they
+        # were aliased before our prepend, so they point to the original method.
+        # See: https://github.com/ged/ruby-pg/blob/master/lib/pg/connection.rb#L870
+        module Connect
+          def new(...)
+            tracer = OpenTelemetry::Instrumentation::PG::Instrumentation.instance.tracer
+            config = OpenTelemetry::Instrumentation::PG::Instrumentation.instance.config
+
+            tracer.in_span('connect', kind: :client) do |span|
+              if block_given?
```
arielvalentin, line 56: Just wanted to confirm that there wasn't a way to simply this. 

Does the block not get forwarded when calling super?

arielvalentin (approved):

instrumentation/pg/lib/opentelemetry/instrumentation/pg/patches/connection.rb
kirs: It does get forwarded, but we want a special path for `block_given?` to be able to set span attributes.


```
                conn = super
                ConnectionHelper.set_connection_attributes(span, conn, config)
                conn
```

^ the above only returns `conn` when there's no block given. When the block is given, it doesn't "leak" the connection object outside of the block.

PR #1766, opened 2025-10-28
ci: Add path-based triggers to CI workflows

### LINKED ISSUE #1753: Proposal to Refine CI Triggers
I would like to propose that only trigger specific CI suites when changes are detected that are related to those specific gems. 

The longest part of the contribution process is waiting for CI to execute the test suite, however there are many cases where running the entire test suite is unnecessary. 

The ["CI Instrumentation With Services"](https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/main/.github/workflows/ci-instrumentation-with-services.yml) suite averages a 18 min runtime that only exercises integrated tests for instrumentation gems that rely on external services like Kafka and MySQL. Waiting on the completion of this test suite slows down the development feedback loop and release process for unrelated changes. 

## What should we isolate?

- CI Instrumentation With Services: This should only be triggered when it detects changes to gems that are listed in the suite
- CI Contrib: Should be subdivided into separate suites and only trigger when changes happen to those gems e.g. a "CI Instrumentation", "CI Processors",  "CI Helpers", "CI Propagators" etc...
- Markdown linters: Should only execute when there are changes in markdown files

## How do we achieve this?

GitHub Actions supports `paths` filters that allow us to use glob patterns to trigger the workflow:

https://docs.github.com/en/enterprise-cloud@latest/actions/reference/workflows-and-actions/workflow-syntax#example-including-paths

The example below demonstrates how we might change one of the workflows:

```yaml
name: CI Instrumentation With Services

on:
  workflow_dispatch:
  push:
    branches:
      - main
     paths:
    - 'instrumentation/bunny/**'
    - 'instrumentation/dalli/**'
    - 'instrumentation/mongo/**'
    - 'instrumentation/mysql2/**'
    - 'instrumentation/pg/**'
    - 'instrumentation/que/**'
    - 'instrumentation/racecar/**'
    - 'instrumentation/rdkafka/**'
    - 'instrumentation/redis/**'
    - 'instrumentation/resque/**'
    - 'instrumentation/ruby_kafka/**'
    - 'instrumentation/sidekiq/**'
    - 'instrumentation/trilogy/**'
  pull_request:
    branches:
      - main
    paths:
      - 'instrumentation/bunny/**'
      - 'instrumentation/dalli/**'
      - 'instrumentation/mongo/**'
      - 'instrumentation/mysql2/**'
      - 'instrumentation/pg/**'
      - 'instrumentation/que/**'
      - 'instrumentation/racecar/**'
      - 'instrumentation/rdkafka/**'
      - 'instrumentation/redis/**'
      - 'instrumentation/resque/**'
      - 'instrumentation/ruby_kafka/**'
      - 'instrumentation/sidekiq/**'
      - 'instrumentation/trilogy/**'
  schedule:
    - cron: "0 0 * * *"
```

## Consequences and Unknowns

My assertion is that these changes will improve our overall development feedback look and help with reducing wait times and save on compute resources in our actions but they do come with added complexity and cost:

- These changes will result in more maintenance and may have some capability issues with the `toys` release process, which waits for checks to resolve before releasing a gem 
- It will be easy to forget that a test suite is only triggered when specific gems are changed
- Adding new instrumentations will require choosing a specific CI file, making it a little more challenging for the instrumentation_generator

I think these trade-offs are worth it considering how much time we spend waiting on test suites to complete just to know that "everything is ok"

arielvalentin:
This should skip running tests for cases where it is not needed

Fixes: https://github.com/open-telemetry/opentelemetry-ruby-contrib/issues/1753

Skipping changes to instrumentation-with-services:

image


.github/workflows/ci-instrumentation-full.yml
```diff
@@ -0,0 +1,195 @@
+name: CI Instrumentation Full
+
+on:
+  workflow_dispatch:
+  push:
+    branches:
+      - main
+    paths:
+      - 'instrumentation/action_mailer/**'
```
xuan-cao-swi, line 10: I think it will still trigger all instrumentation test to run even if I only modified the file inside e.g. `http/`.

Just an idea: the matrix can be modified
```yaml
  detect-changes:
    runs-on: ubuntu-latest
    outputs:
      matrix: ${{ steps.set-matrix.outputs.matrix }}
    steps:
      ...

      - name: Get changed instrumentation files
        id: changed-files
        run: |
          if [[ "${{ github.event_name }}" == "schedule" ]] || [[ "${{ github.event_name }}" == "workflow_dispatch" ]]; then
            echo "run_all=true" >> $GITHUB_OUTPUT
          else
            echo "run_all=false" >> $GITHUB_OUTPUT
            echo "changed_gems=$(git diff --name-only ...)" >> $GITHUB_OUTPUT
          fi
      
      - name: Set matrix
        id: set-matrix
        run: |
          # get CHANGED_GEMS from last step
          echo "matrix={\"gem\":$CHANGED_GEMS,\"os\":[\"ubuntu-latest\"]}" >> $GITHUB_OUTPUT

  instrumentation:
    needs: detect-changes
    if: ${{ needs.detect-changes.outputs.matrix != '' && fromJson(needs.detect-changes.outputs.matrix).gem[0] != null }}
    strategy:
      fail-fast: false
      matrix: ${{ fromJson(needs.detect-changes.outputs.matrix) }}

    name: ${{ matrix.gem }} / ${{ matrix.os }}
    runs-on: ${{ matrix.os }}
    steps:
      - uses: actions/checkout@v5
      - name: "Test Ruby 3.4"
        ...
```

.github/workflows/ci-contrib.yml
```diff
@@ -5,9 +5,33 @@ on:
   push:
     branches:
       - main
+    paths:
+      - 'helpers/**'
```
xuan-cao-swi, line 9: Same as last comment, if I only modified resources/, I probably only want resources/ to run test.

With paths-filter, we can use the output from detect-changes to decide on what jobs to run
```yaml
jobs:
  detect-changes:
    runs-on: ubuntu-latest
    outputs:
      helpers: ${{ steps.filter.outputs.helpers }}
      propagators: ${{ steps.filter.outputs.propagators }}
      ...
    steps:
      - uses: actions/checkout@v5
      - uses: dorny/paths-filter@v3
        id: filter
        with:
          filters: |
            helpers:
              - 'helpers/**'
            propagators:
              - 'propagator/**'
            ...

  helpers:
    needs: detect-changes
    if: ${{ needs.detect-changes.outputs.helpers == 'true' || github.event_name == 'schedule' || github.event_name == 'workflow_dispatch' }}

  propagators:
    ...
```

.github/workflows/ci-instrumentation-full.yml
arielvalentin: Oh I see what you mean. 

If an instrumentation changes, there are too many interdependencies at the moment between helpers, frameworks and HTTP etc.. so I do want them to run together. 

For example, if I change `instrumentation/active_support` I also want to tests for anything that depends on that gem e.g. action_view. 

It is a bit difficult to encode that as a set of dependencies to make sure they all run together. 

Now that I think about it, changes to `instrumentation/base` and `helpers` should also trigger Instrumentation with services since those are critical dependencies for those libraries as well.

arielvalentin:
Reviewed in the SIG and I missed a few paths triggers so I am going to see if I could use anchors to avoid structural duplication

.github/workflows/ci-contrib.yml
```diff
@@ -5,9 +5,22 @@ on:
...
+      - 'propagator/**'
+      - 'resources/**'
+      - 'processor/**'
+      - 'sampler/**'
+      - '.github/workflows/ci-contrib.yml'
+      - '.github/actions/**'
+      - 'Gemfile'
+      - 'Rakefile'
+      - '.rubocop.yml'
+      - 'gemspecs/**'
```
kaylareopelle, line 19: What do these paths give us?

.github/workflows/ci-instrumentation-full.yml
```diff
@@ -0,0 +1,156 @@
+name: CI Instrumentation Full
+
+on:
+  workflow_dispatch:
+  push:
+    branches:
+      - main
+    paths: &path_filters
```
kaylareopelle, line 8: I'm curious -- why would you like to do a filtered run on pushes to main?

kaylareopelle (commented):
Thanks for working on this, @arielvalentin! This seems like a huge improvement. I have a few questions before I approve.

.github/workflows/ci-instrumentation-full.yml
arielvalentin: I wanted it to be symmetric. Basically only run tests on the subset of gems that were merged by the PR, e.g. only gems that will need a release.

.github/workflows/ci-contrib.yml
arielvalentin: This ensures that if the action changes... then the action, in this case `ci-contrib` ends up executing. Otherwise nothing will run tests for this action if it changes. 
```
- '.github/workflows/ci-contrib.yml'
```

The following are shared files across multiple actions. If any of those change then the `ci-contrib` job should also run:
```
      - '.github/actions/**'
```


Top level files, which basically drive when we update Rubocop or for anyone running all tests from the root directory. This `.rubocop.yml` is the root/base file all other rubocops inherit from
```
     - 'Gemfile'
     - 'Rakefile'
     - '.rubocop.yml'
```

These are the files I use to test gem installations. If something changes in these files I want the entire test suite to run because installations are failing. 
```
 - 'gemspecs/**'
 - ```

.github/workflows/ci-instrumentation-full.yml
arielvalentin: That being said... I am not married to it. I could make it so all tests run on main on push.

.github/workflows/ci-contrib.yml
kaylareopelle: That makes sense! Thanks for breaking it down.

.github/workflows/ci-instrumentation-full.yml
kaylareopelle: Makes sense! My vote is to keep the filtered runs on main too for now, and if we end up wanting something else we can change it down the line. The scheduled CI still runs all the tests without a filter, so that should catch things if we missed a suite combo that has a related test failure.

kaylareopelle (approved):

Issue #1779, opened 2025-11-06
Implement Semantic Convention Rules for HTTP method


arielvalentin:
> [1] http.request.method: HTTP request method value SHOULD be “known” to the instrumentation. By default, this convention defines “known” methods as the ones listed in [RFC9110](https://www.rfc-editor.org/rfc/rfc9110.html#name-methods), the PATCH method defined in [RFC5789](https://www.rfc-editor.org/rfc/rfc5789.html) and the QUERY method defined in [httpbis-safe-method-w-body](https://datatracker.ietf.org/doc/draft-ietf-httpbis-safe-method-w-body/?include_text=1).
> 
> If the HTTP request method is not known to instrumentation, it MUST set the http.request.method attribute to _OTHER.
> 
> If the HTTP instrumentation could end up converting valid HTTP request methods to _OTHER, then it MUST provide a way to override the list of known HTTP methods. If this override is done via environment variable, then the environment variable MUST be named OTEL_INSTRUMENTATION_HTTP_KNOWN_METHODS and support a comma-separated list of case-sensitive known HTTP methods (this list MUST be a full override of the default known method, it is not a list of known methods in addition to the defaults).
> 
> HTTP method names are case-sensitive and http.request.method attribute value MUST match a known HTTP method name exactly. Instrumentations for specific web frameworks that consider HTTP methods to be case insensitive, SHOULD populate a canonical equivalent. Tracing instrumentations that do so, MUST also set http.request.method_original to the original value.

https://opentelemetry.io/docs/specs/semconv/http/http-spans/#http-client-span

# Task List

- [ ] Set unknown methods to `_OTHER` e.g. [Fastly `purge`](https://www.fastly.com/documentation/guides/concepts/edge-state/cache/purging/#url-purge)
  - [ ] http
  - [ ] http-client
  - [ ] httpx
  - [ ] faraday
  - [ ] net-http
  - [ ] ethon
  - [ ] excon
  - [ ] ~restclient~ covered by net::http
- [ ] Add support for  `OTEL_INSTRUMENTATION_HTTP_KNOWN_METHODS`
  - [ ] http
  - [ ] http-client
  - [ ] httpx
  - [ ] faraday
  - [ ] net-http
  - [ ] ethon
  - [ ] excon
  - [ ] ~restclient~ covered by net::http

PR #1781, opened 2025-11-06
fix: HTTP unknown methods

arielvalentin:
Ensures that all HTTP client gems support unknown methods as described in semantic conventions: https://opentelemetry.io/docs/specs/semconv/http/http-spans/#http-client-span

Some concessions: 

1. This change **intentionally** introduces _structural and knowledge duplication_ in each client gem as a first step refactoring to extract and move the functionality into a shared helper gem or the semantic conventions gem. 
2. The HttpHelper produces a bit of primitive obsession by adding a struct that is used to copy to semconv attributes
3.  I am _fixing_ the span names for the `_OTHER` method cases by setting the span names to `HTTP`
4. I am preserving the invalid span name generated by the `ethon` gem for `_OTHER` to be `HTTP N/A`
5. I am deferring updating span naming to the latest guidance to a future PR
6. I am deferring supporting `OTEL_INSTRUMENTATION_HTTP_KNOWN_METHODS` environment variable to a future PR

See https://github.com/open-telemetry/opentelemetry-ruby-contrib/issues/1779


instrumentation/ethon/lib/opentelemetry/instrumentation/ethon/http_helper.rb
```diff
@@ -0,0 +1,86 @@
...
+        private_constant :METHOD_CACHE, :OLD_SPAN_NAMES
+
+        # Prepares all span data for the specified semantic convention in a single call
+        # @param method [String, Symbol] The HTTP method
+        # @param semconv [Symbol] The semantic convention to use (:stable or :old)
+        # @return [SpanCreationAttributes] struct containing span_name, normalized_method, and original_method
+        def self.span_attrs_for(method, semconv: :stable)
+          normalized = METHOD_CACHE[method]
+          if normalized
+            span_name = semconv == :old ? OLD_SPAN_NAMES[normalized] : normalized
```
kaylareopelle, line 69: @arielvalentin - I'm concerned about changing the span names and attributes for the "old" instrumentation. It seems to break the contract for the `OTEL_SEMCONV_STABILITY_OPT_IN` var: 
> The default behavior (in the absence of one of these values) is to continue emitting whatever version of the old experimental HTTP and networking conventions the instrumentation was emitting previously.
[(source)](https://opentelemetry.io/docs/specs/semconv/http/)

What do you think about reverting the changes for the `old` modules and leaving the changes for `dup` and `new`?

arielvalentin: Ok that makes sense. I can leave the span name alone in follow up PRs.

Please consider that the existing `old` implementation does not follow the correct conventions for span names and should not have a prefix of `HTTP`: 

> HTTP spans MUST follow the overall [guidelines for span names](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.20.0/specification/trace/api.md#span). HTTP server span names SHOULD be {http.method} {http.route} if there is a (low-cardinality) http.route available. HTTP server span names SHOULD be {http.method} if there is no (low-cardinality) http.route available. HTTP client spans have no http.route attribute since client-side instrumentation is not generally aware of the "route", and therefore HTTP client spans SHOULD use {http.method}. Instrumentation MUST NOT default to using URI path as span name, but MAY provide hooks to allow custom logic to override the default span name.

arielvalentin: _may provide hooks to allow custom logic to override the default span name_ is also something new to me 🤔

kaylareopelle (commented):
I want to get @hannahramadan's opinion on this too, but I'm at the point where I feel comfortable pushing this fix to the old instrumentations. It's an unimplemented feature that we likely would have fixed if we noticed/had the time. The only request I have is to mark this as a breaking change so people can be alerted their span names are changing.

hannahramadan:
Hi @arielvalentin! I was initially a little iffy on the ethon span name change, but after chatting with @kaylareopelle, I agree that if we mark this as a breaking change, then fixing the span name (`HTTP N/A`-> `HTTP _OTHER`) would be a worthwhile improvement. Apologies for the back and forth - I see you already made the change back to the 'broken' state. I'm happy to fix it in a separate followup PR if we want to go ahead and merge this now.

kaylareopelle (approved):

PR #1788, opened 2025-11-14
feat: HTTP Client Semconv v1.17 Span Naming

arielvalentin:
~It corrects the span name for the old semantic conventions where the name should be set to the HTTP Method or the literal `HTTP` as well as adds the `url.template` to the name when using stable conventions.~

> HTTP spans MUST follow the overall [guidelines for span names](https://opentelemetry.io/docs/specs/otel/trace/api/#span).
> 
> HTTP span names SHOULD be {method} {target} if there is a (low-cardinality) target available. If there is no (low-cardinality) {target} available, HTTP span names SHOULD be {method}.
> 
> (see below for the exact definition of the [{method}](https://opentelemetry.io/docs/specs/semconv/http/http-spans/#method-placeholder) and [{target}](https://opentelemetry.io/docs/specs/semconv/http/http-spans/#target-placeholder) placeholders).
> 
> The {method} MUST be {http.request.method} if the method represents the original method known to the instrumentation. In other cases (when {http.request.method} is set to _OTHER), {method} MUST be HTTP.
> 
> The {target} SHOULD be one of the following:
> 
> [http.route](https://opentelemetry.io/docs/specs/semconv/registry/attributes/http/) for HTTP Server spans
> [url.template](https://opentelemetry.io/docs/specs/semconv/registry/attributes/url/) for HTTP Client spans if enabled and available (Development)
> Other value MAY be provided through custom hooks at span start time or later.
> Instrumentation MUST NOT default to using URI path as a {target}.

~More explicitly stated, when using `old` conventions the span name will be one of the known HTTP method (`GET`, `PUT`, `POST`, `PATCH`, `DELETE`...) or `HTTP` if the method is non-standard.~ 

When using `dup` or `stable` conventions the name _may_ contain a `url.template` value if it is known at the time or the request. Users may provide a `url.template` via the `ClientContext` attributes.  

Some concessions:

1. This change intentionally preserves structural and knowledge duplication in each client gem as a secibd step refactoring to extract and move the functionality into a shared helper gem or the semantic conventions gem.
2. I am deferring supporting OTEL_INSTRUMENTATION_HTTP_KNOWN_METHODS environment variable to a future PR

See https://github.com/open-telemetry/opentelemetry-ruby-contrib/issues/1779

## Notice to reviewers

_This is a follow up to_ https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/1781


hannahramadan:
@arielvalentin overall looks great! The one thing that caught my eye are the edits for `old` span names. In a prior PR, I know we fixed old span names that were implemented incorrectly when there was an unknown method: `HTTP N/A` -> `HTTP`.

For this PR, we are changing old, but [old spec compliant](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.17.0/specification/trace/semantic_conventions/http.md#name), spans: `HTTP GET` -> `GET`.

Should we be updating these names?

instrumentation/all/Gemfile
```diff
@@ -33,6 +33,7 @@ group :test do
   Dir.entries('../')
      .select { |entry| File.directory?(File.join('../', entry)) }
      .reject { |entry| excluded_instrumentations.include?(entry) }
+     .reject { |entry| entry.start_with?('.') }
```
kaylareopelle: Is this to exclude the instrumentation-all gem from the Gemfile?

Rakefile
```diff
@@ -36,7 +36,8 @@ task default: [:each]
 
 def foreach_gem(cmd)
   Dir.glob("**/opentelemetry-*.gemspec") do |gemspec|
-    name = File.basename(gemspec, ".gemspec")
+    next if gemspec.start_with?('.')
```
kaylareopelle: What is this protecting against?

kaylareopelle (approved):
THANK YOU for your patience with this review. I have two small questions about some of the gem loop changes, but they're non-blocking.

Rakefile
arielvalentin: It's been so long I don't remember. I think it was from a case where I was performing a local gem install.

I could revert this.

arielvalentin:
```suggestion
```

instrumentation/all/Gemfile
```diff
@@ -32,6 +32,7 @@ group :test do
...
```
arielvalentin:
```suggestion
```

Rakefile
```diff
@@ -36,7 +36,7 @@ task default: [:each]
 
 def foreach_gem(cmd)
   Dir.glob("**/opentelemetry-*.gemspec") do |gemspec|
-    name = File.basename(gemspec, ".gemspec")
+
```
arielvalentin:
```suggestion
    name = File.basename(gemspec, ".gemspec")
```

PR #1796, opened 2025-11-21
fix: Sidekiq spans crashing when enqueued by Exq

### LINKED ISSUE #1795: Sidekiq timestamp incompatibility results in NoMethodError
If the `created_at` field is missing on a Sidekiq message (for example, if the message was created via `Exq` in Elixir), this library will raise `undefined method '/' for nil` and a stack that ends with:
```
 /.../vendor/bundle/ruby/3.3.0/gems/opentelemetry-instrumentation-sidekiq-0.26.1/lib/opentelemetry/instrumentation/sidekiq/middlewares/server/tracer_middleware.rb:78:in `time_from_timestamp'
       /.../vendor/bundle/ruby/3.3.0/gems/opentelemetry-instrumentation-sidekiq-0.26.1/lib/opentelemetry/instrumentation/sidekiq/middlewares/server/tracer_middleware.rb:34:in `call'
       /.../vendor/bundle/ruby/3.3.0/gems/sidekiq-7.1.6/lib/sidekiq/middleware/chain.rb:182:in `traverse'
```
Broken by [this commit](https://github.com/open-telemetry/opentelemetry-ruby-contrib/commit/f1dd08fb15f22ee89eb8c99748b21bb5d6c2fdb4), as it tries to convert a nonexistent `created_at` 

Simplified Reproduction:

1) With a Sidekiq server that's using opentelemetry-instrumentation-sidekiq
2) Add a Sidekiq-compatible job to Redis that doesn't include `created_at` via manual enqueue or Exq or some other method

<sub>**Tip**: [React](https://github.blog/news-insights/product-news/add-reactions-to-pull-requests-issues-and-comments/) with 👍 to help prioritize this issue. Please use comments to provide useful context, avoiding `+1` or `me too`, to help us triage it. Learn more [in our end user docs](https://opentelemetry.io/community/end-user/issue-participation/).</sub>

xathien:
Exq doesn't seem to pass "created_at" with its messages. [Current Sidekiq API](https://github.com/sidekiq/sidekiq/blob/f9e0a026543c93ae49c1d1d41c6592d2877d318a/lib/sidekiq/api.rb#L456) defaults `created_at` to `enqueued_at` or `0` if it's not available, so we can mimic that method for safely defaulting these values.

Fixes https://github.com/open-telemetry/opentelemetry-ruby-contrib/issues/1795

### DIFF
--- a/instrumentation/sidekiq/lib/opentelemetry/instrumentation/sidekiq/middlewares/server/tracer_middleware.rb
+++ b/instrumentation/sidekiq/lib/opentelemetry/instrumentation/sidekiq/middlewares/server/tracer_middleware.rb
@@ -31,13 +31,10 @@ module OpenTelemetry
                           end
 
               extracted_context = OpenTelemetry.propagation.extract(msg)
-              created_at = time_from_timestamp(msg['created_at'] || msg['enqueued_at'] || 0)
-              enqueued_at = time_from_timestamp(msg['enqueued_at'] || 0)
               OpenTelemetry::Context.with_current(extracted_context) do
                 if instrumentation_config[:propagation_style] == :child
                   tracer.in_span(span_name, attributes: attributes, kind: :consumer) do |span|
-                    span.add_event('created_at', timestamp: created_at)
-                    span.add_event('enqueued_at', timestamp: enqueued_at)
+                    add_span_timestamps(span, msg)
                     yield
                   end
                 else
@@ -46,8 +43,7 @@ module OpenTelemetry
                   links << OpenTelemetry::Trace::Link.new(span_context) if instrumentation_config[:propagation_style] == :link && span_context.valid?
                   span = tracer.start_root_span(span_name, attributes: attributes, links: links, kind: :consumer)
                   OpenTelemetry::Trace.with_span(span) do
-                    span.add_event('created_at', timestamp: created_at)
-                    span.add_event('enqueued_at', timestamp: enqueued_at)
+                    add_span_timestamps(span, msg)
                     yield
                   rescue Exception => e # rubocop:disable Lint/RescueException
                     span.record_exception(e)
@@ -68,6 +64,13 @@ module OpenTelemetry
 
             def tracer
               Sidekiq::Instrumentation.instance.tracer
+            end
+
+            def add_span_timestamps(span, msg)
+              created_at = msg['created_at']
+              enqueued_at = msg['enqueued_at']
+              span.add_event('created_at', timestamp: time_from_timestamp(created_at)) if created_at
+              span.add_event('enqueued_at', timestamp: time_from_timestamp(enqueued_at)) if enqueued_at
             end
 
             def time_from_timestamp(timestamp)
--- a/instrumentation/sidekiq/lib/opentelemetry/instrumentation/sidekiq/middlewares/server/tracer_middleware.rb
+++ b/instrumentation/sidekiq/lib/opentelemetry/instrumentation/sidekiq/middlewares/server/tracer_middleware.rb
@@ -31,8 +31,8 @@ module OpenTelemetry
                           end
 
               extracted_context = OpenTelemetry.propagation.extract(msg)
-              created_at = time_from_timestamp(msg['created_at'])
-              enqueued_at = time_from_timestamp(msg['created_at'])
+              created_at = time_from_timestamp(msg['created_at'] || msg['enqueued_at'] || 0)
+              enqueued_at = time_from_timestamp(msg['enqueued_at'] || 0)
               OpenTelemetry::Context.with_current(extracted_context) do
                 if instrumentation_config[:propagation_style] == :child
                   tracer.in_span(span_name, attributes: attributes, kind: :consumer) do |span|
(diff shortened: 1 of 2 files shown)


instrumentation/sidekiq/lib/opentelemetry/instrumentation/sidekiq/middlewares/server/tracer_middleware.rb
```diff
@@ -31,8 +31,8 @@ def call(_worker, msg, _queue)
                           end
 
               extracted_context = OpenTelemetry.propagation.extract(msg)
-              created_at = time_from_timestamp(msg['created_at'])
-              enqueued_at = time_from_timestamp(msg['created_at'])
+              created_at = time_from_timestamp(msg['created_at'] || msg['enqueued_at'] || 0)
+              enqueued_at = time_from_timestamp(msg['enqueued_at'] || 0)
```
arielvalentin: Won't this end up loosing meaning? Would it not be better to omit the span events in these cases?

xathien: Probably! I'll admit I'm not an opentelemetry expert; just trying to fix the crash. 🙂 

Would it be more advisable to let `time_from_timestamp` handle/return `nil` and skip the `add_event` in those cases?

I think letting the `created_at` default to `enqueued_at` when not specified can still be useful, but maybe not so much the `0` default (which I only selected since Sidekiq's API does that).

dmathieu: IMHO, If the value is not set, we shouldn't have any event. Not provide misleading data.

dmathieu: IMHO, If the value is not set, we shouldn't have any event. Not provide misleading data.

instrumentation/sidekiq/test/opentelemetry/instrumentation/sidekiq/middlewares/server/tracer_middleware_test.rb
```diff
@@ -67,6 +67,31 @@
       _(job_span.attributes['messaging.operation']).must_equal 'process'
     end
 
+    it 'traces when enqueued through another system' do
```
dmathieu: I would name this as "when enqueued with minimal data"

instrumentation/sidekiq/test/opentelemetry/instrumentation/sidekiq/middlewares/server/tracer_middleware_test.rb
```diff
@@ -67,6 +67,31 @@
...
+      payload = { 'queue' => 'default', 'args' => [], 'class' => SimpleJob, 'enqueued_at' => Time.current, 'jid' => '4' }
```
dmathieu: Shouldn't enqueued_at be removed here too?

dmathieu (commented):
This will need a changelog entry

instrumentation/sidekiq/lib/opentelemetry/instrumentation/sidekiq/middlewares/server/tracer_middleware.rb
arielvalentin: @xathien think then that it's best to check if the values are present and only add the events if that's the case. I don't think the timestamp helper should tolerate nil values at this time.

instrumentation/sidekiq/test/opentelemetry/instrumentation/sidekiq/middlewares/server/tracer_middleware_test.rb
xathien: Exq supplies this one, but not `created_at` from what I can tell. In addition, `Sidekiq::Client#raw_push` adds `enqueued_at` on its own. I could try doing some shenanigans to push the job directly to Redis, but I'm not sure that effort would be worth the slight test improvement.

instrumentation/sidekiq/lib/opentelemetry/instrumentation/sidekiq/middlewares/server/tracer_middleware.rb
xathien: Sounds good to me! I've updated the change accordingly.

instrumentation/sidekiq/test/opentelemetry/instrumentation/sidekiq/middlewares/server/tracer_middleware_test.rb
dmathieu: IMHO, it's not about Exq. It's about what any custom library could do. If it's possible to enqueue a job and have it processed by sidekiq without an `enqueued_at, then we should handle it.
If sidekiq would auto-add it, or reject processing it then, we're good.

xathien: Seems like manual Redis shenanigans doesn't work properly with Sidekiq in test mode, but the test-mode `Queues.push` gave me a workaround! Added tests to verify each combination behaves as expected.

dmathieu (approved):

arielvalentin (approved):

PR #1818, opened 2025-11-27
feat: Add SQL Comment Propagator

### LINKED ISSUE #479: Adding trace context in sql query as the comments
When executing sql through ORM (e.g. active_record and sequel), I'd like to inject some comments (e.g. trace context) into sql query before hitting the sql database server. 

The most important comment is the `traceparent` that is consisted of trace context information (i.e. `00-%<trace_id>s-%<span_id>s-%<trace_flags>.2d`), which is for trace correlation.

```sql
# unjected query
SELECT "posts".* FROM "posts"

# injected query
SELECT "posts".* FROM "posts" /*action='index',application='SqlcommenterRailsDemo',controller='posts',db_driver='ActiveRecord::ConnectionAdapters::SQLite3Adapter',framework='rails_v6.0.0.rc1',route='/posts',traceparent='00-828f28f7fb3df3dd07ee6478b2016b2a-a52cad0a8d1425ab-01'*/ 

# example from https://github.com/open-telemetry/opentelemetry-sqlcommenter/tree/main/ruby/sqlcommenter-ruby/sqlcommenter_rails_demo#get-posts
```

[Marginalia](https://github.com/basecamp/marginalia) and [sqlcommenter](https://github.com/open-telemetry/opentelemetry-sqlcommenter/tree/main/ruby/sqlcommenter-ruby) should handle the case if both marginalia and sqlcommenter is required from end user. 

Will opentelemetry-ruby integrate the marginalia and sqlcommenter into the codebase in the future so that end user doesn't need to require in their side (as automatic as possible)?

I created the draft pr for opentelemtery integration with marginalia [here](https://github.com/open-telemetry/opentelemetry-sqlcommenter/pull/37).

arielvalentin:
Implements a helper to propagate context via SQL comments as specified here:

> Instrumentations MAY propagate context using SQL commenter by injecting comments into SQL queries before execution. SQL commenter-based context propagation SHOULD NOT be enabled by default, but instrumentation MAY allow users to opt into it.
>
> The instrumentation implementation SHOULD append the comment to the end of the query. Semantic conventions for individual database systems MAY specify different format, which may include different position, encoding, or schema, depending on the specific database system’s requirements or preferences.
>
> The instrumentation SHOULD allow users to pass a propagator to overwrite the global propagator. If no propagator is provided by the user, instrumentation SHOULD use the global propagator.

https://opentelemetry.io/docs/specs/semconv/database/database-spans/#context-propagation

This helper conforms to the propagator inject interface, since it does not ever extract values we are skipping that implementation.

~Once the processors replaces the obfuscator then we can include the propagators in the SQL gems for trilogy, mysql2, and pg~

Fixes https://github.com/open-telemetry/opentelemetry-ruby-contrib/issues/479


.toys/.data/releases.yml
```diff
@@ -300,4 +300,4 @@ gems:
 
   - name: opentelemetry-sampler-xray
     directory: sampler/xray
-    version_constant: [OpenTelemetry, Sampler, XRay, VERSION]
\ No newline at end of file
+    version_constant: [OpenTelemetry, Sampler, XRay, VERSION]
```
arielvalentin, line 301: unrelated but I am fixing a missing new line at the end of this file

instrumentation/mysql2/lib/opentelemetry/instrumentation/mysql2/patches/client.rb
```diff
@@ -18,8 +18,16 @@ def query(sql, options = {})
               _otel_span_name(sql),
               attributes: _otel_span_attributes(sql),
               kind: :client
-            ) do
-              super
+            ) do |_span, context|
+              if propagator && sql.frozen?
+                sql = +sql
```
hannahramadan, line 23: I did not know of this syntax - cool!

hannahramadan (approved):
I think this looks good and perfectly on-spec. Adding the dedicated Vitess propagator compatibility is a nice touch. Are there any other specific system compatibility/ handling we should think about for the future? I don't know quite enough about this (Vitess was new to me)

PR #1854, opened 2025-12-13
ci: Use local git repostiories for development

arielvalentin:
Tools like dependabot and renovate bot do not support updating gems that have a `path` attribute set, however they do support `git` sources.

This change is an attempt to use bundlers built in local git repository features to allow building. 🤞🏼 

https://bundler.io/man/bundle-config.1.html#LOCAL-GIT-REPOS


bin/setup-local-dev
```diff
@@ -0,0 +1,127 @@
...
+echo "[OK] Configured $DIR_COUNT gem directories with git dependencies"
+echo ""
+echo "You can now work on any gem and your changes will be"
+echo "immediately available to dependent gems."
+echo ""
+echo "Example workflow:"
+echo "  1. Edit code in instrumentation/base/"
+echo "  2. cd instrumentation/rails/"
+echo "  3. bundle install"
+echo "  4. bundle exec rake test"
```
kaylareopelle, line 117: Since most of the instrumentation tests need appraisal to run, what do you think about adjusting the instructions to include it?

```suggestion
echo "  3. bundle install"
echo "  4. bundle exec appraisal install"
echo "  5. bundle exec appraisal rake test"
```

bin/setup-local-dev
```diff
@@ -0,0 +1,127 @@
...
+echo ""
+echo "Your local changes from instrumentation/base will be used!"
+echo ""
+echo "To verify your configuration:"
+echo "  cd instrumentation/rails && bundle config local.opentelemetry-instrumentation-base"
```
kaylareopelle, line 122: I can't seem to get this to work on my machine. :( 

Even after running bundle config local for base and also running the global bundle config in the root directory[^1], it's still pulling from the git repository. 

[^1]: `bundle config local.opentelemetry-ruby-contrib /Users/arielvalentin/github/opentelemetry-ruby-contrib`

bin/setup-local-dev
kaylareopelle: Should the config step for base come before all the installing?

guides/local-development.md
```diff
@@ -0,0 +1,117 @@
...
+   bundle config local.opentelemetry-ruby-contrib /Users/arielvalentin/github/opentelemetry-ruby-contrib
+   ```
+
+   Replace the path with the actual path to your local clone.
+
+3. **Install dependencies**:
+
+   ```bash
+   bundle install
+   ```
```
arielvalentin, line 46: This is confusing. Lets remove it
```suggestion
```

PR #1899, opened 2026-01-02
chore: Block ruby uses-with updates

thompson-tomo:
This should hopefully unblock #1861 by ignoring the ruby dependency when in a github action.


arielvalentin:
@thompson-tomo would you kindly resolve the conflicts so that I may review this change?

.github/renovate.json5
```diff
@@ -13,6 +13,10 @@
     {
       "matchUpdateTypes": ["minor", "major"],
       "schedule": ["before 8am on Monday"]
+    },
```
arielvalentin, line 22: I think the actions I am mostly concerned about are the `release` actions. 

How would you feel about limiting the checks to only include those actions?

https://docs.renovatebot.com/configuration-options/#matchfilenames

thompson-tomo: Hadn't thought of that but I agree.

thompson-tomo:
@arielvalentin no worries, conflict resolved & change to rule scope made.

kaylareopelle (approved):

arielvalentin (commented):
We have multiple custom actions that support our release process. We never successfully migrated to using release please so I would ask that you include all release related yaml files part of this.

thompson-tomo:
Ahh, I see the confusion. I wrote the rule based on what renovate is detecting as usages of that dependency. I have update the rule to use a pattern now.

arielvalentin (approved):

PR #1901, opened 2026-01-03
chore: Remove Explicit Bundler Dependency

arielvalentin:
Declaring bundler as a dependency is therefore redundant and our pessemistic versioning constraints blocks Ruby 4 upgrades:
- https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/1885
- https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/1892

Bundler is included with RubyGems by default so declaring it separately seems unecessary


.github/actions/test_gem/action.yml
```diff
@@ -84,21 +84,19 @@ runs:
     # ...but not for appraisals, sadly.
     - name: Install Ruby ${{ inputs.ruby }} with dependencies
       if: "${{ steps.setup.outputs.appraisals == 'false' }}"
-      uses: ruby/setup-ruby@3fee6763234110473bd57dd4595c5199fce2c510 # v1.258.0
+      uses: ruby/setup-ruby@4c24fa5ec04b2e79eb40571b1cee2a0d2b705771 # v1.278.0
```
xuan-cao-swi, line 87: Is that possible to use `ruby/setup-ruby@v1.278`?

arielvalentin: Best security practices is to use a pined sha since the `@vX` versioning scheme is a "reference" like a tag name or branch so I believe renovatebot started to update our actions to use shas with a comment like this.

xuan-cao-swi (approved):

PR #1903, opened 2026-01-03
ci: Add Ruby 4 Tests

arielvalentin:
Ensures that all contrib packages are compatible with Ruby 4

The notable changes are:

- `logger`: The Ruby logger gem has been removed and is no longer automatically loaded. Since our tests make extensive use of a Logger, I have added it to test dependencies 
- `delayed_job`: The included testing library we use in our site requires `ostruct`
- `gruf`: The library has strict versioning Ruby requirement `< 3.5` 
- `rack`: We must use stable versions that are continuing to receive security updates
- `sinatra`: Relies on specific versions of `rack` but the gem requirements are very permissive


instrumentation/gruf/Gemfile
```diff
@@ -19,6 +18,8 @@ group :test do
   gem 'simplecov', '~> 0.22.0'
   gem 'webmock', '~> 3.24'
   gem 'yard', '~> 0.9'
+  # TODO: regenerate protobuf files for protobuf 4.x compatibility
```
kaylareopelle: Do we have an issue for this?

kaylareopelle (commented):
This looks good to me! Thanks for taking this on and for cleaning up bundler. Once the conflicts are resolved, I'll approve.

instrumentation/gruf/Gemfile
arielvalentin: No that was completed here https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/1911

instrumentation/gruf/Gemfile
```diff
@@ -18,6 +18,8 @@ group :test do
...
+  gem 'google-protobuf', '~> 3.25'
```
arielvalentin:
```suggestion
```

xuan-cao-swi (approved):

PR #1904, opened 2026-01-05
feat: Add Rage instrumentation

rsamoilov:
## Summary

This PR adds OpenTelemetry instrumentation for [Rage](https://github.com/rage-rb/rage), a fiber-based Ruby web framework.

The instrumentation leverages Rage's built-in [Telemetry interface](https://rage-rb.dev/docs/telemetry) to instrument:

- HTTP requests - Enriches Rack spans with controller/action information and route patterns
- WebSocket connections - Traces connections, channel subscriptions, and broadcasts with span linking
- Event bus - Instruments event publishing and subscriber processing with context propagation
- Background jobs - Traces job enqueuing and execution with context propagation and retry tracking
- Fiber context propagation - Ensures OpenTelemetry context flows correctly across application-level fibers

**Implementation Details:**

- Uses Semantic Conventions v1.36+
- Implements span linking to tie WebSocket connections back to initial handshakes
- Adds trace/span IDs to Rage logs

~~**Note:** The Rage version compatible with this instrumentation (1.20.0+) has not been released yet. This PR is intended to be merged before the Rage release so that the instrumentation is available immediately when users upgrade.~~


arielvalentin:
Thank you @rsamoilov for your contribution. I see that you are a core contributor of `rage-rb` so I am pleased that you all have taken the time to instrument the framework using OpenTelemetry! 

Our preference would eventually be for library maintainers to maintain first party instrumentations in their respective repositories as opposed to the contrib package. In other words, have the `rage-rb-otel` instrumentation be included in your library and we work with you to publish it in the OTel registry. 

This would give you the flexibility to maintain the instrumentation as a part of your organization and reduce friction trying to get changes merged into the contrib repo. We have limited resources and lack the expertise in the framework so it will be difficult for us to provide you with a thorough review. 

The areas where I think that we will be able to help with code-reviews or unblock you when faced with incompatibilities with the SDK.

Is this something you would be amenable to?

rsamoilov:
Hi @arielvalentin 

This makes sense to me. I will create a repo and get back to you. Thanks!

instrumentation/rage/lib/opentelemetry/instrumentation/rage/handlers/cable.rb
```diff
@@ -0,0 +1,146 @@
...
+          # @param env [Hash] the Rack env
+          # @param action [:connect, :disconnect] the name of the action being processed
+          # @param connection [Rage::Cable::Connection] the connection instance
+          def self.create_connection_span(env:, action:, connection:)
+            handshake_context = env[HANDSHAKE_CONTEXT]
+            handshake_link = env[HANDSHAKE_LINK]
+
+            OpenTelemetry::Context.with_current(handshake_context) do
+              attributes = {
+                SemConv::Incubating::MESSAGING::MESSAGING_SYSTEM => 'rage.cable',
```
thompson-tomo: Are we sure that this is a messaging system? Is this not the websocket framework https://rage-rb.dev/docs/websockets/#action-cable-protocol

instrumentation/rage/lib/opentelemetry/instrumentation/rage/handlers/cable.rb
```diff
@@ -0,0 +1,146 @@
...
+          # @param env [Hash] the Rack env
+          # @param action [Symbol] the name of the action being processed
+          # @param channel [Rage::Cable::Channel] the channel instance
+          def self.create_channel_span(env:, action:, channel:)
+            handshake_context = env[HANDSHAKE_CONTEXT]
+            handshake_link = env[HANDSHAKE_LINK]
+
+            OpenTelemetry::Context.with_current(handshake_context) do
+              attributes = {
+                SemConv::Incubating::MESSAGING::MESSAGING_SYSTEM => 'rage.cable',
```
thompson-tomo: As above

instrumentation/rage/lib/opentelemetry/instrumentation/rage/handlers/cable.rb
```diff
@@ -0,0 +1,146 @@
...
+              ensure
+                span.finish
+              end
+            end
+          end
+
+          # @param stream [String] the name of the stream
+          def self.create_broadcast_span(stream:)
+            attributes = {
+              SemConv::Incubating::MESSAGING::MESSAGING_SYSTEM => 'rage.cable',
```
thompson-tomo: As above

instrumentation/rage/lib/opentelemetry/instrumentation/rage/handlers/deferred.rb
```diff
@@ -0,0 +1,84 @@
...
+        # enqueuing and processing of deferred tasks in spans.
+        class Deferred < ::Rage::Telemetry::Handler
+          handle 'deferred.task.enqueue', with: :create_enqueue_span
+          handle 'deferred.task.process', with: :create_perform_span
+
+          # @param task_class [Class] the class of the deferred task
+          # @param task_context [Hash] the context for the deferred task
+          def self.create_enqueue_span(task_class:, task_context:)
+            attributes = {
+              SemConv::Incubating::MESSAGING::MESSAGING_SYSTEM => 'rage.deferred',
```
thompson-tomo: Are we sure this is a messaging system. Sounds more like a Workflow/task system.

instrumentation/rage/lib/opentelemetry/instrumentation/rage/handlers/deferred.rb
```diff
@@ -0,0 +1,84 @@
...
+
+          # @param task_class [Class] the class of the deferred task
+          # @param task [Rage::Deferred::Task] the deferred task instance
+          # @param task_context [Hash] the context for the deferred task
+          def self.create_perform_span(task_class:, task:, task_context:)
+            otel_context = OpenTelemetry.propagation.extract(task_context)
+
+            OpenTelemetry::Context.with_current(otel_context) do
+              attributes = {
+                SemConv::Incubating::MESSAGING::MESSAGING_SYSTEM => 'rage.deferred',
```
thompson-tomo: As above

instrumentation/rage/lib/opentelemetry/instrumentation/rage/handlers/events.rb
```diff
@@ -0,0 +1,61 @@
...
+          def self.create_subscriber_span(event:, subscriber:)
+            # deferred subscribers will be wrapped into spans by the `Handlers::Deferred` handler
+            return yield if subscriber.class.deferred?
+
+            kind = :consumer
+            attributes = {
+              SemConv::Incubating::MESSAGING::MESSAGING_SYSTEM => 'rage.events',
+              SemConv::Incubating::MESSAGING::MESSAGING_OPERATION_TYPE => 'process',
+              SemConv::Incubating::MESSAGING::MESSAGING_DESTINATION_NAME => event.class.name,
+              SemConv::Incubating::CODE::CODE_FUNCTION_NAME => "#{subscriber.class}#call"
```
thompson-tomo: Is this needed?

instrumentation/rage/lib/opentelemetry/instrumentation/rage/handlers/request.rb
```diff
@@ -0,0 +1,44 @@
...
+          def self.enrich_request_span(controller:, request:)
+            span = OpenTelemetry::Instrumentation::Rack.current_span
+            return yield unless span.recording?
+
+            http_route = request.route_uri_pattern
+            span.name = "#{request.method} #{http_route}"
+
+            attributes = {
+              SemConv::HTTP::HTTP_ROUTE => http_route,
+              SemConv::Incubating::CODE::CODE_FUNCTION_NAME => "#{controller.class}##{controller.action_name}"
```
thompson-tomo: Is this needed? If we are going to keep them, then look at what other frameworks do.

instrumentation/rage/lib/opentelemetry/instrumentation/rage/handlers/request.rb
```diff
@@ -0,0 +1,44 @@
...
+          # @param controller [RageController::API] the controller processing the request
+          # @param request [Rage::Request] the request being processed
+          def self.enrich_request_span(controller:, request:)
+            span = OpenTelemetry::Instrumentation::Rack.current_span
+            return yield unless span.recording?
+
+            http_route = request.route_uri_pattern
+            span.name = "#{request.method} #{http_route}"
+
+            attributes = {
```
thompson-tomo: Is there any additional semconv attributes that could be added?

instrumentation/rage/lib/opentelemetry/instrumentation/rage/handlers/cable.rb
rsamoilov: I guess it depends on your definition of a messaging system. I added this attribute because [socket.io uses it](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/451caaf158d9bf9fa50cccec9bd48b9b58e59efa/packages/instrumentation-socket.io/src/socket.io.ts#L301). 

Do you think `otel.scope.name` would fit better here?

instrumentation/rage/lib/opentelemetry/instrumentation/rage/handlers/deferred.rb
rsamoilov: All Ruby background systems (e.g. [Sidekiq](https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/b9e025b696a9ae1ba5dd305a91d8c764048b308d/instrumentation/sidekiq/lib/opentelemetry/instrumentation/sidekiq/middlewares/client/tracer_middleware.rb#L19) or [Que](https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/b9e025b696a9ae1ba5dd305a91d8c764048b308d/instrumentation/que/lib/opentelemetry/instrumentation/que/middlewares/server_middleware.rb#L53)) use this attribute.

Do you think `otel.scope.name` would fit better here?

instrumentation/rage/lib/opentelemetry/instrumentation/rage/handlers/events.rb
rsamoilov: Yes, I would prefer to leave it as it substitutes `code.namespace` and `code.function` which are widely used in other instrumentations. Is it the correct usage of the attribute?

instrumentation/rage/lib/opentelemetry/instrumentation/rage/handlers/request.rb
rsamoilov: Not that I can think of - these are pretty much the same attributes Action Pack adds to requests. Any recommendations?

instrumentation/rage/lib/opentelemetry/instrumentation/rage/handlers/deferred.rb
```diff
@@ -0,0 +1,84 @@
...
+                SemConv::Incubating::MESSAGING::MESSAGING_OPERATION_TYPE => 'process',
+                SemConv::Incubating::MESSAGING::MESSAGING_DESTINATION_NAME => task_class.name,
+                SemConv::Incubating::CODE::CODE_FUNCTION_NAME => "#{task_class}#perform"
+              }
+
+              attributes['messaging.message.delivery_attempt'] = task.meta.attempts if task.meta.retrying?
```
rsamoilov: @thompson-tomo Any feedback on the name of this attribute? For context, here we are instrumenting background jobs that work similar to those in Sidekiq.

instrumentation/rage/lib/opentelemetry/instrumentation/rage/handlers/request.rb
thompson-tomo: Take a look at https://opentelemetry.io/docs/specs/semconv/http/http-spans/#http-client-span

instrumentation/rage/lib/opentelemetry/instrumentation/rage/handlers/events.rb
thompson-tomo: So in semantic conventions the attributes are not defined on any of the applicable definitions which make means they are opt-in via Config. To me they seem more relevant to logging.

instrumentation/rage/lib/opentelemetry/instrumentation/rage/handlers/deferred.rb
thompson-tomo: That doesn't make it correct. In fact we have run into issues with sidekiq because they have used messaging.

I did attempt to introduce workflows but I needed a project to be formed etc. Anyway the proposal I had was https://github.com/thompson-tomo/semantic-conventions/pull/146

instrumentation/rage/lib/opentelemetry/instrumentation/rage/handlers/cable.rb
thompson-tomo: Based on the fact that semantic conventions doesn't classify signalr or mqtt as messaging systems. I couldn't get an explanation of the requirements/definition.

Looking at the definitions, the span kind also doesn't align with messaging which uses producer/consumer and not server/client. 

I think it is closer to rpc conventions

instrumentation/rage/lib/opentelemetry/instrumentation/rage/handlers/events.rb
rsamoilov: > To me they seem more relevant to logging

Hmm, fair. Let me think about it a bit more.

instrumentation/rage/lib/opentelemetry/instrumentation/rage/handlers/request.rb
rsamoilov: You're referring to a client span for outbound connections. `handlers/request.rb` processes inbound connections.

instrumentation/rage/lib/opentelemetry/instrumentation/rage/handlers/deferred.rb
rsamoilov: Using the `workflow` namespace is fine by me, I guess. The only thing is that your proposal is still in review. How do you recommend we proceed?

instrumentation/rage/lib/opentelemetry/instrumentation/rage/handlers/cable.rb
rsamoilov: > I think it is closer to rpc conventions

Agreed, but RPC and WebSockets are on different abstraction levels.

Same as with `handlers/deferred.rb`, I'm also using the `messaging.operation.type` and `messaging.destination.name` attributes. What is your recommendation regarding those?

instrumentation/rage/lib/opentelemetry/instrumentation/rage/handlers/cable.rb
```diff
@@ -0,0 +1,146 @@
...
+          # @param action [:connect, :disconnect] the name of the action being processed
+          # @param connection [Rage::Cable::Connection] the connection instance
+          def self.create_connection_span(env:, action:, connection:)
+            handshake_context = env[HANDSHAKE_CONTEXT]
+            handshake_link = env[HANDSHAKE_LINK]
+
+            OpenTelemetry::Context.with_current(handshake_context) do
+              attributes = {
+                SemConv::Incubating::MESSAGING::MESSAGING_SYSTEM => 'rage.cable',
+                SemConv::Incubating::MESSAGING::MESSAGING_DESTINATION_NAME => connection.class.name,
```
thompson-tomo:
```suggestion
                'rage.cable.stream.name' => connection.class.name,
```

instrumentation/rage/lib/opentelemetry/instrumentation/rage/handlers/cable.rb
```diff
@@ -0,0 +1,146 @@
...
+          # @param action [Symbol] the name of the action being processed
+          # @param channel [Rage::Cable::Channel] the channel instance
+          def self.create_channel_span(env:, action:, channel:)
+            handshake_context = env[HANDSHAKE_CONTEXT]
+            handshake_link = env[HANDSHAKE_LINK]
+
+            OpenTelemetry::Context.with_current(handshake_context) do
+              attributes = {
+                SemConv::Incubating::MESSAGING::MESSAGING_SYSTEM => 'rage.cable',
+                SemConv::Incubating::MESSAGING::MESSAGING_DESTINATION_NAME => channel.class.name,
```
thompson-tomo:
```suggestion
                'rage.cable.stream.name' => connection.class.name,
```

instrumentation/rage/lib/opentelemetry/instrumentation/rage/handlers/cable.rb
```diff
@@ -0,0 +1,146 @@
...
+          # @param connection [Rage::Cable::Connection] the connection instance
+          def self.create_connection_span(env:, action:, connection:)
+            handshake_context = env[HANDSHAKE_CONTEXT]
+            handshake_link = env[HANDSHAKE_LINK]
+
+            OpenTelemetry::Context.with_current(handshake_context) do
+              attributes = {
+                SemConv::Incubating::MESSAGING::MESSAGING_SYSTEM => 'rage.cable',
+                SemConv::Incubating::MESSAGING::MESSAGING_DESTINATION_NAME => connection.class.name,
+                SemConv::Incubating::CODE::CODE_FUNCTION_NAME => "#{connection.class}##{action}"
```
thompson-tomo:
```suggestion
                'rpc.operation.type' => "{action}"
```

instrumentation/rage/lib/opentelemetry/instrumentation/rage/handlers/cable.rb
```diff
@@ -0,0 +1,146 @@
...
+          # @param channel [Rage::Cable::Channel] the channel instance
+          def self.create_channel_span(env:, action:, channel:)
+            handshake_context = env[HANDSHAKE_CONTEXT]
+            handshake_link = env[HANDSHAKE_LINK]
+
+            OpenTelemetry::Context.with_current(handshake_context) do
+              attributes = {
+                SemConv::Incubating::MESSAGING::MESSAGING_SYSTEM => 'rage.cable',
+                SemConv::Incubating::MESSAGING::MESSAGING_DESTINATION_NAME => channel.class.name,
+                SemConv::Incubating::CODE::CODE_FUNCTION_NAME => "#{channel.class}##{action}"
```
thompson-tomo:
```suggestion
                'rpc.operation.type' => "{action}"
```

instrumentation/rage/lib/opentelemetry/instrumentation/rage/handlers/cable.rb
```diff
@@ -0,0 +1,146 @@
...
+            env[HANDSHAKE_CONTEXT] = OpenTelemetry::Context.current
+            env[HANDSHAKE_LINK] = [OpenTelemetry::Trace::Link.new(span.context)]
+
+            yield
+          end
+
+          # @param env [Hash] the Rack env
+          # @param action [:connect, :disconnect] the name of the action being processed
+          # @param connection [Rage::Cable::Connection] the connection instance
+          def self.create_connection_span(env:, action:, connection:)
```
thompson-tomo, line 46: Is this called from the server or client?

instrumentation/rage/lib/opentelemetry/instrumentation/rage/handlers/cable.rb
```diff
@@ -0,0 +1,146 @@
...
+            handshake_link = env[HANDSHAKE_LINK]
+
+            OpenTelemetry::Context.with_current(handshake_context) do
+              attributes = {
+                SemConv::Incubating::MESSAGING::MESSAGING_SYSTEM => 'rage.cable',
+                SemConv::Incubating::MESSAGING::MESSAGING_DESTINATION_NAME => channel.class.name,
+                SemConv::Incubating::CODE::CODE_FUNCTION_NAME => "#{channel.class}##{action}"
+              }
+
+              attributes[SemConv::Incubating::MESSAGING::MESSAGING_OPERATION_TYPE] = 'receive' unless action == :unsubscribed
```
thompson-tomo:
```suggestion
```

instrumentation/rage/lib/opentelemetry/instrumentation/rage/handlers/cable.rb
```diff
@@ -0,0 +1,146 @@
...
+              ensure
+                span.finish
+              end
+            end
+          end
+
+          # @param env [Hash] the Rack env
+          # @param action [Symbol] the name of the action being processed
+          # @param channel [Rage::Cable::Channel] the channel instance
+          def self.create_channel_span(env:, action:, channel:)
```
thompson-tomo, line 83: Is this called from the server or client?

instrumentation/rage/lib/opentelemetry/instrumentation/rage/handlers/cable.rb
```diff
@@ -0,0 +1,146 @@
...
+              end
+            end
+          end
+
+          # @param stream [String] the name of the stream
+          def self.create_broadcast_span(stream:)
+            attributes = {
+              SemConv::Incubating::MESSAGING::MESSAGING_SYSTEM => 'rage.cable',
+              SemConv::Incubating::MESSAGING::MESSAGING_OPERATION_TYPE => 'publish',
+              SemConv::Incubating::MESSAGING::MESSAGING_DESTINATION_NAME => stream
```
thompson-tomo:
```suggestion
                'rage.cable.stream.name' => connection.class.name,
```

instrumentation/rage/lib/opentelemetry/instrumentation/rage/handlers/cable.rb
```diff
@@ -0,0 +1,146 @@
...
+                span.finish
+              end
+            end
+          end
+
+          # @param stream [String] the name of the stream
+          def self.create_broadcast_span(stream:)
+            attributes = {
+              SemConv::Incubating::MESSAGING::MESSAGING_SYSTEM => 'rage.cable',
+              SemConv::Incubating::MESSAGING::MESSAGING_OPERATION_TYPE => 'publish',
```
thompson-tomo:
```suggestion
              'rpc.operation.type' => 'publish',
```

instrumentation/rage/lib/opentelemetry/instrumentation/rage/handlers/deferred.rb
thompson-tomo: The Workflow proposal should cater for this.

instrumentation/rage/lib/opentelemetry/instrumentation/rage/handlers/events.rb
thompson-tomo:
```suggestion
              'messaging.client.id' => subscriber.class
```

thompson-tomo: See latest thought in https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/1904#discussion_r2675422187

instrumentation/rage/lib/opentelemetry/instrumentation/rage/handlers/request.rb
thompson-tomo: In that case look at the server span.

instrumentation/rage/lib/opentelemetry/instrumentation/rage/handlers/cable.rb
rsamoilov: It's called once a client establishes a WebSocket connection.

instrumentation/rage/lib/opentelemetry/instrumentation/rage/handlers/cable.rb
rsamoilov: It's called once a WebSocket client subscribes to a specific channel.

instrumentation/rage/lib/opentelemetry/instrumentation/rage/handlers/request.rb
rsamoilov: Most of those attribute are already being added to the span inside the Rack instrumentation:

- `http.request.method`
- `server.address`
- `url.scheme`
- `url.path`
- `url.query`
- `user_agent.original`
- `http.route`
- `http.response.status_code`

instrumentation/rage/lib/opentelemetry/instrumentation/rage/handlers/request.rb
rsamoilov: [Other frameworks](https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/90f1f4479d75d87d243b679fe25dd5f4a72e4c31/instrumentation/action_pack/lib/opentelemetry/instrumentation/action_pack/handlers/action_controller.rb#L35-L38) use `code.namespace` and `code.function` which are deprecated in favor of `code.function.name`.

instrumentation/rage/lib/opentelemetry/instrumentation/rage/handlers/cable.rb
thompson-tomo: Ok, but I am more asking is it executed by the application which initiated the connection or the server which was connected to.

instrumentation/rage/lib/opentelemetry/instrumentation/rage/handlers/cable.rb
thompson-tomo: Ok, but I am more asking is it executed by the application which initiated the connection or the server which was connected to.

instrumentation/rage/lib/opentelemetry/instrumentation/rage/handlers/cable.rb
thompson-tomo: > Same as with Action Cable, streams aren't properties of connections - they are being created when the application code executes stream_from. A single client can be associated with zero or multiple streams, so using rage.cable.stream.name when accepting connections/subscriptions doesn't seem semantically correct.
> 
> Using messaging.destination.name with the connection/channel class name identifies the actual destination of the message.

It might "fit" into messaging on first glance but action cable is not messaging hence need another solution.

Perhaps we go with 
```suggestion
                'ruby.action_cable.class.name' => connection.class.name,
```

instrumentation/rage/lib/opentelemetry/instrumentation/rage/handlers/cable.rb
thompson-tomo: As above

instrumentation/rage/lib/opentelemetry/instrumentation/rage/handlers/events.rb
thompson-tomo: > Looking at other implementations and OTel docs, messaging.client.id seems to be used to identify the instance that processes the message (process/host) rather than the code being executed.

You might be right, Howabout 
```suggestion
              'messaging.client.type' => subscriber.class
```

instrumentation/rage/lib/opentelemetry/instrumentation/rage/handlers/cable.rb
thompson-tomo: > Cable is a pub/sub (broadcast to subscribers) system, not RPC (call and wait for response). While channels can be triggered usin RPC actions, there's no return value. Using rpc.operation.type essentially misclassifies the system.

RPC also considers streaming which is not request and response.

instrumentation/rage/lib/opentelemetry/instrumentation/rage/handlers/deferred.rb
thompson-tomo: > Background job queues have all attributes of messaging systems; since workflow conventions are still a proposal, I'd propose using established messaging conventions now and migrating to workflow once it's stable.

But it lacks the key thing, it doesn't satisfy the definition of a messaging system. We should avoid doing things which we know is wrong just because that is how they were previously done and instead try to make it as correct as possible.

rsamoilov: Makes sense. I'm all up for using the `workflow` namespace, but I can't rely on a proposal that may never be accepted. Do you know what the timeline for the workflow proposal is?

instrumentation/rage/lib/opentelemetry/instrumentation/rage/handlers/events.rb
rsamoilov: I don't think I understand the problem we're trying to solve here. Why would we want to invent a new attribute instead of using the standard `code.function.name`, which is specifically supposed to represent the function being called?

instrumentation/rage/lib/opentelemetry/instrumentation/rage/handlers/cable.rb
rsamoilov: It's a bit hard for me to reason about it without understanding what a messaging system is in the OpenTelemetry terms.

Anyway, the [latest semantic conventions](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/rpc/rpc-spans.md#rpc-server-span) state that for streaming RPCs, the server span should cover the full lifetime of the request and/or response streams until they are closed or terminated.

This implies that:

1) RPC is considered request/response, which is not true for WebSockets.
2) The RPC span should cover the entire lifetime of the WebSocket connection. This is unrealistic for WebSockets as connections can be open for days and weeks.

Please let me know if I've misinterpreted it.

instrumentation/rage/lib/opentelemetry/instrumentation/rage/handlers/cable.rb
rsamoilov: The server.

instrumentation/rage/lib/opentelemetry/instrumentation/rage/handlers/cable.rb
rsamoilov: The server.

rsamoilov:
@arielvalentin @thompson-tomo Just checking in on the PR - please let me know how we can proceed. 

Currently, most disagreements are about the Cable instrumentation. I’m happy to delete it and tackle in another PR if you think it makes sense.

instrumentation/rage/lib/opentelemetry/instrumentation/rage/handlers/cable.rb
thompson-tomo: I agree that the defination of rpc covering the entire duration as one span makes it harder, hopefully it will be addressed as part of the rpc stabilisation effort. I do know streaming  was  initially out of scope.

> It's a bit hard for me to reason about it without understanding what a messaging system is in the OpenTelemetry terms.

Agree that having the definition/requirements of things like `messaging.system` documented is important.

instrumentation/rage/lib/opentelemetry/instrumentation/rage/handlers/cable.rb
thompson-tomo: In that case shouldn't the span kind become server?

instrumentation/rage/lib/opentelemetry/instrumentation/rage/handlers/deferred.rb
thompson-tomo: See other comment

instrumentation/rage/lib/opentelemetry/instrumentation/rage/handlers/deferred.rb
thompson-tomo: There needs to be a project spun up etc. Using non semantic conventions is not any different to using development conventions in terms of stability.

instrumentation/rage/lib/opentelemetry/instrumentation/rage/handlers/events.rb
thompson-tomo: Because we aren't talking about source code rather the client. By using messaging.client we are complementing the id attribute, foster re-use etc

instrumentation/rage/lib/opentelemetry/instrumentation/rage/handlers/request.rb
thompson-tomo: Are you saying that those attributes are being added to this span elsewhere or are they on different spans?

instrumentation/rage/lib/opentelemetry/instrumentation/rage/handlers/request.rb
thompson-tomo: I was more thinking non ruby ones. 😉

For instance does url templates help us?

instrumentation/rage/lib/opentelemetry/instrumentation/rage/handlers/request.rb
rsamoilov: They are added by the [Rack](https://github.com/open-telemetry/opentelemetry-ruby-contrib/tree/main/instrumentation/rack) instrumentation.

instrumentation/rage/lib/opentelemetry/instrumentation/rage/handlers/cable.rb
rsamoilov: Hmm, I guess. I initially used `consumer` because the [spec](https://opentelemetry.io/docs/concepts/signals/traces/#span-kind) says the `server` kind represents a "**synchronous** incoming remote call". And with WebSocket messages, the client doesn't wait for the server to respond.
But I guess `server` makes more sense.

instrumentation/rage/lib/opentelemetry/instrumentation/rage/handlers/deferred.rb
rsamoilov: Fair. It looks like the [Hangfire](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/tree/main/src/OpenTelemetry.Instrumentation.Hangfire) instrumentation uses `workflow` attributes. I'll take a closer look.

instrumentation/rage/lib/opentelemetry/instrumentation/rage/handlers/cable.rb
rsamoilov: But you do recommend going with the RPC attributes, right?

rsamoilov:
@thompson-tomo Made some updates:

* Reverted to the `server` span kind for Cable channels
* Removed `code.*` attributes
* Updated Cable instrumentation to use custom `websocket.*` attributes - let me know if you think `rpc.*` or `rage.cable.*` attributes will work better
* Updated Deferred instrumentation to use the `workflow.*` attributes

rsamoilov:
Closed in favor of https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/1975.

PR #1907, opened 2026-01-05
ci: leverage public dictionaries and increase cspell usage

thompson-tomo:
This updates the cspell setup to use public dictionaries rather than custom words to reduce maintenance effort and barriers especially when writing new documents due to the extensive list of worlds available. In the process the checking scope is broadened to include more file types but limited to comments in those documents

### DIFF
--- a/.cspell.yml
+++ b/.cspell.yml
@@ -3,22 +3,14 @@
 # https://www.streetsidesoftware.com/vscode-spell-checker/docs/configuration/
 version: '0.2'
 caseSensitive: false
+useGitignore: true
 import:
   - https://cdn.jsdelivr.net/npm/@cspell/dict-people-names/cspell-ext.json
+  - https://cdn.jsdelivr.net/npm/@cspell/dict-redis/cspell-ext.json
 ignorePaths:
   - 'bin'
-  - '**.conf'
-  - '**.gemspec'
-  - '**.json'
   - '**.proto'
-  - '**.rb'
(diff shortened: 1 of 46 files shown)


.cspell.yml
```diff
@@ -3,6 +3,17 @@
 # https://www.streetsidesoftware.com/vscode-spell-checker/docs/configuration/
 version: '0.2'
 caseSensitive: false
+useGitignore: true
+enableGlobDot: true
+import:
+  - https://cdn.jsdelivr.net/npm/@cspell/dict-people-names/cspell-ext.json
```
arielvalentin, line 9: Is there some way to cache the dictionaries?

arielvalentin (approved):

.cspell.yml
thompson-tomo: So there is 3 options:
- import from http address as done here
- define a package.json file, restore the npm packages prior to running the action.
- explicitly install the npm package prior to running the action.

Let me know if you would like me to go down one of the other options. Note they both require a node environment to run.

arielvalentin: The other options sound too complicated. Let's continue with importing and trust that the CDN has high uptimes.

thompson-tomo: Agree this is the most efficient and it is documented in https://github.com/streetsidesoftware/cspell-dicts/tree/main/dictionaries/people-names

PR #1918, opened 2026-01-06
feat: Add SQL Query Summary - Tokenizer, Parser, Cache

hannahramadan:
This PR adds the framework for a query summary generator, including a tokenizer, parser, and cache. It is meant to generate the `db.query.summary` attribute for database spans.

Query Pipeline: SQL Query → Cache Lookup → Tokenizer → FSM Parser → Summary

**Fast Path & Caching**
- Tiered Processing: High-frequency CRUD queries (e.g., SELECT * FROM table) use pre-compiled regex to bypass tokenization.
- LRU Cache: Thread-safe caching (default size: 1000) prevents redundant processing of identical SQL fingerprints.

**Tokenizer**
  - Breaks SQL into typed tokens (:keyword, :identifier, :operator) using StringScanner.
  - Filters sensitive data like string literals, comments, numeric values
  - Optimization: Uses a hash-based keyword lookup and frequency-ordered scanning to minimize

**Parser**
- 3 States: PARSING_STATE (verb hunting), EXPECT_COLLECTION_STATE (table gathering), and DDL_BODY_STATE (skipping procedure/trigger internal logic).
- Manages implicit aliases, JOIN lists, UNION consolidation, and IF EXISTS DDL patterns.

**Summary Generation**
- Consolidates operations and table names: "SELECT users profiles".
- Truncates at 255 characters
- Automatically removes duplicate table names in complex joins or unions

**Resources:**
- Spec: [Generating a query summary](https://github.com/open-telemetry/semantic-conventions/blob/v1.33.0/docs/database/database-spans.md#generating-a-summary-of-the-query)
- .NET [implementation](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/main/src/Shared/SqlProcessor.cs) and [test cases ](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/non-normative/database-test-cases/db-sql-test-cases.json)


helpers/sql-processor/lib/opentelemetry/helpers/sql_processor/query_summary.rb
```diff
@@ -0,0 +1,46 @@
...
+      # @return [String] The query summary or 'UNKNOWN' if parsing fails
+      #
+      # @api public
+      def generate_summary(query, cache:)
+        cache.fetch(query) do
+          tokens = Tokenizer.tokenize(query)
+          Parser.build_summary_from_tokens(tokens)
+        end
+      rescue StandardError
+        'UNKNOWN'
```
hannahramadan: `db.query.summary` should be used as a database span name, but if the summary is not available, then the name should be `{db.operation.name} {target}`. We should keep that in mind when changing db instrumentation. If the return value of the summary is `UNKNOWN`, then we have more work to do to create the span name.

hannahramadan (commented):
(removed)

PR #1765, opened 2025-10-28
chore: Fix Ruby Version Management

arielvalentin:
Prior to this change, the RUBY_REQUIREMENT file was used to keep Ruby versions in sync, however toys will not detect a change to gem since the gemspec file was unaltered.

Rather than modify toys to be more flexible in this case, it will be easier to update the Ruby requirement in each gemspec using a script.

The script will write out diagnostic output of the gemspecs that were changed. E.g.

```ruby

$> bin/update-ruby-version 3.2

Updating Ruby version requirement to: >= 3.2
Finding gemspec files...
Found 62 gemspec files to update:
Updated: helpers/mysql/opentelemetry-helpers-mysql.gemspec
Updated: helpers/sql/opentelemetry-helpers-sql.gemspec
Updated: helpers/sql-obfuscation/opentelemetry-helpers-sql-obfuscation.gemspec
Updated: helpers/sql-processor/opentelemetry-helpers-sql-processor.gemspec
Updated: instrumentation/action_mailer/opentelemetry-instrumentation-action_mailer.gemspec
...
Updated: resources/google_cloud_platform/opentelemetry-resource-detector-google_cloud_platform.gemspec
Updated: sampler/xray/opentelemetry-sampler-xray.gemspec

Summary:
- Total files found: 62
- Successfully updated: 62

Done! All gemspecs have been updated to require Ruby >= 3.2

```

PR #1923, opened 2026-01-07
chore: Upgrade release system to the toys-release gem

dazuma:
This is a long-awaited upgrade to the release system. It was rewritten pretty much from the ground up to be more testable and maintainable. That said, the OTel-Ruby repos exercise the system more than any other customer, so I'll be around to monitor the next releases closely to ensure things go smoothly.

Some key changes:

* The release system is now distributed as a gem, [toys-release](https://rubygems.org/gems/toys-release). So loading it is less sketchy than depending on a tag in another repository as we've been doing before. This change currently pins the toys-release gem version at 0.4.0 (see the file `.toys/release.rb`).
* Most of the interfaces are mostly the same, but backward compatibility was not a goal here. I updated the release configs (`.toys/.data/releases.yml`) and the tool invocations in the GitHub Actions workflows accordingly.
* I set the GitHub Actions workflows to run on Ruby 4.0 and Toys 0.19.1. In particular, toys-release requires at least Toys 0.19.0. Ruby 4.0 isn't really necessary, but toys-release uses it by default. If people are more comfortable using 3.4, I can change it to that version.

Some of the config has also been updated to work with the toys-release gem. Specifically:

* The `main_branch` key defaults to `main`, so `main_branch: main` is redundant and was removed.
* The `commit_lint` key was dropped. The commit linter in general was removed from toys-release, as it was not really maintained and no one was using it. And it has been disabled in our repos anyway. There are better dedicated conventional commit linter tools available if we want to use one.
* I removed the trailing slash hack for the `opentelemetry-helpers-sql` and `opentelemetry-instrumentation-http` gem configurations. That bug has been fixed, and the hack is no longer necessary.
* I removed `version_rb_path` settings for gems whose configurations did not need it.

There is quite a bit of new functionality and fixed issues in toys-release, that we might choose to take advantage of (again, as a second step after this initial migration). For example:

* There is now a more flexible way to customize which paths signal that a gem has changed, supporting shared files that are used by multiple gems. This should make hacks like https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/1765 unnecessary.
* There is now a way to customize how conventional commit tags (including scopes) map to releases and semver, which Ariel requested in https://github.com/dazuma/toys/issues/215.
* Toys-release is no longer limited to releasing from the main branch. You should be able to release from any branch, by choosing it when triggering the "open release request" workflow.

Finally, I (@dazuma) expect to be able to engage a lot more now on OTel-Ruby. I quit my job at Google a few months ago, in part because it was stifling me and I was not able to make space for open source and community work. Now that I'm out, I've had a chance to, among other things, finish this long-overdue update to the release system. I also should be able to attend the weekly SIGs more often now. So anyone working on this repo, if you have questions, issues, or feature requests regarding the release system and process, please let me know.

I'm keeping this PR in draft until I've had the chance to discuss this update with the other maintainers, and to do some more testing.

Note there is a corresponding pull request in the main opentelemetry-ruby repository: https://github.com/open-telemetry/opentelemetry-ruby/pull/2006

One final note: I notice there is a release-please workflow present in this repository as well. release-please is an alternative release mechanism that could replace toys-release. (In fact, toys-release actually borrows a few of its ideas, such as release pull requests and conventional commit-based changelogs, from release-please.) I have not touched the existing release-please workflow, pending discussion with other maintainers about which direction to go. However, I do have some inside knowledge regarding release-please, so please do not make that decision without discussing with me.


dazuma:
@arielvalentin Let's discuss it in the SIG next week. Sorry I missed yesterday; I was trying to make it, but stuff happened.

Edit: Oh, and accumulating subsequent changes into an existing release PR would definitely be a feature I could add to toys-release.

.github/workflows/release-hook-on-closed.yml
```diff
@@ -19,17 +19,16 @@ jobs:
       - name: Install Ruby
         uses: ruby/setup-ruby@675dd7ba1b06c8786a1480d89c384f5620a42647 # v1.281.0
         with:
-          ruby-version: "3.2"
+          ruby-version: "4.0"
       - name: Checkout repo
         uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # v6.0.1
       - name: Install Toys
-        run: "gem install --no-document toys -v 0.15.5"
+        run: "gem install --no-document toys -v 0.19.1"
```
arielvalentin, line 26: Should we add toys to the project gemfile and use bundle install along with bundle-cache: true?

arielvalentin (approved):
Mostly improve by trust here. Since the interface is very similar and it's using envars in some cases, I don't see anything surprising.

One thought I had was, why not add toys as a bundler dependency in the top level Gemfile? That would allow us to be able to install toys and run commands locally for testing in a more familiar way. 

The actions would then execute installs using Bundler and could potentially cache gemfiles between runs. 

This is all food for thought and not really necessary.

I'll also ask for your help to remove any release-please related code and configs in a follow up PR since we are moving forward with new versions of toys.

PR #1928, opened 2026-01-11
ci: Address issues identified by renovate

thompson-tomo:
This goes through setting explicit versions in the gemfile when they were missing which was resulting in them being unable to be managed by renovate and logging an error on the local cli.

I have added an ignore for the packages which are only needed on ruby 3.4+ but if maintainers happy this should also be set to an explicit version.


instrumentation/all/Gemfile
```diff
@@ -15,8 +15,8 @@ group :test do
   gem 'rubocop-performance', '~> 1.26.0'
   gem 'simplecov', '~> 0.22.0'
   gem 'yard', '~> 0.9'
-  gem 'active_model_serializers'
-  gem 'activesupport'
+  gem 'active_model_serializers', '~> 0.10.16'
```
arielvalentin: can we omit these?

instrumentation/grape/Gemfile
```diff
@@ -18,13 +18,13 @@ group :test do
   gem 'simplecov', '~> 0.22.0'
   gem 'webmock', '~> 3.24'
   gem 'yard', '~> 0.9'
-  gem 'activesupport'
-  gem 'opentelemetry-test-helpers'
-  gem 'rack-test'
-  gem 'rspec-mocks'
+  gem 'activesupport', '~> 8.1.2'
```
arielvalentin: I think these combinations need to be covered by appraisals and not as part of the default set.

instrumentation/racecar/Gemfile
```diff
@@ -9,15 +9,15 @@ source 'https://rubygems.org'
 gemspec
 
 group :test do
-  gem 'activesupport'
+  gem 'activesupport', '~> 8.1.2'
```
arielvalentin: hmm same. These should be in the appraisals.

instrumentation/resque/Gemfile
```diff
@@ -19,7 +19,7 @@ group :test do
   gem 'simplecov', '~> 0.22.0'
   gem 'webmock', '~> 3.24'
   gem 'yard', '~> 0.9'
-  gem 'activejob'
+  gem 'activejob', '~> 8.1.2'
```
arielvalentin: These should be in the appraisals.

instrumentation/grape/Gemfile
thompson-tomo: Currently appraisals only has grape. Are you saying to move the 2x active* dependencies to appraisals.

instrumentation/racecar/Gemfile
thompson-tomo: Same question as the other and note the appraisals only currently has racecar.

instrumentation/resque/Gemfile
thompson-tomo: Same question again.

instrumentation/aws_sdk/example/Gemfile
```diff
@@ -2,10 +2,10 @@
 
 source 'https://rubygems.org'
 
-gem 'aws-sdk-core', '~> 3.240.0'
-gem 'aws-sdk-sns', '~> 1.109.0'
+gem 'aws-sdk-core'
```
arielvalentin, line 5: Why did this remove version numbers when while adding it to others?

instrumentation/delayed_job/Gemfile
```diff
@@ -18,13 +18,13 @@ group :test do
...
+  gem 'rspec-mocks', '~> 3.13.7'
+  gem 'sqlite3', '~> 2.9'
   gem 'opentelemetry-instrumentation-base', path: '../base'
   if RUBY_VERSION >= '3.4'
     gem 'base64'
     gem 'mutex_m'
   end
   # https://github.com/collectiveidea/delayed_job/issues/1245
-  gem 'ostruct' if Gem::Version.new(RUBY_VERSION) >= Gem::Version.new('4')
+  gem 'ostruct', '~> 0.6.3' if Gem::Version.new(RUBY_VERSION) >= Gem::Version.new('4')
```
arielvalentin: Add ostruct to the ignore group like you did with big decimal and mutex

instrumentation/que/Gemfile
```diff
@@ -18,7 +18,7 @@ group :test do
   gem 'rubocop-performance', '~> 1.26.0'
   gem 'simplecov', '~> 0.22.0'
   gem 'yard', '~> 0.9'
-  gem 'pg'
+  gem 'pg', '~> 1.6.3'
```
arielvalentin: I think this should be defined in appraisals and not in the gemfile. I'm surprised we aren't getting warnings or errors about this 
```suggestion
```

instrumentation/aws_sdk/example/Gemfile
thompson-tomo: Because this is just an example of a gemfile and not actually used. The other example files also lack versions.

Hence example gemfile are excluded.

instrumentation/delayed_job/Gemfile
thompson-tomo: Done.

instrumentation/que/Gemfile
thompson-tomo: Have reverted it from here & added it to the other pr.

instrumentation/aws_sdk/example/Gemfile
arielvalentin: I overlooked that. Thank you.

arielvalentin (approved):

kaylareopelle (approved):

PR #1938, opened 2026-01-13
ci: Address issues identified by renovate part. 2

thompson-tomo:
This goes through setting explicit versions in the gemfile when they were missing which was resulting in them being unable to be managed by renovate and logging an error on the local cli.


instrumentation/all/Gemfile
```diff
@@ -15,8 +15,8 @@ group :test do
   gem 'rubocop-performance', '~> 1.26.0'
   gem 'simplecov', '~> 0.22.0'
   gem 'yard', '~> 0.9'
-  gem 'active_model_serializers'
-  gem 'activesupport'
+  gem 'active_model_serializers', '~> 0.10.16'
+  gem 'activesupport', '~> 8.1.2'
```
scbjans: Given we didn't define a version constraint before, do we want to define this as a pessimistic minor version constraint?

arielvalentin (commented):
These gem dependencies are open ended otherwise they cause conflicts with the appraisal generated gemfiles; although I am now wondering why these are included in the first place.

Would you mind deleting them instead of pinning them to see if what the result is on the test suite?

arielvalentin (approved):

PR #1945, opened 2026-01-15
ci: add renovate config validation to the ci checks

thompson-tomo:
This adds a renovate config validation check which will produce errors like shown below if the config is invalid:
image

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -28,3 +28,5 @@ instrumentation/active_record/db
 
 tags
 !**/*/.gitkeep
+
+node_modules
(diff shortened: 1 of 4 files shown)


.gitignore
```diff
@@ -28,3 +28,5 @@ instrumentation/active_record/db
 
 tags
 !**/*/.gitkeep
+
+node_modules
```
arielvalentin, line 32: Why does this have be configured in a subdirectory?

thompson-tomo: It's configured in the root, just appears to be in instrumentation/active_record/db as that is another directory being ignored. ;)

arielvalentin: I see. This was an idiosyncrasy of the mobile UI that made me think this it was a duplicate of the root gitignore file

arielvalentin (approved):

.gitignore
thompson-tomo: No worries, I've made that mistake a couple of times myself.

PR #1948, opened 2026-01-17
chore: Removed duplicate entry in the release configs

dazuma:
Should fix the error messages coming from the release system.


arielvalentin (approved):
🤦🏻‍♂️

PR #1952, opened 2026-01-19
chore: run release workflows when files change

thompson-tomo:
This ensures that the release workflows run when the definition is updated. To prevent issues the final step is skipped when triggered via a pr.


dazuma:
@thompson-tomo Can you describe again what you want to accomplish here? I'm kind of confused. (In particular, some of these workflows are definitely not meant to be triggered in the manner that I think is being done here.)

.github/workflows/release-hook-on-closed.yml
```diff
@@ -3,6 +3,8 @@ name: "[release hook] Process release"
 on:
   pull_request:
     types: [closed]
+    paths: &path_filters
+      - ".github/workflows/release-hook-on-closed.yml"
```
dazuma: I don't think this does what we want. It seems like we'd want this workflow to run if type=closed (i.e. normal runs) OR path matches this file (CI runs), but it looks like adding the paths in this way simply adds a filter, i.e. the workflow runs if type=closed AND path matches. That would break normal operation, and does not give us the CI runs either (as evidenced by the fact that this workflow isn't included in the checks for this current PR).

thompson-tomo: Thanks for spoting that, I have addressed it.

Note path filters are not apply to closed events as there is no paths hence this should always run on closed and filtered otherwise.

dazuma: As I understand it, removing `types` altogether will revert it to the default, which is `[opened, synchronize, reopened]`. (Ref: https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#pull_request). So now I expect this does not run on `closed` at all.

dazuma (commented):
Just a reminder here: if #1954 is merged, we should add `.toys/release/Gemfile` to the pull request paths being watched.

.github/workflows/release-hook-on-closed.yml
thompson-tomo: Ahh, I thought default was broader hence have explicitly set it.

thompson-tomo:
> These seem messy and precarious. For example, this PR, I believe, already breaks one of the workflows as it stands.

I think this breakage has been fixed. In almost all cases the dry-run would be run on pr's.

> Although it seems that depending on renovate to update setup-ruby in time has already failed once in the case of #1947, so I'm not sure I trust it

The issue is the setup-ruby action is in the queue to update and with this change the pr to update ruby would've failed until the setup-ruby had been merged.

> What cases would not be caught that the current PR would catch?

In current form it catch issues with the gemfile, with the real benefit coming with the dry-run option.

> The current dry-run mode, for example, doesn't actually do a release, but still does all the input checks to make sure the requested release is viable, does all the builds, etc.

That sounds exactly like what I would want the dry-run to do. This way when an update to toys occurs via renovate it undergoes all the necessary checks so that we have confidence in merging it.

.github/workflows/release-hook-on-closed.yml
dazuma: I tested this in a test repository, and unfortunately it doesn't look like it works. The `paths` filter *does* apply to the `closed` type: if a pull request is merged or closed that doesn't touch the given path, the workflow is not triggered.

.github/workflows/release-hook-on-closed.yml
```diff
@@ -28,6 +30,7 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           GEM_HOST_API_KEY: ${{ secrets.RUBYGEMS_API_KEY }}
+        if: github.event.pull_request.state == 'closed'
```
dazuma: This should be `if: github.event.action == 'closed'`. As written, it will trigger if a closed pull request is edited, which we don't want.

.github/workflows/release-hook-on-closed.yml
thompson-tomo: Interesting as a number of searches says that it would've so thank you for testing that out. I have made further changes which should hopefully work. 🤞

dazuma:
Note: I opened #1959, an alternative for discussion purposes. I included what I see as the pros and cons in its description.

PR #1958, opened 2026-01-20
fix: Handle `Self` field in X-Amzn-Trace-Id header

whi-tw:
AWS' load balancers add a `Self` field to the X-Amzn-Trace-Id header [^1] if the header exists, and a `Root` field is already present. This breaks the regex used to parse the header and leads to the traces not being associated with each other.

Allow for an optional Self field at the start of the header. Don't capture this.

[^1]: https://docs.aws.amazon.com/elasticloadbalancing/latest/application/load-balancer-request-tracing.html

### DIFF
--- a/propagator/xray/lib/opentelemetry/propagator/xray/text_map_propagator.rb
+++ b/propagator/xray/lib/opentelemetry/propagator/xray/text_map_propagator.rb
@@ -18,11 +18,23 @@ module OpenTelemetry
       # Propagates context in carriers in the xray single header format
       class TextMapPropagator
         XRAY_CONTEXT_KEY = 'X-Amzn-Trace-Id'
-        XRAY_CONTEXT_REGEX = /\ARoot=(?<trace_id>([a-z0-9\-]{35}))(?:;Parent=(?<span_id>([a-z0-9]{16})))?(?:;Sampled=(?<sampling_state>[01d](?![0-9a-f])))?(?:;(?<trace_state>.*))?\Z/ # rubocop:disable Lint/MixedRegexpCaptureTypes
         SAMPLED_VALUES = %w[1 d].freeze
         FIELDS = [XRAY_CONTEXT_KEY].freeze
 
-        private_constant :XRAY_CONTEXT_KEY, :XRAY_CONTEXT_REGEX, :SAMPLED_VALUES, :FIELDS
+        # Header parsing constants
+        KV_PAIR_DELIMITER = ';'
+        KEY_AND_VALUE_DELIMITER = '='
+        TRACE_ID_KEY = 'Root'
+        PARENT_ID_KEY = 'Parent'
+        SAMPLED_FLAG_KEY = 'Sampled'
+        TRACE_ID_LENGTH = 35
+        SPAN_ID_LENGTH = 16
+        VALID_SAMPLED_VALUES = %w[0 1 d].freeze
+
+        private_constant :XRAY_CONTEXT_KEY, :SAMPLED_VALUES, :FIELDS,
+                         :KV_PAIR_DELIMITER, :KEY_AND_VALUE_DELIMITER,
+                         :TRACE_ID_KEY, :PARENT_ID_KEY, :SAMPLED_FLAG_KEY,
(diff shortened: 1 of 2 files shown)


whi-tw:
Note: the python implementation doesn't use regex: it extracts all the K=V pairs with `split()`, which means it's not dependent on the header attribute order being consistent: https://github.com/open-telemetry/opentelemetry-python-contrib/blob/be56431caa05be1cd8ccd610ea13f7584f995acd/propagator/opentelemetry-propagator-aws-xray/src/opentelemetry/propagators/aws/aws_xray_propagator.py#L172. This may be a better solution going forward for this gem

propagator/xray/lib/opentelemetry/propagator/xray/text_map_propagator.rb
```diff
@@ -90,11 +102,56 @@ def inject(carrier, context: Context.current, setter: Context::Propagation.text_
...
+              trace_id = value if valid_trace_id?(value)
+            when PARENT_ID_KEY
+              span_id = value if valid_span_id?(value)
+            when SAMPLED_FLAG_KEY
+              sampling_state = value if valid_sampling_state?(value)
+            when 'Self'
+              # Ignore Self field added by load balancers
+              next
+            else
+              # Collect other fields as potential tracestate
```
xuan-cao-swi, line 126: Why we ignore `Self` here? I saw there is case that only `Self` present.
```
If the X-Amzn-Trace-Id header is present and has a Self field, the load balancer updates the value of the Self field.
```

whi-tw: So, there should never just be `Self` present - unless an application has specifically done that.

1. The load balancer adds `Root` if there's no existing header
2. It adds `Self` when `Root` is already present in the header
3. If the application has created a custom header with `Root`, it still just adds a `Self` header, but leaves all existing K:V pairs.

I can't work out exactly what `Self` would map to in an OTEL context. LBs add it for request tracing, but that's distinct from X-Ray (it lets people manually trace requests, I guess). I don't think capturing it would actually have any benefit to end users. I suppose we could add it to tracestate, but that may just confuse. In theory, the propagator will link the traces together: `Self` would 'just' allow someone to identify an ALB request from logs. X-Ray doesn't make any links to those, though.

xuan-cao-swi (approved):
LGTM: the strategy of decomposing regex patterns for individual processing is effective.

propagator/xray/lib/opentelemetry/propagator/xray/text_map_propagator.rb
```diff
@@ -90,11 +102,56 @@ def inject(carrier, context: Context.current, setter: Context::Propagation.text_
...
+
+          {
+            'trace_id' => trace_id,
+            'span_id' => span_id,
+            'sampling_state' => sampling_state,
+            'trace_state' => trace_state_parts.empty? ? nil : trace_state_parts.join(KV_PAIR_DELIMITER)
+          }
+        end
+
+        def valid_trace_id?(value)
```
xuan-cao-swi, line 141: Is this based on python `_validate_trace_id`?
It might be worth adding a comment to clarify the origin of the `1-` and `10`, as their source could be confusing.

whi-tw: Yep, good catch. I thought I'd recreated and used all the constants, but absent-minded these ones. I'll sort them out Monday morning.

propagator/xray/lib/opentelemetry/propagator/xray/text_map_propagator.rb
jterapin: Your reasoning makes sense @whi-tw, since I also feel that end users will not entirely benefit from this addition for now.

jterapin (approved):
Approved from the AWS SDK side, given @xuan-cao-swi has validated above.

whi-tw:
Hi @arielvalentin it looks like you queued this up for merge, but it fell out the queue again? I've rebased this against main, so any up-to-date gates should now pass.

arielvalentin (approved):

PR #1954, opened 2026-01-20
chore: use a gemfile to install toys and toys-release during release workflows

dazuma:
Use the setup-ruby action to install toys and toys-release via a Gemfile during release workflows, instead of an explicit step that runs `gem install`. This will:

* Allow renovate to manage the versions of the tools being used
* Allow setup-ruby to cache the bundle installs

Note the Gemfile is kept in the release subdirectory of the toys directory hierarchy, so it doesn't interfere with bundler integration for non-release-related tools.

This change is included in https://github.com/open-telemetry/opentelemetry-ruby/pull/2006 in the main repo.

Ref: @thompson-tomo [comments in https://github.com/open-telemetry/opentelemetry-ruby/pull/2006](https://github.com/open-telemetry/opentelemetry-ruby/pull/2006#discussion_r2704502519)
Ref: @arielvalentin [comments in https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/1923](https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/1923#pullrequestreview-3658301732)

PR #1959, opened 2026-01-20
chore: canary test for release workflows

dazuma:
This is an alternative to #1952, for discussion.

#1952 modifies the existing release workflows so they partially run during pull request CI if the workflows are changed (e.g. when Renovate updates them). This is meant to catch problems like #1947 which silently broke the workflows. However, the changes are pretty invasive to the workflows themselves, because the logic needed to trigger the workflow in a normal release context *or* a CI context, and the logic needed to distinguish the two cases, is sometimes tricky to implement.

This pull request is a simpler alternative with some trade-offs. It introduces a new "canary" workflow that runs only in the CI context and duplicates the installations in the other release workflows. It does not modify any of the other workflows. The idea is, when Renovate does any updates, it should open a pull request that affects all the workflows, including this one. Any installation breakages should manifest at CI time in this workflow run without needing to run the others. This should catch Renovate-driven cases like #1947, but may not catch other cases such as someone modifying individual workflows without making corresponding changes to this one.

So to summarize the pros and cons.

Arguments for #1952:

* Can catch individual changes (e.g. not driven by Renovate) to release workflows that do not also touch this "canary" workflow.
* ~Framework for potential future invocation of CLIs in dry-run mode to catch breakages that would manifest only when running the CLI itself. (Does not provide this yet, but can in the future.)~ Edit: now that I think about it, such "dry-run test invocations" of the CLI itself could simply be added to this canary workflow, so this is not a *distinguishing* argument for 1952.

Arguments for the current PR:

* Non-invasive in the existing release workflows, does not increase their complexity or require more third-party dependencies.

### DIFF
--- a/.github/workflows/release-test.yml
+++ b/.github/workflows/release-test.yml
@@ -18,7 +18,7 @@ jobs:
       - name: Install Ruby
         uses: ruby/setup-ruby@675dd7ba1b06c8786a1480d89c384f5620a42647 # v1.281.0
         with:
-          ruby-version: "4.0.1"
+          ruby-version: "4.0"
       - name: Checkout repo
         uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # v6.0.1
       - name: Install Toys
--- a/.github/workflows/release-test.yml
+++ b/.github/workflows/release-test.yml
@@ -1,0 +1,27 @@
+name: Test for release workflows
+
+on:
+  pull_request:
+    branches:
+      - main
+    paths: &path_filters
+      - ".github/workflows/release-test.yml"
+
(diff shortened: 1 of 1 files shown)


.github/workflows/release-test.yml
```diff
@@ -0,0 +1,31 @@
+name: Test for release workflows
+
+on:
+  pull_request:
+    branches:
+      - main
+    paths: &path_filters
+      - ".github/workflows/release-test.yml"
```
dazuma, line 8: Note: if #1954 is merged, then ".toys/release/Gemfile" should be added to this list. (And the changes from #1954 should be applied to the job specification below.)

PR #1970, opened 2026-01-31
fix: Move checkout to be before install

thompson-tomo:
This will help fix the issue reported in comments in #1954


kaylareopelle (approved):
🚀

PR #1975, opened 2026-02-01
feat: Add Rage instrumentation to instrumentation_all

rsamoilov:
This PR adds OpenTelemetry instrumentation for the [Rage framework](https://github.com/rage-rb/opentelemetry-instrumentation) to `instrumentation_all`.


rsamoilov:
@arielvalentin Thank you for the response. I share the concerns about the `all` gem and agree a more explicit dependency model is healthier for the ecosystem long-term.

That being said, I'm happy to close this PR and instead focus on improving the developer experience for users who adopt `opentelemetry-instrumentation-rage` directly.

PR #1983, opened 2026-02-04
ci: Commit lock files

### LINKED ISSUE #1962: Commit gemfile.lock to repo
To assist with implementing reproducible builds and simplifying the fossa Workflow we should commit the lock files except for examples. We can then have renovate update these files weekly and additionally filter the fossa Workflow to only run when the lock file changes. This reduces un-necessary uploads.

thompson-tomo:
Closes #1962 

This goes through and commits the gem lockfiles.

By committing the lockfiles this enables the fossa scan to be simplified to only upload pre-generated files rather than needing to generate them on the go. It also enables better deterministic builds as everyone will always be using the same lock file.

We do not need to worry about maintaining the lockfile as renovate will raise a weekly pr to bump the contents.

Note i have also scheduled a weekly upload of the files to fossa.


robbkidd:
> It also enables better deterministic builds as everyone will always be using the same lock file.

Were the `Gemfile.lock`s to be checked in, consumers of these gems would not be using them. Only version constraints declared in each gem's `.gemspec` affect dependency resolution occurring within consuming projects. Deterministic builds aren't really a thing controllable by a Ruby library.

PR #1984, opened 2026-02-04
feat: Add release tag into source code url of gem metadata

### LINKED ISSUE #1961: Use commit id in source code url property of gemspec
To enable removal of the source code of packages no longer supported, it would be useful if the published gems pointed to the commit of the source code used. Ideally the file path is also set.

On looking into it we should be able to add the following:

> commit = ENV["GITHUB_SHA"] || "main"

Then update the url

> spec.metadata["source_code_uri"] =
  "https://github.com/open-telemetry/opentelemetry-ruby-contrib/tree/#{commit}/instrumentation/ruby_kafka"

This way the commit id is used on github otherwise main is used.

thompson-tomo:
Closes #1961

This tweaks the gemspec files so that the source code url now contains the release tag. By including this tag, we can as part of deprecation work remove the source files while having atleast the latest published version of gem pointing to the code used and a readme containing the deprecation

Note the only other change was to update `resources/aws/opentelemetry-resource-detector-aws.gemspec` so that it was the same metadata as the other gems.

### DIFF
--- a/.instrumentation_generator/templates/gemspec.tt
+++ b/.instrumentation_generator/templates/gemspec.tt
@@ -4,6 +4,7 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 
+commit = ENV["GITHUB_SHA"] || "main"
 lib = File.expand_path('lib', __dir__)
 $LOAD_PATH.unshift(lib) unless $LOAD_PATH.include?(lib)
 require 'opentelemetry/instrumentation/<%= instrumentation_name %>/version'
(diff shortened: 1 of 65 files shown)


scbjans (approved):
LGTM

thompson-tomo:
Both achieve the same result and experience, tags are nicer for users using github but we would need to extend the CI to be adding tags and using them. The nice thing about commits is that they are always available to both ci/users and can not be tampered with.

arielvalentin (approved):

PR #1991, opened 2026-02-08
feat: instrument `PG::Connection`'s pipeline methods

zergov:
We use query pipelining in our application, and I expected the `pg` otel plugin to instrument some of the pipeline methods. 

This PR instruments the following methods on the `pg` gem's `PG::Connection` class:
  - `enter_pipeline_mode`
  - `pipeline_sync`
  - `exit_pipeline_mode`

### DIFF
--- a/instrumentation/pg/lib/opentelemetry/instrumentation/pg/patches/connection.rb
+++ b/instrumentation/pg/lib/opentelemetry/instrumentation/pg/patches/connection.rb
@@ -138,6 +138,24 @@ module OpenTelemetry
             end
           end
 
+          define_method :enter_pipeline_mode do |*args|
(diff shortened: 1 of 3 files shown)


zergov:
Hey @thompson-tomo, thanks for the feedback.

>I feel that we Would be better off creating a span when we enter pipeline mode (kind: Internal) and ending it when we exit pipeline mode.

I think what you suggest is the cleanest solution. I hesitated doing this because the `#enter_pipeline_mode` method doesn't accept a block, so I'd need to handle the lifecycle manually. The thing that is not clear to me is the condition to stop the span when `#exit_pipeline_mode` is not called. We probably need to finish the span if the connection closes or the user resets the connection while in a pipeline. 

I also like the idea of using events on the pipeline span 👌 I'll work on it this week, thanks!

PR #1997, opened 2026-02-11
chore: Update to toys-release 0.6.0

dazuma:
Fixes #1996. The releaser was getting confused by `VERSION = '1.2.3'` forms with single rather than double quotes. It now supports both.

Also includes several additional improvements that have been requested:

* Support for retaining the issue/PR numbers that GitHub appends to the end of merge/squash commit messages in the changelog entries, and even linkifying them in the changelog. This is controlled by releases.yml (the default behavior is the current behavior that removes them from changelog entries). The maintainers can decide how we want to configure these moving forward.
* Support for configuring which bullet (asterisk or hyphen) is used in changelog entries. This is also controlled by the releases.yml config, and the default is the current behavior of asterisk. The maintainers can decide if we want to change this.
* Changelog entries are stripped of trailing whitespace

This PR also:

* Removes the `version_constant` setting from releases.yml, which has been deprecated and unused in recent versions of toys-release.
* Removes the explicit dependency on base64 from the Gemfile, since toys-release 0.6.0 now properly lists it as a transitive dependency.

PR #2002, opened 2026-02-14
chore: Trim whitespace from changelogs

thompson-tomo:
This remedies the whitespace issue with toys-release which has since been resolved with #1997

This also removes empty lines from lists


arielvalentin (commented):
These change log entries are autogenerated by `toys`. 

@dazuma Is some of the white space intentional?

dazuma (approved):
Thanks for cleaning this up!

PR #2031, opened 2026-02-27
ci: Update gemspec dependencies using renovate

thompson-tomo:
Add in custom manager which detects dependencies in gemspecs and then ensure we have rules which updates them like bundler

### DIFF
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -55,13 +55,6 @@
     {
       matchUpdateTypes: ["major"],
       matchCategories: ["ruby", "docker", "custom"],
-      matchDepTypes: ["development_dependency"],
-      schedule: ["before 8am on Monday"],
-    },
-    {
-      matchUpdateTypes: ["major"],
-      matchCategories: ["ruby", "docker", "custom"],
-      matchDepTypes: ["!development_dependency"],
       dependencyDashboardApproval: true,
     },
     {
@@ -90,6 +83,20 @@
       matchUpdateTypes: ["major"],
       matchDepNames: ["mysql"],
       minimumReleaseAge: "1765 days",
+    },
+    {
+      matchCategories: ["custom"],
+      matchDepTypes: ["!dependency"],
+      matchFileNames: ["**/*.gemspec"],
+      enabled: false,
+    },
+    {
+      matchFileNames: [
+        ".github/actions/test_gem/action.yml",
+        ".github/workflows/installation-tests.yml",
+        ".github/workflows/release-*.yml",
+      ],
+      ignoreDeps: ["ruby"],
     },
   ],
   customManagers: [
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -86,6 +86,7 @@
     },
     {
       matchCategories: ["custom"],
+      matchDepNames: ["!ruby"],
       matchDepTypes: ["!dependency"],
       matchFileNames: ["**/*.gemspec"],
       enabled: false,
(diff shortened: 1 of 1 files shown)


.github/renovate.json5
```diff
@@ -107,6 +114,18 @@
       extractVersionTemplate: "^{{depVersion}}-alpine(?<version>\\d+(?:\\.\\d+)?(?:\\.\\d+)?)$",
       depNameTemplate: "{{packageName}}-alpine",
     },
+    {
+      customType: "regex",
+      description: "Update dependencies in gemspecs",
+      managerFilePatterns: ["**/*.gemspec"],
+      matchStrings: [
+        ".add_(?<depType>.*dependency)\\s'(?<packageName>.+?)',\\s'\\S+\\s(?<depVersion>[0-9.]+)'",
```
arielvalentin: Does renovate not update gempsecs by default or does it only do it for Gemfiles. 

Either way, we only want to keep runtime dependencies in the gem spec so `add_dependency` is sufficient without needing depType.

Development dependencies should be declared in Gemfiles. 

If you find an inconsistency please update the Gemspec to use a Gemfile instead.

thompson-tomo: > Does renovate not update gempsecs by default or does it only do it for Gemfiles.

Correct no support for gemspecs out of the box as per https://github.com/open-telemetry/opentelemetry-ruby-contrib/issues/1803.

> Either way, we only want to keep runtime dependencies in the gem spec so add_dependency is sufficient without needing depType.

DepType is the name of regex capture group which appears in pr's and can be used in rules.

> Development dependencies should be declared in Gemfiles.
> 
> If you find an inconsistency please update the Gemspec to use a Gemfile instead.

I don't believe there were many if any in contrib, however in `otel-ruby` dev dependencies are widely used and my intention was once these pr's done we/I monitor the behaviour for a couple of weeks before syncing the renovate rules and managers across.

arielvalentin: Removing development dependencies from gemspecs is something I'd like to do in the core repo as well. Hopefully that will make the repos a bit more symmetric.

thompson-tomo: Sure, I agree that the more consistent the repos are the better so all for updating core. The question is, do I need to remove the package rule which allows dev dependencies to bypass manual approval to be created?

arielvalentin: I think so. It makes it simpler for me to read and know will communicate to others that renovate will only update `add_dependency` declarations and ignore undesirable  `add_runtime_dependency` and `add_development_dependency`

thompson-tomo: Ok, so based on that description I have removed the rule which handled dev dependencies in a different manner to non dev and instead added a rule which explicitly disabled non dependency updates. This way it is clearer that they have been disabled.

arielvalentin: Am I misunderstanding  something? The capture group is still there.

thompson-tomo: You are correct the capture group is still there to avoid hard coding it. The custom manager just enables the dependency to be detected and listed in the dashboard. 

The package rules are what control if update pr's are raised and in the case here we have a rule which explicitly blocks all updates to `gemfiles` if they are not a `dependency` update. This means that dependencies added via `add_development_dependency` won't be updated but a `add_dependency` would be.

thompson-tomo:
The match string can be tested by doing a regex search over the files. The only way I have found to test the config is by using the cli and running

> LOG_LEVEL=debug npx renovate --platform=local > test.log

arielvalentin (approved):

PR #2032, opened 2026-02-27
ci: Manage min sdk using renovate

thompson-tomo:
Bump min gemspec ruby sdk version on approval within renovate. This ensures the gemspec and github actions are mantained.

This also improves the renovate dashboard by grouping the min sdk updates into the pending status check before moving to approval once previous is EoL.

### DIFF
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -26,8 +26,6 @@
     },
     {
       description: "Separate ruby updates so that patches can be installed after new minor has been released.",
-      groupName: "Min-Ruby-sdk",
-      dependencyDashboardCategory: "Min Ruby SDK",
       matchDepNames: ["ruby"],
       separateMultipleMinor: true,
       separateMinorPatch: true,
@@ -83,14 +81,6 @@
       matchUpdateTypes: ["major"],
       matchDepNames: ["mysql"],
       minimumReleaseAge: "1765 days",
-    },
-    {
-      groupName: "Min-Ruby-sdk",
-      dependencyDashboardCategory: "Min Ruby SDK",
-      matchCategories: ["custom"],
-      matchDepTypes: ["gemspec.required_ruby_version"],
-      separateMultipleMinor: true,
-      dependencyDashboardApproval: true,
     },
   ],
   customManagers: [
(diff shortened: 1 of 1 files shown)


.github/renovate.json5
```diff
@@ -66,6 +67,14 @@
       schedule: ["before 8am on Monday"],
       dependencyDashboardApproval: false,
     },
+    {
+      groupName: "Min-Ruby-sdk",
```
arielvalentin: The group name is throwing me off a little. This is for the minimum supported Ruby Language version right? Not be OTel SDK min version?

.github/renovate.json5
```diff
@@ -107,6 +116,22 @@
       extractVersionTemplate: "^{{depVersion}}-alpine(?<version>\\d+(?:\\.\\d+)?(?:\\.\\d+)?)$",
       depNameTemplate: "{{packageName}}-alpine",
     },
+    {
+      customType: "regex",
```
arielvalentin, line 150: I don't quite understand how this is supposed to work. Do you have a guide that I may review to understand the implications of this change?

.github/renovate.json5
thompson-tomo: Correct this is the min ruby version.

.github/renovate.json5
thompson-tomo: The best doc I have found is https://docs.renovatebot.com/configuration-options/#custommanagers

In effect it locates files which match managerFilePatterns. For each file it does a regex search for each matchString from which the current version is extracted and in this case depname is hard coded. It then uses the datasource to find all new versions. We descard all non x.y.0 updates.

There is a package rule which says if the update matches this depType then it needs approval to be created.

In effect it automates the update of all the gemspecs after an approval.

thompson-tomo:
Putting this to draft till #2031 lands as the rule introduced there will need to be extended to allow these updates through.

.github/renovate.json5
thompson-tomo: Group name has been updated to "Min Ruby Runtime".

arielvalentin (approved):

PR #2033, opened 2026-02-27
ci: Renovate manage appraisals

thompson-tomo:
This enables renovate to detect dependencies in appraisals. The updates are as follows:

- `~>` do nothing as it is pinned
- `>=` update like any other ruby dependency

Also based on testing this will only lead to 1 package update based on current appraisals. 

In the future this current setup does enable us to specify the min version of gems for latest appraisals and then having renovate weekly bump them. This would give us visibility into new versions/features and also increases exposure to abandoned/deprecated gems.

### DIFF
--- a/instrumentation/anthropic/Appraisals
+++ b/instrumentation/anthropic/Appraisals
@@ -12,6 +12,6 @@ if Gem::Version.new(RUBY_VERSION) < Gem::Version.new('4')
   end
 
   appraise 'anthropic-1.9' do
-    gem 'anthropic', '~> 1.9'
+    gem 'anthropic', '~> 1.9.0'
   end
 end
--- a/instrumentation/anthropic/Appraisals
+++ b/instrumentation/anthropic/Appraisals
@@ -11,7 +11,7 @@ if Gem::Version.new(RUBY_VERSION) < Gem::Version.new('4')
     gem 'anthropic'
   end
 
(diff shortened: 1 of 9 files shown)


instrumentation/anthropic/Appraisals
```diff
@@ -11,7 +11,7 @@ if Gem::Version.new(RUBY_VERSION) < Gem::Version.new('4')
     gem 'anthropic'
   end
 
-  appraise 'anthropic-1.9.0' do
-    gem 'anthropic', '~> 1.9.0'
+  appraise 'anthropic-1.9' do
+    gem 'anthropic', '~> 1.9'
```
arielvalentin: I intentionally use pessimistic versioning scoped to bug fix versions and not minor versions.

As new minor versions are released, I add them to the list.

The changes like this one will run the latest minor version e.g. 1.12.3, but claim to be running anthropic-1.9 in the title. 

Unless there is a compelling reason, we should continue to have open ended bug fix scope and not minor versions.

thompson-tomo: You are right, I got the versioning all twisted up in my head. 🤣

kaylareopelle (commented):
Thanks for opening this! Is there a way I can test it out?

arielvalentin (approved):

kaylareopelle (approved):

PR #2034, opened 2026-02-27
ci: Remove dependabot from tooling/processes

thompson-tomo:
This removes the dependabot configuration which is un-necessary due to adoption of renovate and migration of the last of the config in #2031


arielvalentin (approved):
🎉

PR #2046, opened 2026-03-03
test: update test for Excon 1.4.0

hannahramadan:
Excon 1.4.0 changed how sockets are initialized, now requiring a `resolver_factory` parameter that didn't exist before. 

https://github.com/excon/excon/pull/897/changes#diff-0719771aa28e5707a8ce035f72d9b535169d6075f5d3a99f839cb9f73668e3d4R63-R64

Tests that directly initialize a socket now conditionally use a new `excon_socket_options` helper that includes `resolver_factory` when `Excon::ResolverFactory` is defined.


arielvalentin (approved):

instrumentation/excon/test/test_helper.rb
```diff
@@ -29,3 +29,11 @@ def with_sampler(sampler)
 ensure
   OpenTelemetry.tracer_provider.sampler = previous_sampler
 end
+
+# Excon 1.4.0+ requires resolver_factory parameter
+def excon_socket_options(hostname:, port:)
```
arielvalentin, line 35: Anything else we can do to give us a hint that we should drop this helper when we boost min support to 1.4?

arielvalentin (approved):

kaylareopelle (approved):

instrumentation/excon/test/test_helper.rb
hannahramadan: Nice call out! Added a TODO https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/2046/commits/01c43dbdfe3d1f4375b59a0df16e66c5509b994e

arielvalentin (approved):

PR #2051, opened 2026-03-06
feat!: Default to stable HTTP semantic conventions

hannahramadan:
Use stable HTTP semantic conventions by default for HTTP instrumentations and update README:
- http
- http_client
- httpx
- net_http
- rack  
- restclient
- excon
- faraday
- ethon

Emit warnings for users who have set `http/dup` or `old` as the value for `OTEL_SEMCONV_STABILITY_OPT_IN`.

Set the final transition date as **April 15, 2026**

Update README documentation for action_pack and sinatra, which depend on rack.

### DIFF
--- a/instrumentation/action_pack/lib/opentelemetry/instrumentation/action_pack/handlers/action_controller.rb
+++ b/instrumentation/action_pack/lib/opentelemetry/instrumentation/action_pack/handlers/action_controller.rb
@@ -37,7 +37,11 @@ module OpenTelemetry
               OpenTelemetry::SemanticConventions::Trace::CODE_FUNCTION => String(payload[:action])
             }
             attributes[OpenTelemetry::SemanticConventions::Trace::HTTP_ROUTE] = http_route if http_route
-            attributes[OpenTelemetry::SemanticConventions::Trace::HTTP_TARGET] = request.filtered_path if request.filtered_path != request.fullpath
+
+            if request.filtered_path != request.fullpath
+              filtered_query = request.filtered_path.split('?', 2)[1]
+              attributes['url.query'] = filtered_query if filtered_query
+            end
 
             if @span_naming == :semconv
               span.name = if http_route
--- a/instrumentation/excon/lib/opentelemetry/instrumentation/excon/instrumentation.rb
+++ b/instrumentation/excon/lib/opentelemetry/instrumentation/excon/instrumentation.rb
@@ -34,12 +34,18 @@ module OpenTelemetry
           values = stability_opt_in.split(',').map(&:strip)
 
           if values.include?('http/dup')
+            emit_old_semconv_deprecation_warning('http/dup')
             'dup'
-          elsif values.include?('http')
-            'stable'
-          else
+          elsif values.include?('old')
+            emit_old_semconv_deprecation_warning('old')
             'old'
+          else
+            'stable'
           end
+        end
+
+        def emit_old_semconv_deprecation_warning(option)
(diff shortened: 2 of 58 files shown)


instrumentation/action_pack/lib/opentelemetry/instrumentation/action_pack/handlers/action_controller.rb
```diff
@@ -37,7 +37,11 @@ def start(_name, _id, payload)
               OpenTelemetry::SemanticConventions::Trace::CODE_FUNCTION => String(payload[:action])
             }
             attributes[OpenTelemetry::SemanticConventions::Trace::HTTP_ROUTE] = http_route if http_route
-            attributes[OpenTelemetry::SemanticConventions::Trace::HTTP_TARGET] = request.filtered_path if request.filtered_path != request.fullpath
+
+            if request.filtered_path != request.fullpath
```
hannahramadan, line 41: With stable semantic conventions, url.target is split into two attributes, url.path and url.query, set by Rack. Rack sets url.query to the raw query string but we want to override it with the filtered query string when Rails parameter filtering is applied.

arielvalentin: If only this method were public: https://api.rubyonrails.org/classes/ActionDispatch/Http/FilterParameters.html#method-i-filtered_query_string

Would there be anything wrong with using URI parsing instead here?

```ruby
uri = URI.parse(request.filtered_path)
attributes['url.query'] = uri.query 
```

hannahramadan: I didn't see anything wrong with using URI! Checked out the Rails method you linked and they use a string spilt approach. That got me thinking about why they choose that. I ran both through a benchmark and looks like string split is faster 

```
Comparison:
    split('?', 2)[1]:   683895.7 i/s
     URI.parse.query:    35134.9 i/s - 19.46x  slower
```

arielvalentin (approved):

instrumentation/excon/lib/opentelemetry/instrumentation/excon/instrumentation.rb
```diff
@@ -34,14 +34,20 @@ def determine_semconv
           values = stability_opt_in.split(',').map(&:strip)
 
           if values.include?('http/dup')
+            emit_old_semconv_deprecation_warning('http/dup')
             'dup'
-          elsif values.include?('http')
-            'stable'
-          else
+          elsif values.include?('old')
+            emit_old_semconv_deprecation_warning('old')
```
kaylareopelle, line 40: Should the include check here be for `http/old`? The README points to that string as the value. 

If it's changed here, a lot of the other instrumentations need to be updated too.

If it should be the way it is, we may want to update the READMEs.
```suggestion
          elsif values.include?('http/old')
            emit_old_semconv_deprecation_warning('old')
```

kaylareopelle (changes requested):
I saw one docs/code mismatch I want to work through before we merge. Otherwise, looks great! Thank you @hannahramadan!

instrumentation/excon/lib/opentelemetry/instrumentation/excon/instrumentation.rb
hannahramadan: Ah great catch! The docs are wrong, code is right. I will update docs

kaylareopelle (approved):

kaylareopelle (approved):

PR #2052, opened 2026-03-06
ci: Run installation tests against latest

thompson-tomo:
This extends the installation tests so that they are always performed against the latest version.

This latest version will be updated by renovate alongside the version used by release tooling.

### DIFF
--- a/.github/workflows/installation-tests.yml
+++ b/.github/workflows/installation-tests.yml
@@ -5,6 +5,16 @@ name: Installation Tests
 
 on:
   workflow_dispatch:
+  push:
+    branches:
+      - main
+    paths: &path_filters
+      - ".github/workflows/installation-tests.yml"
+      - "releases/**"
+  pull_request:
+    branches:
+      - main
+    paths: *path_filters
   schedule:
     # Everyday at 2 PM UTC
     - cron: "0 14 * * *"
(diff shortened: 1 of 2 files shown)


.github/workflows/installation-tests.yml
```diff
@@ -32,3 +43,17 @@ jobs:
       - name: "Install Latest Gem Versions on ${{ matrix.ruby-version }}"
         working-directory: releases
         run: ./run.sh
+  installation-latest-tests:
```
arielvalentin, line 46: What am I missing? 

How is this different from the matrix install besides fail fast setting?

arielvalentin (commented):
Thank you. I can't believe I missed that last time.

.github/workflows/installation-tests.yml
thompson-tomo: The difference is that the ruby-version is explicitly set to 4.0.1 rather than being defined in the matrix. This way renovate can keep it up to date whenever a new version of ruby is released.

arielvalentin: Are we unable to customize renovate to keep the matrix up to date?

thompson-tomo: I think It might be possible but would take a bit of testing etc, I have raised #2053 to explore this automation later but we would need to decide versioning strategy etc. ie do we only keep latest patch of each minor, do we keep every version. Hence would be good to have this as a starting point and then we can optimise as time goes by.

arielvalentin (approved):

PR #2054, opened 2026-03-07
ci: Auto label pr's based on files changed

thompson-tomo:
This incorporates a labeler action into the ci workflow which adds labels based on the gems which have changed. By labelling the pr's we are increasing visibility as to what gems are being changed and simplify filtering etc.

The next step would be to use the label data to populate the gem matrix property so that we can decrease un-necessary ci testing on pr's to only those in scope.

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -1,0 +1,253 @@
+helpers-mysql:
+  - changed-files:
+      - any-glob-to-any-file:
+          - "helpers/mysql/**"
+helpers-sql:
+  - changed-files:
+      - any-glob-to-any-file:
+          - "helpers/sql/**"
+helpers-sql-obfuscation:
+  - changed-files:
+      - any-glob-to-any-file:
+          - "helpers/sql-obfuscation/**"
+helpers-sql-processor:
+  - changed-files:
+      - any-glob-to-any-file:
+          - "helpers/sql-processor/**"
+
+instrumentation-action_mailer:
+  - changed-files:
+      - any-glob-to-any-file:
+          - "instrumentation/action_mailer/**"
+instrumentation-action_pack:
+  - changed-files:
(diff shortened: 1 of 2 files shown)


.github/workflows/auto-label.yml
```diff
@@ -0,0 +1,14 @@
+name: Auto Label PRs
+
+on:
```
arielvalentin, line 3: Would you mind also adding an option to run the workflow manually using workflow dispatch?

https://docs.github.com/en/actions/how-tos/manage-workflow-runs/manually-run-a-workflow

It would be nice to retroactively apply these changes to existing PRs or test the workflow after config changes.

thompson-tomo: Done

.github/labeler.yml
```diff
@@ -0,0 +1,253 @@
...
+helpers-sql-obfuscation:
+  - changed-files:
+      - any-glob-to-any-file:
+          - "helpers/sql-obfuscation/**"
+helpers-sql-processor:
+  - changed-files:
+      - any-glob-to-any-file:
+          - "helpers/sql-processor/**"
+
+instrumentation-action_mailer:
```
arielvalentin, line 18: Something else occurred to me my second time reviewing this PR.

We have an instrumentation generator script that try's to update any CI files to include the new instrumentation. 

Ideally I'd to see like a follow up PR that will update this config when new gems are added to the repo but I don't think it's something required to merge this PR.

arielvalentin (approved):

.github/labeler.yml
thompson-tomo: Agree have raised https://github.com/open-telemetry/opentelemetry-ruby-contrib/issues/2060 so we don't forget.

PR #2055, opened 2026-03-07
chore: Run lint with reduced customizations

thompson-tomo:
This helps with consistency by enforcing more lint rules

### DIFF
--- a/.markdownlint.yaml
+++ b/.markdownlint.yaml
@@ -1,13 +1,12 @@
 default: true
 
-emphasis-style: false
-line-length: false
-link-fragments: false
-list-marker-space: false
-no-emphasis-as-heading: false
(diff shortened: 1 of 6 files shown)


.markdownlint.yaml
```diff
@@ -1,13 +1,12 @@
...
-no-emphasis-as-heading: false
-no-hard-tabs: false
+ul-style:
+  style: dash
+emphasis-style:
+  style: underscore
+strong-style:
+  style: asterisk
+line-length:
+  line_length: 500
```
thompson-tomo, line 10: I know this is high but any lower and it will trigger text to be updated. This line break formatting can come in seperate PR.

arielvalentin (approved):

PR #2070, opened 2026-03-09
fix: resolve labeller issue

thompson-tomo:
Updated the workflow to use Ubuntu 24.04 and added permissions for labels.

Note I attempted to encorporate automated testing but https://github.com/actions/labeler?tab=readme-ov-file#recommended-permissions informs me that it is not possible.

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -1,4 +1,4 @@
-ci:
+keep:
   - changed-files:
       - any-glob-to-any-file:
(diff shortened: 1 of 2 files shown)


arielvalentin (approved):

arielvalentin (approved):
@kaylareopelle et.al.

I'll be a little more diligent in reviewing actions changes. It's difficult to test changes without actually merging them.

PR #1087, opened 2024-07-25
feat: collect pg db.collection_name attribute

hannahramadan:
The [db.collection.name](https://opentelemetry.io/docs/specs/semconv/database/database-spans/#:~:text=Required-,db.collection.name,-string) attribute is conditionally required for Database spans. This PR uses regex to detect the collection name (called table name in PG) and report it as the `db.collection.name` attribute for PG instrumentation.

closes https://github.com/open-telemetry/opentelemetry-ruby-contrib/issues/1059

PR #2076, opened 2026-03-14
feat: draft - PG semantic convention stability migration

hannahramadan:

Changes (Old -> Stable):   
  - db.system → db.system.name                                               
  - db.name → db.namespace                                                   

  Removed in stable:
  - db.user
  - db.instance.id
  - peer.service
  - db.collection.name - this really should never have been added ([PR](https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/1087)). "The collection name SHOULD NOT be extracted from db.query.text, when the database system supports query text with multiple collections in non-batch operations." ([spec](https://github.com/open-telemetry/semantic-conventions/blob/v1.33.0/docs/attributes-registry/db.md#db-collection-name))

  Added in stable:
  - server.port
  - error.type                             
  - db.response.status_code


instrumentation/pg/README.md
toncid, line 56: I know that it is out of scope of this PR, but would you be willing to also document the `obfuscation_limit: 2000` parameter in this file, please?

The description would be something like:
> When enabled by `db_statement`, this instrumentation will not obfuscate SQL queries longer than 2000 characters. The default limit can be overridden by setting the `obfuscation_limit` to another integer value.

toncid: The config was released in [v0.25.0](https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/main/instrumentation/pg/CHANGELOG.md#v0250--2023-05-25) (on 2023-05-25), but never documented it seems.

arielvalentin: It's going to be a while before this is ready for review and merged. 

@toncid would you be amenable to submitting a PR that documents the option?

toncid: Sure, no problem. Thanks for the quick answer.

toncid: @arielvalentin Done - #2122. 🚀

PR #2086, opened 2026-03-16
ci: Remove hourly pr limits

thompson-tomo:
Remove limit on the number of pr's created per hour. This is based on the fact that still pr's are not raised in a timely manner due to rate limiting hence this does away with the hourly limit.

This setting is described in https://docs.renovatebot.com/configuration-options/#prhourlylimit

### DIFF
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -6,6 +6,7 @@
     ":separateMultipleMajorReleases",
   ],
   prConcurrentLimit: 12,
+  prHourlyLimit: 0,
   packageRules: [
     {
       groupName: "all patch versions",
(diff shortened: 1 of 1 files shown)


kaylareopelle (approved):

.github/renovate.json5
```diff
@@ -6,6 +6,7 @@
     ":separateMultipleMajorReleases",
   ],
   prConcurrentLimit: 12,
+  prHourlyLimit: 0,
```
arielvalentin: What am I missing? I took a brief look the docs because I do not see an explanation of what the behavior is when set to 0. 

https://docs.renovatebot.com/configuration-options/#prhourlylimit

I also do not want to end up abusing our usage in the API. How many per hour do we expect it to open?

thompson-tomo: We are limiting ourselves to 12 open concurrently. I saw the message in the logs but will change it to 12 for certainty.

arielvalentin (approved):

Issue #1459, opened 2025-03-27
Caught AR validation errors are resulting in `error` spans


renchap:
Our app uses ActiveRecord validations ([example](https://github.com/mastodon/mastodon/blob/main/app/models/user.rb#L115)), and uses some `create!` calls in the code (example [controller action](https://github.com/mastodon/mastodon/blob/main/app/controllers/api/v1/accounts_controller.rb#L34), and [`create!`](https://github.com/mastodon/mastodon/blob/main/app/services/app_sign_up_service.rb#L24)).

Then a record fails the validation, an exception is thrown, and is then caught and returned as an HTTP error, with the failed validations in the body ([here](https://github.com/mastodon/mastodon/blob/main/app/controllers/concerns/api/error_handling.rb#L7)).

This causes the OTEL instrumentation to generate traces where the `User.create!` span has an error attached, but the root span does not:

![Image](https://github.com/user-attachments/assets/807073b6-0cdd-47cf-af37-c2ca7564f464)

This causes tools like Sentry or Datadog to create a new entry in their error tracking system, even if everything is handled in the app and the error is not actionable.

Using exceptions for such control flow is a common pattern in Rails, and as an application developer or operator, I expect that errors are things that needs to be reviewed and fixed, which is not the case here.

Note: there is [a relevant discussion in the Slack channel](https://cloud-native.slack.com/archives/C01NWKKMKMY/p1742995231280499)

Possible discussed solutions are marking some exceptions (such as `ActiveRecord::RecordInvalid` or `ActiveRecord:: RecordNotFound`) as non-exceptional in the AR Validations instrumentation, or having a setting to prevent those spans to be errors entirely, allowing the app developer to explicitly indicate that those exceptions are handled correctly by the app (and if thats not the case, the root span should be an error anyway).

UnderpantsGnome:
@arielvalentin can you point me in the right direction to look into this? We just started using OTEL and this "issue" generates a lot of noise in our app, as it reports ALL rescued exceptions.

Thanks

harm-matthias-harms:
We encounter the same issue, not only with AR validations, but with various errors. All errors are caught outside of the span created by one of the instrumentation packages. Has anyone found a workaround for this issue?

aviralgarg05:
I would like to solve this issue please if someone could assign me

arielvalentin:
@kaylareopelle @aviralgarg05 this conversation came up during the SIG. 

[...]

PR #2087, opened 2026-03-16
fix(active-record): avoid error spans for handled RecordInvalid

aviralgarg05:
## Summary
This PR fixes issue #1459 by preventing low-signal error spans for app-handled ActiveRecord bang persistence exceptions.

## What changed
- save!, create!, and update! now close spans and re-raise configured handled exceptions outside the span.
- Default handled exception list remains narrow: [ActiveRecord::RecordInvalid].
- Added optional configuration: handled_exceptions, which can include entries like ActiveRecord::RecordNotFound when teams explicitly want that behavior.

## Why this is in scope
- Default behavior still targets the original issue for RecordInvalid.
- Reviewer feedback about configurability is implemented without broadening defaults.

## Tests
- RecordInvalid is suppressed for save!, create!, and update!.
- Non-handled exceptions still emit exception events.
- Configured handled exception behavior is covered by tests.


instrumentation/active_record/test/instrumentation/active_record/patches/persistence_test.rb
```diff
@@ -30,14 +30,13 @@
       _(save_span).wont_be_nil
     end
 
-    it 'adds an exception event if it raises' do
+    it 'does not add an exception event if it raises a handled validation error' do
```
dmathieu, line 46: You should still test that raising any other error reports an exception.

dmathieu (commented):
What about other errors that usually don't raise either, such as RecordNotFound?
Should the list of exceptions be configurable?

arielvalentin (changes requested):
Thank you for this contribution and patience. We are low on reviewing capacity recently so its taking me a while to get to changes like yours for a review. 

I think that the way Rails allows flow and control through exception handling is a bit at odds with the guidance in the spec which recommends we _should_ record an error if it is unhandled and set the span status to Error. 

https://opentelemetry.io/docs/specs/otel/trace/exceptions/#recording-an-exception

Adding allow lists to a library in this way and then interrupting the flow control and re-raising seems like something we would want to centralized in the `in_span` helper _or_ these methods should inline the `in_span` helper and only record unexpected exceptions.

I think we can also avoid the unnecessary comparisons in the `handle_exception` method if we explicitly handled them in rescue blocks.

kaylareopelle:
Hi @aviralgarg05, thank you for opening this PR. We discussed this during the SIG meeting yesterday. 

In this case, we need to defer [to the spec](https://opentelemetry.io/docs/specs/otel/trace/exceptions/#recording-an-exception) and record the error/update the span status if it's unhandled.

Fortunately, OpenTelemetry has other options to allow you to customize this behavior for your application. One option could be to ignore these errors using the [collector transformprocessor.](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/processor/transformprocessor)

PR #2088, opened 2026-03-16
feat: optimize extract_statement_type regex and encoding in helpers-mysql

schlubbi:
## Summary

`extract_statement_type` is called on every MySQL query to determine the span name. Two inefficiencies make this more expensive than necessary:

### 1. Trailing `.*` in `QUERY_NAME_REGEX`

The regex at `helpers/mysql/lib/opentelemetry/helpers/mysql.rb:45` has a trailing `.*` that forces the regex engine to scan the **entire** SQL string after finding the keyword. Only the first capture group (the SQL keyword like `select`, `insert`, etc.) is ever used — the `.*` serves no purpose.

**Fix:** Remove the trailing `.*`, making regex cost O(1) in query length instead of O(n).

### 2. Unconditional `utf8_encode` with `binary: true` on every query

`extract_statement_type` calls `OpenTelemetry::Common::Utilities.utf8_encode(sql, binary: true)` on every invocation. The `binary: true` flag forces a full `string.encode('UTF-8', 'binary', ...)` re-encoding **every time**, even though the vast majority of SQL strings are already valid UTF-8 (Ruby string literals default to UTF-8, and MySQL connections typically use `utf8mb4`).

**Fix:** Add a fast-path check — skip encoding entirely when the string is already valid UTF-8. Fall back to the existing `binary: true` encoding only for non-UTF-8 or invalid byte sequences.

## Benchmarks

<details>
<summary>Regex improvement (Part 1)</summary>

Measured with `benchmark-ips` — `QUERY_NAME_REGEX.match(sql)` across representative query shapes:

| Query shape | Length | Original | Fixed | Speedup |
|---|---|---|---|---|
| Simple SELECT | 53 chars | 930 ns | 677 ns | 1.4× |
| With marginalia comment | 143 chars | 1.60 µs | 1.25 µs | 1.3× |
| Multi-JOIN | 768 chars | 6.03 µs | 1.01 µs | **6.0×** |
| Large IN clause | 2,406 chars | 14.24 µs | 742 ns | **19.2×** |

</details>

<details>
<summary>Encoding fast-path (Part 2)</summary>

Measured cost of `utf8_encode` with different strategies on a typical UTF-8 SQL string:

| Strategy | Cost | Notes |
|---|---|---|
| `utf8_encode(sql, binary: true)` | 821 ns | Current — always re-encodes |
| `utf8_encode(sql)` (no binary flag) | 126 ns | Uses existing UTF-8 fast path in Common::Utilities |
| Encoding check + skip | ~55 ns | New fast path — avoids method call entirely |

</details>

<details>
<summary>Combined pipeline (end-to-end extract_statement_type)</summary>

Full `extract_statement_type` call including encoding + regex:

| Query shape | Original | Optimized | Speedup |
|---|---|---|---|
| Short SELECT (53c) | 1.91 µs | 833 ns | **2.3×** |
| With comment (143c) | 2.60 µs | 1.38 µs | **1.9×** |
| Multi-JOIN (768c) | 6.97 µs | 1.13 µs | **6.2×** |
| Large IN clause (2,406c) | 15.19 µs | 886 ns | **17.1×** |

</details>

## Changes

- `helpers/mysql/lib/opentelemetry/helpers/mysql.rb`:
  - Remove trailing `.*` from `QUERY_NAME_REGEX` (line 45)
  - Add UTF-8 fast-path check in `extract_statement_type` — skip `utf8_encode` when string is already valid UTF-8
  - Early return for `nil` input (clearer control flow)

- `helpers/mysql/test/helpers/mysql_test.rb`:
  - Expanded test coverage for `extract_statement_type` from 7 to 35+ test cases
  - Added coverage for: all 15 supported query names, case insensitivity, leading whitespace, long queries (IN clauses, JOINs), prepended comments, encoding handling (UTF-8, ASCII, binary, invalid bytes, frozen strings), nil/empty/whitespace inputs, unrecognized statements, and partial keyword non-matching

### DIFF
--- a/helpers/mysql/lib/opentelemetry/helpers/mysql.rb
+++ b/helpers/mysql/lib/opentelemetry/helpers/mysql.rb
@@ -42,7 +42,7 @@ module OpenTelemetry
       # Ignore query names that might appear in comments prepended to the
       # statement.
       PREPENDED_COMMENTS_REGEX = %r{(?:\/\*.*?\*\/)}m
-      QUERY_NAME_REGEX = Regexp.new("^\s*(?:#{PREPENDED_COMMENTS_REGEX})?\s*\\b(#{QUERY_NAMES.join('|')})\\b.*", Regexp::IGNORECASE)
+      QUERY_NAME_REGEX = Regexp.new("^\s*(?:#{PREPENDED_COMMENTS_REGEX})?\s*\\b(#{QUERY_NAMES.join('|')})\\b", Regexp::IGNORECASE)
 
       # This is a span naming utility intended for use in MySQL database
       # adapter instrumentation.
@@ -68,9 +68,13 @@ module OpenTelemetry
 
       # @api private
       def extract_statement_type(sql)
-        sql = OpenTelemetry::Common::Utilities.utf8_encode(sql, binary: true)
+        return unless sql
 
-        QUERY_NAME_REGEX.match(sql) { |match| match[1].downcase } unless sql.nil?
+        unless sql.encoding == ::Encoding::UTF_8 && sql.valid_encoding?
+          sql = OpenTelemetry::Common::Utilities.utf8_encode(sql, binary: true)
+        end
+
+        QUERY_NAME_REGEX.match(sql) { |match| match[1].downcase }
       rescue StandardError => e
         OpenTelemetry.handle_error(message: 'Error extracting SQL statement type', exception: e)
         nil
--- a/helpers/mysql/lib/opentelemetry/helpers/mysql.rb
+++ b/helpers/mysql/lib/opentelemetry/helpers/mysql.rb
@@ -70,9 +70,7 @@ module OpenTelemetry
       def extract_statement_type(sql)
         return unless sql
 
-        unless sql.encoding == ::Encoding::UTF_8 && sql.valid_encoding?
-          sql = OpenTelemetry::Common::Utilities.utf8_encode(sql, binary: true)
-        end
+        sql = OpenTelemetry::Common::Utilities.utf8_encode(sql, binary: true) unless sql.encoding == ::Encoding::UTF_8 && sql.valid_encoding?
 
         QUERY_NAME_REGEX.match(sql) { |match| match[1].downcase }
       rescue StandardError => e
--- a/helpers/mysql/test/helpers/mysql_test.rb
+++ b/helpers/mysql/test/helpers/mysql_test.rb
@@ -102,55 +102,168 @@ describe OpenTelemetry::Helpers::MySQL do
       end
     end
 
-    describe 'when sql contains invalid byte sequences' do
-      let(:sql) { "SELECT * from users where users.id = 1 and users.email = 'test@test.com\255'" }
-
-      it 'extracts the statement' do
-        assert_equal('select', extract_statement_type)
+    describe 'recognizes all supported query names' do
+      {
+        'SELECT 1' => 'select',
+        'INSERT INTO users VALUES (1)' => 'insert',
+        'UPDATE users SET name = "a"' => 'update',
+        'DELETE FROM users' => 'delete',
+        'BEGIN' => 'begin',
+        'COMMIT' => 'commit',
+        'ROLLBACK' => 'rollback',
+        'SAVEPOINT sp1' => 'savepoint',
+        'RELEASE SAVEPOINT sp1' => 'release savepoint',
+        'EXPLAIN SELECT 1' => 'explain',
+        'DROP DATABASE test_db' => 'drop database',
+        'DROP TABLE users' => 'drop table',
+        'CREATE DATABASE test_db' => 'create database',
+        'CREATE TABLE users (id INT)' => 'create table',
+        "SET NAMES 'utf8mb4'" => 'set names'
+      }.each do |query, expected|
+        it "extracts '#{expected}' from: #{query[0..40]}" do
+          assert_equal(expected, OpenTelemetry::Helpers::MySQL.extract_statement_type(query))
+        end
       end
     end
 
-    describe 'when sql contains unknown query statement' do
-      let(:sql) { 'DESELECT 1' }
+    describe 'case insensitivity' do
+      it 'handles uppercase' do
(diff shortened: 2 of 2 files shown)


helpers/mysql/lib/opentelemetry/helpers/mysql.rb
```diff
@@ -68,9 +68,11 @@ def database_span_name(sql, operation, database_name, config)
 
       # @api private
       def extract_statement_type(sql)
-        sql = OpenTelemetry::Common::Utilities.utf8_encode(sql, binary: true)
+        return unless sql
 
-        QUERY_NAME_REGEX.match(sql) { |match| match[1].downcase } unless sql.nil?
+        sql = OpenTelemetry::Common::Utilities.utf8_encode(sql, binary: true) unless sql.encoding == ::Encoding::UTF_8 && sql.valid_encoding?
```
arielvalentin, line 73: Is this a case where we are "holding it wrong" and something should change in either how we are using the helper and/or the helper itself needs to change for a more optimized solution? 

https://github.com/open-telemetry/opentelemetry-ruby/blob/575edf4fad5d38547a372fdc273b9c1640b33e28/common/lib/opentelemetry/common/utilities.rb#L60

helpers/mysql/lib/opentelemetry/helpers/mysql.rb
```diff
@@ -68,9 +68,11 @@ def database_span_name(sql, operation, database_name, config)
...
+
+        QUERY_NAME_REGEX.match(sql) { |match| match[1].downcase }
```
thompson-tomo, line 75: as per https://opentelemetry.io/docs/specs/semconv/db/database-spans/#name in particular 

> The operation name SHOULD NOT be extracted from db.query.text, when the database system supports query text with multiple operations in non-batch operations.

hence if multiple matches we should return null.

schlubbi: Good catch on the spec reference. I agree this is worth considering, but it's orthogonal to this PR — the "match first keyword" behavior is pre-existing and unchanged here. We only removed the trailing .* from the regex and added a UTF-8 fast path.

  [Trilogy doesn't enable MULTI_STATEMENTS by default](https://github.com/trilogy-libraries/trilogy/blob/9f8ba20e639690a823a786bb5aa857179d15c7b8/contrib/ruby/ext/trilogy-ruby/cext.c#L702-L703) and [neither does mysql2](https://github.com/brianmario/mysql2/blob/b009d7e114729cbae5bef069a1033dd78acf7745/lib/mysql2/client.rb#L14) — both require an explicit multi_statement option — so in practice I'd argue the regex only ever sees one logical statement per call.

  If the maintainers want to revisit the extraction semantics to align more closely with the spec, I'd be happy to help — but I think that deserves its own PR since it would be a behavioral change rather than a performance optimization

helpers/mysql/lib/opentelemetry/helpers/mysql.rb
schlubbi: Yes, the original code was "holding it wrong" — it passes `binary: true` unconditionally, which bypasses [the fast path that already exists in utf8_encode for UTF-8 strings](https://github.com/open-telemetry/opentelemetry-ruby/blob/575edf4fad5d38547a372fdc273b9c1640b33e28/common/lib/opentelemetry/common/utilities.rb#L63-L64). The `binary: true` flag forces a full `string.encode('UTF-8', 'binary', ...)` [re-encoding on every call](https://github.com/open-telemetry/opentelemetry-ruby/blob/575edf4fad5d38547a372fdc273b9c1640b33e28/common/lib/opentelemetry/common/utilities.rb#L60-L62), even when the string is already valid UTF-8.

  This could also be fixed upstream in opentelemetry-ruby by adding the same fast-path check to the binary branch of utf8_encode:

```ruby
   if binary
     return string if string.encoding == ::Encoding::UTF_8 && string.valid_encoding?
     string.encode('UTF-8', 'binary', invalid: :replace, undef: :replace, replace: '')
```

That would fix it for all callers passing `binary: true`, and `extract_statement_type` could go back to a plain `utf8_encode(sql, binary: true)` without its own guard. Happy to open that PR against opentelemetry-ruby as well. In the meantime, the check here works today without waiting for an upstream release — and would become redundant but harmless once the upstream fix lands.

helpers/mysql/lib/opentelemetry/helpers/mysql.rb
kaylareopelle: Thank you for looking at this with an eye for semantic conventions, @thompson-tomo . I think this is something that @hannahramadan may be working on in a separate PR with her efforts to introduce the `SEMCONV_STABILITY_OPT_IN` environment variable in DB libraries.

hannahramadan: I'm working on the database semantic conventions [migration](https://opentelemetry.io/docs/specs/semconv/non-normative/db-migration/) and think this would be a good to consider revisiting for the move to stable semantic conventions. 

`db.operation.name` also says it should be captured: "If readily available and if there is a single operation name that describes the database call."

I think we could also argue that `db.operation.name` isn't readily available. For strict spec adherence, we probably shouldn't capture by default, but I could see potentially a config if folks were keen to keep this behavior as an option. In any case, I agree that this change good to go as a performance optimization!

Edit: thanks @kaylareopelle! The page hadn't refreshed to see your comment ◡̈

helpers/mysql/lib/opentelemetry/helpers/mysql.rb
kaylareopelle: Thanks for breaking that down. I think a contribution to opentelemetry-ruby with this change would be helpful 🙂

kaylareopelle (approved):
These changes look helpful to me. Thank you for the benchmarks! Before we merge, the conventional commit prefix needs to change to one that will initiate a release. Could you change it to `feat` or `fix` (whichever you think fits better)?

arielvalentin (approved):

PR #2089, opened 2026-03-16
feat: cache static span attributes in Trilogy instrumentation

schlubbi:
## Summary

The Trilogy instrumentation rebuilds a hash of static span attributes on every `query()` call. Attributes like `db.system`, `net.peer.name`, `db.name`, `db.user`, and `peer_service` never change after connection initialization, yet a new hash is allocated, populated, and then merged on every query. This PR caches these static attributes at initialization time and uses `Hash#dup` on the hot path.

### Changes

**1. Cache base attributes at initialization**

A new `populate_base_attributes` method runs during `initialize` and builds a frozen hash of static attributes (`db.system`, `net.peer.name`, `db.name`, `db.user`, `peer_service`). On each `query()` call, `client_attributes` now calls `@_otel_base_attributes.dup` instead of constructing a new hash from scratch.

**2. Cache `database_name` and `database_user` as instance variables**

Previously, `database_name` and `database_user` performed a hash lookup on `connection_options` on every call. These values are now cached as `@_otel_database_name` and `@_otel_database_user` during initialization.

**3. Single `Trilogy.attributes` lookup per query**

The `query()` method previously called `OpenTelemetry::Instrumentation::Trilogy.attributes` twice — once to extract `DB_OPERATION` for span naming, and again to merge into span attributes. It now reads the context attributes once into a local variable and passes it to both uses.

### Dynamic attributes preserved

- `db.instance.id` (`@connected_host`) is added per-call since it can change after connection (e.g., after failover)
- `db.statement` is inherently per-query
- `Trilogy.attributes` (context propagation) is per-call

## Benchmarks

<details>
<summary>client_attributes micro-benchmark</summary>

Measured with `Process.clock_gettime` over 500k iterations. Test setup: `Trilogy.allocate` with typical connection options (host, database, username), `db_statement: :omit` config.

| Method | Before | After | Speedup |
|---|---|---|---|
| `client_attributes` (no sql) | 864 ns | 163 ns | **5.3×** |
| `client_attributes` (with sql, omit) | 946 ns | 235 ns | **4.0×** |
| `database_name` | 318 ns | 58 ns | **5.5×** |
| `database_user` | 193 ns | 59 ns | **3.3×** |

</details>

<details>
<summary>Hash construction comparison</summary>

Comparing strategies for building the per-query attributes hash:

| Strategy | Cost |
|---|---|
| Hash literal (4 keys) — current approach | 186 ns |
| `Hash#dup` from frozen cached hash — new approach | 104 ns |

The cached `dup` approach is ~45% faster for hash construction alone, but the real win comes from eliminating the repeated method calls (`database_name`, `database_user`, `config[:peer_service]`, `connection_options.fetch`) that populated the hash on every call.

</details>

## Test coverage

- All 59 existing integration tests pass (tested against MySQL 8.0)
- Added new unit test file (`patches/client_attributes_test.rb`) with 19 tests covering:
  - All static attribute keys (db.system, net.peer.name, db.name, db.user, peer_service, db.instance.id)
  - Nil/missing connection options (unknown sock fallback, omitted keys)
  - Hash independence (each call returns a distinct hash)
  - SQL statement handling (include, omit, obfuscate modes)
  - `database_name` and `database_user` accessor behavior

### DIFF
--- a/instrumentation/trilogy/lib/opentelemetry/instrumentation/trilogy/patches/client.rb
+++ b/instrumentation/trilogy/lib/opentelemetry/instrumentation/trilogy/patches/client.rb
@@ -15,7 +15,8 @@ module OpenTelemetry
         module Client
           def initialize(options = {})
             @connection_options = options # This is normally done by Trilogy#initialize
-            populate_base_attributes
+            @_otel_database_name = connection_options&.dig(:database)
+            @_otel_base_attributes = _build_otel_base_attributes.freeze
 
             tracer.in_span(
               'connect',
@@ -66,20 +67,16 @@ module OpenTelemetry
 
           private
 
-          def populate_base_attributes
-            @_otel_database_name = connection_options&.dig(:database)
-            @_otel_database_user = connection_options&.dig(:username)
-            @_otel_base_attributes = build_base_attributes.freeze
-          end
+          def _build_otel_base_attributes
+            database_user = connection_options&.dig(:username)
 
-          def build_base_attributes
             attributes = {
               ::OpenTelemetry::SemanticConventions::Trace::DB_SYSTEM => 'mysql',
               ::OpenTelemetry::SemanticConventions::Trace::NET_PEER_NAME => connection_options&.fetch(:host, 'unknown sock') || 'unknown sock'
             }
 
             attributes[::OpenTelemetry::SemanticConventions::Trace::DB_NAME] = @_otel_database_name if @_otel_database_name
-            attributes[::OpenTelemetry::SemanticConventions::Trace::DB_USER] = @_otel_database_user if @_otel_database_user
+            attributes[::OpenTelemetry::SemanticConventions::Trace::DB_USER] = database_user if database_user
             attributes[::OpenTelemetry::SemanticConventions::Trace::PEER_SERVICE] = config[:peer_service] unless config[:peer_service].nil?
             attributes
           end
@@ -100,14 +97,6 @@ module OpenTelemetry
             end
 
             attributes
-          end
-
-          def database_name
-            @_otel_database_name
-          end
-
-          def database_user
-            @_otel_database_user
           end
 
           def tracer
--- a/instrumentation/trilogy/lib/opentelemetry/instrumentation/trilogy/patches/client.rb
+++ b/instrumentation/trilogy/lib/opentelemetry/instrumentation/trilogy/patches/client.rb
@@ -15,6 +15,7 @@ module OpenTelemetry
         module Client
           def initialize(options = {})
             @connection_options = options # This is normally done by Trilogy#initialize
+            populate_base_attributes
 
             tracer.in_span(
               'connect',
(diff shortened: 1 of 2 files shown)


instrumentation/trilogy/lib/opentelemetry/instrumentation/trilogy/patches/client.rb
```diff
@@ -92,11 +103,11 @@ def client_attributes(sql = nil)
           end
 
           def database_name
```
arielvalentin: This question applies to `database_name` and `database_user.

Do we need these helper methods anymore since you are referencing them using direct instance variable names elsewhere?

instrumentation/trilogy/lib/opentelemetry/instrumentation/trilogy/patches/client.rb
```diff
@@ -67,15 +66,27 @@ def query(sql)
...
+          def build_base_attributes
             attributes = {
               ::OpenTelemetry::SemanticConventions::Trace::DB_SYSTEM => 'mysql',
               ::OpenTelemetry::SemanticConventions::Trace::NET_PEER_NAME => connection_options&.fetch(:host, 'unknown sock') || 'unknown sock'
             }
 
-            attributes[::OpenTelemetry::SemanticConventions::Trace::DB_NAME] = database_name if database_name
-            attributes[::OpenTelemetry::SemanticConventions::Trace::DB_USER] = database_user if database_user
+            attributes[::OpenTelemetry::SemanticConventions::Trace::DB_NAME] = @_otel_database_name if @_otel_database_name
+            attributes[::OpenTelemetry::SemanticConventions::Trace::DB_USER] = @_otel_database_user if @_otel_database_user
```
arielvalentin: Now that I am looking at this a bit closer, why did we treat the DB name and user separately from the `net.peer.name` attribute before this change 🤔?

Do we use the ivars for any other reason than to populate these attributes?

instrumentation/trilogy/lib/opentelemetry/instrumentation/trilogy/patches/client.rb
```diff
@@ -67,15 +66,27 @@ def query(sql)
 
           private
 
-          def client_attributes(sql = nil)
+          def populate_base_attributes
```
arielvalentin: populate an build are only called together in the constructor right?

Later on the code only uses the ivar https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/2089/changes#diff-6ba5c9beb3eb94883042f29a069003cab7f6296395e612e75b0abe00c98abbb3R88

Can we inline this all and only keep the has around once with all of the static attributes in it?

instrumentation/trilogy/lib/opentelemetry/instrumentation/trilogy/patches/client.rb
```diff
@@ -15,6 +15,8 @@ module Patches
         module Client
           def initialize(options = {})
             @connection_options = options # This is normally done by Trilogy#initialize
+            @_otel_database_name = connection_options&.dig(:database)
+            @_otel_base_attributes = _build_otel_base_attributes.freeze
```
kaylareopelle, line 19: I'm curious -- why did you add an underscore before the names of these instance variables?

kaylareopelle (approved):
This is great! Thanks for your contribution, @schlubbi! 

The tests and benchmarks are really helpful.

I don't believe our release tooling will pick up `perf` as a conventional commit to initiate a release. `feat` or `fix` may better suit this change.

instrumentation/trilogy/lib/opentelemetry/instrumentation/trilogy/patches/client.rb
schlubbi: I think the reasoning here was to avoid collisions with any instance variable added to Trilogy itself. Happy to rename them as the chances that there will be an actual collision are rather small!

schlubbi: @kaylareopelle do you want me to change the naming and drop the `_` prefix?

arielvalentin: @kaylareopelle lets keep the prefixes. If that is not already mentioned in the instrumentation contribution guide I will add it.

kaylareopelle: Thank you, @schlubbi! I agree with @arielvalentin, let's keep them. I just was curious about the motivation and your explanation makes sense 😄

arielvalentin (approved):

PR #2095, opened 2026-03-18
feat: Trilogy attribute semantic convention stability

hannahramadan:

Changes (Old -> Stable):   
  - db.system → db.system.name                                               
  - db.name → db.namespace                                                   

  Removed in stable:
  - db.user
  - db.instance.id
  - peer.service

  Added in stable:
  - server.port
  - error.type                             
  - db.response.status_code


instrumentation/trilogy/lib/opentelemetry/instrumentation/trilogy/patches/stable/client.rb
```diff
@@ -0,0 +1,133 @@
...
+              @_otel_base_attributes = _build_otel_base_attributes.freeze
+
+              tracer.in_span(
+                'connect',
+                attributes: client_attributes.merge!(OpenTelemetry::Instrumentation::Trilogy.attributes),
+                kind: :client,
+                record_exception: config[:record_exception]
+              ) do |span|
+                super
+              rescue StandardError => e
```
hannahramadan, line 29: This is new, to capture `error.type` and `db.status.status_code`. The same pattern exists in the `ping` and `query` methods.

instrumentation/trilogy/lib/opentelemetry/instrumentation/trilogy/patches/stable/client.rb
```diff
@@ -0,0 +1,133 @@
...
+            def _build_otel_base_attributes
+              mysql_host = connection_options&.fetch(:host, nil) || 'unknown sock'
+              mysql_port = connection_options&.dig(:port)
+
+              attributes = {
+                'db.system.name' => 'mysql',
+                'server.address' => mysql_host
+              }
+
+              attributes['server.port'] = mysql_port if mysql_port
```
hannahramadan, line 89: `server.port` is a new attribute. I debated always adding this as it is only required when different from the default, and originally first checked if it was different from the default before adding. I am happy to add this condition back in. 

[10] server.port: When observed from the client side, and when communicating through an intermediary, server.port SHOULD represent the server port behind any intermediaries, for example proxies, if it's available.

[11] server.port: If using a port other than the default port for this DBMS and if server.address is set.

helpers/mysql/lib/opentelemetry/helpers/mysql.rb
```diff
@@ -66,6 +67,31 @@ def database_span_name(sql, operation, database_name, config)
...
+      #
+      # Note: Per spec, db.operation.name SHOULD NOT be extracted from db.query.text.
+      # The operation should only be used if explicitly provided by the application
+      # (e.g., via with_attributes).
+      #
+      # @param operation [String] The database operation (db.operation.name), if provided by the application.
+      # @param database_name [String] The name of the database (db.namespace).
+      # @return [String] The span name.
+      # @api private
+      def stable_database_span_name(operation, database_name)
```
hannahramadan: I would love to hear opinions on the span name change. I believe this as-is follows the spec, but want to make sure users are okay with this change. Note: it removes config[:span_name], so users wouldn't be able to set it anymore. I'm happy to revert this - just trying to follow spec to get started.

**Span Naming Changes for Stable Semantic Conventions**   
  The stable database semantic conventions spec defines a specific [span naming convention](https://github.com/open-telemetry/semantic-conventions/blob/v1.33.0/docs/database/database-spans.md#name):                                                               
  1. {db.query.summary} if available
  2. {db.operation.name} {target} if operation is available
  3. {target} alone (e.g., db.namespace)
  4. {db.system.name} as fallback (e.g., mysql)

  Additionally, the spec states:

>   "The operation name SHOULD NOT be extracted from db.query.text"     
  
  Stable/Dup semconv (new behavior):
  - Span name follows the spec, ignoring config[:span_name]
  - Operation is not extracted from SQL
  - Operation is only used if explicitly provided via with_attributes
  - Fallback chain: {operation} {db.namespace} → {db.namespace} → mysql

We have a [PR](https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/1918) to generate a summary query, which I believe it was discussed to put that behind a config because of potential overhead. This could be added later as an option for span name.

instrumentation/trilogy/lib/opentelemetry/instrumentation/trilogy/patches/stable/client.rb
```diff
@@ -0,0 +1,131 @@
...
+            def initialize(options = {})
+              @connection_options = options # This is normally done by Trilogy#initialize
+              @_otel_database_name = connection_options&.dig(:database)
+              @_otel_base_attributes = _build_otel_base_attributes.freeze
+
+              tracer.in_span(
+                'connect',
+                attributes: client_attributes.merge!(OpenTelemetry::Instrumentation::Trilogy.attributes),
+                kind: :client,
+                record_exception: config[:record_exception]
```
thompson-tomo, line 26: Should we avoid recording exceptions on the span for stable given span event deprecation?

instrumentation/trilogy/lib/opentelemetry/instrumentation/trilogy/patches/stable/client.rb
```diff
@@ -0,0 +1,131 @@
...
+              ) do |span|
+                super
+              rescue StandardError => e
+                set_error_attributes(span, e)
+                raise
+              end
```
thompson-tomo, line 32: Should we also capture a `db.client.operation.exception` event if recording exceptions is enabled.

instrumentation/trilogy/lib/opentelemetry/instrumentation/trilogy/patches/stable/client.rb
```diff
@@ -0,0 +1,131 @@
...
+
+              attributes = {
+                'db.system.name' => 'mysql',
+                'server.address' => mysql_host
+              }
+
+              attributes['server.port'] = mysql_port if mysql_port
+
+              attributes['db.namespace'] = @_otel_database_name if @_otel_database_name
+              attributes[::OpenTelemetry::SemanticConventions::Trace::PEER_SERVICE] = config[:peer_service] unless config[:peer_service].nil?
```
thompson-tomo: I would suggest leaving peer.service out of stable as it is not stable and is likely to be renamed again

helpers/mysql/lib/opentelemetry/helpers/mysql.rb
thompson-tomo: I don't see a new need to define config for span name.

instrumentation/trilogy/lib/opentelemetry/instrumentation/trilogy/patches/stable/client.rb
```diff
@@ -0,0 +1,131 @@
...
+            end
+
+            def client_attributes(sql = nil)
+              attributes = @_otel_base_attributes.dup
+
+              if sql
+                case config[:db_statement]
+                when :obfuscate
+                  attributes['db.query.text'] =
+                    OpenTelemetry::Helpers::SqlProcessor.obfuscate_sql(sql, obfuscation_limit: config[:obfuscation_limit], adapter: :mysql)
```
thompson-tomo, line 102: Is there a reason for not using the global value length limit setting https://opentelemetry.io/docs/specs/otel/common/#configurable-parameters

instrumentation/trilogy/lib/opentelemetry/instrumentation/trilogy/patches/stable/client.rb
robbkidd: 👍🏻 I'm in favor on recording the port, even if it's the default. 

It's the trade off between:
* some extra compute (memory, data transfer, and storage)
* downstream interpreters of telemetry—human or bot—being left to make assumptions in the absence of explicit data

Maybe we'll get complaints and can add Yet Another Config Option, but I say we optimize for the consumers of telemetry here.

instrumentation/trilogy/lib/opentelemetry/instrumentation/trilogy/patches/stable/client.rb
robbkidd: > Should we **avoid recording exceptions** on the span for stable given span event deprecation? [emphasis added]

I vote: continue to support this parameter in our `tracer.in_span()` convenience method to activate recording exceptions within spans it creates.

I think [the span eevnt deprecation plan OTEP](https://github.com/open-telemetry/opentelemetry-specification/blob/fd43145dde7e5192ebc59a20992d98a3e6af5553/oteps/4430-span-event-api-deprecation-plan.md) still has some open questions still about handing exceptions within spans. Once we start tackling the span event deprecation, we would update `in_span()` to use either (1) an updated `span.record_exception()` that actually emits an exception-flavored, trace-participating log record or (2) use the `logger.on_emit()` interface directly.

… or confusingly still emit span events: [Sending (log-based) exceptions and Events as Span Events](https://github.com/open-telemetry/opentelemetry-specification/blob/fd43145dde7e5192ebc59a20992d98a3e6af5553/oteps/4430-span-event-api-deprecation-plan.md). I suspect the span event code paths in API and SDKs are going to continue to exist for a long time, even if with deprecation notices recommending using logs.

helpers/mysql/lib/opentelemetry/helpers/mysql.rb
arielvalentin: "Should not" isn't the same as "must not".

That leaves the decision to instrumentation authors to decide how to handle it.

My vote is to keep using the parsed operation name in place until the summary processor is ready to replace it.

instrumentation/trilogy/lib/opentelemetry/instrumentation/trilogy/patches/stable/client.rb
hannahramadan: Thank you, removed! https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/2095/changes/769b3e2c53a4ce6f14c05904a112f71a02af0b7a

helpers/mysql/lib/opentelemetry/helpers/mysql.rb
hannahramadan: If using sql summary for span name is planned to be behind a config, perhaps we leave the naming option for the config? Then we would just add another case statement to the current. 

```ruby
      def database_span_name(sql, operation, database_name, config)
        case config[:span_name]
        when :statement_type
          extract_statement_type(sql)
        when :db_name
          database_name
        when :db_operation_and_name
          db_operation_and_name(operation, database_name)
        when :sql_summary
           OpenTelemetry::Helpers::SQLProcessor.summary(sql) # for example
        end || 'mysql'
      end
```

instrumentation/trilogy/lib/opentelemetry/instrumentation/trilogy/patches/stable/client.rb
hannahramadan: I hadn't noted this before. It seems like the `obfuscation_limit` is for not passing too much into the regex as a performance guard. It sets a default of 2000. While the attribute limit is global and unlimited by default. It seems they are complimentary.

instrumentation/trilogy/lib/opentelemetry/instrumentation/trilogy/patches/stable/client.rb
hannahramadan: I think we do this, when config[:record_exception] is passed to [in_span](https://github.com/open-telemetry/opentelemetry-ruby/blob/edfc008e2dfdbc82aa78607797b1bb79c6d2ac73/api/lib/opentelemetry/trace/tracer.rb#L34), the SDK records an [exception event](https://github.com/open-telemetry/opentelemetry-ruby/blob/edfc008e2dfdbc82aa78607797b1bb79c6d2ac73/sdk/lib/opentelemetry/sdk/trace/span.rb#L195) (with exception.type, exception.message, exception.stacktrace).

instrumentation/trilogy/lib/opentelemetry/instrumentation/trilogy/patches/stable/client.rb
hannahramadan: Thank you both! It sounds like we can keep this as-is for now and take care of this update separately later

instrumentation/trilogy/lib/opentelemetry/instrumentation/trilogy/patches/stable/client.rb
hannahramadan: "Yet Another Config Option" 😂 Right lets keep this, I feel the same way about recording even if its the default

instrumentation/trilogy/lib/opentelemetry/instrumentation/trilogy/patches/stable/client.rb
thompson-tomo: Haven't those open questions been resolved based on we now use domain specific event Names and there is a convince logger method.

This PR is about stable semconv and given that https://github.com/open-telemetry/semantic-conventions/pull/3256 now being merged shouldn't we be following that as well? Note the api deprecation is handled via https://github.com/open-telemetry/opentelemetry-specification/pull/4825/ 

The key thing for me is

> It SHOULD stop using these, instead recording exceptions and events using the Logs API.

If we don't make this change now, then at a later point in time we need to introduce yet another migration process and also we would be required to bump the major version.

instrumentation/trilogy/lib/opentelemetry/instrumentation/trilogy/patches/stable/client.rb
thompson-tomo: I was more referring to following https://github.com/open-telemetry/semantic-conventions/blob/main/docs/db/database-exceptions.md

instrumentation/trilogy/lib/opentelemetry/instrumentation/trilogy/patches/stable/client.rb
```diff
@@ -0,0 +1,130 @@
...
+                when :include
+                  attributes['db.query.text'] = sql
+                end
+              end
```
thompson-tomo, line 106: Reading the spec and the trilogy api,

> Non-parameterized query text SHOULD NOT be collected by default unless there is sanitization that excludes sensitive data, e.g. by redacting all literal values present in the query text. See Sanitization of db.query.text. Parameterized query text SHOULD be collected by default (the query parameter values themselves are opt-in, see db.query.parameter.<key>).

Given that the execute method doesn't support parameterised queries, shouldn't we always be running the obfuscation and simply offer the ability to exclude the attribute?

instrumentation/trilogy/README.md
thompson-tomo: Can we update this.

thompson-tomo, line 56: Can we add a configuration options section here.

instrumentation/trilogy/lib/opentelemetry/instrumentation/trilogy/patches/stable/client.rb
hannahramadan: We would need to make changes to the Logs API and include it as a dependency for each of these libraries in  order to do this. The api deprecation [ticket](https://github.com/open-telemetry/opentelemetry-specification/pull/4825) still has a number of open items, so I'm not reading that as ready yet either.

If others are okay with this, I would prefer to leave this as-is and open an issue to track the stability of recording exceptions on logs and make changes when ready. 

The `OTEL_SEMCONV_STABILITY_OPT_IN` variable is intended to help users transition to an _initial_ stable version, so we should be okay to make these updates later on. The agent is still on an experimental version, so we shouldn't need to do any major version bumping.

thompson-tomo: But shouldn't we also be moving the instrumentation to stable?

As a middle ground why don't we remove span events from stable but keep it in the other 2 options. That way stable doesn't emit deprecated.

robbkidd: I would prioritize getting most of the attribute names aligned in this PR and address the exception implementation in a separate PR. To defend the deferral, I'll use the "database semantic conventions went stable in semconv v1.33.0 which is what this PR targets and that version of semconv still has exceptions as span_events" gambit.

helpers/mysql/lib/opentelemetry/helpers/mysql.rb
robbkidd: I think I've gotten confused on the current proposal. Does the case statement preclude fallback behavior?

instrumentation/trilogy/lib/opentelemetry/instrumentation/trilogy/patches/stable/client.rb
```diff
@@ -0,0 +1,130 @@
...
+
+              attributes
+            end
+
+            def set_error_attributes(span, error)
+              span.set_attribute('error.type', error.class.name)
+              span.set_attribute('db.response.status_code', error.error_code.to_s) if error.error_code
```
robbkidd: 🤔 Since a `rescue StandardError => e` can bring us here, I think this line needs an additional guard for when the `error` isn't a type that has an `error_code` (i.e. isn't a Trilogy exception).

```ruby
❯ irb
irb(main):001:0> e = RuntimeError.new
=> #<RuntimeError: RuntimeError>
irb(main):002:0> e.class.name
=> "RuntimeError"
irb(main):003:0> e.error_code
(irb):3:in `<main>': undefined method `error_code' for #<RuntimeError: RuntimeError> (NoMethodError)
```


```suggestion
              span.set_attribute('db.response.status_code', error.error_code.to_s) if error.respond_to?(:error_code) && error.error_code
```

instrumentation/trilogy/README.md
robbkidd: Maybe something like … 

| stable | old | Description |
| - | - | - |
| `db.namespace` | `db.name` | Database name from connection_options |
| `db.query.text` | `db.statement` | The database query being executed; set according to the `db_statement` config option |
| `db.response.status_code` | — | The Trilogy error code, if available |
| `db.system.name` | `db.system` | DBMS product identifier; always `mysql` |
| `error.type` | — | The exception class name when the operation fails |
| `server.address` | `net.peer.name` | Database host from connection_options |
| `server.port` | — | Database port from connection_options |
| — | `db.instance.id` | Connected host, e.g. result of `SELECT @@hostname` |
| — | `db.user` | Database username from connection_options |
| — | `peer.service` | Configured via the `peer_service` config option |

… (does not include `db.query.summary` because 🔜)  and move the new "Database semantic convention stability" section to here or below here.

robbkidd: Here or/and as [a doc comment like we did in the Sidekiq instrumentation](https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/beb078fb168e7ed1de91cb1a07b8c7ca4da935ab/instrumentation/resque/lib/opentelemetry/instrumentation/resque/instrumentation.rb#L10-L64) for the options to appear in [the (questionably-useful) doc site](https://www.rubydoc.info/gems/opentelemetry-instrumentation-sidekiq/OpenTelemetry/Instrumentation/Sidekiq/Instrumentation).

As a doc comment: more likely to get updated since the documentation is near to the option declarations.

As a README section: more likely to actually be found and read.

instrumentation/trilogy/lib/opentelemetry/instrumentation/trilogy/patches/stable/client.rb
hannahramadan: Ah thank you for sharing that link! I've been avoiding implementing anything that isn't marked as `stable`.

instrumentation/trilogy/lib/opentelemetry/instrumentation/trilogy/patches/stable/client.rb
hannahramadan: This is a great catch!

instrumentation/trilogy/README.md
hannahramadan: Love the table idea: https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/2095/changes/db681245275a0aab66f9850dc4b400ce07e1c25d

hannahramadan: Added both https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/2095/changes/76d80f4a34a84ff41a0a984bbcc2877802501f1e

Note: leaving this unresolved as a reminder to update `span_name` and potentially `db_statement`, since those changes are under current discussion and may change in stable

instrumentation/trilogy/lib/opentelemetry/instrumentation/trilogy/patches/stable/client.rb
```diff
@@ -0,0 +1,130 @@
...
+                attributes: client_attributes(sql).merge!(context_attributes),
+                kind: :client,
+                record_exception: config[:record_exception]
+              ) do |span, context|
+                if propagator && sql.frozen?
+                  sql = +sql
+                  propagator.inject(sql, context: context)
+                  sql.freeze
+                elsif propagator
+                  propagator.inject(sql, context: context)
```
thompson-tomo, line 68: Shouldn't this be as below so that we can eliminate the config option and just using the global propagation as it means we support any propagator.

```suggestion
                  OpenTelemetry.propagation.inject(sql, context: context)
                  sql.freeze
                elsif propagator
                  OpenTelemetry.propagation.inject(sql, context: context)
```

instrumentation/trilogy/lib/opentelemetry/instrumentation/trilogy/instrumentation.rb
thompson-tomo, line 142: This doesn't seem right

robbkidd: Can you say more about what seems off?

thompson-tomo: We have a propagator called trace context which usually implies w3c trace context (a listed propagator on https://opentelemetry.io/docs/specs/otel/context/api-propagators/#propagators-distribution) but in fact it is calling sqlcommenter.

robbkidd: It's been a minute since I was deep in propagation, but if memory serves, using sqlcommenter is the implementation for conveying tracecontext into the database service itself. Because the protocol between database client and database server is not HTTP (generally), the W3C propagator does not really apply. Sqlcommenter is how we can propagate traceparent and tracestate from the database client into the database service: via query comments.

So, I believe this case statement is correct.

thompson-tomo: Technically all trace propagators are responsible for inject & extracting the trace context when traversing process boundaries with the carrier dictating how it is shared.

Note this point would be null if we just used the global propagator and eliminated the config option. 😁

robbkidd: My preference is for this PR to remain scoped to stable semantic conventions for the attributes it writes to spans. If there are other issues, let's open and address them separately.

hannahramadan: Thanks for taking such a thorough look @thompson-tomo. This PR needs that with all the new files and duplications. Because there is so much going on, I also prefer to keep the scope of this to attributes and make tickets for other issues we note.

helpers/mysql/lib/opentelemetry/helpers/mysql.rb
hannahramadan: @robbkidd this is a bit stale 🥲 Apologies for not keeping this updated. I am hoping to chat about this more in SIG and will re-update this PR when ready!

hannahramadan: @thompson-tomo @arielvalentin @robbkidd After discussion in the spec reviewers meeting on 4/14, I'd like to leave the span name untouched for now and consider this PR focused on attributes. I've reverted changes to the span names and created a [ticket](https://github.com/open-telemetry/opentelemetry-ruby-contrib/issues/2243) to followup.

instrumentation/trilogy/lib/opentelemetry/instrumentation/trilogy/patches/stable/client.rb
hannahramadan: This is my understanding. Right now, the default is to obfuscate and users can decide if they want to send raw sql. I'd like to use this PR as an opportunity to update the attributes and introduce the stability environment variable, with updates such as this to follow. I want to lean on the "SHOULD" language here to hold on this change @thompson-tomo

instrumentation/trilogy/lib/opentelemetry/instrumentation/trilogy/patches/stable/client.rb
hannahramadan: I'd like to pause on these type of changes. Would you like me to make a ticket to revisit this @thompson-tomo

instrumentation/trilogy/lib/opentelemetry/instrumentation/trilogy/patches/dup/client.rb
```diff
@@ -0,0 +1,156 @@
...
+                  attributes['db.query.text'] = sql
+                end
+              end
+
+              attributes
+            end
+
+            def set_error_attributes(span, error)
+              span.set_attribute('error.type', error.class.name)
+              span.set_attribute('db.response.status_code', error.error_code.to_s) if error.error_code
```
robbkidd: Ope. I made [that previous suggestion](https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/2095#discussion_r3075986513) only on the stable variant and missed that it's here, too! 🤦🏻 

```suggestion
              span.set_attribute('db.response.status_code', error.error_code.to_s) if error.respond_to?(:error_code) && error.error_code
```

instrumentation/trilogy/lib/opentelemetry/instrumentation/trilogy/patches/dup/client.rb
```diff
@@ -0,0 +1,156 @@
...
+              # Database name (old: db.name, stable: db.namespace)
+              if @_otel_database_name
+                attributes[::OpenTelemetry::SemanticConventions::Trace::DB_NAME] = @_otel_database_name
+                attributes['db.namespace'] = @_otel_database_name
+              end
+
+              # db.user (old only - removed in stable)
+              attributes[::OpenTelemetry::SemanticConventions::Trace::DB_USER] = database_user if database_user
+
+              # peer.service (same in both)
```
robbkidd:
```suggestion
              # peer.service (old only - not stable and not a db attribute)
```

instrumentation/trilogy/lib/opentelemetry/instrumentation/trilogy/patches/stable/client.rb
robbkidd: Documentation for the config options have added since this initial comment. I think those docs address the question.

> … shouldn't we always be running the obfuscation and simply offer the ability to exclude the attribute? 

The default is `:obfuscate` so we're meeting the "unless there is sanitization" clause for default behavior. The docs for `:include` states that the value would then be raw query text. And then there's `:omit` for excluding the attribute. While `:omit` isn't present in the case statement, the result of using it is fall-through with no `'db.query.text'` attribute set.

instrumentation/trilogy/lib/opentelemetry/instrumentation/trilogy/instrumentation.rb
```diff
@@ -7,7 +7,70 @@
...
+      #
+      # ### `:obfuscation_limit`
+      #
+      # Maximum length of the obfuscated SQL statement. Statements exceeding this limit
+      # are truncated. Default is `2000`.
+      #
+      # ### `:peer_service`
+      #
+      # Sets the `peer.service` attribute on spans. Only applies when using old semantic
+      # conventions. Default is `nil`.
```
robbkidd:
```suggestion
      # Sets the `peer.service` attribute on spans. Default is `nil`.
      # Only applies when using old semantic conventions. Deprecated with no replacement. 
```

instrumentation/trilogy/.simplecov
```diff
@@ -0,0 +1,16 @@
+# frozen_string_literal: true
+
+require 'digest'
+
+digest = Digest::MD5.new
+digest.update('test')
+digest.update(ENV.fetch('BUNDLE_GEMFILE', 'gemfile'))
```
hannahramadan: This is a copy of the repo-level .simplecov, minus a guard:

`digest.update(ENV.fetch('BUNDLE_GEMFILE', 'gemfile')) if ENV['APPRAISAL_INITIALIZED']`

Usually appraisals are testing the same test file with different versions. They are overwriting each other's test coverage but it doesn't matter because it still looks like one file is well tested. 

But now different appraisals are testing different source files and still overwriting each other's name, so it looks like code is missing tests.

The fix gives each run a unique name so all 6 results are saved and combined. When we are done with all these duplicate files, we can simply delete this file

hannahramadan: I'm just seeing https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/2272. If that is merged, i'll revert this change, merge that change in, and see if that fixes things

hannahramadan: That fix worked and this new file was removed.

robbkidd (approved):
Is it time?

instrumentation/trilogy/lib/opentelemetry/instrumentation/trilogy/patches/dup/client.rb
```diff
@@ -0,0 +1,156 @@
...
+
+              # peer.service (old only - not stable and not a db attribute)
+              attributes[::OpenTelemetry::SemanticConventions::Trace::PEER_SERVICE] = config[:peer_service] unless config[:peer_service].nil?
+              attributes
+            end
+
+            def client_attributes(sql = nil)
+              attributes = @_otel_base_attributes.dup
+
+              attributes['db.instance.id'] = @connected_host unless @connected_host.nil?
```
kaylareopelle, line 115: I love the comments throughout dup to differentiate the stable from old conventions. I noticed this one is missing a comment. Here's a suggestion, but write whatever feels right to you

```suggestion
              # Old convention
              attributes['db.instance.id'] = @connected_host unless @connected_host.nil?
```

instrumentation/trilogy/test/opentelemetry/instrumentation/trilogy/patches/dup/instrumentation_test.rb
```diff
@@ -0,0 +1,660 @@
...
+        expect do
+          client.query('SELECT INVALID')
+        end.must_raise Trilogy::Error
+
+        _(span.attributes['error.type']).must_equal 'Trilogy::ProtocolError'
+      end
+
+      it 'sets db.response.status_code when error has error_code' do
+        expect do
+          client.query('SELECT INVALID')
```
kaylareopelle, line 373: It's great to see new tests for the error span attributes. What do you think about adding error tests for the `ping` and `intitialize` methods since they can record errors now too?

helpers/mysql/lib/opentelemetry/helpers/mysql.rb
```diff
@@ -2,7 +2,8 @@
 
 # Copyright The OpenTelemetry Authors
 #
-# SPDX-License-Identifier: Apache-2.0module OpenTelemetry
+# SPDX-License-Identifier: Apache-2.0
```
kaylareopelle: What do you think about separating this into another PR so it doesn't trigger a release for the MySQL Helpers gem? That PR could have a `chore` conventional commit to avoid the release.

kaylareopelle (changes requested):

helpers/mysql/lib/opentelemetry/helpers/mysql.rb
hannahramadan: Great call! I've removed this change and will open a new PR for this clean up

instrumentation/trilogy/test/opentelemetry/instrumentation/trilogy/patches/dup/instrumentation_test.rb
hannahramadan: Added new tests for these ◡̈

kaylareopelle (approved):

PR #2116, opened 2026-03-23
ci: Block docker updates based on age

thompson-tomo:
This blocks major updates for docker services until the previous major is EoL. Dates are based on https://endoflife.date/

Reason for this change is that the current solution would still raise them just to have them closed manually as we want to remain on the LTS version. The Dashboard gaind the ability to trigger the updates.

Official sources:
- https://www.postgresql.org/support/versioning/
- https://blogs.oracle.com/mysql/introducing-mysql-innovation-and-longterm-support-lts-versions follow 5 years of premier support minus 3 months to bump to the major with it bumped to latest minor


arielvalentin (commented):
Did you pick the ages based on a official maintenance policy or based on EoL history?

arielvalentin (approved):

PR #2122, opened 2026-03-25
chore: [PG] Update README with obfuscation_limit details

toncid:
Document the `obfuscation_limit` config in the README.

The config was released in [v0.25.0](https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/main/instrumentation/pg/CHANGELOG.md#v0250--2023-05-25) (on 2023-05-25), but it was never documented, so this PR fixes it.


xuan-cao-swi (approved):
Would be nice to also document `:propagator`; and also to have this doc for mysql2 and trilogy too.

PR #2127, opened 2026-03-25
feat: draft Mongo semantic stability

hannahramadan:
Span name
- {collection}.{operation} -> {operation} {collection}

Changes:
- db.system → db.system.name
- db.name → db.namespace
- db.operation → db.operation.name
- net.peer.port → server.port
- db.mongodb.collection → db.collection.name 

Removed:
- peer.service

Added:
- error.type
- db.response.status_code


hannahramadan:
Closing because Mongo support is deprecated, so the instrumentation will not need updating.

PR #2130, opened 2026-03-26
fix(rack): fiber-safe context detachment in `EventHandler`

rsamoilov:
## Summary

Fixes context detachment errors in fiber-based environments.

**Problem:** When using `Rack::Events`, the `on_finish` callback can be invoked from a different fiber than `on_start` (e.g., when streaming response bodies). This causes `OpenTelemetry::Context.detach(token)` to fail with:

```
ERROR -- : OpenTelemetry error: calls to detach should match corresponding calls to attach
```

**Solution:** Store the fiber reference alongside the token and span. On `detach_context`:

- Always finish the span (trace data is never lost)
- Only detach context if running in the same fiber
- Log a debug message when detachment is skipped

**Note on fiber pooling:** Skipping `detach` could leave stale context in the original fiber. However, servers that reuse fibers across requests would be fundamentally incompatible  with any Ruby feature relying on fiber-locals (e.g., `ActiveSupport::CurrentAttributes`), so this isn't something we should solve at the instrumentation level.

## Changes

- Changed env key from `otel.rack.token_and_span` to `otel.context_info`
- Changed storage format from `[token, span]` to `[fiber, token, span]`
- Updated `detach_context` to check fiber identity before detaching
- Applied changes to all three convention variants (stable, old, dup)

## Test plan

- [x] Added unit tests for same-fiber detachment
- [x] Added unit tests for cross-fiber detachment (verifies span finishes without error)
- [x] All existing tests pass across all 12 configurations (4 Rack versions × 3 conventions)


instrumentation/rack/lib/opentelemetry/instrumentation/rack/middlewares/dup/event_handler.rb
```diff
@@ -209,11 +210,17 @@ def request_span_attributes(env)
...
-              return nil unless request.env[OTEL_TOKEN_AND_SPAN]
+              otel_context_info = request.env[OTEL_CONTEXT_INFO]
+              return unless otel_context_info
 
-              token, span = request.env[OTEL_TOKEN_AND_SPAN]
+              original_fiber, token, span = otel_context_info
               span.finish
-              OpenTelemetry::Context.detach(token)
+
+              if Fiber.current.equal?(original_fiber)
```
arielvalentin, line 219: Could you share more details about how this could happen? 

What application server are you using that invokes the rack events in different fibers?

arielvalentin (commented):
I am concerned about passing around references to fibers and want to better understand if we are leaking context objects or detaching them out of order. 

Could you share more information or a small script that demonstrates the problem?

rsamoilov:
I'm using [Rage](https://github.com/rage-rb/rage), which runs each request in a separate fiber. But the issue isn't specific to Rage - it affects any setup where the response body is closed in a different fiber than the one that handled the request.  

Here's a minimal reproduction that demonstrates the problem using just Puma + a simple middleware: https://gist.github.com/rsamoilov/c01b3923af0a35dd75b34056062abf0a

`WrapInFiber` simulates the fiber boundary. `Rack::Events` calls `on_start` inside the fiber (where context is attached), but calls `on_finish` when the body is closed - back in the original fiber. Since OTel context is fiber-local, the detach fails.

The span always finishes regardless of which fiber we're in. The only thing skipped is the `Context.detach` call when it would error anyway.

rsamoilov:
Commenting to remove the stale status, but let me know if you're not comfortable with this change and the PR should be closed.

kaylareopelle:
Hi @rsamoilov, I'll add this to our next SIG agenda. I hope to have more for you after that meeting.

kaylareopelle:
Hi @rsamoilov, thanks for your patience while we reviewed your proposal. The PR raises a real concern with fiber-based frameworks that I wasn't aware of. Thank you for the reproduction script. With the growth of Fibers in Ruby, I'd like to see our Rack instrumentation better support them.

After you opened this, we received a bug report from a Sinatra user (https://github.com/open-telemetry/opentelemetry-ruby-contrib/issues/2425) about correctly passing context in a scenario where multiple apps are mounted. I opened https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/2476 to fix this using a last-in, first-out context stack. 

That solution falls short if we're running the code in a fiber context, since fibers can attach and detach out of order. 

To resolve both of our issues, I suggest we consider using the Rack request itself as a reference and switching context management to a hash so we can always detach the context specifically associated with the request. This does still require instrumentation for libraries to correctly pass context for this to work.

I made a branch off your work with the proposal. I wanted to draft a PR to merge into this branch, but I can't seem to find your fork. Here's the comparison in the meantime: https://github.com/open-telemetry/opentelemetry-ruby-contrib/compare/main...kaylareopelle:opentelemetry-ruby-contrib:fiber-safe-context-with-multi-app-fix?expand=1

Could you take a look and let me know what you think? I'm also happy to open this as a PR in the main repo if that's easier for you to review.

rsamoilov:
Hi @kaylareopelle,

One fix to tackle both problems at once - love it 😄 

Your approach looks great, and I can confirm the error is fixed with these changes.

I've granted you access to [my fork](https://github.com/rage-rb/opentelemetry-ruby-contrib/tree/fix/rack-fiber-safe-context-detachment) just in case. Will be happy to review your changes either here or in another PR.

kaylareopelle:
Hi @rsamoilov, thank you for the access! I've opened the PR to merge my branch into this one: https://github.com/rage-rb/opentelemetry-ruby-contrib/pull/1

rsamoilov:
Hey @kaylareopelle ,

While looking at your changes, I realised there was another problem my initial fix missed. 

I opened https://github.com/rage-rb/opentelemetry-ruby-contrib/pull/2 - could you please give it a look?

kaylareopelle:
> Hey @kaylareopelle ,
> 
> While looking at your changes, I realised there was another problem my initial fix missed.
> 
> I opened [rage-rb#2](https://github.com/rage-rb/opentelemetry-ruby-contrib/pull/2) - could you please give it a look?

Thank you, @rsamoilov! I've been pulled away by other things this week. I will review this tomorrow!

instrumentation/rack/test/opentelemetry/instrumentation/rack/middlewares/old/event_handler_test.rb
```diff
@@ -483,4 +483,119 @@ class EventHandlerError < StandardError
...
+
+      _(finished_spans.size).must_equal 1
+      _(rack_span.name).must_equal 'HTTP GET'
+    end
+  end
+
+  # This might happen if one Sinatra app is called by another
+  # See: https://github.com/open-telemetry/opentelemetry-ruby-contrib/issues/2425
+  describe 'when nested inside another instance of this EventHandler on the same env' do
+    let(:inner_handler) { OpenTelemetry::Instrumentation::Rack::Middlewares::Stable::EventHandler.new }
```
mwear: This is a `Stable::EventHandler` in the `old` test. It should nest an `Old::Handler` instead.
```suggestion
    let(:inner_handler) { OpenTelemetry::Instrumentation::Rack::Middlewares::Old::EventHandler.new }
```

instrumentation/rack/test/opentelemetry/instrumentation/rack/middlewares/old/event_handler_test.rb
```diff
@@ -483,4 +483,119 @@ class EventHandlerError < StandardError
...
+
+      _(outer_span).wont_be_nil
+      _(inner_span).wont_be_nil
+      _(inner_span.parent_span_id).must_equal outer_span.span_id
+    end
+  end
+
+  # Combines the fiber bug reported in #2130 with the multi-app Sinatra fix in #2476
+  describe 'when nested inside another instance of this EventHandler on the same env' do
+    let(:inner_handler) { OpenTelemetry::Instrumentation::Rack::Middlewares::Stable::EventHandler.new }
```
mwear:
```suggestion
    let(:inner_handler) { OpenTelemetry::Instrumentation::Rack::Middlewares::Old::EventHandler.new }
```

instrumentation/rack/test/opentelemetry/instrumentation/rack/middlewares/dup/event_handler_test.rb
```diff
@@ -508,4 +508,120 @@ class EventHandlerError < StandardError
...
+
+      _(finished_spans.size).must_equal 1
+      _(rack_span.name).must_equal 'GET'
+    end
+  end
+
+  # This might happen if one Sinatra app is called by another
+  # See: https://github.com/open-telemetry/opentelemetry-ruby-contrib/issues/2425
+  describe 'when nested inside another instance of this EventHandler on the same env' do
+    let(:inner_handler) { OpenTelemetry::Instrumentation::Rack::Middlewares::Stable::EventHandler.new }
```
mwear: This is a `Stable::EventHandler` in the dup test. It should nest a `Dup::Handler` instead.
```suggestion
    let(:inner_handler) { OpenTelemetry::Instrumentation::Rack::Middlewares::Dup::EventHandler.new }
```

instrumentation/rack/test/opentelemetry/instrumentation/rack/middlewares/dup/event_handler_test.rb
```diff
@@ -508,4 +508,120 @@ class EventHandlerError < StandardError
...
+
+      _(outer_span).wont_be_nil
+      _(inner_span).wont_be_nil
+      _(inner_span.parent_span_id).must_equal outer_span.span_id
+    end
+  end
+
+  # Combines the fiber bug reported in #2130 with the multi-app Sinatra fix in #2476
+  describe 'when nested inside another instance of this EventHandler on the same env' do
+    let(:inner_handler) { OpenTelemetry::Instrumentation::Rack::Middlewares::Stable::EventHandler.new }
```
mwear:
```suggestion
    let(:inner_handler) { OpenTelemetry::Instrumentation::Rack::Middlewares::Dup::EventHandler.new }
```

instrumentation/rack/test/opentelemetry/instrumentation/rack/middlewares/stable/event_handler_test.rb
```diff
@@ -479,4 +479,119 @@ class EventHandlerError < StandardError
...
+      handler.on_finish(rack_request, response)
+
+      _(finished_spans.size).must_equal 1
+      _(rack_span.name).must_equal 'GET'
+    end
+  end
+
+  # This might happen if one Sinatra app is called by another
+  # See: https://github.com/open-telemetry/opentelemetry-ruby-contrib/issues/2425
+  describe 'when nested inside another instance of this EventHandler on the same env' do
```
mwear: This test and the one on line 560 have the same name. Can we differentiate them?
```suggestion
  describe 'when nested inside another EventHandler in the same fiber' do
```

instrumentation/rack/test/opentelemetry/instrumentation/rack/middlewares/stable/event_handler_test.rb
```diff
@@ -479,4 +479,119 @@ class EventHandlerError < StandardError
...
+      inner_span = finished_spans.find { |s| s.parent_span_id != OpenTelemetry::Trace::INVALID_SPAN_ID }
+
+      _(outer_span).wont_be_nil
+      _(inner_span).wont_be_nil
+      _(inner_span.parent_span_id).must_equal outer_span.span_id
+    end
+  end
+
+  # Combines the fiber bug reported in #2130 with the multi-app Sinatra fix in #2476
+  describe 'when nested inside another instance of this EventHandler on the same env' do
```
mwear:
```suggestion
  describe 'when nested inside another EventHandler across a fiber boundary' do
```

instrumentation/rack/test/opentelemetry/instrumentation/rack/middlewares/dup/event_handler_test.rb
```diff
@@ -508,4 +508,120 @@ class EventHandlerError < StandardError
...
+      handler.on_finish(rack_request, response)
+
+      _(finished_spans.size).must_equal 1
+      _(rack_span.name).must_equal 'GET'
+    end
+  end
+
+  # This might happen if one Sinatra app is called by another
+  # See: https://github.com/open-telemetry/opentelemetry-ruby-contrib/issues/2425
+  describe 'when nested inside another instance of this EventHandler on the same env' do
```
mwear: This test and the one on line 590 have the same name. Can we differentiate them?
```suggestion
  describe 'when nested inside another EventHandler in the same fiber' do
```

instrumentation/rack/test/opentelemetry/instrumentation/rack/middlewares/dup/event_handler_test.rb
```diff
@@ -508,4 +508,120 @@ class EventHandlerError < StandardError
...
+      inner_span = finished_spans.find { |s| s.parent_span_id != OpenTelemetry::Trace::INVALID_SPAN_ID }
+
+      _(outer_span).wont_be_nil
+      _(inner_span).wont_be_nil
+      _(inner_span.parent_span_id).must_equal outer_span.span_id
+    end
+  end
+
+  # Combines the fiber bug reported in #2130 with the multi-app Sinatra fix in #2476
+  describe 'when nested inside another instance of this EventHandler on the same env' do
```
mwear:
```suggestion
  describe 'when nested inside another EventHandler across a fiber boundary' do
```

instrumentation/rack/test/opentelemetry/instrumentation/rack/middlewares/old/event_handler_test.rb
```diff
@@ -483,4 +483,119 @@ class EventHandlerError < StandardError
...
+      handler.on_finish(rack_request, response)
+
+      _(finished_spans.size).must_equal 1
+      _(rack_span.name).must_equal 'HTTP GET'
+    end
+  end
+
+  # This might happen if one Sinatra app is called by another
+  # See: https://github.com/open-telemetry/opentelemetry-ruby-contrib/issues/2425
+  describe 'when nested inside another instance of this EventHandler on the same env' do
```
mwear: This test and the one on line 564 have the same name. Can we differentiate them?
```suggestion
  describe 'when nested inside another EventHandler in the same fiber' do
```

instrumentation/rack/test/opentelemetry/instrumentation/rack/middlewares/old/event_handler_test.rb
```diff
@@ -483,4 +483,119 @@ class EventHandlerError < StandardError
...
+      inner_span = finished_spans.find { |s| s.parent_span_id != OpenTelemetry::Trace::INVALID_SPAN_ID }
+
+      _(outer_span).wont_be_nil
+      _(inner_span).wont_be_nil
+      _(inner_span.parent_span_id).must_equal outer_span.span_id
+    end
+  end
+
+  # Combines the fiber bug reported in #2130 with the multi-app Sinatra fix in #2476
+  describe 'when nested inside another instance of this EventHandler on the same env' do
```
mwear:
```suggestion
  describe 'when nested inside another EventHandler across a fiber boundary' do
```

instrumentation/rack/lib/opentelemetry/instrumentation/rack/middlewares/dup/event_handler.rb
```diff
@@ -43,7 +43,7 @@ module Dup
           class EventHandler
             include ::Rack::Events::Abstract
 
-            OTEL_TOKEN_AND_SPAN = 'otel.rack.token_and_span'
+            OTEL_CONTEXT_INFO = 'otel.context_info'
```
mwear: Should we keep the current convention and keep rack in the key name:


```suggestion
            OTEL_CONTEXT_INFO = 'otel.rack.context_info'
```

mwear (commented):
The approach looks good to me. I found some issues in the tests that I left line comments for. It's really 2 issues that are repeated in the 3 test variants.

instrumentation/rack/lib/opentelemetry/instrumentation/rack/middlewares/dup/event_handler.rb
rsamoilov: Thanks for the review! Updated the PR.

instrumentation/rack/lib/opentelemetry/instrumentation/rack/middlewares/dup/event_handler.rb
robbkidd: Rage and a reproduction with puma have been mentioned in comments below.

robbkidd:
Should the Rack instrumentation's `on_commit` and `on_error` handlers use the same context lookup mechanism? They both still rely on `Rack.current_span` unlike the new lookup added to `on_finish`.

rsamoilov:
> Should the Rack instrumentation's `on_commit` and `on_error` handlers use the same context lookup mechanism? They both still rely on `Rack.current_span` unlike the new lookup added to `on_finish`.

Only if for consistency, but normally I wouldn't - `on_start`, `@app.call`, `on_commit`, and `on_error` all run synchronously inside `Rack::Events#call`. It's only `on_finish` that's deferred through `Rack::BodyProxy#close`.

robbkidd:
Fair 'nuff!

robbkidd (approved):

kaylareopelle (approved):

PR #2131, opened 2026-03-26
feat: LMDB semantic stability

hannahramadan:
Changed:
  - db.system → db.system.name
  - Span name drops key from name
  - Query obfuscated by default

Removed in stable:
  - peer.service
                                                     
Added in stable:
  - db.operation.name
  - error.type

This PR is mostly new files (the old/stable/dup patch split), so the standard file diff shows everything as added and makes it hard to see what actually differs between the three variants. 

To make it easier to review, you could consider copying and pasting the changed files (stable/dup) over the current file into your local IDE. Then that diff will show what changes were made.

----------


instrumentation/lmdb/lib/opentelemetry/instrumentation/lmdb/patches/stable/database.rb
```diff
@@ -0,0 +1,107 @@
...
+
+            def get(key)
+              statement = formatted_statement('GET', "GET #{key}")
+              attributes = {
+                'db.system.name' => 'lmdb',
+                'db.operation.name' => 'GET'
+              }
+              attributes['db.query.text'] = statement if config[:db_statement] == :include
+
+              tracer.in_span("GET #{key}", attributes: attributes, kind: :client) do |span|
```
thompson-tomo, line 27: Shouldn't these be internal as there is no server but rather just a local file?

instrumentation/lmdb/lib/opentelemetry/instrumentation/lmdb/patches/stable/database.rb
```diff
@@ -0,0 +1,107 @@
...
+          # Module to prepend to LMDB::Database for instrumentation
+          module Database
+            STATEMENT_MAX_LENGTH = 500
+
+            def get(key)
+              statement = formatted_statement('GET', "GET #{key}")
+              attributes = {
+                'db.system.name' => 'lmdb',
+                'db.operation.name' => 'GET'
+              }
```
thompson-tomo, line 24: Can we add the `db.namespace` attribute and set it to the `env.path`?

thompson-tomo, line 24: Can we add the `db.collection.name` attribute and set it to the `env.database`?

instrumentation/lmdb/lib/opentelemetry/instrumentation/lmdb/patches/stable/database.rb
```diff
@@ -0,0 +1,107 @@
...
+              attributes['db.query.text'] = statement if config[:db_statement] == :include
```
thompson-tomo, line 25: Can we not always include this attribute given it is parameterised?

instrumentation/lmdb/lib/opentelemetry/instrumentation/lmdb/patches/stable/database.rb
thompson-tomo, line 24: Can we add `db.query.summary`?

instrumentation/lmdb/lib/opentelemetry/instrumentation/lmdb/patches/stable/database.rb
```diff
@@ -0,0 +1,107 @@
...
+              tracer.in_span("GET #{key}", attributes: attributes, kind: :client) do |span|
+                super
+              rescue StandardError => e
+                set_error_attributes(span, e)
+                raise
+              end
+            end
+
+            def delete(key, value = nil)
+              statement = formatted_statement('DELETE', "DELETE #{key} #{value}".strip)
```
thompson-tomo, line 36: Values should be left out

instrumentation/lmdb/lib/opentelemetry/instrumentation/lmdb/patches/stable/database.rb
```diff
@@ -0,0 +1,107 @@
...
+              }
+              attributes['db.query.text'] = statement if config[:db_statement] == :include
+
+              tracer.in_span("DELETE #{key}", attributes: attributes, kind: :client) do |span|
+                super
+              rescue StandardError => e
+                set_error_attributes(span, e)
+                raise
+              end
+            end
```
thompson-tomo, line 50: Same changes as get

instrumentation/lmdb/lib/opentelemetry/instrumentation/lmdb/patches/stable/database.rb
```diff
@@ -0,0 +1,107 @@
...
+              }
+              attributes['db.query.text'] = statement if config[:db_statement] == :include
+
+              tracer.in_span("PUT #{key}", attributes: attributes, kind: :client) do |span|
+                super
+              rescue StandardError => e
+                set_error_attributes(span, e)
+                raise
+              end
+            end
```
thompson-tomo, line 67: Same changes as delete

instrumentation/lmdb/lib/opentelemetry/instrumentation/lmdb/patches/stable/database.rb
```diff
@@ -0,0 +1,107 @@
...
+              }
+              attributes['db.query.text'] = 'CLEAR' if config[:db_statement] == :include
+
+              tracer.in_span('CLEAR', attributes: attributes, kind: :client) do |span|
+                super
+              rescue StandardError => e
+                set_error_attributes(span, e)
+                raise
+              end
+            end
```
thompson-tomo, line 83: Same changes as get

instrumentation/lmdb/lib/opentelemetry/instrumentation/lmdb/patches/stable/database.rb
```diff
@@ -0,0 +1,107 @@
...
+
+require 'opentelemetry'
+
+module OpenTelemetry
+  module Instrumentation
+    module LMDB
+      module Patches
+        module Stable
+          # Module to prepend to LMDB::Database for instrumentation
+          module Database
```
thompson-tomo, line 15: What about the drop & has operations?

instrumentation/lmdb/lib/opentelemetry/instrumentation/lmdb/patches/stable/environment.rb
```diff
@@ -0,0 +1,48 @@
...
+    module LMDB
+      module Patches
+        module Stable
+          # Module to prepend to LMDB::Environment for instrumentation
+          module Environment
+            def transaction(*args)
+              attributes = {
+                'db.system.name' => 'lmdb',
+                'db.operation.name' => 'TRANSACTION'
+              }
```
thompson-tomo, line 21: Can we add the `db.namespace` attribute and set it to the `env.path`?

thompson-tomo, line 21: Can we add the `db.collection.name` attribute and set it to the `env.database`?

instrumentation/lmdb/lib/opentelemetry/instrumentation/lmdb/patches/stable/database.rb
hannahramadan: I'd to keep the scope to updating the current instrumentation to stable attributes. These PRs are so big as is 🥲 I consider these updates a feature improvement, not a semconv migration.

instrumentation/lmdb/lib/opentelemetry/instrumentation/lmdb/patches/stable/database.rb
hannahramadan: Added! https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/2131/changes/0035308e8c1f1207b7d816523cbcd3ecf40f2253

hannahramadan: I looked into this and I don't think `env.database` will give us what we want. Thats a method that takes an optional name, opens that table and returns a new Database object. So calling it doesnt tell us the name of the database, it just opens a new one.

hannahramadan: Also leaving this off since it is build off the collection_name.

instrumentation/lmdb/lib/opentelemetry/instrumentation/lmdb/patches/stable/database.rb
hannahramadan: Are you thinking that `db.query.text` should be obfuscated by default?

instrumentation/lmdb/lib/opentelemetry/instrumentation/lmdb/patches/stable/database.rb
hannahramadan: I think I'd like to keep client for now. I'm not seeing other related instrumentations by us or other languages that use internal.

I see that "MAY" language was [added](https://github.com/open-telemetry/semantic-conventions/pull/1506) to the spec to allow for flexibility, but it doesn't appear to be a necessity.

thompson-tomo: I think the key part is 

> It MAY be set to INTERNAL on spans representing in-memory database calls. It’s RECOMMENDED to use CLIENT kind when database system being instrumented usually runs in a different process than its client or when database calls happen over instrumented protocol such as HTTP.

If you look at the list of db systems there is maybe 1 inmemory provider with no dedicated refinements.

For me it would be better to have it internal as that way we can explain the absence of multiple recommended & conditionally required attributes and obviously it would be useful to have the definition in semconv.

instrumentation/lmdb/lib/opentelemetry/instrumentation/lmdb/patches/stable/database.rb
thompson-tomo: Yes as we are constructing a query like string

hannahramadan: I can flip the default to obfuscate. You mentioned deleting the `values` from the recorded queries. Wouldn't we still want those in the :include option?

instrumentation/lmdb/lib/opentelemetry/instrumentation/lmdb/patches/stable/database.rb
hannahramadan: I don't know that internal is needed to explain absent of certain attributes. `server.address`, `network.peer.address` and `network.peer.port` are all recommended to include when available, and they aren't here. `server.port` is Conditionally Required "if using a port other than the default port for this DBMS and if server.address is set", which is false. So a client LMDB span isn't missing anything against the span definition.

I'm not convinced based on the spec that a change to an internal span is necessary. But also happy to change it if others feel it is important. 

Perhaps we can get some others to weigh in?

instrumentation/lmdb/lib/opentelemetry/instrumentation/lmdb/patches/stable/database.rb
thompson-tomo: I don't see the need to have the values included in the query text, if you want the values that is where https://opentelemetry.io/docs/specs/semconv/registry/attributes/db/ comes in.

instrumentation/lmdb/lib/opentelemetry/instrumentation/lmdb/patches/stable/database.rb
hannahramadan: @thompson-tomo We chatted about this in SIG and came to the decision to change this to internal! Thank you for bringing this up and I'll make this update.

instrumentation/lmdb/lib/opentelemetry/instrumentation/lmdb/patches/stable/database.rb
hannahramadan: Since this is a key-value store and the value is part of the command, I'm inclined to keep it in when the user wants it included. redis and dalli both include it under `:include` too. I agree `db.query.parameter.<key>` is the good place for values, but it's still marked Development and this PR is the stable migration, so I'd rather add it as a follow-up. Flipped the default to `:obfuscate` in the meantime.

kaylareopelle (dismissed):

instrumentation/lmdb/lib/opentelemetry/instrumentation/lmdb/patches/old/database.rb
```diff
@@ -0,0 +1,91 @@
...
+          # Module to prepend to LMDB::Database for instrumentation
+          module Database
+            STATEMENT_MAX_LENGTH = 500
+
+            def get(key)
+              statement = formatted_statement('GET', "GET #{key}")
+              attributes = {
+                'db.system' => 'lmdb'
+              }
+              attributes['db.statement'] = statement unless config[:db_statement] == :omit
```
kaylareopelle: I don't think we should change the include/omit behavior in `old`, only in `dup` and `stable`

instrumentation/lmdb/lib/opentelemetry/instrumentation/lmdb/patches/stable/database.rb
```diff
@@ -0,0 +1,116 @@
...
+              statement = case config[:db_statement]
+                          when :obfuscate
+                            operation + (' ?' * args.compact.length)
+                          when :include
+                            [operation, *args.compact].join(' ')
+                          end
+              return unless statement
+
+              statement = OpenTelemetry::Common::Utilities.truncate(statement, STATEMENT_MAX_LENGTH)
+              OpenTelemetry::Common::Utilities.utf8_encode(statement, binary: true)
```
kaylareopelle: Why are we adding `binary: true` to these calls?

instrumentation/lmdb/lib/opentelemetry/instrumentation/lmdb/patches/stable/database.rb
```diff
@@ -0,0 +1,116 @@
...
+            rescue StandardError => e
+              OpenTelemetry.logger.debug("non formattable LMDB statement for #{operation}: #{e}")
```
kaylareopelle: nit
```suggestion
              OpenTelemetry.logger.debug("unformattable LMDB statement for #{operation}: #{e}")
```

instrumentation/lmdb/lib/opentelemetry/instrumentation/lmdb/patches/dup/database.rb
```diff
@@ -0,0 +1,144 @@
...
+              when :include
+                raw_statement(operation, *args)
+              end
+            end
+
+            def truncate_and_encode(statement, operation)
+              statement = OpenTelemetry::Common::Utilities.truncate(statement, STATEMENT_MAX_LENGTH)
+              OpenTelemetry::Common::Utilities.utf8_encode(statement, binary: true)
+            rescue StandardError => e
+              OpenTelemetry.logger.debug("non formattable LMDB statement for #{operation}: #{e}")
```
kaylareopelle: nit:
```suggestion
              OpenTelemetry.logger.debug("unformattable LMDB statement for #{operation}: #{e}")
```

instrumentation/lmdb/lib/opentelemetry/instrumentation/lmdb/patches/stable/database.rb
thompson-tomo: Created https://github.com/open-telemetry/opentelemetry-ruby-contrib/issues/2563

instrumentation/lmdb/lib/opentelemetry/instrumentation/lmdb/patches/stable/database.rb
thompson-tomo: Have created https://github.com/open-telemetry/opentelemetry-ruby-contrib/issues/2564 

could we for stable always obfuscate and soon add the opt-in setting to capture the value. This way we avoid having settings in stable mode that will be replaced.

instrumentation/lmdb/lib/opentelemetry/instrumentation/lmdb/patches/stable/database.rb
```diff
@@ -0,0 +1,116 @@
...
+        module Stable
+          # Module to prepend to LMDB::Database for instrumentation
+          module Database
+            STATEMENT_MAX_LENGTH = 500
+
+            def get(key)
+              attributes = {
+                'db.system.name' => 'lmdb',
+                'db.operation.name' => 'GET',
+                'db.namespace' => env.path
```
thompson-tomo, line 22: I have taken a fresh look at the lmdb docs and while this is useful perhaps we should change this to
```suggestion
                'db.namespace' => "#{env.path}/unknown_db"
```
As env.path is only giving us the path to the directory but there can be multiple db's in that directory. We could even potentially use the value of the env.databases if only 1 is returned.

thompson-tomo (commented):
Looks better but after reviewing the lmdb docs, have 1 clarification of db.namespace.

instrumentation/lmdb/lib/opentelemetry/instrumentation/lmdb/patches/stable/database.rb
hannahramadan: For including `unknown_db`, the `db.namespace` spec says missing components "and their associated separators" SHOULD be omitted, so I think `env.path` on its own is already okay.

[env.databases](https://rubydoc.info/gems/lmdb/LMDB/Environment#databases-instance_method) isn't reliable because it only lists named databases so the default unnamed database never appears in it. So an app using the default database + one named database gets (ie 'users') will have every default-database operation tagged with 'users'.

thompson-tomo: In that case let's document the definition of db.namespace ie it only includes the environment as recommended in semconv.

PR #2136, opened 2026-03-28
test: run simplecov on all frameworks using collated reports

thompson-tomo:
This ensures test coverage is meet on all frameworks to avoid potential issues when bumping min frameworks. This also addresses that the min coverage is not actually currently being enforced due to coverage only being set after the tests have run.

To Address failures due to insufficient coverage the min coverage is being overwritten using a yml file in the working directory. This ensures the definition of that gem and all required inputs resides within the gem folder and pr's can be appropriately labelled.

Here is an example of where the min coverage was not meet (increased for testing to 99) but check still passed: https://github.com/open-telemetry/opentelemetry-ruby-contrib/actions/runs/23683365227/job/68998870234?pr=2136.
image

See after changes, with the key being the % is now correct
image


kaylareopelle (approved):
Thanks for noticing this and fixing it!

hannahramadan (approved):

thompson-tomo:
@arielvalentin could you take a look at this. Thanks.

.github/workflows/ci-contrib.yml
arielvalentin, line 1: Does this not need a "Set Environment Variables" step as well?

.github/actions/test_gem/action.yml
```diff
@@ -155,10 +150,9 @@ runs:
 
     - name: Coverage
       shell: bash
-      if: "${{ inputs.coverage == 'true' }}"
       # This starts a new simplecov run which tracks nothing of its own,
       # but merges with the existing coverage reports generated during testing.
-      run: 'bundle exec ruby -e ''require "simplecov"; SimpleCov.minimum_coverage(${{ inputs.minimum_coverage }})'''
+      run: 'bundle exec ruby -e ''require "simplecov"; SimpleCov.minimum_coverage(${{ inputs.minimum_coverage || 85 }}); SimpleCov.collate Dir["coverage/**/.resultset.json"];'''
```
arielvalentin, line 155: as an enhancement, can you add the final coverage report as a summary or artifact?

thompson-tomo: Tbh, I would rather do that via a seperate PR so we can discuss if we should  instead switch to Codecov like quite a bit of otel has done. that way we can get this fix in sooner and we can ensure that the coverage fixes are in fact meeting our requirements.

arielvalentin: Yes, a separate PR was my intention by referring to it "as an enhancement".  
Also TIL we Codecov was the "standard" for the community.

thompson-tomo: Captured as https://github.com/open-telemetry/opentelemetry-ruby-contrib/issues/2250

.github/workflows/ci-contrib.yml
thompson-tomo: It doesn't need it as no gems tested as part of the contrib Workflow needs it.

PR #2145, opened 2026-03-30
chore: Improve developer experience

thompson-tomo:
This extends the contributing guide to talk about adding a package rule for external services when adding one.

It also implements that the local checks are done in ci when one of the tools changes, this way we can ensure that when the local tool is updated that it will still pass when run locally.


instrumentation/net_ldap/.rubocop.yml
```diff
@@ -1 +1,6 @@
 inherit_from: ../../.rubocop.yml
+
+Style/RedundantLineContinuation:
+  Enabled: true
+  Exclude:
+    - "test/opentelemetry/instrumentation/net/ldap/instrumentation_test.rb"
```
thompson-tomo: @scbjans this is needed as on windows https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/main/instrumentation/net_ldap/test/opentelemetry/instrumentation/net/ldap/instrumentation_test.rb#L17 reports a violation

scbjans: Instead of adding this as an exception, can't we rewrite that line into the following?
```
let(:ldap) do
    Net::LDAP.new host: 'test.mocked.com', port: 636,
                  auth: {
                    method: :simple,
                    username: 'test_user',
                    password: 'test_password'
                  },
                  force_no_page: true
  end
```

instrumentation/CONTRIBUTING.md
```diff
@@ -380,6 +380,22 @@ instrumentation_silver:
...
+      description: "Wait until current major postgres is EoL before updating",
+      dependencyDashboardCategory: "Min Docker service",
+      matchUpdateTypes: ["major"],
+      matchDepNames: ["postgres"],
+      minimumReleaseAge: "1460 days",
+    },
+```
+
+The days is to be calculated based on the expected age of the major version when it becomes the lowest major version which is not end of life.
+In the above example, we want until a major version has been available for 1460 days (4 years) which is calculated based on each major version of Postgres being supported for 5 years with a new major each year.
```
scbjans:
```suggestion
In the above example, we wait until a major version has been available for 1460 days (4 years) which is calculated based on each major version of Postgres being supported for 5 years with a new major each year.
```

instrumentation/net_ldap/.rubocop.yml
thompson-tomo: Good suggestion @scbjans I didn't think of that. Change has been made.

instrumentation/CONTRIBUTING.md
```diff
@@ -380,6 +380,22 @@ instrumentation_silver:
       # ...
 ```
 
+The final step in adding a new service is to define a renovate package rule which manages the min major version of the service tested against.
```
scbjans:
```suggestion
The final step in adding a new service is to define a renovate package rule to `.github/renovate.json5` which manages the min major version of the service tested against.
```

scbjans (approved):

instrumentation/CONTRIBUTING.md
```diff
@@ -380,6 +380,22 @@ instrumentation_silver:
...
+
+```json
+    {
+      description: "Wait until current major postgres is EoL before updating",
+      dependencyDashboardCategory: "Min Docker service",
+      matchUpdateTypes: ["major"],
+      matchDepNames: ["postgres"],
+      minimumReleaseAge: "1460 days",
+    },
+```
```
kaylareopelle, line 393: Is this something we could add to the instrumentation generator?

instrumentation/CONTRIBUTING.md
```diff
@@ -380,6 +380,22 @@ instrumentation_silver:
...
+
+The days is to be calculated based on the expected age of the major version when it becomes the lowest major version which is not end of life.
```
kaylareopelle:
```suggestion
The `minimumReleaseAge` days value should be calculated based on the expected age of the major version when it becomes the lowest major version which is not end of life.
```

CONTRIBUTING.md
```diff
@@ -163,11 +163,19 @@ And prior to submitting a pr run the following to perform all the checks:
...
 npm run check
-bundle exec rake rubocop
 ```
 
 which will help to ensure your code complies and it passes all the test required of a pull request.
 
+If you would like to perform automated fixes where possible of the issues identified by the check command, you can run:
+
+```sh
+npm run write
```
kaylareopelle, line 173: Is this the same as running `bundle exec rubocop`? 

If I'm reading things correctly, it looks like only `check:source` would run. That script only checks cops from the Bundler, Gemspec, Metrics, and Security departments.

instrumentation/CONTRIBUTING.md
thompson-tomo: It only needs to be added when using a service not already supported.

CONTRIBUTING.md
thompson-tomo: So the write command ends up running the following: 

```sh
jsonlint --config .jsonlintrc.yml --in-place && 
prettier **/*.json --write &&
markdownlint-cli2 '**/*.md' --fix &&
prettier **/*.md --write &&
rubocop --only Lint,Naming,Style -- --autocorrect-all &&
rubocop --only Layout -- --autocorrect-all &&
prettier . --write &&
rubocop --only Bundler,Gemspec,Metrics,Security -- --autocorrect-all
```

Which means that it will do auto-correction of Lint,Naming,Style as well as Layout and also Bundler,Gemspec,Metrics,Security. Based on https://docs.rubocop.org/rubocop/latest/index.html the command was only missing Migration which I have now added.

kaylareopelle (commented):
Thanks for working on this! I have a few suggestions.

CONTRIBUTING.md
kaylareopelle: Thanks for validating!

kaylareopelle (approved):

arielvalentin (approved):

PR #2149, opened 2026-03-31
ci: only build when gem changes

thompson-tomo:
This ensures that the all instrumentation gem is only built when it changes as opposed to always.

### DIFF
--- a/.github/workflows/ci-instrumentation.yml
+++ b/.github/workflows/ci-instrumentation.yml
@@ -5,6 +5,11 @@ on:
   push:
     branches:
       - main
+    paths: &path_filters
+      - "instrumentation/all/**"
+      - ".github/workflows/ci-instrumentation.yml"
+      - ".github/actions/**"
+      - "gemspecs/**"
   pull_request:
     branches:
       - main
--- a/.github/workflows/ci-instrumentation.yml
+++ b/.github/workflows/ci-instrumentation.yml
@@ -13,6 +13,7 @@ on:
   pull_request:
     branches:
       - main
+    paths: *path_filters
   merge_group:
   schedule:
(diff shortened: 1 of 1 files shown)


.github/workflows/ci-instrumentation.yml
```diff
@@ -5,9 +5,15 @@ on:
   push:
     branches:
       - main
+    paths: &path_filters
+      - "instrumentation/all/**"
+      - ".github/workflows/ci-instrumentation.yml"
+      - ".github/actions/**"
+      - "gemspecs/**"
```
arielvalentin, line 13: Did we reverse our decision and committed the gemspecs?

thompson-tomo: This is for the post install message and consistency with the other workflows.

This folder contains the file referenced by  https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/main/instrumentation/all/opentelemetry-instrumentation-all.gemspec#L74

.github/workflows/ci-instrumentation.yml
```diff
@@ -5,9 +5,15 @@ on:
   push:
     branches:
       - main
+    paths: &path_filters
+      - "instrumentation/all/**"
```
arielvalentin: Building the "all" gem is required at the moment before merging.

Are we looking to remove this requirement?

thompson-tomo: Ahh, I had missed that. Yes we should turn that off.

Reason being is we don't need to be building all when we aren't changing either instrumentation or a helper. Note I have updated the path filter to match this.

arielvalentin (approved):

PR #2163, opened 2026-04-05
ci: apt packages in Config file

thompson-tomo:
This moves the definition of required apt packages out of the Workflow and into an apt-packages.txt file which is defined per gem.

This approach ensures if the packages are changed, the pr is labelled accordingly and ensure workflows remain generic.


xuan-cao-swi (approved):

instrumentation/active_storage/test/apt-packages.txt
arielvalentin, line 1: This should go into the `test` directory and not the instrumentation directory.

thompson-tomo: Done.

arielvalentin (approved):

PR #2173, opened 2026-04-07
ci: Run performance and layout checks in RuboCop

kaylareopelle:
The refactor to run Rubocop using NPM in #1974 missed a few Departments: Layout and Performance. This PR makes sure those departments are run as part of the checks.


.github/workflows/ci-markdown-checks.yml
```diff
@@ -108,6 +108,16 @@ jobs:
           set -e # forces exit code to propagate
           npm run check:source:ruby -- --format github
 
+      - name: "Check all ruby files for layout"
+        run: |
+          set -e # forces exit code to propagate
+          npm run check:format:ruby -- --format github
```
thompson-tomo: This would already be covered as part of https://github.com/kaylareopelle/opentelemetry-ruby-contrib/blob/rubocop-performance-layout/.github/workflows/ci-markdown-checks.yml#L167-L168

package.json
```diff
@@ -11,6 +11,7 @@
     "check:lint:markdown": "markdownlint-cli2 '**/*.md'",
     "check:lint:renovate": "renovate-config-validator",
     "check:lint:ruby": "rubocop --only Lint,Naming,Style",
+    "check:performance:ruby": "rubocop --only Performance",
```
thompson-tomo, line 14: We should also have a write command.

And also add performance to both the check & write command.

thompson-tomo (commented):
Sorry for missing performance when I added the commands.

package.json
kaylareopelle: What do you think about adding performance to one of the existing checks rather than making it its own check? We don't have anything other performance checking tools to group it with. Could it fit with `check:source:ruby`?

thompson-tomo: Sure that works for me.

thompson-tomo:
@kaylareopelle are you happy for this to be closed as I have added performance to the check source as discussed.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "scripts": {
-    "check": "npm run check:lint && npm run check:format && npm run check:links && npm run check:spelling && npm run check:source",
+    "check": "npm run check:lint && npm run check:format && npm run check:links && npm run check:performance:ruby npm run check:spelling && npm run check:source",
```
thompson-tomo:
```suggestion
    "check": "npm run check:lint && npm run check:format && npm run check:links && npm run check:performance:ruby && npm run check:spelling && npm run check:source",
```

kaylareopelle:
Hi @thompson-tomo, thank you for closing this and for implementing it in a separate PR. I'm sorry I lost track of the notification for your comment.

PR #2174, opened 2026-04-07
test: instrumentation/http - improve test coverage to meet 85% threshold

### LINKED ISSUE #2157: Improve http instrumentation test coverage
The coverage of the http instrumentation doesn't meet the expected 85% as calculated by simplecov.

aabhinavvvvvvv:
## Summary

Fixes #2157 — the `http` instrumentation test coverage was below the required 85% threshold due to two silent bugs.

### Root causes

**1. Two connection test files were misnamed**

`stable/connection.rb` and `old/connection.rb` in the test directory did not match the Rakefile's `FileList['test/**/*_test.rb']` glob, so they were never executed.

**2. `old/connection_test.rb` was broken**

The `before` block called `instrumentation.install({})` without setting `OTEL_SEMCONV_STABILITY_OPT_IN=old`. This caused `determine_semconv` to return `'stable'`, prepending `Patches::Stable::Connection` instead of `Patches::Old::Connection`.


kaylareopelle (dismissed):
Thank you for catching this! 🚀

arielvalentin (approved):

thompson-tomo:
Thanks for this @aabhinavvvvvvv could you take another look as coverage is currently being calculated as 61%. Thanks

thompson-tomo:
My apologies @aabhinavvvvvvv it looks like the ci wasn't calculating the coverage as expected hence the discrepancy.

thompson-tomo (approved):

hannahramadan (approved):

PR #2178, opened 2026-04-08
fix: deprecation of sql obfuscation

thompson-tomo:
This excludes sql obfuscation from renovate updates and rolls back the ruby 3.3 update.

To help avoid issue again, changes to deprecated will be labelled accordingly.


kaylareopelle (approved):
Thank you!

arielvalentin (approved):

PR #2179, opened 2026-04-08
test(rack): run coverage tests in every appraisal

### LINKED ISSUE #2161: Improve rack instrumentation test coverage
The coverage of the rack instrumentation doesn't meet the expected 85% as calculated by simplecov.

aviralgarg05:
## Summary

Fixes [#2161](https://github.com/open-telemetry/opentelemetry-ruby-contrib/issues/2161) by restoring rack test coverage in the appraisal matrix.

The issue here was not missing rack instrumentation behavior. The real problem was that several semconv-specific rack tests were skipping themselves during the CI-style appraisal runs, so large parts of the suite never executed. That caused coverage to drop far below the enforced threshold even though the code paths already had tests written for them.

This change makes those tests run in the appraisals they were meant to cover, so coverage is measured against the full rack suite again.

## What changed

- removed appraisal-name skip guards that prevented rack tests from running in CI
- set `OTEL_SEMCONV_STABILITY_OPT_IN` explicitly in tests that need `old` or `http/dup`
- kept the change limited to the rack test suite
- left production code unchanged

## Why this fixes #2161

[#2161](https://github.com/open-telemetry/opentelemetry-ruby-contrib/issues/2161) is about the rack package not meeting the expected coverage bar and relying on coverage exceptions instead of exercising the intended test paths.

This PR fixes that at the source:

- the `stable` tests now run in stable appraisals
- the `old` tests now run in old appraisals
- the `dup` tests now run in dup appraisals
- shared rack tests no longer skip unnecessarily

With those tests actually running, the appraisal matrix covers the intended semconv branches and the coverage threshold is satisfied without changing the instrumentation implementation.

## Validation

I verified this locally with the rack package workflow:

- ran the full rack appraisal loop across all generated appraisals
- confirmed all appraisals pass with `0 failures`, `0 errors`, and `0 skips`
- confirmed coverage returns to `97.25% (708 / 728)` during the appraisal runs
- ran `bundle exec rake rubocop` in `instrumentation/rack`
- ran `bundle exec rake yard` in `instrumentation/rack`

## Notes

`yard` still reports the same pre-existing documentation warnings already present on the current base branch. This PR does not add or change production behavior, so those warnings were left out of scope.

## Impact

This closes [#2161](https://github.com/open-telemetry/opentelemetry-ruby-contrib/issues/2161) by making rack coverage enforcement meaningful again and ensuring the existing semconv test suite is fully exercised in CI.


xuan-cao-swi (approved):

hannahramadan (approved):
Thank you for fixing this! These changes look good both locally and on CI.

hannahramadan:
Hi @aviralgarg05, we implemented a broad https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/2272 to how simplecov was reading coverage files. It was overwriting files instead of combining them. After that fix, Rack reported coverage exceeds the threshold from, https://github.com/open-telemetry/opentelemetry-ruby-contrib/issues/2161, so most of the changes in this PR are no longer needed.

However, there's one change worth keeping with an edit: 

```diff
# instrumentation_stable_test.rb‎
- skip unless ENV['BUNDLE_GEMFILE'].include?('http')
+ skip unless ENV['BUNDLE_GEMFILE'].include?('stable')
```

This was silently always true, since no Gemfile contains `http`. 

If you are open reverting all changes and to making that edit, we can accept this PR.

aviralgarg05:
Sure will do it!

aviralgarg05:
Reverted the broader coverage/appraisal changes and kept only the instrumentation_stable_test.rb guard fix, changing the skip check from 'http' to 'stable'.

kaylareopelle (approved):

simi (approved):

kaylareopelle:
Thank you for your contribution, @aviralgarg05! We appreciate your patience with our review!

PR #2182, opened 2026-04-08
chore: cleanup gemfiles

thompson-tomo:
This removes the duplicate references to the logger gem in the gemfiles which are currently generating warnings.


arielvalentin:
I added these because of a warning that it would get removed from gems. Does this mean something else is requiring logger now?

thompson-tomo:
Logger has ended up being added twice to the gemfiles. Most of the time it is 2 lines higher/lower.

arielvalentin (approved):

PR #2190, opened 2026-04-09
ci: skip anthropic instrumentation on JRuby CI

arielvalentin:
## Summary

Add `anthropic` to the JRuby skip list in CI, fixing nightly build failures.

## Motivation

The upstream `anthropic` gem (v1.31.0) has a JRuby load-order bug where `Anthropic::Beta` is not yet defined when `BetaManagedAgentsFileResourceParams` references it at require time, causing `NameError: uninitialized constant Anthropic::Beta`.

This is the same pattern used for 19 other JRuby-incompatible instrumentations.

## Testing

CI-only change — no instrumentation code modified.

## Rollout / Risks

None. Can be reverted if the upstream gem adds JRuby support.

Closes #2187

### DIFF
--- a/.github/workflows/ci-instrumentation-full.yml
+++ b/.github/workflows/ci-instrumentation-full.yml
@@ -158,6 +158,7 @@ jobs:
           [[ "${{ matrix.gem }}" == "rails"                    ]] && echo "skip=true" >> $GITHUB_OUTPUT
           [[ "${{ matrix.gem }}" == "grpc"                     ]] && echo "skip=true" >> $GITHUB_OUTPUT
           [[ "${{ matrix.gem }}" == "gruf"                     ]] && echo "skip=true" >> $GITHUB_OUTPUT
+          [[ "${{ matrix.gem }}" == "anthropic"                 ]] && echo "skip=true" >> $GITHUB_OUTPUT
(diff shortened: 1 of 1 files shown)


.github/workflows/ci-instrumentation-full.yml
```diff
@@ -158,6 +158,7 @@ jobs:
           [[ "${{ matrix.gem }}" == "rails"                    ]] && echo "skip=true" >> $GITHUB_OUTPUT
           [[ "${{ matrix.gem }}" == "grpc"                     ]] && echo "skip=true" >> $GITHUB_OUTPUT
           [[ "${{ matrix.gem }}" == "gruf"                     ]] && echo "skip=true" >> $GITHUB_OUTPUT
+          [[ "${{ matrix.gem }}" == "anthropic"                 ]] && echo "skip=true" >> $GITHUB_OUTPUT
```
thompson-tomo, line 161:
```suggestion
          [[ "${{ matrix.gem }}" == "anthropic"                ]] && echo "skip=true" >> $GITHUB_OUTPUT
```

thompson-tomo (approved):

robbkidd (approved):
choosing to ignore that

PR #2193, opened 2026-04-09
fix(excon): enable SimpleCov coverage merging across appraisal runs

### LINKED ISSUE #2155: Improve excon instrumentation test coverage
The coverage of the excon instrumentation doesn't meet the expected 85% as calculated by simplecov.

nicknikolakakis:
## Summary

- Set a unique `SimpleCov.command_name` per appraisal gemfile so coverage data from each semantic convention mode (stable, old, dup) is preserved and merged rather than overwritten
- This brings combined test coverage from ~58% to 99.19%, meeting the required 85% threshold

### Root cause

The excon instrumentation tests 3 semantic convention modes via separate appraisal gemfiles. Each test run loads code for all 3 modes but only exercises one (the others are skipped based on `BUNDLE_GEMFILE`). Without a unique `command_name`, each run overwrites the previous one's entry in SimpleCov's `.resultset.json`, so coverage never reflects the full test suite.

### Fix

One line in `test_helper.rb` derives a unique `command_name` from `BUNDLE_GEMFILE`:

```ruby
SimpleCov.command_name(ENV.fetch('BUNDLE_GEMFILE', 'default').split('/').last.sub('.gemfile', ''))
```

This produces names like `excon_latest_stable`, `excon_1.3.1_old`, etc., allowing SimpleCov to merge results across all appraisal runs.

### Coverage

| Metric | Before | After |
|--------|--------|-------|
| Per-run coverage | ~58% | ~58% (unchanged per run) |
| Merged coverage | ~58% (overwrites) | **99.19%** (proper merge) |

The 3 remaining uncovered lines (0.81%) are the `rescue StandardError => e` safety nets in each middleware's `handle_response`.

Fixes #2155

## Test plan

- [x] Verified locally: ran all 6 appraisals and confirmed SimpleCov merges coverage to 99.19%
- [x] Verified `SimpleCov.minimum_coverage(85)` passes after merge
- [x] RuboCop passes on the changed file


thompson-tomo:
Thanks for that valuable tip @nicknikolakakis at what could be causing this apparent low coverage. As this is likely to be affecting other gems I have proposed via #2272 to set it at the CI level hence took the pr off the queue. This way all gems benefit from the fix.

instrumentation/excon/test/test_helper.rb
```diff
@@ -5,6 +5,7 @@
 # SPDX-License-Identifier: Apache-2.0
 
 require 'simplecov'
+SimpleCov.command_name(ENV.fetch('BUNDLE_GEMFILE', 'default').split('/').last.sub('.gemfile', ''))
```
thompson-tomo:
```suggestion
```
Would be good to set it at the CI level so all gems benefit aka #2272

thompson-tomo (changes requested):

thompson-tomo (approved):

PR #2196, opened 2026-04-10
fix(Instrumentation-HttpClient): improve test coverage

### LINKED ISSUE #2158: Improve http_client instrumentation test coverage
The coverage of the http_client instrumentation doesn't meet the expected 85% as calculated by simplecov.

GRISONRF:
Fixes #2158 

### Changes:

- Removed `skip unless ENV['BUNDLE_GEMFILE'].include?('stable')` guard from `instrumentation_test.rb`.

- Removed `#install accepts argument` test as it caused intermittent double-prepend failures depending on test execution order (this is my first time coding in Ruby, and if I understood correctly, Ruby's `prepend` is permanent within a process, right? So calling install in this test file, which runs alongside other test files that also call install,  can stack patches multiple times depending on random seed order, leading to unpredictable failures -please someone correct me if I'm wrong)

- Added explicit `determine_semconv` tests to cover the `dup`, `old`, and `stable` branching logic, which was previously only covered indirectly.

Verified locally and all 3 appraisals pass with 0 failures, 0 errors, and coverage of 100%, `SimpleCov.minimum_coverage(85)` passes, and `RuboCop` passes on the changed file.


instrumentation/http_client/test/instrumentation/http_client/instrumentation_test.rb
```diff
@@ -22,9 +20,23 @@
     _(instrumentation.version).wont_be_empty
   end
 
-  describe '#install' do
-    it 'accepts argument' do
-      instrumentation.install({})
+  describe 'determine_semconv' do
+    it 'returns "dup" when OTEL_SEMCONV_STABILITY_OPT_IN includes http/dup' do
+      OpenTelemetry::TestHelpers.with_env('OTEL_SEMCONV_STABILITY_OPT_IN' => 'http/dup') do
+        _(instrumentation.send(:determine_semconv)).must_equal('dup')
```
arielvalentin, line 26: @hannahramadan Is there another way to address this? It seems to me that breaking encapsulation for the purposes of testing is probably not the best way to validate this or increase coverage.

hannahramadan: I'm not sure test coverage is an issue any longer after merging #https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/2272. Here is a [test run](https://github.com/open-telemetry/opentelemetry-ruby-contrib/actions/runs/25022222670/job/73285177187?pr=2293) with a PR that bumps up the coverage to 85% for http_client.

@thompson-tomo as the author of this ticket, can you confirm?

thompson-tomo: That is correct, it is no longer an issue. I was letting the pr's be reviewed as any chance to improve coverage is worthwhile.

It looks like with the latest calculation it is already 100%. So happy for this PR to be closed.

GRISONRF: Ok, I will close this PR then!
Thanks @thompson-tomo @hannahramadan

hannahramadan: Thank you @GRISONRF for helping out here in the first place. It is appreciated.

PR #2198, opened 2026-04-10
feat(active-job): Add performance metrics to process span

bdewater-thatch:
Record cpu_time, idle_time, gc_time, and allocations as span attributes on the `perform.active_job` process span, mirroring the metrics that ActiveSupport::Notifications::Event captures [internally](https://github.com/rails/rails/blob/main/activesupport/lib/active_support/notifications/instrumenter.rb). 

We don't get these for free because these are only captured on the EventObject path (`ActiveSupport::Notifications.subscribe('perform.active_job') { |event| ... }`) where OTel uses the Evented path instead ([fanout.rb](https://github.com/rails/rails/blob/main/activesupport/lib/active_support/notifications/fanout.rb)).

### DIFF
--- a/instrumentation/active_job/lib/opentelemetry/instrumentation/active_job/handlers/perform.rb
+++ b/instrumentation/active_job/lib/opentelemetry/instrumentation/active_job/handlers/perform.rb
@@ -37,6 +37,37 @@ module OpenTelemetry
             { span: span, ctx_token: attach_consumer_context(span, parent_context) }
           end
 
+          # Overrides `Default#start` to also snapshot performance metrics at job start
+          #
+          # @param name [String] of the Event
+          # @param id [String] of the event
+          # @param payload [Hash] containing job run information
+          # @return [Hash] the payload passed as a method argument
+          def start(name, id, payload)
+            payload.merge!(__otel: start_span(name, id, payload), __otel_metrics: snapshot_metrics)
+          rescue StandardError => e
+            OpenTelemetry.handle_error(exception: e)
+          end
+
+          # Overrides `Default#finish` to record performance metrics on the span
+          #
+          # @param _name [String] of the Event (unused)
+          # @param _id [String] of the event (unused)
+          # @param payload [Hash] containing job run information
+          def finish(_name, _id, payload)
+            otel = payload.delete(:__otel)
+            metrics_start = payload.delete(:__otel_metrics)
+            span = otel&.fetch(:span)
+            token = otel&.fetch(:ctx_token)
+
+            record_metrics(span, metrics_start) if span && metrics_start
+            on_exception(payload[:error] || payload[:exception_object], span)
+          rescue StandardError => e
+            OpenTelemetry.handle_error(exception: e)
+          ensure
+            finish_span(span, token)
+          end
+
           # This method attaches a span to multiple contexts:
           # 1. Registers the ingress span as the top level ActiveJob span.
           #    This is used later to enrich the ingress span in children, e.g. setting span status to error when a child event like `discard` terminates due to an error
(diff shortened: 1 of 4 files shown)


thompson-tomo (commented):
As a first point of discussion I am not sure if we should be adding them as attributes on the span. Creating them as metrics seems more appropriate.

bdewater-thatch:
@thompson-tomo thanks for having a look. ~IMO it is useful to know for a specific job execution what these numbers are, as well as seeing them in aggregate.~ TIL about exemplars

I can ~add (or complete~ change over to metrics, but I didn't see any examples in this repo to work from. I did find a bunch of old pull requests but I'll need to know what the desired direction is.

arielvalentin:
@bdewater-thatch thanks for this contribution. 

This is not the first time someone has requested or submitted a PR to add Rails internal measurements as span attributes.

Although I agree that the Rails internal measurements  should be derived from existing spans, I find there is mixed messaging in the spec about recording metric type values on span attributes.

E.g. database batch size and response return rows or HTTP body size, resend count but you can also record any arbitrary HTTP header.

They seem to draw the line on latency attributes, which would very likely be more interesting and familiar to Rails users.

I used to be much more opinionated that these Rails measured performance values should not be added as span attributes but not I'm not so sure. 

As a compromise here I'd like to offer up a suggestion that allows users the option to opt-in  level:

https://opentelemetry.io/docs/specs/semconv/general/attribute-requirement-level/

I haven't looked at how other SDKs are implementing opt-in yet but I think this is a good use case for us to consider with the caveat that there may be a need to have more complex configurations and these attributes would be absent by default.

bdewater-thatch:
On today's SIG call we discussed this PR and agreed to an opt-in, I'll work on that this week.

kaylareopelle:
Hi @bdewater-thatch! Have you had a chance to pick this work back up?

bdewater-thatch:
@kaylareopelle I had this morning :) please take another look!

instrumentation/active_job/lib/opentelemetry/instrumentation/active_job/handlers/perform.rb
```diff
@@ -50,6 +83,70 @@ def attach_consumer_context(span, parent_context)
...
+
+            duration    = finish[:monotonic_time] - start[:monotonic_time]
+            cpu_time    = finish[:cpu_time] - start[:cpu_time]
+            idle_time   = [duration - cpu_time, 0.0].max
+            gc_time     = (finish[:gc_time] - start[:gc_time]) / 1_000_000.0
+            allocations = finish[:allocations] - start[:allocations]
+
+            span.set_attribute('messaging.active_job.job.cpu_time', cpu_time)
+            span.set_attribute('messaging.active_job.job.idle_time', idle_time)
+            span.set_attribute('messaging.active_job.job.gc_time', gc_time)
```
thompson-tomo: similiar to above

```suggestion
            span.set_attribute('rails.gc.time', gc_time)
```

instrumentation/active_job/lib/opentelemetry/instrumentation/active_job/handlers/perform.rb
```diff
@@ -50,6 +83,70 @@ def attach_consumer_context(span, parent_context)
...
+          def record_metrics(span, start)
+            finish = snapshot_metrics
+
+            duration    = finish[:monotonic_time] - start[:monotonic_time]
+            cpu_time    = finish[:cpu_time] - start[:cpu_time]
+            idle_time   = [duration - cpu_time, 0.0].max
+            gc_time     = (finish[:gc_time] - start[:gc_time]) / 1_000_000.0
+            allocations = finish[:allocations] - start[:allocations]
+
+            span.set_attribute('messaging.active_job.job.cpu_time', cpu_time)
```
thompson-tomo: I am thinking we

```suggestion
            span.set_attribute('rails.cpu.time', cpu_time)
```
That way we can have a metric of the same name with a rails.component.name attribute to identify active_job.

instrumentation/active_job/lib/opentelemetry/instrumentation/active_job/handlers/perform.rb
```diff
@@ -50,6 +83,70 @@ def attach_consumer_context(span, parent_context)
...
+            span.set_attribute('messaging.active_job.job.idle_time', idle_time)
+            span.set_attribute('messaging.active_job.job.gc_time', gc_time)
+            span.set_attribute('messaging.active_job.job.allocations', allocations)
```
thompson-tomo: Would this be

```suggestion
            span.set_attribute('rails.memory.allocations', allocations)
```

instrumentation/active_job/lib/opentelemetry/instrumentation/active_job/handlers/perform.rb
```diff
@@ -50,6 +83,70 @@ def attach_consumer_context(span, parent_context)
...
+            finish = snapshot_metrics
+
+            duration    = finish[:monotonic_time] - start[:monotonic_time]
+            cpu_time    = finish[:cpu_time] - start[:cpu_time]
+            idle_time   = [duration - cpu_time, 0.0].max
+            gc_time     = (finish[:gc_time] - start[:gc_time]) / 1_000_000.0
+            allocations = finish[:allocations] - start[:allocations]
+
+            span.set_attribute('messaging.active_job.job.cpu_time', cpu_time)
+            span.set_attribute('messaging.active_job.job.idle_time', idle_time)
```
thompson-tomo: Do we need this attribute?

If we do then perhaps change it to be a utilisation % like other metrics & complements others.

```suggestion
            span.set_attribute('rails.cpu.utilization', idle_time)
```

bdewater-thatch:
We've had our own version running of this in our app and in the end it didn't help us much figuring out the memory growth problem we had. I think this is better implemented with metrics from `GC.stat`, using for example https://www.speedshop.co/blog/a-guide-to-gc-stat/ to know what data to use.

PR #2219, opened 2026-04-13
docs: Extended developer guideline doc

### LINKED ISSUE #2206: Update development guide
The developer documentation should be updated to cover:

- Usage of the labeler
- Adding new services
- How to freeze the development of a gem
- when gem source can be removed after freeze

thompson-tomo:
Closes #2206

This focuses on providing a detailed developer guide which covers adding, deprecating and removal of gems including the steps needed.


xuan-cao-swi:
A couple of thoughts:
- Do we actually need both instrumentation/CONTRIBUTING.md and the main CONTRIBUTING.md?
It might make sense to merge them into a single document and include a table of contents for easier navigation. ([example](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/main/CONTRIBUTING.md))
- Similarly, README.md and datadog-porting-guide.md seem to overlap in purpose. Maybe they could be consolidated or reorganized into a more unified structure.

thompson-tomo:
I agree about merging the contributing guides but would rather treat them separately to avoid making this PR too big.

In relation to dd migration guide, I am inclined to leave it as is as I suspect it would not often be read. I will think if it can be better incorporated.

PR #2222, opened 2026-04-13
ci: Use 3 digit version for otel libs

thompson-tomo:
It emerged via #2212 that only a few locations were using three digit versions, this ensures all otel are using 3 digits. The beneift of 3 digits is that we are more titghly locking down of dependencies for frozen gems. We will still get renovate pr to bump it when they are updated.

It also reduces the chance that a new minor could cause failing pr checks as it would be caught via the bump pr.


arielvalentin:
@thompson-tomo Take a look at the instrumentation generator templates as well. 

Let's make sure that new gems follow the same pessimistic locking scope.

thompson-tomo:
Already done https://github.com/thompson-tomo/opentelemetry-ruby-contrib/blob/consistent_dependencies/.instrumentation_generator/templates/Gemfile 😉

arielvalentin (approved):

arielvalentin:
Always 2 steps ahead of me
I see

PR #2223, opened 2026-04-13
chore: Remove grpc mock

thompson-tomo:
This removes the grpc-mock dependency from gruf due to being abandoned as per #1803 with the last release occurring 2022-07-24


arielvalentin:
I don't see any code changes. Does it mean it wasn't using this library at all in tests?

thompson-tomo:
Correct that is what it looks like to me as well. I couldn't find references to it & the tests still passed without it.

arielvalentin (approved):

PR #2224, opened 2026-04-13
chore: remove unused `yard-doctest`

thompson-tomo:
This removes the unused deprected yard-doctest dependency from the offending gemfiles as shown in https://github.com/open-telemetry/opentelemetry-ruby-contrib/issues/1803 with the last release being 2019-09-12.

### DIFF
--- a/helpers/mysql/Gemfile
+++ b/helpers/mysql/Gemfile
@@ -17,7 +17,6 @@ group :test do
   gem 'rubocop-performance', '~> 1.26.0'
   gem 'simplecov', '~> 0.22.0'
   gem 'yard', '~> 0.9'
-  gem 'yard-doctest', '~> 0.1.6'
   if RUBY_VERSION >= '3.4'
     gem 'base64'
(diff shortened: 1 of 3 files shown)


arielvalentin:
Similarly, was yardoc-test not being invoked as part of the suite?

thompson-tomo:
I don't believe it is based on usage instructions in https://github.com/p0deje/yard-doctest

arielvalentin (approved):
Thanks for finding and cleaning up all of these unused dependencies.

arielvalentin:
@thompson-tomo would you mind updating the cspell dictionary? I don't want releases to fail because of "doctest"

thompson-tomo:
No worries @arielvalentin I have updated the title so it will pass. Note text within `` is always ignored. 😉

robbkidd (approved):
I love yard doc tests, but if we don't have any, then it doesn't make sense to keep this dep around.

PR #2231, opened 2026-04-14
fix: Handle missing config in middleware when OTEL_SDK_DISABLED=true

### LINKED ISSUE #2230: Dup::TracerMiddleware crashes with NoMethodError: undefined method 'include?' for nil when OTEL_SDK_DISABLED=true
Title: Dup::TracerMiddleware crashes with NoMethodError: undefined method 'include?' for nil when OTEL_SDK_DISABLED=true

Gem: opentelemetry-instrumentation-rack v0.30.0
Ruby SDK: opentelemetry-sdk v1.10.0

Description
-----------
When OTEL_SDK_DISABLED=true is set, every incoming request raises:

    NoMethodError: undefined method 'include?' for nil

originating from Dup::TracerMiddleware#untraced_request?. This contradicts the SDK's
documented contract that OTEL_SDK_DISABLED is a seamless no-op swap.

Root Cause
----------
SDK.configure returns immediately when OTEL_SDK_DISABLED=true, so
c.use "OpenTelemetry::Instrumentation::Rack" is never called. The instrumentation is
lazily instantiated but never configured, so Rack::Instrumentation.instance.config
returns {}.

In Dup::TracerMiddleware#initialize:

    @untraced_endpoints = config[:untraced_endpoints]  # {} -> nil (key absent, no defaults applied)

Then on every request in untraced_request?:

    return true if @untraced_endpoints.include?(env['PATH_INFO'])  # nil.include? -> crash

When the SDK is enabled, c.use triggers full configuration which applies defaults
(untraced_endpoints: []). The disabled path skips this entirely, leaving
@untraced_endpoints = nil.

Proposed Fix
------------
Either guard the assignment in initialize:

    @untraced_endpoints = config[:untraced_endpoints] || []

Or use safe navigation in untraced_request?:

    return true if @untraced_endpoints&.include?(env['PATH_INFO'])

Workaround
----------
Skip inserting the middleware when the SDK is disabled:

    unless ENV["OTEL_SDK_DISABLED"] == "true"
      require "opentelemetry/instrumentation/rack/middlewares/dup/tracer_middleware"
      config.middleware.insert_before 0, OpenTelemetry::Instrumentation::Rack::Middlewares::Dup::TracerMiddleware
    end

hannahramadan:
When `OTEL_SDK_DISABLED=true`, the SDK skips configuration and Rack instrumentation's config is an empty hash. When attempting to access nil values, a `NoMethodError` was raised. This same issue seemed to exist in at least excon, grpc, and gruf.

Now, in `Instrumentation::Base#initialize`, `@config` returns declared option defaults for missing keys while staying empty:
                                                                        
```ruby
defaults = (@options || []).to_h { |opt| [opt[:name], opt[:default]] }
@config = Hash.new { |_, k| defaults[k] }
```
                                                                                 
`config[:untraced_endpoints]` returns `[]` instead of `nil`, and `config.empty?` still returns `true`. After `install()`, `@config` is replaced as usual.                   

grpc and gruf were updated to guard with `installed?` instead of `config.empty?`

closes #2230

### DIFF
--- a/instrumentation/base/lib/opentelemetry/instrumentation/base.rb
+++ b/instrumentation/base/lib/opentelemetry/instrumentation/base.rb
@@ -201,9 +201,9 @@ module OpenTelemetry
         @install_blk = install_blk
         @present_blk = present_blk
         @compatible_blk = compatible_blk
-        @config = {}
-        @installed = false
         @options = options
+        @config = config_options({})
+        @installed = false
         @tracer = OpenTelemetry::Trace::Tracer.new
       end
       # rubocop:enable Metrics/ParameterLists
--- a/instrumentation/base/lib/opentelemetry/instrumentation/base.rb
+++ b/instrumentation/base/lib/opentelemetry/instrumentation/base.rb
@@ -202,7 +202,7 @@ module OpenTelemetry
         @present_blk = present_blk
         @compatible_blk = compatible_blk
         @options = options
-        defaults = (@options || []).each_with_object({}) { |opt, h| h[opt[:name]] = opt[:default] }
+        defaults = (@options || []).to_h { |opt| [opt[:name], opt[:default]] }
         @config = Hash.new { |_, k| defaults[k] }
         @installed = false
         @tracer = OpenTelemetry::Trace::Tracer.new
--- a/instrumentation/base/lib/opentelemetry/instrumentation/base.rb
+++ b/instrumentation/base/lib/opentelemetry/instrumentation/base.rb
@@ -202,7 +202,8 @@ module OpenTelemetry
         @present_blk = present_blk
         @compatible_blk = compatible_blk
         @options = options
-        @config = config_options({})
+        defaults = (@options || []).each_with_object({}) { |opt, h| h[opt[:name]] = opt[:default] }
+        @config = Hash.new { |_, k| defaults[k] }
         @installed = false
         @tracer = OpenTelemetry::Trace::Tracer.new
       end
--- a/instrumentation/base/lib/opentelemetry/instrumentation/base.rb
+++ b/instrumentation/base/lib/opentelemetry/instrumentation/base.rb
@@ -202,6 +202,7 @@ module OpenTelemetry
         @present_blk = present_blk
         @compatible_blk = compatible_blk
         @options = options
+        # Empty hash that falls back to option defaults for missing keys
         defaults = (@options || []).to_h { |opt| [opt[:name], opt[:default]] }
         @config = Hash.new { |_, k| defaults[k] }
         @installed = false
--- a/instrumentation/grpc/lib/opentelemetry/instrumentation/grpc/interceptors/client_tracer.rb
+++ b/instrumentation/grpc/lib/opentelemetry/instrumentation/grpc/interceptors/client_tracer.rb
@@ -29,7 +29,7 @@ module OpenTelemetry
           private
 
           def call(type:, requests: nil, call: nil, method: nil, metadata: nil)
-            return yield if instrumentation_config.empty?
+            return yield unless Grpc::Instrumentation.instance.installed?
 
             method_parts = method.to_s.split('/')
             service = method_parts[1]
--- a/instrumentation/rack/test/opentelemetry/instrumentation/rack/middlewares/dup/tracer_middleware_test.rb
+++ b/instrumentation/rack/test/opentelemetry/instrumentation/rack/middlewares/dup/tracer_middleware_test.rb
@@ -415,4 +415,21 @@ describe OpenTelemetry::Instrumentation::Rack::Middlewares::Dup::TracerMiddlewar
       _(first_span.status.code).must_equal OpenTelemetry::Trace::Status::ERROR
     end
   end
+
+  # When OTEL_SDK_DISABLED=true, the SDK skips installation and config remains empty.
+  describe 'when config is empty' do
+    let(:empty_config_rack_builder) { Rack::Builder.new }
+
+    before do
+      instrumentation.instance_variable_set(:@config, {})
+      described_class.send(:clear_cached_config)
+      empty_config_rack_builder.run app
+      empty_config_rack_builder.use described_class
+    end
+
+    it 'handles requests without raising an error' do
+      response = Rack::MockRequest.new(empty_config_rack_builder).get('/ping', env)
+      _(response.status).must_equal 200
+    end
+  end
 end
--- a/instrumentation/rack/test/opentelemetry/instrumentation/rack/middlewares/dup/tracer_middleware_test.rb
+++ b/instrumentation/rack/test/opentelemetry/instrumentation/rack/middlewares/dup/tracer_middleware_test.rb
@@ -416,19 +416,18 @@ describe OpenTelemetry::Instrumentation::Rack::Middlewares::Dup::TracerMiddlewar
     end
   end
 
-  # When OTEL_SDK_DISABLED=true, the SDK skips installation and config remains empty.
-  describe 'when config is empty' do
-    let(:empty_config_rack_builder) { Rack::Builder.new }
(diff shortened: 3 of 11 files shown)


arielvalentin:
@hannahramadan is this not something that should be handled in Base?

Also, any chance that you could add some test coverage for this?

arielvalentin (dismissed):

hannahramadan:
> @hannahramadan is this not something that should be handled in Base?
> 
> Also, any chance that you could add some test coverage for this?

Base could make sense because it looks like this same issue could occur in a few other instrumentations. So we'd either need the `Array()` approach in a few additional spots or add it to Base, which fixes all cases and safe guards this issue in the future. I'll look into this.

instrumentation/grpc/lib/opentelemetry/instrumentation/grpc/interceptors/client_tracer.rb
```diff
@@ -29,7 +29,7 @@ def bidi_streamer(requests: nil, call: nil, method: nil, metadata: nil, &)
           private
 
           def call(type:, requests: nil, call: nil, method: nil, metadata: nil)
-            return yield if instrumentation_config.empty?
+            return yield unless Grpc::Instrumentation.instance.installed?
```
hannahramadan, line 32: This check to see if instrumentation was installed, along with the checks in `gruf/interceptors/client` and `gruf/interceptors/server` aren't technically equivalent to the previous behavior of checking if the configs were empty.

Is `installed?` the right check here, or should we handle this differently? I'm unsure about the original intent of using an config empty vs installed to guard.

instrumentation/base/lib/opentelemetry/instrumentation/base.rb
```diff
@@ -201,9 +201,9 @@ def initialize(name, version, install_blk, present_blk,
         @install_blk = install_blk
         @present_blk = present_blk
         @compatible_blk = compatible_blk
-        @config = {}
-        @installed = false
         @options = options
+        @config = config_options({})
```
hannahramadan: As this is a change to base, I would love some extra eyes here. I noted one downstream impact of checking for empty configs not really being able to exist anymore, since now they wont ever really be.

instrumentation/grpc/lib/opentelemetry/instrumentation/grpc/interceptors/client_tracer.rb
arielvalentin: @AS-AlStar would you mind sharing more information about your intent here?

arielvalentin:
@hannahramadan why not check the envar here? https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/main/instrumentation/base/lib/opentelemetry/instrumentation/base.rb#L216

hannahramadan:
> @hannahramadan why not check the envar here? https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/main/instrumentation/base/lib/opentelemetry/instrumentation/base.rb#L216

@arielvalentin ah because if we have `OTEL_SDK_DISABLED=true`, it [returns early](https://github.com/open-telemetry/opentelemetry-ruby/blob/main/sdk/lib/opentelemetry/sdk.rb#L64) and the `install` method is never reached

instrumentation/grpc/lib/opentelemetry/instrumentation/grpc/interceptors/client_tracer.rb
hannahramadan: @AS-AlStar hoping to learn more about the intent here - wouldn't want to negatively impact anything

arielvalentin: @hannahramadan this feels more "right" but does this mean `instrumentation_config`  will always have a value after installation?

arielvalentin: @AS-AlStar if we do not receive a reply from you we will assume you are no longer able to maintain this instrumentation going forward.

We will merge this change because tests are passing but will have to deprecate the gem and stop releasing it.

AS-AlStar: As I see it, the code value will always remain the same after installation. Each option has a default value, so everything will work correctly. All existing options allow for more precise customization, but they don't affect the basic behavior in any way.

hannahramadan: @arielvalentin - yes, with the `config_options({})` approach, `instrumentation_config` will never be empty.

I can see how maybe that's not ideal? Idk how that could affect other things down the line/ be maybe unexpected. If we don't like that, Here's alt approach that would preserve the `config.empty?` check and then we wouldn't need to make this change to grpc:

```ruby
defaults = (@options || []).each_with_object({}) { |opt, h| h[opt[:name]] = opt[:default] }
@config = Hash.new { |_, k| defaults[k] }
```

Before install, @config is an empty hash with a fallback, so when we ask for values before install, we will get the declared default value. But the config will remain empty. When instrumentation is eventually installed, the @config will re-point to `config_options(config)` as normal:

```ruby
def install(config = {})
  return true if installed?

  @config = config_options(config)
  ...
end
```

Thoughts?

kaylareopelle: Since this specific file is in the gRPC instrumentation, @michal-kazmierczak do you have any input?

instrumentation/rack/test/opentelemetry/instrumentation/rack/middlewares/dup/tracer_middleware_test.rb
```diff
@@ -415,4 +415,20 @@ class SimulatedError < StandardError
...
+      instrumentation.instance_variable_set(:@installed, false)
+      described_class.send(:clear_cached_config)
+      disabled_rack_builder.run app
+      disabled_rack_builder.use described_class
+    end
+
+    it 'handles requests without raising an error' do
+      response = Rack::MockRequest.new(disabled_rack_builder).get('/ping', env)
+      _(response.status).must_equal 200
+    end
```
kaylareopelle: How does this test behave if you use `OpenTelemetry::TestHelpers.with_env` to set the `OTEL_SDK_DISABLED` environment variable?

instrumentation/grpc/lib/opentelemetry/instrumentation/grpc/interceptors/client_tracer.rb
michal-kazmierczak: I believe replacing 
```ruby
if instrumentation_config.empty?
```
(which essentially is `if Grpc::Instrumentation.instance.config.empty?`)

with 
```ruby
unless Grpc::Instrumentation.instance.installed?
```
is the right move. 

I think the intent was to use an empty config as an indirect proxy to detect if the grpc instrumentation was initialized, `Grpc::Instrumentation.instance.installed?` better serves this purpose since it verifies the success of the installation.

hannahramadan: Thank you @michal-kazmierczak, we will keep this change.

instrumentation/rack/test/opentelemetry/instrumentation/rack/middlewares/dup/tracer_middleware_test.rb
hannahramadan: I've updated tests to use this strategy. I think this is more accurate since before, this wasn't actually testing the OTEL_SDK_DISABLED full path

kaylareopelle (approved):

PR #2234, opened 2026-04-16
ci: pin OS matrix

thompson-tomo:
This ensures all os runners are pinned including those in the matrix which can be achieved due to a new custom manager.


.github/workflows/ci-instrumentation.yml
```diff
@@ -29,8 +29,12 @@ jobs:
           - all
         os:
           - ubuntu-latest
+        include: # TODO: remove once the required check has been removed
+          - gem: all
+            os: ubuntu-24.04
+            name: ubuntu-latest
```
thompson-tomo, line 35: I have prepared #2197 which introduces a new check which can become the required check. And once changed this can be removed as part of #2235

.github/renovate.json5
```diff
@@ -179,6 +179,20 @@
       depNameTemplate: "ruby",
       depTypeTemplate: "gemspec.required_ruby_version",
     },
+    {
+      customType: "regex",
+      description: "Update os matrix",
+      managerFilePatterns: [".github/**/**"],
+      matchStrings: [
```
arielvalentin, line 186: The regular expressions are a little difficult to reason about but I think I understand them. 

I only have one suggestion which is to minimize greedy matchers here so that it hits the first match and then stops scanning till the EoF.

arielvalentin (approved):

.github/renovate.json5
thompson-tomo: The issue would be some of our files have multiple matches. 😉 The key thing is, in the case here due to being recursive it looks for the 1st match string ie matrix followed by os string and then the next :. for each match it match it runs the second match string and then on the third it gets what it wants.

PR #2251, opened 2026-04-19
ci: Restrict link checking

### LINKED ISSUE #2247: linkspector check failures
## The Problem 

Linkspector runs as part of an OS matrix validating links multiple times.

They are also fragile and dependent on the remote servers to return a HTTP 2xx response. 

The examples below demonstrate those problems, one where too many jobs running in parallel or requests made in a short time period lead to a HTTP 429 Too Many Requests or a HTTP 504 Gateway timeout:

https://github.com/open-telemetry/opentelemetry-ruby-contrib/actions/runs/24605467121/job/71950871705

> propagator\xray\README.md:63:1: 🚫 https://cloud-native.slack.com/archives/C01NWKKMKMY Status:429 Cannot reach link
> - Checking propagator\xray\README.md...
> 💥 Error: Some hyperlinks in the specified files are invalid.

https://github.com/open-telemetry/opentelemetry-ruby-contrib/actions/runs/24605467121/job/71950871705

> instrumentation\logger\README.md:60:1: 🚫 https://github.com/ruby/logger Status:504 Cannot reach link
> - Checking instrumentation\logger\README.md...
> instrumentation\lmdb\README.md:43:1: 🚫 https://github.com/minad/lmdb Status:504 Cannot reach link
> - Checking instrumentation\lmdb\README.md...
> - Checking instrumentation\lmdb\README.md...
> instrumentation\lmdb\README.md:47:1: 🚫 https://github.com/open-telemetry/community#ruby-sig Status:504 Cannot reach link
> instrumentation\lmdb\README.md:48:1: 🚫 https://github.com/open-telemetry/community#community-meetings Status:504 Cannot reach link
> - Checking instrumentation\lmdb\README.md...
> - Checking instrumentation\lmdb\README.md...
> instrumentation\lmdb\README.md:50:1: 🚫 https://github.com/open-telemetry/opentelemetry-ruby/discussions Status:504 Cannot reach link
> - Checking instrumentation\koala\README.md...
> instrumentation\koala\README.md:36:34: 🚫 https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/main/instrumentation/koala/example/trace_demonstration.rb Status:504 Cannot reach link
> - Checking instrumentation\koala\README.md...
> instrumentation\koala\README.md:55:1: 🚫 https://github.com/open-telemetry/opentelemetry-ruby/discussions Status:504 Cannot reach link
> - Checking instrumentation\grpc\README.md...
> instrumentation\grpc\README.md:6:48: 🚫 https://github.com/grpc/grpc/tree/master/src/ruby Status:504 Cannot reach link
> 💥 Error: Some hyperlinks in the specified files are invalid.

## What should we do?

I believe these checks are redundant since they should not change as a result of being run on a different OS. We should reduce the. number checks to a single OS and only run once. 

We should also skip running linkspector tests as part of merge groups unless relevant files change.

thompson-tomo:
Closes: #2247

This restricts checking all links to just the action, the local checks workflow when triggered by the cloud will check just 1 doc with 1 link hence addressing issues we are seeing.

Note we should evaluate replacing the link checker due to it not being widely used/supported and key features missing needed to be worked around or broken ie run on modified only, lack of cli arguments hence config file switching.

Note this has disabled link checking on Windows.


thompson-tomo:
@arielvalentin i believe I have finally arrived at a situation where this false failing link checks should be resolved.

bin/link-check.cjs
```diff
@@ -0,0 +1,22 @@
+// This script is necessary as we can't pass environment variables to
+// npm scripts in a cross platform manner.
+
+const os = require("node:os");
+
+// Detect Windows and exit immediately
+// Workaround for https://github.com/nodejs/node/issues/56645
+if (os.platform() === "win32") {
```
arielvalentin, line 8: Any reason why we need to run these across macOS?

I think that most if not all of the linters only need to run in Ubuntu and don't need to be run as part of a matrix.

thompson-tomo: It will only check a single link in the cloud but this way we can confirm that the command we mention in the readme works regardless of what the user is developing on.

A key scenario would be ensuring line endings are not causing issues like we saw in main recently.

arielvalentin (approved):

PR #2271, opened 2026-04-21
chore: Add custom coverage threshold for httpx

thompson-tomo:
Set minimum coverage threshold in GitHub CI


thompson-tomo:
@arielvalentin once this all passes, it would be good to get this in as it will resolve the failing test we have atm.

thompson-tomo:
@arielvalentin all tests are now passing.

arielvalentin (approved):

PR #2272, opened 2026-04-21
fix: Set simplecov command name based on appraisal name

thompson-tomo:
This sets a name for the simplecov command via environment so that collate has a collection of reports to collate rather than the last one. This is necessary because otherwise each run gets the same guid.

### DIFF
--- a/.github/actions/test_gem/action.yml
+++ b/.github/actions/test_gem/action.yml
@@ -109,9 +109,9 @@ runs:
         if [[ -f "Appraisals" ]]; then
           for i in `bundle exec appraisal list | sed 's/-/_/g' `; do
             echo "::group::🔎 Appraising ${i}"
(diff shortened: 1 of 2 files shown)


arielvalentin (approved):

arielvalentin:
@thompson-tomo I prematurely approved... it was more of a hopeful approval

thompson-tomo:
Yeap, I was hopeful to. Will return to this later today when I am actually at a pc which will hopefully make it easier to test. 🤞

thompson-tomo:
@arielvalentin and it is now ready. Just as I was about to call it a night, I remembered the .simplecov file. 🤣 the issue was that file was overriding it. It should be good to merge now. The key thing is checking for lines like below

> Coverage report generated for appraisal_httpclient_latest_dup, appraisal_httpclient_latest_old, appraisal_httpclient_latest_stable to /home/runner/work/opentelemetry-ruby-contrib/opentelemetry-ruby-contrib/instrumentation/http_client/coverage.

List multiple files, previously it was only the 1 guid.

arielvalentin (approved):

PR #2276, opened 2026-04-22
ci: automated renovate dry-run

thompson-tomo:
This runs a renovate dry-run when the renovate config file changes or the Workflow changes.

### DIFF
--- a/.github/workflows/test-renovate.yml
+++ b/.github/workflows/test-renovate.yml
@@ -5,7 +5,7 @@ on:
     branches:
       - main
     paths:
-      - ".github/workflows/renovate.yml"
+      - ".github/workflows/test-renovate.yml"
       - ".github/renovate.json5"
   workflow_dispatch:
 
--- a/.github/workflows/test-renovate.yml
+++ b/.github/workflows/test-renovate.yml
@@ -1,4 +1,4 @@
-name: Renovate
+name: Test / Renovate
 
 on:
   pull_request:
@@ -36,4 +36,4 @@ jobs:
           RENOVATE_TOKEN: "${{ secrets.GITHUB_TOKEN }}"
           RENOVATE_GITHUB_COM_TOKEN: "${{ secrets.GITHUB_TOKEN }}"
         run: |
-          npm run execute:renovate:dryrun
+          npm run test:renovate:dryrun
--- a/.github/workflows/test-renovate.yml
+++ b/.github/workflows/test-renovate.yml
@@ -18,8 +18,7 @@ concurrency:
 
 jobs:
   automated-checks:
-        if: ${{ github.repository == 'open-telemetry/opentelemetry-ruby-contrib' }}
-
+    if: ${{ github.repository == 'open-telemetry/opentelemetry-ruby-contrib' }}
     strategy:
       fail-fast: false
     name: Dry-run
--- a/.github/workflows/test-renovate.yml
+++ b/.github/workflows/test-renovate.yml
@@ -12,8 +12,14 @@ on:
 permissions:
   contents: read
 
+concurrency:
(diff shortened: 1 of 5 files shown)


.github/workflows/test-renovate.yml
```diff
@@ -0,0 +1,39 @@
...
+    branches:
+      - main
+    paths:
+      - ".github/workflows/test-renovate.yml"
+      - ".github/renovate.json5"
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
```
arielvalentin, line 14:
```suggestion

concurrency:
  group: ${{ github.workflow }}-${{ github.event_name }}-${{ github.ref }} # Ensure that only one instance of this workflow is running per event type and ref
  cancel-in-progress: true # Cancel any previous runs of this workflow for the same event type and ref

```

.github/workflows/test-renovate.yml
```diff
@@ -0,0 +1,39 @@
...
+jobs:
+  automated-checks:
```
arielvalentin, line 20:
```suggestion
  automated-checks:
        if: ${{ github.repository == 'open-telemetry/opentelemetry-ruby-contrib' }}

```

package.json
```diff
@@ -14,6 +14,7 @@
     "check:source": "npm run check:source:ruby",
     "check:source:ruby": "rubocop --only Bundler,Gemspec,Metrics,Migration,Minitest,Performance,Rake,RSpec,Security",
     "check:spelling": "cspell -c .cspell.yml .",
+    "test:renovate:dryrun": "renovate --platform=local",
```
arielvalentin: https://docs.renovatebot.com/modules/platform/local/

> Experimental features might be changed or even removed at any time.

I would prefer if we wait till features like this are GA.

arielvalentin (changes requested):

package.json
thompson-tomo: @arielvalentin command has been switched and produces desired outcome.

package.json
```diff
@@ -14,6 +14,7 @@
     "check:source": "npm run check:source:ruby",
     "check:source:ruby": "rubocop --only Bundler,Gemspec,Metrics,Migration,Minitest,Performance,Rake,RSpec,Security",
     "check:spelling": "cspell -c .cspell.yml .",
+    "test:renovate:dryrun": "renovate --dry-run=full --autodiscover",
```
arielvalentin: I don't understand what the autodiscover flag is doing here. 

> When you enable autodiscover, by default, Renovate runs on every repository that the bot account can access. You can limit which repositories Renovate can access by using the autodiscoverFilter config option.

That does not seem ideal and reads to me like it's going to try to run on all repos in the org. Am I misunderstanding this?

Also for whatever reason these options are documented under self hosted. I don't know what that means for us. 

Is this only available for certain kinds of customers? Is CNCF OTel self hosted?

https://docs.renovatebot.com/self-hosted-configuration/

thompson-tomo: Autodiscover is scanning of the file system for repository config.

Based on the logs (see below), it is only finding 1 repository:

> 2026-05-01T03:34:02.5660223Z DEBUG: Autodiscovering GitHub repositories
2026-05-01T03:34:02.7381108Z DEBUG: Autodiscovered 1 repositories
2026-05-01T03:34:02.7390334Z  INFO: Autodiscovered repositories
2026-05-01T03:34:02.7390803Z        "length": 1,
2026-05-01T03:34:02.7391436Z        "repositories": ["open-telemetry/opentelemetry-ruby-contrib"]

Note the logs are on the check.

The cli tool = self-hosted as you can use that tool to create your pr's for a collection of repos etc.

thompson-tomo: To be extra certain I have added a filter so even if it did somehow end up finding other repos, it will still just run on this one.

arielvalentin (approved):

PR #2278, opened 2026-04-22
ci: (actions) interpolate into env vars

robbkidd:
Move ${{ ... }} interpolations from 'run:' scripts to env vars to avoid shell injection shenanigans.

Similar to the updates to workflows in #2269, but this is for the custom `test_gem` action used _within_ workflows.

### DIFF
--- a/.github/actions/test_gem/action.yml
+++ b/.github/actions/test_gem/action.yml
@@ -36,9 +36,13 @@ runs:
     - name: Setup
       id: setup
       shell: bash
+      env:
+        GEM_INPUT: ${{ inputs.gem }}
+        RUBY_INPUT: ${{ inputs.ruby }}
+        LATEST_INPUT: ${{ inputs.latest }}
       run: |
         # 🛠️ Setup 🛠️
-        dir=$(find . -iname '${{ inputs.gem }}.gemspec' -exec dirname {} \;)
+        dir=$(find . -iname "${GEM_INPUT}.gemspec" -exec dirname {} \;)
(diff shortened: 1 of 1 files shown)


.github/actions/test_gem/action.yml
```diff
@@ -36,9 +36,13 @@ runs:
     - name: Setup
       id: setup
       shell: bash
+      env:
+        GEM_INPUT: ${{ inputs.gem }}
+        RUBY_INPUT: ${{ inputs.ruby }}
+        LATEST_INPUT: ${{ inputs.latest }}
```
thompson-tomo, line 42: Is it possible to set these at the composite action level rather than under each job?

arielvalentin (approved):

.github/actions/test_gem/action.yml
robbkidd: Not in Action YAML syntax. Composite `runs:` only has `using:` and `steps:` keys, [according to the docs](https://docs.github.com/en/actions/reference/workflows-and-actions/metadata-syntax#runs-for-composite-actions). 

We could set these only once in this first Setup step by [writing them to `$GITHUB_ENV`](https://docs.github.com/en/actions/reference/workflows-and-actions/metadata-syntax#runsstepsenv). Subsequent steps would have these env vars set … but _every_ subsequent step would have them, even those in a calling workflow after this action is invoked.

I think the duplication here is minimal. It makes the intent and use explicit instead of magical and potentially surprising.

PR #2280, opened 2026-04-23
ci: Use gem env variables to specify if Jruby is not supported

thompson-tomo:
This moves the control of if jruby is not supported out of the Workflow file & into the gem environment variables file.

This ensure workflows are generic with any specifics for a. Gem defined via environment variables listed alongside the source code.

This improves readability and maintainable due to not needing to alter ci when adding/removing a gem. The appropriate code owners can be identified when adding/removing interpreters. Also helps with only building/testing what has changed.

Note this is resulting in more gems being tested on jruby as only those needing an exclusion were added rather than those which had needed it in the past.

Related pr's:
- #2203 
- #2282

### DIFF
--- a/.github/workflows/ci-instrumentation-full.yml
+++ b/.github/workflows/ci-instrumentation-full.yml
@@ -153,7 +153,7 @@ jobs:
           build: true
           minimum_coverage: ${{ env.minimum_coverage }}
       - name: "Test JRuby"
-        if: "${{ startsWith(matrix.os, 'ubuntu') && !contains(env.unsupportedInterpreters, split(matrix.ruby, '-')[0]) }}"
+        if: "${{ startsWith(matrix.os, 'ubuntu') &&  !contains(env.unsupportedInterpreters, 'jruby') }}"
         uses: ./.github/actions/test_gem
         with:
           gem: "opentelemetry-instrumentation-${{ matrix.gem }}"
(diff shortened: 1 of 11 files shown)


arielvalentin (commented):
Where are the `.github-ci.yaml` files loaded? 

Everything seems to be passing. Does that mean that JRuby works for all of these? 

https://github.com/open-telemetry/opentelemetry-ruby-contrib/actions/runs/24869007966/job/72811264185?pr=2280

Image

thompson-tomo:
The CI config file is loaded in https://github.com/thompson-tomo/opentelemetry-ruby-contrib/blob/jruby-optimisation/.github/workflows/ci-instrumentation-full.yml#L127-L135

Yes the current tests report jruby is compatible. What I did was run it for all gems and only excluded when necessary.

arielvalentin:
> The CI config file is loaded in https://github.com/thompson-tomo/opentelemetry-ruby-contrib/blob/jruby-optimisation/.github/workflows/ci-instrumentation-full.yml#L127-L135
> 
> Yes the current tests report jruby is compatible. What I did was run it for all gems and only excluded when necessary.

I see! I picked one of the wrong gems and see this working. I guess I am still confused because the tests are passing on this branch and forgot that the config files were added in an [earlier PR](https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/2136). 

image

arielvalentin (approved):

PR #2282, opened 2026-04-24
ci: Move env variables out of workflow

thompson-tomo:
This moves environment variables used by tests out of the composite action which is exposed to all tests to now be defined in the environment variables file within the gem directory.

This ensures that the workflows/actions are generic with any customisations defined within the scope of the gem directory. Hence reduces redundant build checks due to only necessary gems being built/tested

In the future we could even automate the test helpers to set the variables to match the definition file reducing the number of locations which values are defined.

Related pr's:
- #2203 
- #2280


arielvalentin (commented):
Nit: I think these CI configuration files should be in a test subdirectory as opposed to the root?

I think naming the files `.config.yaml` or something of that nature may also be something to consider. 

Would it be useful to set these options in unit tests and docker as well so they match?

thompson-tomo:
The reason I didn't put it under tests as it also can define environment variables used by simplecov etc. The extension will be to define if jruby etc is supported.

In relation to setting them in local unit tests, I agree that would be nice and in fact I was thinking about the best option for that. Note this is intended to be a like for like of what we already had.

arielvalentin (approved):

PR #1701, opened 2025-09-27
feat!: Min Ruby Version 3.2 and Rails 7.1

arielvalentin:
Ruby 3.1 is no longer receiving updates as of (31 Mar 2025)
Rails 7.0 is no longer supported as of (01 Apr 2025)
JRuby 10 supports newer versions of Ruby
Update Style to Match Ruby 3.2

PR #2283, opened 2026-04-24
feat!: Min Rails 7.1 (enforced this time)

hmcguire-shopify:
[feat!: bump runtime MINIMUM_VERSION to Rails 7.1](https://github.com/open-telemetry/opentelemetry-ruby-contrib/commit/bb5116b1b8df503442a5d39b33833dae156d55ce) 

PR https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/1701 dropped Rails 7.0 from the Appraisals test matrix and declared
Rails 7.1 as the new minimum, but did not update the MINIMUM_VERSION
constants used by each instrumentation's compatible { ... } check at
runtime. As a result, consumers on Rails 7.0 would still have these
instrumentations install successfully despite being on an unsupported
and untested Rails version.

Bump MINIMUM_VERSION to 7.1 across all Rails-family instrumentations
(rails, action_pack, action_view, action_mailer, active_job,
active_record, active_support, active_storage), mirroring the pattern
established in PR https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/1231 (the 6.1 -> 7.0 drop). active_storage previously
used the '7.0.0' format; it is normalized to '7.1' for consistency
(Gem::Version treats '7.1' and '7.1.0' as equal).

---

[refactor(active_record): simplify querying patch](https://github.com/open-telemetry/opentelemetry-ruby-contrib/commit/7ba2a53b882ba1bd78662cdc11ef53e6d4e532aa) 

The querying patch chose between :_query_by_sql (Rails 7+) and
:find_by_sql (Rails 6.1) via a runtime Gem::Version comparison and
define_method. Since the minimum supported Rails version is now 7.1,
the :find_by_sql branch is unreachable and the dynamic dispatch is
unnecessary.

Replace define_method with a plain def, and use (...) forwarding to
match the convention used elsewhere in the repo.

### DIFF
--- a/instrumentation/action_mailer/lib/opentelemetry/instrumentation/action_mailer/instrumentation.rb
+++ b/instrumentation/action_mailer/lib/opentelemetry/instrumentation/action_mailer/instrumentation.rb
@@ -55,7 +55,7 @@ module OpenTelemetry
       #     })
       #   end
       class Instrumentation < OpenTelemetry::Instrumentation::Base
-        MINIMUM_VERSION = Gem::Version.new('7')
+        MINIMUM_VERSION = Gem::Version.new('7.1')
         EMAIL_ATTRIBUTE = %w[email.to.address email.from.address email.cc.address email.bcc.address].freeze
 
         install do |_config|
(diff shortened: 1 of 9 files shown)


robbkidd (approved):
You are technically correct. The best kind of correct.

PR #2291, opened 2026-04-27
chore(deps): Bump mysql and min trilogy

### LINKED ISSUE #2044: Upgrade mysql tests to use 8.4 or higher
When I was working on https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/2019 I had attempted to consolidate the mysql version on to 8.4 but I couldn't get the CI to pass.

The error which I was getting without any changes was and affected both mysql2 and trilogy:

> Error:  2026-02-25T04:20:16.191787Z 0 [ERROR] [MY-000067] [Server] unknown variable 'default-authentication-plugin=mysql_native_password'.

This error can be addressed by changing the servicd command from

> exec docker-entrypoint.sh mysqld --default-authentication-plugin=mysql_native_password

To

> exec docker-entrypoint.sh mysqld --mysql-native-password=ON

The resolves the ci issues for mysql2 but not trilogy which was producing the following error:

> Trilogy::BaseConnectionError: trilogy_auth_recv: caching_sha2_password requires either TCP with TLS or a unix socket: TRILOGY_UNSUPPORTED

To attempt to address this I have tried the following but no success or any change in the error:

- mounting an Init script see appendix 1 as a volume using `./mysql-init:/docker-entrypoint-initdb.d`. Issue is volumes are apparently not supported for github action services?
- Adding a post install step to change permissions. See appendix 2
- read through https://github.com/trilogy-libraries/trilogy/issues/26

In summary the issues are:

• cannot mount init scripts
• cannot use Unix sockets
• cannot inject TLS certs

Which all appear related to https://dev.mysql.com/doc/refman/8.4/en/caching-sha2-pluggable-authentication.html

My suggested solution would be to switch to mariadb.

------------

Appendix 1

```sql
ALTER USER 'root'@'%' IDENTIFIED WITH mysql_native_password BY 'root';
ALTER USER 'root'@'localhost' IDENTIFIED WITH mysql_native_password BY 'root';
FLUSH PRIVILEGES;
```

Appendix 2

```yaml
- name: Fix MySQL auth plugin for Trilogy
  run: |
    sudo apt-get update && sudo apt-get install -y mysql-client
    until mysql -h 127.0.0.1 -uroot -proot -e "SELECT 1"; do
      echo "Waiting for MySQL..."
      sleep 2
    done
    mysql -h 127.0.0.1 -uroot -proot <<'EOF'
      ALTER USER 'root'@'%' IDENTIFIED WITH mysql_native_password BY 'root';
      ALTER USER 'root'@'localhost' IDENTIFIED WITH mysql_native_password BY 'root';
      FLUSH PRIVILEGES;
    EOF

```

thompson-tomo:
Closes #2044

Bumps mysql and changes trilogy connection to work with new versions. This has meant the min trilogy version is bumped from 2.3.0 to 2.11.


arielvalentin:
@composerinteralia any tips ?

instrumentation/trilogy/test/opentelemetry/instrumentation/trilogy/instrumentation_test.rb
```diff
@@ -21,7 +21,8 @@
       username: username,
       password: password,
       database: database,
-      ssl: false
+      ssl_mode: Trilogy::SSL_VERIFY_CA,
+      sslca: File.expand_path('../../../../../../.docker/infra/mysql/certs/ca.pem', __dir__)
```
composerinteralia: typo?

```suggestion
      ssl_ca: File.expand_path('../../../../../../.docker/infra/mysql/certs/ca.pem', __dir__)
```

composerinteralia:
But also Trilogy v2.11.0 supports caching_sha2_password without tls, so it's possible we don't need to do all this if we bump the version of Trilogy used in tests.

instrumentation/trilogy/test/opentelemetry/instrumentation/trilogy/instrumentation_test.rb
thompson-tomo: Let's try. 🤞

thompson-tomo:
Omg that did it. Thanks very much for your input @composerinteralia 😄 

@arielvalentia when you have a chance can you take a look please.

.docker/infra/mysql/certs/ca-key.pem
arielvalentin, line 1: do we have to commit private signed keys? can we generate these as part of the test suite and ignore them in git?

arielvalentin: or bump the minimum version of trilogy we support? https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/2291#issuecomment-4353079364

composerinteralia: Trilogy uses this [script to generate test keys](https://github.com/trilogy-libraries/trilogy/blob/main/test/mysql/docker-entrypoint-initdb.d/generate_keys.sh), if you need inspiration.

thompson-tomo: If we bump it we would only be testing on trilogy released since march 23, 2026 instead of October 2024. I don't think that is old enough. https://rubygems.org/gems/trilogy/versions

I can look to incorporate running the generation script before compose during the day and drop those files out.

arielvalentin: @composerinteralia Does trilogy have an explicit EoL or version support policy?

composerinteralia: No explicit policy. In practice we've only applied bug fixes, etc. to the latest release, and don't backport to older releases. So older releases may or may not work with newer versions of Ruby, MySQL, etc.

Another approach might be to continue informally supporting older Trilogy versions but only test on more recent versions (unless there are specific compatibility issues that we need to test).

arielvalentin: @thompson-tomo in that case I think that we update the readme and inform users we are no longer supporting older versions of Trilogy and add a min version constraint to the instrumentation.

thompson-tomo: Thanks very much @composerinteralia & @arielvalentin changes have been made as suggested. All tests are still passing with minimal changes.

arielvalentin (approved):

PR #2312, opened 2026-05-01
chore: Split examples out of docker compose

thompson-tomo:
Progresses: #2320

This splits the definition of services of the examples into dedicated compose files residing within the corresponding gem example folder.

This is a purely mechanical pr with the next intended pr to be adding in of automatic running of the ruby examples on start to confirm they are runable. Further to that exploration on if it is possible to detect error in the container & if so propogate it.

None of that is currently possible hence why it can come later & up until #2147 we weren't even checking if the services were valid.

The only functional change is a dedicated compose file for examples which is used for ci. This is only for until a future step where this build is on a per gem basis.

Next steps blocked until this is merged:
- ensure examples run
- switch examples to base ruby image rather than custom
- bumping dev environment to latest ruby


docker-compose.yml
```diff
@@ -7,6 +7,21 @@ include:
...
+  - ./instrumentation/excon/example/compose.yml
+  - ./instrumentation/faraday/example/compose.yml
+  - ./instrumentation/mysql2/example/compose.yml
+  - ./instrumentation/net_http/example/compose.yml
+  - ./instrumentation/net_ldap/example/compose.yml
+  - ./instrumentation/pg/example/compose.yml
+  - ./instrumentation/que/example/compose.yml
+  - ./instrumentation/rack/example/compose.yml
+  - ./instrumentation/restclient/example/compose.yml
+  - ./instrumentation/sinatra/example/compose.yml
```
thompson-tomo: Does this bring any value or would it be OK to remove them?

kaylareopelle (approved):

PR #2313, opened 2026-05-01
chore: Remove un-necessary test service

thompson-tomo:
This removes the baggage processor docker service which is just running the tests without any external services.

No other gem has a service like this especially as it doesn't use any other docker services hence opted to remove for consistency.


arielvalentin:
@robbkidd @MikeGoldsmith did you all add this docker service to test or verify the baggage processor?

thompson-tomo:
Also note we now have the base docker service, which is setup identical and only difference being the working directory + need to manual call the commands.

MikeGoldsmith:
Yeah, it was added when the predicate func for the processor was added in #990. I think it's okay to remove so long as the base docker service calls the test instead of having a custom variant.

simi (approved):

kaylareopelle (approved):

PR #2314, opened 2026-05-01
ci: Use dotenv to set environment variables

thompson-tomo:
This switches all of the tests across to using dotenv to set environment variables rather than custom ci solution. This ensures that both cloud & local use the same setup. Also switched is the examples which have workspace services to use the same approach

Have also added linting to ci workflow

Reason file is in the root of the gem is so that they can be reused for examples, a file just in test or example folder is only for test or example, whereas being in parent means it applies to both.


arielvalentin (commented):
I thought we were converging in the ci yaml files?

thompson-tomo:
The CI yaml file enabled a like for like migration for what we had previously without touching any source code. This takes it a step further by enabling the env variables to be automatically set for local tests & examples, not just the CI hence further simplifying/reducing duplication.

Tbh, I was expecting to have to script something to set them locally hence why I held off but looked into it some more after https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/2282#issuecomment-4357270351 and I found this solution.

Note The CI yml will still be used for variables which are needed for the ci logic ie is jruby supported but we can we can use the .env.test for the variables only needed within the ruby process.

arielvalentin:
I'm not sure where we are going here. 

I'm going to ask that you create an issue that maps out the final state of where you want to CI, dev, and workflows with all of these changes. It would also help for us to follow along if you would file issues describing the challenges and problems you have encountered and want to address. 

Changes like this are a little harder to review because I'm not sure where we are going and if there is a benefit to shifting configs around from multiple configurations (actions yaml and code) to a different set of multiple configurations (gem ci yaml and env files).

thompson-tomo:
@arielvalentin i have raised https://github.com/open-telemetry/opentelemetry-ruby-contrib/issues/2318 which seeks to describe what I am progressing towards with the key goal being increased efficiency in build process while reducing maintenance and increasing coverage to cover more os etc without decreasing usability of the checks in pr's.

Note the idea is eliminate duplication of config hence it is either `.env.test` or `github-ci.yml` and not both. The deciding factor is, does the variable need to be exposed outside the ruby process or not.

instrumentation/bunny/example/.env
```diff
@@ -0,0 +1,2 @@
+RABBITMQ_URL=amqp://guest:guest@localhost:5672
+OTEL_TRACES_EXPORTER=console
```
kaylareopelle: Just to follow-up on the linter's concern
```suggestion
OTEL_TRACES_EXPORTER=console
RABBITMQ_URL=amqp://guest:guest@localhost:5672
```

thompson-tomo: Thanks, i hadn't noticed the message.

kaylareopelle (approved):

PR #2339, opened 2026-05-12
ci: Add a check so toys-release runs on gemfile change

thompson-tomo:
This has the toys-release gem run on changes to the gemfile with toys-release just showing the help which is the lowest impact command option. The hope here is that it would catch scenarios where toys-release could not be loaded as well as where.

Ideally we could use existing workflows and just but cli into a dryrun mode using env variables.


kaylareopelle (approved):
I like this idea. I'm going to leave #2336 open until we can try this in the context of that Renovate PR.

PR #2344, opened 2026-05-16
chore: ensure markdownlint in gh action & cli behave the same

thompson-tomo:
Add default and showConfig options to markdownlint config


thompson-tomo:
Consistency can be achieved by removing the explicit setting of fix to false. Issue will be fixed in gh action, hence closing.

PR #2347, opened 2026-05-17
feat: only emit span events by default on old semconv

thompson-tomo:
This switches span events to only be enabled by default on old semconv due to span events undergoing deprecation, the event name is not accurate & the definition is not stable.

Coverage report before change:

> Line Coverage: 98.31% (233 / 237)

Coverage report after change:

> Line Coverage: 98.32% (234 / 238)


hannahramadan:
Hi @thompson-tomo! I was reading the [blog post](https://opentelemetry.io/blog/2026/deprecating-span-events/) about this and it seems like the guidance is to move the signal to logs, not drop it, and to keep existing stable majors behaviorally compatible for now. If I remember correctly, the Logs API isn't quite ready for this change. I'm worried this is a step too soon, but I could be missing something. What do you think?

thompson-tomo:
Yes the intention would be to add log events however this doesn't drop them but instead only emits them by default on the old conventions hence users need to optin to keep them when they switch to stable.

> existing stable majors behaviorally compatible for now.

Doesn't apply here as we don't have a major release which is what I want to get us to a position to release.

In regards to if the logs api is ready I think it might be functionally ready after @kaylareopelle recent pr to add event names. That being said, it's not a stable package. So more work would be needed there.

PR #2359, opened 2026-05-19
refactor: integrate httpx instrumentation with the tracing plugin

HoneyryderChuck:
`httpx` v1.7.4 ships with the `:tracing` plugin, which ships the whole boilerplate around tracing when a request starts and finishes, thereby enabling the removal of most of the current code.

It has been since then the foundation for the `:datadog` integration too.


thompson-tomo (dismissed):
Can you ensure that the tests are passing to avoid regressions.

HoneyryderChuck:
@thompson-tomo tests are passing now

HoneyryderChuck:
@thompson-tomo any updates?

instrumentation/httpx/Appraisals
```diff
@@ -13,10 +13,10 @@ semconv_stability = %w[dup stable old]
 
 semconv_stability.each do |mode|
   appraise "httpx-1-#{mode}" do
-    gem 'httpx', '~> 1.6'
+    gem 'httpx', '>= 1.7.4', '< 2'
   end
 
   appraise "httpx-latest-#{mode}" do
-    gem 'httpx'
+    gem 'httpx', '>= 1.7.4'
```
thompson-tomo:
```suggestion
    gem 'httpx'
```
Not necessary as it will auto use latest

instrumentation/httpx/Appraisals
```diff
@@ -13,10 +13,10 @@ semconv_stability = %w[dup stable old]
 
 semconv_stability.each do |mode|
   appraise "httpx-1-#{mode}" do
-    gem 'httpx', '~> 1.6'
+    gem 'httpx', '>= 1.7.4', '< 2'
```
thompson-tomo:
```suggestion
    gem 'httpx', '~ 1.7'
```

instrumentation/httpx/lib/opentelemetry/instrumentation/httpx/old/plugin.rb
thompson-tomo, line 1: These changes need to be replicated to stable & dup.

HoneyryderChuck: they were already. unless I misunderstood, can you be clear about what exactly needs to be replicated?

HoneyryderChuck: @thompson-tomo any feedback?

kaylareopelle (commented):
I think this looks good, @HoneyryderChuck. Thanks for adding the content to dup/old/stable too. One question -- do we need any new tests for the changes?

HoneyryderChuck:
@kaylareopelle thx. not really, this is essentially a rewrite of the integration using an internal httpx plugin I added specifically to share common parts across other similar tracing providers (the datadog plugin uses the same internal layer) towards diminishing the required amount of code at the "edge".

kaylareopelle (approved):
Thanks for your patience!

PR #2360, opened 2026-05-20
fix: Make logs bridge broadcast suppression fiber-local

bdewater-thatch:
We're running a patched copy of the bridge logger in our app, during code review thread safety concerns were raised about the shared @skip_otel_emit instance variable. This PR contains the changes we made to address that, we have been running in production successfully for the past couple of weeks.


thompson-tomo (commented):
If we moved to defining a logger formatter or an output device would we still need the patch?

bdewater-thatch:
Yes, let me rework this in a new PR. I think it'll be much cleaner if we assume Rails 7.1+ as the baseline and have a OTel bridge logger as a sink for the broadcast logger, and reserve a thread-safe monkey patch for older versions of Rails.

thompson-tomo:
I think assuming 7.1 or higher is fine given that is already the requirement in other gems.

PR #2361, opened 2026-05-20
feat(active_job): add spans for Continuation

bdewater-thatch:
[Rails 8.1 ](https://rubyonrails.org/2025/10/22/rails-8-1)added Active Job Continuation for interruptible and resumable jobs. This PR adds instrumentation for the events emitted when the `step` DSL is used: https://edgeguides.rubyonrails.org/active_support_instrumentation.html#active-job

### DIFF
--- a/instrumentation/active_job/lib/opentelemetry/instrumentation/active_job/handlers/step.rb
+++ b/instrumentation/active_job/lib/opentelemetry/instrumentation/active_job/handlers/step.rb
@@ -1,0 +1,63 @@
+# frozen_string_literal: true
+
+# Copyright The OpenTelemetry Authors
+#
+# SPDX-License-Identifier: Apache-2.0
+
+module OpenTelemetry
+  module Instrumentation
+    module ActiveJob
+      module Handlers
+        # Handles step.active_job to generate child spans for continuable job steps
+        class Step < Default
+          # Overrides the `Default#start_span` method to create a child span
+          # for a continuable job step
+          #
+          # @param name [String] of the Event
+          # @param id [String] of the event
+          # @param payload [Hash] containing job run information
+          # @return [Hash] with the span and generated context tokens
+          def start_span(name, _id, payload)
+            job = payload.fetch(:job)
+            job_name = job.class.name
+            step = payload.fetch(:step)
+            step_name = step.name.to_s
+
+            attributes = @mapper.call(payload).merge(
+              'messaging.active_job.step.name' => step_name,
+              'messaging.active_job.step.state' => step.resumed? ? 'resumed' : 'started'
+            )
+
+            span = tracer.start_span("#{step_name} #{job_name}", attributes: attributes)
+            token = OpenTelemetry::Context.attach(OpenTelemetry::Trace.context_with_span(span))
+
+            { span: span, ctx_token: token }
+          end
+
+          # Overrides the `Default#finish` method to record step-specific
+          # attributes before closing the span
+          #
+          # @param _name [String] of the Event (unused)
+          # @param _id [String] of the event (unused)
+          # @param payload [Hash] containing job run information
+          def finish(_name, _id, payload)
+            otel = payload.delete(:__otel)
+            span = otel&.fetch(:span)
+            token = otel&.fetch(:ctx_token)
+
+            step = payload.fetch(:step)
+            span&.set_attribute('messaging.active_job.step.result', 'interrupted') if payload[:interrupted]
+            span&.set_attribute('messaging.active_job.step.cursor', step.cursor.to_s) if step.cursor
+
+            # Continuation::Interrupt is control flow, not a real error — skip recording it
+            on_exception(payload[:error] || payload[:exception_object], span) unless payload[:interrupted]
+          rescue StandardError => e
+            OpenTelemetry.handle_error(exception: e)
+          ensure
+            finish_span(span, token)
+          end
+        end
+      end
+    end
+  end
+end
--- a/instrumentation/active_job/test/opentelemetry/instrumentation/active_job/handlers/step_test.rb
+++ b/instrumentation/active_job/test/opentelemetry/instrumentation/active_job/handlers/step_test.rb
@@ -1,0 +1,127 @@
+# frozen_string_literal: true
+
+# Copyright The OpenTelemetry Authors
+#
+# SPDX-License-Identifier: Apache-2.0
+
+require 'test_helper'
+
+require_relative '../../../../../lib/opentelemetry/instrumentation/active_job'
+
+require 'active_job/continuation/test_helper' if defined?(ActiveJob::Continuable)
+
+describe OpenTelemetry::Instrumentation::ActiveJob::Handlers::Step do
+  let(:instrumentation) { OpenTelemetry::Instrumentation::ActiveJob::Instrumentation.instance }
+  let(:config) { { propagation_style: :link, span_naming: :queue } }
+  let(:exporter) { EXPORTER }
+  let(:spans) { exporter.finished_spans }
+  let(:process_span) { spans.find { |s| s.name == 'default process' } }
+  let(:step_spans) { spans.select { |s| s.name.start_with?('first_step', 'second_step', 'process_items') } }
+
+  before do
+    skip 'Requires ActiveJob::Continuable (Rails 8.1+)' unless defined?(ActiveJob::Continuable)
+
+    OpenTelemetry::Instrumentation::ActiveJob::Handlers.unsubscribe
+    instrumentation.instance_variable_set(:@config, config)
+    instrumentation.instance_variable_set(:@installed, false)
+
+    instrumentation.install(config)
+
+    exporter.reset
+  end
+
+  describe 'basic step tracing' do
+    it 'creates a span for each step' do
+      ContinuableJob.perform_now
+
+      _(step_spans.length).must_equal 2
+    end
+
+    it 'names spans using the step name' do
+      ContinuableJob.perform_now
+
+      first_step_span = spans.find { |s| s.name == 'first_step ContinuableJob' }
+      second_step_span = spans.find { |s| s.name == 'second_step ContinuableJob' }
+
+      _(first_step_span).wont_be_nil
+      _(second_step_span).wont_be_nil
+    end
+
+    it 'creates step spans as children of the process span' do
+      ContinuableJob.perform_now
+
+      step_spans.each do |step_span|
+        _(step_span.parent_span_id).must_equal process_span.span_id
+      end
+    end
+  end
+
+  describe 'step attributes' do
+    it 'includes the step name' do
+      ContinuableJob.perform_now
+
+      first_step_span = spans.find { |s| s.name == 'first_step ContinuableJob' }
+      _(first_step_span.attributes['messaging.active_job.step.name']).must_equal 'first_step'
+    end
+
+    it 'includes state as started for first execution' do
+      ContinuableJob.perform_now
+
+      first_step_span = spans.find { |s| s.name == 'first_step ContinuableJob' }
+      _(first_step_span.attributes['messaging.active_job.step.state']).must_equal 'started'
+    end
+
+    it 'includes standard messaging attributes' do
+      ContinuableJob.perform_now
+
+      first_step_span = spans.find { |s| s.name == 'first_step ContinuableJob' }
+      _(first_step_span.attributes['code.namespace']).must_equal 'ContinuableJob'
+      _(first_step_span.attributes['messaging.system']).must_equal 'active_job'
+      _(first_step_span.attributes['messaging.destination']).must_equal 'default'
+    end
+
+    it 'does not mark the step as interrupted' do
+      ContinuableJob.perform_now
+
+      first_step_span = spans.find { |s| s.name == 'first_step ContinuableJob' }
+      assert_nil(first_step_span.attributes['messaging.active_job.step.result'])
+    end
+  end
+
+  describe 'cursor tracking' do
+    it 'includes the cursor value when present' do
+      ContinuableWithCursorJob.perform_now
+
+      step_span = spans.find { |s| s.name == 'process_items ContinuableWithCursorJob' }
+      _(step_span.attributes['messaging.active_job.step.cursor']).must_equal '3'
+    end
+  end
+
+  describe 'interrupted step' do
+    before do
+      singleton_class.include ActiveJob::Continuation::TestHelper
+      ActiveJob::Base.queue_adapter = :test
+    end
+
+    after do
+      ActiveJob::Base.queue_adapter = :inline
+    end
+
+    it 'does not record an error on the step span' do
+      ContinuableWithCursorJob.perform_later
+      interrupt_job_during_step(ContinuableWithCursorJob, :process_items, cursor: 2) { perform_enqueued_jobs }
+
+      step_span = spans.find { |s| s.name == 'process_items ContinuableWithCursorJob' }
+      _(step_span.status.code).must_equal OpenTelemetry::Trace::Status::OK
+      _(step_span.events).must_be_nil
+    end
+
+    it 'marks the step as interrupted' do
+      ContinuableWithCursorJob.perform_later
+      interrupt_job_during_step(ContinuableWithCursorJob, :process_items, cursor: 2) { perform_enqueued_jobs }
+
+      step_span = spans.find { |s| s.name == 'process_items ContinuableWithCursorJob' }
+      _(step_span.attributes['messaging.active_job.step.result']).must_equal 'interrupted'
+    end
+  end
+end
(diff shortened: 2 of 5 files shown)


instrumentation/active_job/lib/opentelemetry/instrumentation/active_job/handlers/interrupt.rb
```diff
@@ -0,0 +1,38 @@
...
+          def start(_name, _id, payload)
+            span = @parent_span_provider.current_span
+            return unless span.recording?
+
+            attributes = {
+              'messaging.active_job.continuation.reason' => payload[:reason].to_s,
+              'messaging.active_job.continuation.description' => payload[:description]
+            }
+
+            span.add_event('interrupt', attributes: attributes)
```
thompson-tomo: I would prefer to avoid adding new span events given they are undergoing a deprecation process. The replacement is to use logger.

instrumentation/active_job/lib/opentelemetry/instrumentation/active_job/handlers/interrupt.rb
```diff
@@ -0,0 +1,38 @@
...
+
+          # @param _name [String] of the Event (unused)
+          # @param _id [String] of the event (unused)
+          # @param payload [Hash] containing job run information
+          def start(_name, _id, payload)
+            span = @parent_span_provider.current_span
+            return unless span.recording?
+
+            attributes = {
+              'messaging.active_job.continuation.reason' => payload[:reason].to_s,
```
thompson-tomo:
```suggestion
              'rails.active_job.continuation.reason' => payload[:reason].to_s,
```

instrumentation/active_job/lib/opentelemetry/instrumentation/active_job/handlers/interrupt.rb
```diff
@@ -0,0 +1,38 @@
...
+              'messaging.active_job.continuation.description' => payload[:description]
```
thompson-tomo:
```suggestion
              'rails.active_job.continuation.description' => payload[:description]
```

instrumentation/active_job/lib/opentelemetry/instrumentation/active_job/handlers/resume.rb
```diff
@@ -0,0 +1,40 @@
...
+          # @param payload [Hash] containing job run information
+          def start(_name, _id, payload)
+            span = @parent_span_provider.current_span
+            return unless span.recording?
+
+            job = payload.fetch(:job)
+
+            attributes = {
+              'messaging.active_job.continuation.resumptions' => job.resumptions,
+              'messaging.active_job.continuation.description' => payload[:description]
```
thompson-tomo:
```suggestion
              'rails.active_job.continuation.description' => payload[:description]
```

instrumentation/active_job/lib/opentelemetry/instrumentation/active_job/handlers/resume.rb
```diff
@@ -0,0 +1,40 @@
...
+          # @param _id [String] of the event (unused)
+          # @param payload [Hash] containing job run information
+          def start(_name, _id, payload)
+            span = @parent_span_provider.current_span
+            return unless span.recording?
+
+            job = payload.fetch(:job)
+
+            attributes = {
+              'messaging.active_job.continuation.resumptions' => job.resumptions,
```
thompson-tomo:
```suggestion
              'rails.active_job.continuation.resumptions' => job.resumptions,
```

instrumentation/active_job/lib/opentelemetry/instrumentation/active_job/handlers/step_skipped.rb
```diff
@@ -0,0 +1,35 @@
...
+
+          # @param _name [String] of the Event (unused)
+          # @param _id [String] of the event (unused)
+          # @param payload [Hash] containing job run information
+          def start(_name, _id, payload)
+            span = @parent_span_provider.current_span
+            return unless span.recording?
+
+            span.add_event('step_skipped', attributes: {
+                             'messaging.active_job.step.name' => payload[:step].to_s
```
thompson-tomo:
```suggestion
                             'rails.active_job.step.name' => payload[:step].to_s
```

instrumentation/active_job/lib/opentelemetry/instrumentation/active_job/handlers/step.rb
```diff
@@ -0,0 +1,63 @@
...
+          # @param id [String] of the event
+          # @param payload [Hash] containing job run information
+          # @return [Hash] with the span and generated context tokens
+          def start_span(name, _id, payload)
+            job = payload.fetch(:job)
+            step = payload.fetch(:step)
+            step_name = step.name.to_s
+
+            attributes = @mapper.call(payload).merge(
+              'messaging.active_job.step.name' => step_name,
```
thompson-tomo, line 27:
```suggestion
              'rails.active_job.step.name' => step_name,
```

instrumentation/active_job/lib/opentelemetry/instrumentation/active_job/handlers/step.rb
```diff
@@ -0,0 +1,63 @@
...
+              'messaging.active_job.step.resumed' => step.resumed?
```
thompson-tomo:
```suggestion
              'rails.active_job.step.resumed' => step.resumed?
```

Perhaps even

```suggestion
              'rails.active_job.step.state' => step.resumed? 'resumed' : 'started'
```

Ideally we would either use `rails.active_job.continuation.resumptions` so that it can be connected or drop the attribute and use just the event. I think the first option would be better.

instrumentation/active_job/lib/opentelemetry/instrumentation/active_job/handlers/step.rb
```diff
@@ -0,0 +1,63 @@
...
+          # @param _name [String] of the Event (unused)
+          # @param _id [String] of the event (unused)
+          # @param payload [Hash] containing job run information
+          def finish(_name, _id, payload)
+            otel = payload.delete(:__otel)
+            span = otel&.fetch(:span)
+            token = otel&.fetch(:ctx_token)
+
+            step = payload.fetch(:step)
+            span&.set_attribute('messaging.active_job.step.interrupted', payload[:interrupted]) if payload.key?(:interrupted)
```
thompson-tomo:
```suggestion
            span&.set_attribute('rails.active_job.step.state', 'interrupted') if payload.key?(:interrupted)
```

Or

```suggestion
            span&.set_attribute('rails.active_job.step.result', 'interrupted') if payload.key?(:interrupted)
```


That way we are avoiding boolean & we could even have a completed state.

instrumentation/active_job/lib/opentelemetry/instrumentation/active_job/handlers/step.rb
```diff
@@ -0,0 +1,63 @@
...
+            span&.set_attribute('messaging.active_job.step.cursor', step.cursor.to_s) if step.cursor
```
thompson-tomo, line 50:
```suggestion
            span&.set_attribute('rails.active_job.step.cursor', step.cursor.to_s) if step.cursor
```

instrumentation/active_job/lib/opentelemetry/instrumentation/active_job/handlers/interrupt.rb
bdewater-thatch: I don't feel strongly about it but https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/404b2a5b51b6fd805bd49679971a30e02794e98b/instrumentation/active_job/lib/opentelemetry/instrumentation/active_job/mappers/attribute.rb#L30-L33 uses `messaging.active_job`?

thompson-tomo: Am aware that a couple of old attributes use the messaging namespace when active job is actually task/Workflow co-ordination system. We should decide if we can switch them or need to have it configurable.

Note active_job wouldn't be accepted in semconv as a messaging system, I couldn't even get mqtt Brooker or signalr accepted.

Ideally we would have a generic Workflow definition but that requires sig to be formed etc.

bdewater-thatch: Gotcha, thanks for that context!

instrumentation/active_job/lib/opentelemetry/instrumentation/active_job/handlers/step.rb
bdewater-thatch: Not entirely following this one - are you suggesting to keep `state` and add `resumptions`, or replace the former by the latter? Resumptions are counted at the job level, not the step level.

thompson-tomo (dismissed):
Telemetry definition LGTM, thanks for making the changes. Will like for someone else to have another review of especially the coding patterns.

instrumentation/active_job/lib/opentelemetry/instrumentation/active_job/handlers/step.rb
bdewater-thatch: Something like `attributes['rails.active_job.resumptions'] = job.resumptions if job.respond_to?(:resumptions)` in the perform handler?

thompson-tomo: What I was thinking was rather than having a state attribute, could we have an attribute to track which resumption it is being run as a part of. That way we can differentiate between different runs.

Thinking some more, maybe let's stick with state as we can get the run info from the parent.

instrumentation/active_job/test/opentelemetry/instrumentation/active_job/handlers/step_test.rb
```diff
@@ -0,0 +1,140 @@
...
+    skip 'Requires ActiveJob::Continuable (Rails 8.1+)' unless defined?(ActiveJob::Continuable)
+
+    OpenTelemetry::Instrumentation::ActiveJob::Handlers.unsubscribe
+    instrumentation.instance_variable_set(:@config, config)
+    instrumentation.instance_variable_set(:@installed, false)
+
+    instrumentation.install(config)
+
+    exporter.reset
+  end
```
kaylareopelle, line 31: Do you have any testing that looks at how the instrumentation behaves if ActiveJob::Continuable isn't defined? (ex. does it fail/skip gracefully?)

instrumentation/active_job/test/opentelemetry/instrumentation/active_job/handlers/step_test.rb
```diff
@@ -0,0 +1,140 @@
...
+    it 'includes standard messaging attributes' do
+      ContinuableJob.perform_now
+
+      first_step_span = spans.find { |s| s.name == 'default first_step' }
+      _(first_step_span.attributes['code.namespace']).must_equal 'ContinuableJob'
+      _(first_step_span.attributes['messaging.system']).must_equal 'active_job'
+      _(first_step_span.attributes['messaging.destination']).must_equal 'default'
+    end
+
+    it 'does not marks the step as interrupted' do
```
hannahramadan:
```suggestion
    it 'does not mark the step as interrupted' do
```

hannahramadan (commented):
Thank you, Bart ◡̈ 

I'd like to revert the prefixes back to `messaging` vs `rails`. I think we should be consistent until https://github.com/open-telemetry/opentelemetry-ruby-contrib/issues/2368 is settled. If it is decided that we want the new naming, then we can make that update across the board. 

cc @thompson-tomo

thompson-tomo:
Actually @hannahramadan that is more broad of a change. Even if we were to stick with messaging the naming convention in this PR currently is what is needed as semconv doesn't prefix vendor specific attributes with the domain.

kaylareopelle:
> Actually @hannahramadan that is more broad of a change. Even if we were to stick with messaging the naming convention in this PR currently is what is needed as semconv doesn't prefix vendor specific attributes with the domain.

I agree with @hannahramadan, the README documents the current attributes across the library:
https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/b1fe9681a1ea563f409737e0a00c3bdae49876b2/instrumentation/active_job/README.md#L81-L89

These attributes may not be aligned with the current semantic conventions, but they are consistent. If we add in the `rails.*` namespace now, we're breaking from the naming convention in the rest of the attributes. Having an inconsistent prefix may make queries more complicated for our users.

I recommend we use `messaging.*` on the new attributes on this PR. If we decide in #2368 the `rails.*` prefix is the right choice, we'll move all the attributes to that naming convention in a separate PR. We could consider taking a semconv-opt-in approach and keeping both values around for a while and then drop the old attributes after a set period of time.

Since we're split, maybe we can get some other points of view to chime in. @simi, @xuan-cao-swi what do you think?

simi:
> I recommend we use messaging.* on the new attributes on this PR. If we decide in https://github.com/open-telemetry/opentelemetry-ruby-contrib/issues/2368 the rails.* prefix is the right choice, we'll move all the attributes to that naming convention in a separate PR. We could consider taking a semconv-opt-in approach and keeping both values around for a while and then drop the old attributes after a set period of time.

👍

thompson-tomo:
For reference the collector which runs into this problem all the time follow the policy of following semconv rules for new attributes & existing attributes remain untouched. This enables the reduction of future breaking changes and additional work for users. The collector then At a later point functionality is added to enable the user to opt-in to the new naming for existing attributes. Hence not sure if the inconsistency in namespace is a big issue for users as I would've expected it to be raised as an issue there.

> If we decide in #2368 the rails.* prefix is the right choice

I am not following #2368 is about changing from using the messaging domain (including span types) as it doesn't fit the expectations of messaging. Even if we stick with messaging for the time being, the active_job attributes need to come out of that namespace.

As mid ground, why don't we first add a flag use_rails_namespace which if true switches the active_job attributes from messaging to rails and new attributes are always rails.*. That way for the user who needs consistency they can get it via opt-in and we reduce future breaking changes.

instrumentation/active_job/test/opentelemetry/instrumentation/active_job/handlers/step_test.rb
bdewater-thatch: No explicit testing, but it will be handled gracefully. In any version of Rails the step handler is created and subscribes to the topic in `OpenTelemetry::Instrumentation::ActiveJob::Handlers.subscribe`, but older versions of Rails won't emit notifications for it.
https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/e7065038f02d1dc2230112a86e7addafe057b589/instrumentation/active_job/lib/opentelemetry/instrumentation/active_job/handlers.rb#L61-L63

bdewater-thatch:
I have named the attributes back to `messaging.active_job.*`, updated the readme, and squashed the commits.

instrumentation/active_job/lib/opentelemetry/instrumentation/active_job/handlers/step.rb
```diff
@@ -0,0 +1,63 @@
...
+            job = payload.fetch(:job)
+            step = payload.fetch(:step)
+            step_name = step.name.to_s
+
+            attributes = @mapper.call(payload).merge(
+              'messaging.active_job.step.name' => step_name,
+              'messaging.active_job.step.state' => step.resumed? ? 'resumed' : 'started'
+            )
+
+            span_name = span_name(job, step_name)
```
bdewater-thatch: I wanted to flag this as a last question for the group: does this still make sense when it's a child span? We use SLA-based queues, so by default this would generate a name like `within_5_minutes step_one`. Steps always happen within a job so to me it feels like the title should be `FooJob step_one`.

kaylareopelle: Good call-out. `FooJob step_one` seems more appropriate to me too for a step. 

Looking [at the SemConv](https://opentelemetry.io/docs/specs/semconv/messaging/messaging-spans/#span-name), span names are `{messaging.operation.name} {destination}`. That might shake out to something more like `step_one FooJob`, but I don't think that reads as well.

kaylareopelle:
@thompson-tomo - I hear what you're saying with the collector, I just don't think that's what Ruby users expect. If anyone is viewing this conversation and thinks otherwise, I'd love to hear about their experience.

We haven't solidified what we want new conventions to be for this library, or even if we want to change them. Until #2368 is decided, I do not want to change the attribute naming strategy for Active Job. I also don't want that discussion to hold up support for Active Job Continuations. Let's move the rest of our discussion about attribute naming for this instrumentation to that issue.

instrumentation/active_job/lib/opentelemetry/instrumentation/active_job/handlers/step.rb
bdewater-thatch: Thanks for the link, and actually I do like that: I'll probably be viewing a job's span first and seeing the step names on child spans first is more valuable IMO, especially if they're short in duration and there's not a lot of screen space. I've changed it to `step_one FooJob`.

kaylareopelle (approved):

Issue #969, opened 2024-05-08
Improve config documentation for `disallowed_notification_payload_keys` and `notification_payload_transform`


kaylareopelle:
The new Action Mailer instrumentation (https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/887) and existing instrumentation for Action View include configuration options named `disallowed_notification_payload_keys` and `notification_payload_transform`. They may also be present in other Rails-related instrumentation. 

These options don't have a lot of documentation. 

We should consider:
* Updating the README to describe these options
* Updating the class comments in the file that contains the `options` method call to include descriptions

Try to apply the documentation updates to all instrumentation with these configuration options.

PR #2363, opened 2026-05-21
docs: document payload-key and payload-transform options in action_mailer and active_storage READMEs

SAY-5:
Adds a Configuration Options section to the action_mailer and active_storage READMEs covering disallowed_notification_payload_keys, disallowed_process_payload_keys, notification_payload_transform, process_payload_transform, email_address, key, and url. Mirrors the format already used by the action_view README.

Calls out that user-supplied transform procs receive the payload after the built-in ECS-style transform (action_mailer) and the active_storage. prefix transform (active_storage) have been applied, which is the part the existing class comments leave implicit.

Refs #969

### DIFF
--- a/instrumentation/action_mailer/README.md
+++ b/instrumentation/action_mailer/README.md
@@ -36,6 +36,35 @@ OpenTelemetry::SDK.configure do |c|
 end
 ```
 
+## Configuration Options
+
+The instrumentation supports the following configuration options:
+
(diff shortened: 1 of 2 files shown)


xuan-cao-swi (approved):

kaylareopelle (approved):
This is great, @SAY-5! Thanks for picking up the issue. I like the extra detail you added to the configuration descriptions.

PR #2370, opened 2026-05-24
fix(bunny): use consumer channel attribute instead of queue.channel

kivanio:
## What this changes

`Patches::Consumer#call` currently dereferences `queue.channel` to get the
`Bunny::Channel` to pass into `with_process_span`. That works when the
consumer was registered through `Bunny::Queue#subscribe` (where `queue` is a
`Bunny::Queue`), but fails when the consumer is registered via
`Bunny::Channel#basic_consume(queue_name, ...)` — where `queue` is a plain
`String`.

This is exactly how [Hutch](https://github.com/gocardless/hutch) and similar
Bunny wrappers subscribe, and `Bunny::Consumer#initialize` itself documents
both types as valid:

```ruby
# @param [Bunny::Queue,String] queue Queue messages will be consumed from
def initialize(channel, queue, ...)
  @queue = queue || raise(ArgumentError, "queue is nil")
  ...
end
```

When a String-queue consumer received a message, the patch raised:

```
NoMethodError: undefined method 'channel' for an instance of String
  @ opentelemetry-instrumentation-bunny-0.25.0/lib/opentelemetry/
      instrumentation/bunny/patches/consumer.rb:14:in
      'OpenTelemetry::Instrumentation::Bunny::Patches::Consumer#call'
```

Bunny logged the exception as ``Uncaught exception from consumer #<Bunny::Consumer ...>``
and the delivery was neither processed nor ack'd. The consumer tag stayed
registered with the broker (so the channel looked alive from the broker's
side), but messages piled up indefinitely.

In production we observed 30k+ such errors in 6h on Hutch consumers,
requiring manual rolling restarts to drain the backlog every few hours.

## The fix

`Bunny::Consumer` exposes `channel` as an `attr_reader` — the same
`Bunny::Channel` instance that `queue.channel` would have returned in the
Queue-object path. Using it directly works for both cases (queue as Queue
object and queue as String) without changing semantics for existing callers.

```diff
- with_process_span(queue.channel, tracer, delivery_info, properties) do
+ with_process_span(channel,       tracer, delivery_info, properties) do
```

## Test

Added a regression test in `consumer_test.rb` that subscribes via
`channel.basic_consume(queue.name)` (String) to exercise the Hutch-style
scenario end-to-end. The test passes after the fix and fails before it with
the `NoMethodError` shown above.

The existing test (`queue.subscribe`-based path) continues to pass.

## Checklist

- [x] Conventional commit message (`fix:`)
- [x] Branch from fork, not from `main`
- [x] No public API changes
- [x] Regression test added for the failing scenario

🤖 Generated with [Claude Code](https://claude.com/claude-code)


instrumentation/bunny/test/opentelemetry/instrumentation/bunny/patches/consumer_test.rb
```diff
@@ -71,4 +71,31 @@
...
+
+    # Wait until the publish message reached the consumer
+    sleep 1.0
+
+    channel.basic_cancel(consumer.consumer_tag)
+
+    process_span = spans.find { |span| span.name == "#{topic}.ruby.news process" }
+    _(process_span).wont_be_nil
+    _(process_span.kind).must_equal(:consumer)
+  end
```
thompson-tomo, line 100: Can you verify that the queue name is being correctly set as an attribute for this case.

Th3-M4jor:
Any chance this fix can get merged? I just got hit by this bug and it caused issues for us too.

lairtonlelis (approved):

thompson-tomo (changes requested):
Currently, i am not comfortable with how queue name is being handled in the process span in particular the lack of ability to confirm the correct queue name is set. The current tests & functionality are such that if u pass nil in, they would pass so it fixes the error but data is still wrong.

What i would like to see is 2 fold:

- Basic attributes added to the process span using the existing helper. This captures the info about the connection.
- tests which check the attributes to confirm in both cases the queue name has been set as expected etc.

Perhaps @Th3-M4jor, @lairtonlelis or @kivanio would like to pick this up.

instrumentation/bunny/test/opentelemetry/instrumentation/bunny/patches/consumer_test.rb
kaylareopelle: @thompson-tomo, after doing a bit of digging, I don't think attributes are currently set on any process spans. The method called by the Consumer module doesn't assign any attributes: 
https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/e2463549c278ba871ae6bc315be2c388ae6ce4fc/instrumentation/bunny/lib/opentelemetry/instrumentation/bunny/patch_helpers.rb#L22-L29

We should probably fix this, because [Process spans have their own semantic conventions](https://opentelemetry.io/docs/specs/semconv/messaging/rabbitmq/#process-span). Since this PR doesn't remove any attributes with the fix, I'm leaning toward merging this PR and opening a different ticket to add attributes to process spans.

What do you think? Am I missing something?

thompson-tomo: My concern comes from how do we know the consume span is correct given currently we just check type & name. At the same time, if it is merged how useful is it really given it lacks any degree of detail.

If you would rather merge this as is and add more details later ok, but we should have atleast an issue to track adding the info.

PR #2390, opened 2026-06-04
chore: remove update registry

thompson-tomo:
Remove the update registry rake file as it is not used instead otel.io now scrapes the repo.


simi (approved):

kaylareopelle (approved):
Thanks! Came up in https://github.com/open-telemetry/opentelemetry.io/issues/9920

PR #2397, opened 2026-06-10
ci: Explicitly install apt packages to speed up grpc build

thompson-tomo:
This speeds up the grpc build by explicitly installing the apt packages before build.

prior to change ruby 4.0 build was taking over 16mins


thompson-tomo:
Installing the apt packages has not had the desired affect, hence closing as out of ideas to improve performance of the check.

PR #2405, opened 2026-06-16
chore: Update renovate schedule

kaylareopelle:
We don't have any hard dependencies on our libraries, so I would like to turn down the frequency of renovate pull requests.

Every four weeks on Monday:
* patch versions
* digests
* ruby 
* minor updates

Every two weeks on Monday:
* major version updates

Every Monday:
* actions/setup-node, ruby/setup-ruby (to make sure we're testing against the latest Ruby versions)

Every Tuesday:
* lock file maintenance

### DIFF
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -12,18 +12,18 @@
       groupName: "all patch versions",
       matchDepNames: ["!ruby"],
       matchUpdateTypes: ["patch"],
-      schedule: ["before 6am every weekday"],
+      schedule: ["before 6am every 4 weeks on Monday"],
     },
     {
       groupName: "all digest versions",
       matchDepNames: ["!ruby"],
       matchUpdateTypes: ["digest"],
-      schedule: ["before 6am every weekday"],
+      schedule: ["before 6am every 4 weeks on Monday"],
     },
     {
       matchDepNames: ["ruby"],
       matchUpdateTypes: ["digest", "patch"],
-      schedule: ["before 6am every weekday"],
+      schedule: ["before 6am every 4 weeks on Monday"],
     },
     {
       description: "Separate ruby updates so that patches can be installed after new minor has been released.",
@@ -57,11 +57,11 @@
     },
     {
       matchUpdateTypes: ["minor"],
-      schedule: ["before 7am on Monday"],
+      schedule: ["before 7am every 4 weeks on Monday"],
     },
     {
       matchUpdateTypes: ["major"],
-      schedule: ["before 8am on Monday"],
+      schedule: ["before 8am every 2 weeks on Monday"],
     },
     {
       matchDepNames: ["actions/setup-node", "ruby/setup-ruby"],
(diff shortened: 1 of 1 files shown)


simi (approved):

.github/renovate.json5
```diff
@@ -12,18 +12,18 @@
...
       matchDepNames: ["!ruby"],
       matchUpdateTypes: ["digest"],
-      schedule: ["before 6am every weekday"],
+      schedule: ["before 6am every 4 weeks on Monday"],
     },
     {
       matchDepNames: ["ruby"],
       matchUpdateTypes: ["digest", "patch"],
-      schedule: ["before 6am every weekday"],
+      schedule: ["before 6am every 4 weeks on Monday"],
```
thompson-tomo, line 26: This doesn't match the description. With this we can be waiting 4 weeks to test the latest ruby.

.github/renovate.json5
```diff
@@ -12,18 +12,18 @@
       groupName: "all patch versions",
       matchDepNames: ["!ruby"],
       matchUpdateTypes: ["patch"],
-      schedule: ["before 6am every weekday"],
+      schedule: ["before 6am every 4 weeks on Monday"],
```
thompson-tomo, line 15: 4 weeks is a long time to wait for patches.

.github/renovate.json5
```diff
@@ -57,15 +57,15 @@
...
     },
     {
       matchUpdateTypes: ["major"],
-      schedule: ["before 8am on Monday"],
+      schedule: ["before 8am every 2 weeks on Monday"],
     },
     {
       matchDepNames: ["actions/setup-node", "ruby/setup-ruby"],
-      schedule: ["before 6am every weekday"],
+      schedule: ["before 6am on Monday"],
```
thompson-tomo, line 68: This is every weekday so that we can always be testing the latest ruby.

thompson-tomo (commented):
I don't see this change having a noticeable affect on the qty of pr's we get. In fact I can see it having the opposite effect where in which we can now be waiting months to get updates.

To decrease the qty what I would suggest rather than changing the frequency, we instead increase update Grouping. For instance have a minor ruby group rather than individual prs & do the same for ci updates. That will bring the qty down without introducing a potential bottle neck causing lengthy delays.

We could also merge digest into patch group to decrease further.

xuan-cao-swi (approved):

mwear (approved):

thompson-tomo:
@kaylareopelle please take a look at https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/2444 as i have made changes to the groups which should deliver a noticeable reduction in qty without needing to delay the updates much longer than they are now. It also adds special handling for renovate due to that being a noisy dependency.

kaylareopelle:
Closing this. It looks like there are more precise options for Renovate we can explore first. The TC/GC are also looking into automating some renovate merges: https://github.com/open-telemetry/community/issues/3635

PR #2407, opened 2026-06-17
fix(grape): detect code namespace for API instance endpoints

bdewater-thatch:
Direct `Grape::API::Instance` subclasses expose `endpoint.options[:for]` as the endpoint class itself rather than a generated mount instance with `@base` set. The previous instrumentation only read `@base`, which produced nil and caused the SDK to log invalid `code.namespace` attribute value errors.

We're subclassing `Grape::API::Instance` directly because we're using Sorbet in our Rails app, with our `grape_sorbet` gem for the signatures. The reason it works this way is documented here: https://github.com/thatch-health/grape_sorbet#subclassing-from-grapeapiinstance-instead-of-grapeapi


simi (approved):

kaylareopelle (approved):
Thank you, @bdewater-thatch!

PR #2410, opened 2026-06-19
feat(pg): add returned and affected rows attributes

bdewater-thatch:
Postgres counterpart of the MySQL Trilogy adapter: https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/1290

A semconv proposal for affected_rows is open at https://github.com/open-telemetry/semantic-conventions/issues/3094 (and already added for Azure Cosmos specifically). I've added it as a separate commit so it can be dropped if the maintainers prefer to wait for that go through the process, but IMO it's very useful to have. I'm using the `AFFECTED_ROWS_COMMANDS` array because cmd_tuples [always returns an integer](https://github.com/ged/ruby-pg/blob/633e16f9cde849976d6a235767423d19186a6b28/ext/pg_result.c#L1102-L1128), also for commands outside of the proposal's definition, and defaults to 0. An alternative would be to pass through the value from the `pg` gem unaltered, like [Active Record does](https://github.com/rails/rails/blob/fa8f0812160665bff083a089d2bb2fc1817ea03e/activerecord/lib/active_record/connection_adapters/postgresql/database_statements.rb#L174-L175).

### DIFF
--- a/instrumentation/pg/lib/opentelemetry/instrumentation/pg/patches/connection.rb
+++ b/instrumentation/pg/lib/opentelemetry/instrumentation/pg/patches/connection.rb
@@ -79,7 +79,7 @@ module OpenTelemetry
           PG::Constants::EXEC_ISH_METHODS.each do |method|
             define_method method do |*args, &block|
               span_name, attrs = span_attrs(:query, *args)
-              tracer.in_span(span_name, attributes: attrs, kind: :client) do |_span, context|
+              tracer.in_span(span_name, attributes: attrs, kind: :client) do |span, context|
                 # Inject propagator context into SQL if propagator is configured
                 if propagator && args[0].is_a?(String)
                   sql = args[0]
@@ -93,10 +93,14 @@ module OpenTelemetry
                   end
                 end
 
+                result = super(*args)
+                response_attrs = db_response_attributes(result, attrs)
+                span.add_attributes(response_attrs) unless response_attrs.empty?
+
                 if block
-                  block.call(super(*args))
+                  block.call(result)
                 else
-                  super(*args)
+                  result
                 end
               end
             end
@@ -105,7 +109,7 @@ module OpenTelemetry
           PG::Constants::PREPARE_ISH_METHODS.each do |method|
             define_method method do |*args|
               span_name, attrs = span_attrs(:prepare, *args)
-              tracer.in_span(span_name, attributes: attrs, kind: :client) do |_span, context|
+              tracer.in_span(span_name, attributes: attrs, kind: :client) do |span, context|
                 # Inject propagator context into SQL if propagator is configured
                 # For prepare, the SQL is in args[1]
                 if propagator && args[1].is_a?(String)
@@ -120,7 +124,10 @@ module OpenTelemetry
                   end
                 end
 
-                super(*args)
+                result = super(*args)
+                response_attrs = db_response_attributes(result, attrs)
+                span.add_attributes(response_attrs) unless response_attrs.empty?
+                result
               end
             end
           end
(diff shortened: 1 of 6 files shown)


instrumentation/pg/lib/opentelemetry/instrumentation/pg/patches/connection.rb
```diff
@@ -158,6 +171,31 @@ def config
...
+            OpenTelemetry.handle_error(message: 'Error setting DB response attributes', exception: e)
+            {}
+          end
+
+          def db_response_returned_rows?(attrs)
+            return false unless config[:db_response_returned_rows]
+            return false if attrs.key?('db.response.returned_rows')
+
+            true
+          end
```
xuan-cao-swi, line 189: nit: could be more concise with `config[:db_response_returned_rows] && !attrs.key?('db.response.returned_rows')` (but it's up to your design choice).

curious: why do you anticipate the attrs will already include the `db.response.returned_rows` (e.g. from other propagators?)

bdewater-thatch: It's to allow setting it via `PG.with_attributes` and have it not get overwritten, following this:
https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/a407830531cd06da5ee6fc5ff5129c1751ded938/instrumentation/pg/README.md?plain=1#L33-L42

hannahramadan (approved):
Recognizing that we are mixing semconv and a bit early on `db.response.affected_rows`, I'm still good with this given both new attributes are opt-in. I think the opt-in, experimental nature gives us more flexibility and these attributes seem useful for users.

instrumentation/pg/lib/opentelemetry/instrumentation/pg/patches/connection.rb
```diff
@@ -158,6 +171,31 @@ def config
...
+
+          def db_response_affected_rows?(result, attrs)
+            return false unless config[:db_response_affected_rows]
+            return false if attrs.key?('db.response.affected_rows')
+            return false unless AFFECTED_ROWS_COMMANDS.include?(result.cmd_status.to_s.split.first)
```
kaylareopelle, line 194: any nil guards we should be mindful of here?

instrumentation/pg/test/opentelemetry/instrumentation/pg/instrumentation_test.rb
```diff
@@ -214,6 +218,118 @@
...
+      create_temp_table
+      client.exec('INSERT INTO otel_affected_rows (id) VALUES (1), (2), (3)')
+
+      _(last_span.attributes['db.response.affected_rows']).must_be_nil
+    end
+
+    describe 'when db_response_returned_rows is enabled' do
+      let(:config) { { db_statement: :include, db_response_returned_rows: true } }
+
+      %i[exec query sync_exec async_exec].each do |method|
```
kaylareopelle: Since we're starting to use this array in more tests, I wonder if it makes sense to add a constant to the test helper? I'm concerned that if we need to add a method in the future, we'll lose test coverage.

kaylareopelle (commented):
@bdewater-thatch this looks great! two small questions

instrumentation/pg/lib/opentelemetry/instrumentation/pg/patches/connection.rb
bdewater-thatch: According [to the documentation](https://github.com/ged/ruby-pg/blob/4323a8927f675af2040e222ccc336e33bb89c30f/ext/pg_result.c#L1087-L1100) it always returns a String, but in case `cmd_status = nil` the `to_s` makes this return `false` and not blow up.

instrumentation/pg/test/opentelemetry/instrumentation/pg/instrumentation_test.rb
bdewater-thatch: Done!

thompson-tomo (approved):

kaylareopelle (approved):

PR #2421, opened 2026-06-29
chore: Address inability to test on jruby

thompson-tomo:
This removes some of the jruby exclusions by adjusting the appraisals to use different adapters etc.


instrumentation/active_record/Appraisals
```diff
@@ -4,21 +4,32 @@
...
   end
 end
 
-appraise 'activerecord-latest' do
-  gem 'sqlite3', '>= 2.1'
-  gem 'activerecord'
+if RUBY_ENGINE == 'jruby'
+  %w[80.0.pre1].each do |version|
+    appraise "jruby_activerecord-#{version}" do
+      gem 'activerecord-jdbcsqlite3-adapter', "~> #{version}"
```
kaylareopelle, line 32: What do you think about adding the equivalent of `activerecord-latest` for the JRuby tests? (And any other appraisals with a new `RUBY_ENGINE == 'jruby'` condition).

I don't believe Appraisals are updated by Renovate, so we'll have to remember to increase the tested version without a `latest` block.

instrumentation/net_ldap/lib/opentelemetry/instrumentation/net/ldap/instrumentation_service.rb
```diff
@@ -32,7 +32,8 @@ def instrument(event, payload)
               'ldap.auth.username' => auth[:username].to_s,
               'ldap.operation.type' => operation_type,
               'ldap.request.message' => begin
-                payload.to_json
+                raw = payload.to_json
+                utf8_clean?(raw) ? raw : nil
```
kaylareopelle: Since this is actually changing the code of the library and not just the test, can we break it into a separate PR so that it's released as expected?

kaylareopelle (changes requested):
Thanks for opening this, @thompson-tomo!

instrumentation/net_ldap/lib/opentelemetry/instrumentation/net/ldap/instrumentation_service.rb
thompson-tomo: Done. Note this only changes the behaviour on jruby.

instrumentation/active_record/Appraisals
thompson-tomo: Done.

thompson-tomo:
No worries @kaylareopelle, i have made the requested changes and raised seperate pr #2428 for net-ldap & for gems not compatible I have #2429 which limits the gem.

simi (approved):

kaylareopelle (approved):

PR #2428, opened 2026-07-03
fix: ensure valid json message attribute value

thompson-tomo:
This ensures that the message attribute is consistently encoded across jruby and ruby.

This is achieved by adding a value check for the message value which is automatically done on ruby but not jruby.

### DIFF
--- a/instrumentation/net_ldap/lib/opentelemetry/instrumentation/net/ldap/instrumentation_service.rb
+++ b/instrumentation/net_ldap/lib/opentelemetry/instrumentation/net/ldap/instrumentation_service.rb
@@ -32,7 +32,8 @@ module OpenTelemetry
               'ldap.auth.username' => auth[:username].to_s,
               'ldap.operation.type' => operation_type,
               'ldap.request.message' => begin
-                payload.to_json
+                raw = payload.to_json
+                utf8_clean?(raw) ? raw : nil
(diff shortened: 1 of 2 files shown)


instrumentation/net_ldap/lib/opentelemetry/instrumentation/net/ldap/instrumentation_service.rb
```diff
@@ -99,6 +100,19 @@ def annotate_span_with_response(span, response)
...
+          def utf8_clean?(value)
+            case value
+            when String
+              value.scrub == value
+            when Hash
+              value.all? { |k, v| utf8_clean?(k) && utf8_clean?(v) }
+            when Array
+              value.all? { |v| utf8_clean?(v) }
+            else
+              true
```
scbjans: Shouldn't we set this to `false`, and drop anything which is not a `String`?

thompson-tomo: Hmm, in this scenario, yes it can probs be simplified as the additional scenarios were from when i was doing it in a different location.

scbjans (approved):
LGTM

simi (approved):

kaylareopelle (approved):
Thank you!

PR #2446, opened 2026-07-10
feat: Extend rake span representation

0robustus1:
During our adoption of OpenTelemetry we noticed that the Span Names weren't exactly useful as primary filters for Rake Tasks, due to them only including `rake.execute` or `rake.invoke` respectively. When generating Span Metrics from them we cannot easily distinguish between them without extracting the `rake.task` attribute.

This PR does the following:

* Adds a `rake.execution.type` attribute, which is either `execute` or `invoke` depending on the span
* Adds a `span_name` option to the instrumentation to enable consumers to switch to a `execution_type_and_task_name` span name.

### DIFF
--- a/instrumentation/rake/lib/opentelemetry/instrumentation/rake/instrumentation.rb
+++ b/instrumentation/rake/lib/opentelemetry/instrumentation/rake/instrumentation.rb
@@ -24,6 +24,8 @@ module OpenTelemetry
           gem_version >= MINIMUM_VERSION
         end
 
+        option :span_name, default: :execution_type_only, validate: %I[execution_type_only execution_type_and_task_name]
+
         private
 
         def gem_version
--- a/instrumentation/rake/lib/opentelemetry/instrumentation/rake/patches/task.rb
+++ b/instrumentation/rake/lib/opentelemetry/instrumentation/rake/patches/task.rb
@@ -7,7 +7,8 @@ module OpenTelemetry
         # Module to prepend to Rask::Task for instrumentation
         module Task
           def invoke(*args)
-            tracer.in_span('rake.invoke', attributes: { 'rake.task' => name, 'rake.execution.type' => 'invoke' }) do
+            span_name = rake_span_name('invoke', name)
+            tracer.in_span(span_name, attributes: { 'rake.task' => name, 'rake.execution.type' => 'invoke' }) do
               super
             end
           ensure
@@ -15,7 +16,8 @@ module OpenTelemetry
           end
 
           def execute(args = nil)
-            tracer.in_span('rake.execute', attributes: { 'rake.task' => name, 'rake.execution.type' => 'execute' }) do
+            span_name = rake_span_name('execute', name)
+            tracer.in_span(span_name, attributes: { 'rake.task' => name, 'rake.execution.type' => 'execute' }) do
               super
             end
           ensure
@@ -26,6 +28,18 @@ module OpenTelemetry
 
           def tracer
             Rake::Instrumentation.instance.tracer
+          end
+
+          def config
+            OpenTelemetry::Instrumentation::Rake::Instrumentation.instance.config
+          end
+
+          def rake_span_name(execution_type, task_name)
+            if config[:span_name] == :execution_type_and_task_name
+              "rake.#{execution_type} #{task_name}"
+            else
+              "rake.#{execution_type}"
+            end
           end
 
           def force_flush
--- a/instrumentation/rake/lib/opentelemetry/instrumentation/rake/patches/task.rb
+++ b/instrumentation/rake/lib/opentelemetry/instrumentation/rake/patches/task.rb
@@ -7,7 +7,7 @@ module OpenTelemetry
         # Module to prepend to Rask::Task for instrumentation
         module Task
           def invoke(*args)
-            tracer.in_span('rake.invoke', attributes: { 'rake.task' => name }) do
+            tracer.in_span('rake.invoke', attributes: { 'rake.task' => name, 'rake.execution.type' => 'invoke' }) do
               super
             end
           ensure
@@ -15,7 +15,7 @@ module OpenTelemetry
           end
 
           def execute(args = nil)
-            tracer.in_span('rake.execute', attributes: { 'rake.task' => name }) do
+            tracer.in_span('rake.execute', attributes: { 'rake.task' => name, 'rake.execution.type' => 'execute' }) do
               super
             end
           ensure
--- a/instrumentation/rake/test/opentelemetry/instrumentation/rake/patches/task_test.rb
+++ b/instrumentation/rake/test/opentelemetry/instrumentation/rake/patches/task_test.rb
@@ -17,12 +17,16 @@ describe OpenTelemetry::Instrumentation::Rake::Patches::Task do
   let(:task_name) { :test_rake_instrumentation }
   let(:task) { Rake::Task[task_name] }
 
-  let(:invoke_span) { spans.find { |s| s.name == 'rake.invoke' } }
-  let(:execute_span) { spans.find { |s| s.name == 'rake.execute' && s.attributes['rake.task'] == task_name.to_s } }
+  let(:invoke_span) { spans.find { |s| s.name.start_with?('rake.invoke') } }
+  let(:execute_span) { spans.find { |s| s.name.start_with?('rake.execute') && s.attributes['rake.task'] == task_name.to_s } }
+  let(:rake_span) { spans.find { |s| s.attributes.key?('rake.task') } }
+  let(:config) { {} }
 
   before do
     exporter.reset
-    instrumentation.install
+    # simulate a fresh install:
+    instrumentation.instance_variable_set(:@installed, false)
+    instrumentation.install(config)
 
     Rake::Task.define_task(task_name)
     task.reenable
@@ -37,6 +41,22 @@ describe OpenTelemetry::Instrumentation::Rake::Patches::Task do
       _(execute_span.kind).must_equal :internal
       _(execute_span.name).must_equal 'rake.execute'
       _(execute_span.attributes['rake.task']).must_equal 'test_rake_instrumentation'
+      _(execute_span.attributes['rake.execution.type']).must_equal 'execute'
+    end
+
+    describe 'with execution_type_and_task_name span_name' do
+      let(:config) { { span_name: :execution_type_and_task_name } }
+
+      it 'creates a properly named span' do
+        task.execute
+
+        _(spans.size).must_equal 1
+
+        _(rake_span.kind).must_equal :internal
+        _(rake_span.name).must_equal 'rake.execute test_rake_instrumentation'
+        _(rake_span.attributes['rake.task']).must_equal 'test_rake_instrumentation'
+        _(rake_span.attributes['rake.execution.type']).must_equal 'execute'
+      end
     end
   end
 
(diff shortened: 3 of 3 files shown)


instrumentation/rake/lib/opentelemetry/instrumentation/rake/instrumentation.rb
```diff
@@ -24,6 +24,8 @@ class Instrumentation < OpenTelemetry::Instrumentation::Base
           gem_version >= MINIMUM_VERSION
         end
 
+        option :span_name, default: :execution_type_only, validate: %I[execution_type_only execution_type_and_task_name]
+
```
thompson-tomo, line 28: What are people's thoughts about making this a boolean option 'use_semconv'?

instrumentation/rake/lib/opentelemetry/instrumentation/rake/patches/task.rb
```diff
@@ -28,6 +31,21 @@ def tracer
...
+
+          def rake_span_name(execution_type, task_name)
+            case config[:span_name]
+            when :execution_type_only
+              "rake.#{execution_type}"
+            when :execution_type_and_task_name
+              "rake.#{execution_type} #{task_name}"
+            else
+              "rake.#{execution_type}"
+            end
```
thompson-tomo: We could shorten this to an if statement.

instrumentation/rake/lib/opentelemetry/instrumentation/rake/patches/task.rb
```diff
@@ -7,15 +7,18 @@ module Patches
...
             end
           ensure
             force_flush
           end
 
           def execute(args = nil)
-            tracer.in_span('rake.execute', attributes: { 'rake.task' => name }) do
+            task_name = name
+            span_name = rake_span_name('execute', task_name)
+            tracer.in_span(span_name, attributes: { 'rake.task' => task_name, 'rake.execution.type' => 'execute' }) do
```
thompson-tomo:
```suggestion
            span_name = rake_span_name('execute', name)
            tracer.in_span(span_name, attributes: { 'rake.task' => name, 'rake.execution.type' => 'execute' }) do
```

instrumentation/rake/lib/opentelemetry/instrumentation/rake/patches/task.rb
```diff
@@ -28,6 +31,21 @@ def tracer
...
+          def config
+            OpenTelemetry::Instrumentation::Rake::Instrumentation.instance.config
+          end
+
+          def rake_span_name(execution_type, task_name)
+            case config[:span_name]
+            when :execution_type_only
+              "rake.#{execution_type}"
+            when :execution_type_and_task_name
+              "rake.#{execution_type} #{task_name}"
```
thompson-tomo, line 39: We can simplify this to

```suggestion
              "#{execution_type} #{task_name}"
```

instrumentation/rake/lib/opentelemetry/instrumentation/rake/patches/task.rb
thompson-tomo: With the use_semconv option we can then name the attributes:

- ruby.rake.task.name
- ruby.rake.task.run.type

Or ideally and i prefer

- ruby.rake.action.name on execute spans
- ruby.rake.task.name on invoke spans


This would make it similar to the cicd workflows and avoid the unusual situation where the task name sometimes differs in the child span. If you want to filter you can do so based on attribute presence.

thompson-tomo (commented):
I like the use case but i wonder could we use the opportunity to switch semconv style and potentially define it in semconv to provide pathway to stable.

See my thoughts with key suggestion being the attribute naming.

instrumentation/rake/lib/opentelemetry/instrumentation/rake/patches/task.rb
0robustus1: We can, but then it doesn't mention rake in the Span Name anymore.
Originally I wanted to keep it as an opt-in change (without changing the current defaults for anyone else). But with suggestion of using/supporting Rake via Semantic Conventions we'd have changes in Span Names and Attribute names anyway, the question is just whether that should be opt-in initially (I would assume: yes?).

0robustus1:
Using SemConv style sounds great. I just steered away from that to not make the change unnecessarily big, but we could define the attributes and Span naming patterns in SemConv and use that here.

Happy to make the necessary changes if there is agreement.

instrumentation/rake/lib/opentelemetry/instrumentation/rake/patches/task.rb
thompson-tomo: Correct it would be opt-in for atleast 6 months after which we can change the default.

thompson-tomo:
I would normally agree but this is a very straight forward instrumentation and the proposed solution avoids adding an option which wouldn't be needed with the opt-in scenario.

In fact if you put the if around the patch loading, the existing file wouldn't need to be touched and no conditionally logic is added to the hot path.

Let's get confirmation from others before switching.

instrumentation/rake/lib/opentelemetry/instrumentation/rake/instrumentation.rb
kaylareopelle: It's an interesting idea, but I worry about promoting off-convention options. I would rather see the convention permit multiple name types and use a configuration name that's more consistent with other instrumentation. 

[Trilogy](https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/6e5784bd4347e653de214d3cf065d80df9e81f8f/instrumentation/trilogy/lib/opentelemetry/instrumentation/trilogy/instrumentation.rb#L91) and [MySQL2](https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/3a39a4481d23c0988381067f4ab6318b19ed7a85/instrumentation/mysql2/lib/opentelemetry/instrumentation/mysql2/instrumentation.rb#L25) both have the configuration option `:span_name` with a default list of allowed types. This might be a good pattern to follow here. 

There is also the option in [GraphQL](https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/bc01e67b15f9ce98c26360a7e15d7d312da723a2/instrumentation/graphql/lib/opentelemetry/instrumentation/graphql/instrumentation.rb#L73) and [ActionView](https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/64565684d8c093b9bbd7f248a2377ef169ba5cdd/instrumentation/action_view/lib/opentelemetry/instrumentation/action_view/instrumentation.rb#L62) instrumentation to have something along the lines of `:legacy_span_names`, and have a boolean config, but if we want to keep both span name options long term, this doesn't seem like the best solution. 

HTTP instrumentation goes another route and has a callable for the span names that can be set with the [option `:span_name_formatter`](https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/ee9095946a5148eb7a84f468e909bf1c68873804/instrumentation/http/lib/opentelemetry/instrumentation/http/instrumentation.rb#L22). I don't think that's a great fit for this instrumentation.

kaylareopelle:
Hi @0robustus1! Thanks for opening this, and thank you @thompson-tomo for your review.

I like the idea of crafting a semantic convention specifically for Rake. 

Though span names are more helpful when they include info like the task within them, that can also create high cardinality span names, which might not be a great fit for some environments. For the convention, I think it's worth having a span name option, so users can easily name their spans with and without the task name. 

My only hesitation is that I don't know how common it is for span names in semantic conventions to have multiple options. That might be worth discussing with the semantic conventions SIG.

I don't think we need to document the semantic conventions before pushing this change if the default name remains the same, but it seems helpful to do sometime soon.

thompson-tomo:
> My only hesitation is that I don't know how common it is for span names in semantic conventions to have multiple options.

I have never seen a scenario where it is configurable if the target is used in the name. It is common to specify the following

> The span name SHOULD be {action} {target} if there is a (low-cardinality) target available. If the target is not available or is likely to have high cardinality, then the span name SHOULD be {action}.

The only exception being http which mentions callbacks for determining the target.

0robustus1:
I'm not sure if we really should consider the inclusion of the task name as being of high cardinality. Higher than 2 (`rake.execute` and `rake.invoke`) most certainly, yes.
But Rake Tasks are generally defined explicitly in code and are therefore of limited nature, e.g. similar to Routes in a Rails App.

And most other implementations either offer an option or by default include the `<verb> <object>` pattern, HTTP is the most common example, where cardinality is intentionally reduced not just by limiting to the HTTP Verb, but by including a templated (lower cardinality) version of the URL whenever available:

> HTTP span names SHOULD be {method} {target} if there is a (low-cardinality) target available. [...]
> The {target} SHOULD be one of the following:
>
>    [http.route](https://opentelemetry.io/docs/specs/semconv/registry/attributes/http/) for HTTP Server spans
>    [url.template](https://opentelemetry.io/docs/specs/semconv/registry/attributes/url/) for HTTP Client spans if enabled and available (Development)
>    Other value MAY be provided through custom hooks at span start time or later.

Yes, as a fallback it does say to use just `{method}`. (see [HTTP Span Names](https://opentelemetry.io/docs/specs/semconv/http/http-spans/#name))

There is also Messaging:

> The span name SHOULD be {messaging.operation.name} {destination} (see below for the exact definition of the [{destination}](https://opentelemetry.io/docs/specs/semconv/messaging/messaging-spans/#destination-placeholder) placeholder).
> [...]
> The {destination} SHOULD describe the entity that the operation is performed against and SHOULD adhere to one of the following values, provided they are accessible:
>
>    messaging.destination.template SHOULD be used when it is available.
>    messaging.destination.name SHOULD be used when the destination is known to be neither [temporary nor anonymous](https://opentelemetry.io/docs/specs/semconv/messaging/messaging-spans/#temporary-and-anonymous-destinations).
>    server.address:server.port SHOULD be used only for operations not targeting any specific destination(s).

I think Rake Task names are already inherently closer in cardinality and design to templates. Now including optional arguments: that would be increasing the cardinality into a potentially immeasurable degree.

If we define this in SemConv I would propose to only define/document the `{execution_type} {task_name}` variant.

instrumentation/rake/lib/opentelemetry/instrumentation/rake/instrumentation.rb
thompson-tomo: For me i don't think that mysql2/trilogy is a good example to follow given that the semconv lacks a configuration option. In fact if we generated query summaries we wouldn't need it there as the summary would be the name.

I am struggling to think of a reason why to make it configurable which is not caused by an issue with the spec.

thompson-tomo:
I tend to also agree that it is not high cardinality. If you look at https://opentelemetry.io/docs/specs/otel/trace/api/#span i would think if we don't have the task name it would be too general as for each of the 2 span types, we would have only 1 value.

kaylareopelle:
That works for me! Thanks for sharing your thoughts, @0robustus1 & @thompson-tomo. What do we need to do to unblock this? Is it a matter of opening the semantic conventions PR?

instrumentation/rake/lib/opentelemetry/instrumentation/rake/instrumentation.rb
kaylareopelle: Do we want to support two names long-term, or just move to one standard name? @0robustus1 what are your thoughts?

thompson-tomo: See related comment https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/2446#discussion_r3559920115

0robustus1:
As we seem to agree on a way forward I'm happy to prepare a Semantic Conventions PR. Will link it here once it's created.

thompson-tomo:
Note the root namespace will need to be ruby as we are the owners of that namespace. Suggest opening it as draft first & let us know if u have any questions.

PR #2465, opened 2026-07-16
chore: Ruby patching via renovate

thompson-tomo:
This addresses the race condition with setup-ruby and ruby updates by grouping them together. The complication is avoiding buming both the min sdk used and the max used in the same pr. As such we will have setup-ruby split across 2 pr's with the pr's only auto being created once ruby is also available to update. It may result in multiple versions of setup-ruby being used but prefered over forcing both min and max ruby sdk in to the same updates.

This will reduce both qty of updates as well as in some cases the qty concurrent updates.

More changes in:

- https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/2444
- https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/2466


simi (approved):

.github/renovate.json5
```diff
@@ -21,8 +21,38 @@
...
+      groupName: "Latest-Ruby-sdk [Patches]",
+      branchTopic: "latest-ruby-sdk",
+      commitMessageTopic: "Latest-Ruby-sdk",
+      commitMessageExtra: "[Patches]",
+      matchDepNames: ["ruby/setup-ruby", "ruby"],
+      matchUpdateTypes: ["!digest"],
+      matchFileNames: [
+        ".github/workflows/fossa.yml",
+        ".github/workflows/installation-tests.yml",
+        ".github/workflows/release-*.yml",
```
kaylareopelle, line 34: Why are these the file names being matched?

thompson-tomo: It is so that we have a min ruby update group & a latest group

kaylareopelle (approved):

kaylareopelle:
Hi @thompson-tomo, I think there's a merge conflict based on the Appraisals PR. Can you take a look?

thompson-tomo:
All resolved @kaylareopelle

PR #2466, opened 2026-07-16
chore: Renovate housekeeping

thompson-tomo:
This addresses the warning on the renovate dashboard by restructing how ruby digest updates are handled. In the process it should fix the updates we have missed

This change can result in a reduction of concurrent pr's due to ruby digest handling being the same as all other docker images.

More changes in:

- https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/2444
- https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/2465


.github/renovate.json5
```diff
@@ -101,14 +99,31 @@
...
+        "/(^|/|\\.)Dockerfile$/",
+        "/(^|/)Dockerfile\\.[^/]*$/",
+      ],
+      matchStrings: [
+        "FROM\\s+(?<packageName>.+):(?<depVersion>[0-9.]+)-alpine(?<alpineVersion>[0-9.]+)@(?<currentDigest>sha256:[a-f0-9]{40,64})\\s",
+        "FROM\\s+(?<packageName>.+):(?<depVersion>[0-9.]+)-alpine(?<alpineVersion>[0-9.]+)\\s",
+      ],
+      datasourceTemplate: "docker",
+      versioningTemplate: "docker",
+      currentValueTemplate: "{{depVersion}}",
```
simi, line 123: Could you verify the new Docker regex manager is not applying a digest resolved from the short Ruby version tag? In my local Renovate dry run, the native dockerfile manager produced a digest for 3.3.12-alpine3.23, but the new regex manager produced a different digest for 3.3.12.

thompson-tomo: Yes it has been verified and that difference is why digest updates for the custom manager has been disabled.

Note the logs show that update will be a patch & not a digest.

PR #2476, opened 2026-07-23
fix: Use last in, first out Rack context stack

### LINKED ISSUE #2425: Sinatra instrumentation throws errors when using multiple Sinatra app classes
## Description of the bug

If you build a Sinatra app that uses other Sinatra apps as middleware (a documented pattern for dividing endpoints into multiple classes), the default OpenTelemetry instrumentation throws errors every time an endpoint is called.

See "Using Sinatra as Middleware" here: https://sinatrarb.com/intro.html

## Share details about your runtime

Operating system details: MacOS, Tahoe 26.5.1
RUBY_ENGINE: "ruby"
RUBY_VERSION: "3.3.4"
RUBY_DESCRIPTION: "ruby 3.3.4 (2024-07-09 revision be1089c8ec) [arm64-darwin23]"

## Share a simplified reproduction if possible

MCVE here: https://github.com/isaaclyman/opentelemetry-instrumentation-sinatra-detach-error-mcve

- Install Sinatra and OpenTelemetry
- Set up the OpenTelemetry SDK with `use_all`
- Create two classes that inherit from `Sinatra::Base`, as described in "Using Sinatra as Middleware"
- Add a basic endpoint to the first class
- Have the second class `use` the first one
- Send a request to the endpoint

You see the following in the application console:

```
W, [...] WARN -- : Calling finish on an ended Span.
E, [...] ERROR -- : OpenTelemetry error: calls to detach should match corresponding calls to attach.
```

Additionally, the generated trace has a missing parent span.

The problem disappears if you have only one class that inherits from `Sinatra::Base` and don't `use` any other classes as middleware.

It also disappears if you pass `install_rack: false` as an option to `OpenTelemetry::Instrumentation::Sinatra`, but then traces are not produced.

<sub>**Tip**: [React](https://github.blog/news-insights/product-news/add-reactions-to-pull-requests-issues-and-comments/) with 👍 to help prioritize this issue. Please use comments to provide useful context, avoiding `+1` or `me too`, to help us triage it. Learn more [in our end user docs](https://opentelemetry.io/community/end-user/issue-participation/).</sub>

kaylareopelle:
This resolves #2425, where the SDK would throw errors when using multiple, nested Sinatra app classes.

This was caused because the Rack env hash only had a single slot for context. When multiple Rack apps existed in the same env, the nested app's call to on_start would overwrite the context token, removing the context for the outer app. When #finish was eventually called the "Calling finish" and "detach should match" errors would be raised and the parent span would disappear.

Now, the context lives in an array, so the most recent context received in the env will be the first one finished.

### DIFF
--- a/instrumentation/rack/lib/opentelemetry/instrumentation/rack/middlewares/dup/event_handler.rb
+++ b/instrumentation/rack/lib/opentelemetry/instrumentation/rack/middlewares/dup/event_handler.rb
@@ -62,7 +62,7 @@ module OpenTelemetry
               span = create_span(parent_context, request)
               span_ctx = OpenTelemetry::Trace.context_with_span(span, parent_context: parent_context)
               rack_ctx = OpenTelemetry::Instrumentation::Rack.context_with_span(span, parent_context: span_ctx)
-              request.env[OTEL_TOKEN_AND_SPAN] = [OpenTelemetry::Context.attach(rack_ctx), span]
+              (request.env[OTEL_TOKEN_AND_SPAN] ||= []) << [OpenTelemetry::Context.attach(rack_ctx), span]
             rescue StandardError => e
               OpenTelemetry.handle_error(exception: e)
             end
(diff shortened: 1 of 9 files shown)


xuan-cao-swi (approved):

kaylareopelle:
This is just a bandaid on our Rack context issues. #2130 raises a different issue with fiber-based web frameworks that isn't resolved by this PR. 

I'll leave this open for now, but I think we'll eventually want to move to a new PR that resolves both this issue and #2130's by using an identifier (perhaps the request itself) to identify which context to detach. 

One way we might achieve that is here (branched off #2130): https://github.com/open-telemetry/opentelemetry-ruby-contrib/compare/main...kaylareopelle:opentelemetry-ruby-contrib:fiber-safe-context-with-multi-app-fix?expand=1

kaylareopelle:
Closing in favor of #2130 with the intention that the fix that solves both of these issues will eventually be merged into it.

PR #2493, opened 2026-08-04
feat(rack): add public_request config option

hwu1001:
From this post https://github.com/open-telemetry/opentelemetry-ruby-contrib/discussions/1985

> For some endpoints with a public ingress there may be a `traceparent` header on the incoming request, but the request's parent trace originates from somewhere that will never be ingested in the receiving system. The result for the receiving system is that when they store their trace in a backend the parent trace will never be found so the parent trace name could be ambiguous or other effects depending on the backend.
> 
> As a reference to another language SDK, Go's [otelhttp](https://pkg.go.dev/go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp) package has a [`WithPublicEndpointFn`](https://pkg.go.dev/go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp#WithPublicEndpointFn) option that has the following in the docs:
> 
> > WithPublicEndpointFn runs with every request, and allows conditionally configuring the Handler to link the span with an incoming span context. If this option is not provided or returns false, then the association is a child association instead of a link.

This change adds a new configuration option for the rack instrumentation called `public_request` that mirrors the behavior in the Go SDK above. By default, this option has the same behavior the rack instrumentation has currently, which is to create a child association to the incoming span. I called the option `public_request` instead of something like `public_endpoint` because the configuration runs for each request and a given endpoint may be used publicly and privately, but it depends on the request to determine whether it's one or the other. I didn't see a previous example of documenting an option for a `callable` so added some information that also mirrors the Go otelhttp SDK, but can update that to match what's expected from the maintainers.


instrumentation/rack/lib/opentelemetry/instrumentation/rack/instrumentation.rb
```diff
@@ -31,6 +31,14 @@ class Instrumentation < OpenTelemetry::Instrumentation::Base
...
         # This option is only valid for applications using Rack 2.0 or greater
         option :use_rack_events,          default: true, validate: :boolean
+        # public_request runs with every request, and allows conditionally configuring
+        # to link the span with an incoming span context. If this option is not provided or returns
+        # false, then the association is a child association instead of a link. The callable's signature is
+        # as follows:
+        #
+        # @param [Rack::Request::Env] The incoming HTTP request environment
+        # @return [Boolean]
+        option :public_request, default: nil, validate: :callable
```
xuan-cao-swi, line 41: Does the option have to be `callable`? From the test case, it seems only testing `true`/`false` (most time options tend to avoid use callable).

hwu1001: I was thinking that in some cases the endpoint might have both public and private ingress so based on some headers or other information from the request the configuration could determine how to create the span.

thompson-tomo (commented):
I am wondering if this is not something we should be raising at the spec level for a consistent experience.

The nicest solution for me is if we could have a remoteContextAnalyzer which is triggered when creating a span. The analyzer would decide based on context & attributes if it should be a parent, link or dropped.

This would enable all instrumentation to benefit etc.

instrumentation/rack/lib/opentelemetry/instrumentation/rack/instrumentation.rb
xuan-cao-swi: In that case, I am thinking maybe it's better to have an array that include list of allowing public_request e.g. [untraced_hosts](https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/opentelemetry-instrumentation-net_http/v0.29.0/instrumentation/net_http/lib/opentelemetry/instrumentation/net/http/instrumentation.rb#L30)

hwu1001: Not sure I follow, would the array have a list of where the request originated from or a list of endpoints to determine how to create the span?

In case it helps - to add the use case I was thinking before is that for different Rack applications they might configure what is considered a public vs. private request on the URL path, headers, or other request information so having the `callable` seemed the most flexible way to meet the various ways the public/private distinction could come up. Is the concern with the `callable` that it's more difficult to configure?

xuan-cao-swi: Yes, that's the only concern since callable gives too much freedom.
Although, it doesn't mean it can't be callable.
Could you add more test case where shows the callable is necessary.

hwu1001: > Yes, that's the only concern since callable gives too much freedom.

Gotcha, I was trying to think of another way to handle this case, but since it can depend on a lot of different parts of the request I'm not sure there's another option in the instrumentation itself. It could be something that users of the Rack instrumentation are meant to handle with their own middleware though - like something that runs before OTel's Rack middleware and removes the `traceparent` / `tracestate` headers in the Rack request environment so that it will or won't create a new root span and then they store the remote context in some custom field in the Rack request environment so they can do the linking themselves. I'm not sure that simplifies the overall process, but would move it out of OTel having to handle it.

>  Could you add more test case where shows the callable is necessary.

I added some tests showing additional logic on headers/path in [d8a4bba](https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/2493/commits/d8a4bbabf8c9c164c5b5563b333449f6eab58420) - how does that look?

hwu1001:
@thompson-tomo Sorry for the slow response. I was thinking about something general for the spec that could work for a remote context analyzer, but I'm not sure how it would work in practice. The common cases I can think of for incoming remote spans would be:
1. Incoming HTTP requests
2. Incoming gRPC requests
3. Messaging consumers (Kafka, AWS SQS, RabbitMQ, etc.)
4. Cloud/event triggers (server-less lambda type operations)

When thinking through these I think one difficulty is per the semantic conventions most of the attributes we'd probably use are in different namespaces like `http`, `grpc`, etc. so the logic in the remote context analyzer would centralize how the process works for span creation, but I'm not sure it would be much different than having the logic in each instrumentation package. I can't think of a way to get around having to example attributes like [`http.request.header.<key>`](https://opentelemetry.io/docs/specs/semconv/registry/attributes/http/#http-request-header) or [`http.route`](https://opentelemetry.io/docs/specs/semconv/registry/attributes/http/#http-route) in the incoming HTTP request scenario because it's usually deployment specific for the application adding instrumentation. 

For example, if there's an istio sidecar sitting in front of an application HTTP server that has OTel instrumentation the istio sidecar might adjust some of the incoming request headers or other request metadata which is how the application server knows how to attach remote contexts for spans or create new roots. Since that part is always left up to the application, I'm not sure how to work with that best in the OTel instrumentation for handling the logic.

In the case of [messaging consumer spans specifically the OTel docs](https://opentelemetry.io/docs/specs/semconv/messaging/messaging-spans/#consumer-spans) recommend the following:

> A single “Process” or “Receive” span can account for a single message, for a batch of messages, or for no message at all (if it is signalled that no messages were received). For each message it accounts for, the “Process” or “Receive” span SHOULD link to the message’s creation context.

So I'm not sure how much a remote context analyzer in the spec would help in that scenario. Did you have any specific scenarios in mind on how the spec would handle different contexts (http, grpc, etc) with the different attributes?

instrumentation/rack/lib/opentelemetry/instrumentation/rack/instrumentation.rb
xuan-cao-swi: Thanks for the new test case. They look good.

xuan-cao-swi (approved):
LGTM (fyi [Links between spans](https://github.com/open-telemetry/opentelemetry-specification/blob/041241bf911cf0d74fa8e7249f57f27323760a62/specification/overview.md#links-between-spans))

thompson-tomo:
> When thinking through these I think one difficulty is per the semantic conventions most of the attributes we'd probably use are in different namespaces like http, grpc, 

I don't think that is an issue, as we have solutions for logs & samplers that works regardless of domain.

> but I'm not sure it would be much different than having the logic in each instrumentation package.

The solution should be quite different, the current approach is that each instrumentation must build and maintain their own solution. This results in alot more maintenance and functionality inconsistencies.

> In the case of messaging consumer spans specifically the OTel docs recommend the following:
> 
> > A single “Process” or “Receive” span can account for a single message, for a batch of messages, or for no message at all (if it is signalled that no messages were received). For each message it accounts for, the “Process” or “Receive” span SHOULD link to the message’s creation context.
> 
> So I'm not sure how much a remote context analyzer in the spec would help in that scenario.

I have no issue with that & in fact that example doesn't need anything as the context should always be linked & never parent-child.

> Did you have any specific scenarios in mind on how the spec would handle different contexts (http, grpc, etc) with the different attributes?

Yes the nicest solution i can think of is to do it similar to samplers. What i mean by this is when the sdk calls receives a call to create a span, if it contains a context; the context, attributes, span name & instrumentation scope is passed to the analyzer. This analyser decides if it should be linked, dropped or become the parent.

This can be configured just like the logger filter ie define what to match on & the result to return.

I really don't like callable as there is no way to configure it using file based config which means that the solution only works in manually instrumented applications and a change in config means a new application version.

hwu1001:
That's a good point, I hadn't thought about the similarity to samplers having access to all the same information to make similar decisions. I searched through the specification repo to see if there had been any previous discussions on this similar topic and found this one https://github.com/open-telemetry/opentelemetry-specification/issues/4814

That one sounds like a somewhat different use case, but similar idea of wanting to configure how links work when spans are created. I think between using something like a remote context analyzer on the receiving side and adding metadata on the sending side, I'd prefer the remote context analyzer on the receiving side. Mostly because the metadata on the sending side isn't that useful for the example from this PR

> For some endpoints with a public ingress there may be a `traceparent` header on the incoming request, but the request's parent trace originates from somewhere that will never be ingested in the receiving system.

Since the sender may have incorrect metadata or not include it, it seems like we would want the receiving system to be able to configure how the links are created on their spans.

> I really don't like callable as there is no way to configure it using file based config which means that the solution only works in manually instrumented applications and a change in config means a new application version.

Is it okay to include a change like this while also asking for more guidance in the specification repo? Or we'd prefer to not include this kind of change in the instrumentation packages until the specification has more information on how to handle these cases?

For handling this scenario in the meantime the only options I can think of is some configuration in the instrumentation or adding custom middleware similar to what [this comment mentions](https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/2493#discussion_r3804631848) for changing the Rack request environment's traceparent/tracestate along with storing the span context to link after the OTel instrumentation runs

thompson-tomo:
> Since the sender may have incorrect metadata or not include it, it seems like we would want the receiving system to be able to configure how the links are created on their spans.

Agree logic needs to be residing in the receiver.

> Is it okay to include a change like this while also asking for more guidance in the specification repo? Or we'd prefer to not include this kind of change in the instrumentation packages until the specification has more information on how to handle these cases?

My feeling would be to raise it to spec with details and a proposal. Actively seek feedback via an OTEP etc.

In the mean time, my gut feeling is that it would be worthwhile to replace the callable with a config properties. Maybe 3 options:
- drop_context
- link_context
- parent_context

With these options being arrays of keyvalue pairs. If span attribute key + value exists in the array use the corresponding context propagation method.

kaylareopelle:
@hwu1001 @thompson-tomo  - Checking in on this one! What are the next steps? @hwu1001, are you reaching out to the specification repo?

thompson-tomo:
My thinking is we should switch from callable to configurable options and separately raise it to spec etc.

hwu1001:
@kaylareopelle Yeah, I can reach out to the specification repo and see what they say there.




@thompson-tomo - To understand this part a bit more

> 
> In the mean time, my gut feeling is that it would be worthwhile to replace the callable with a config properties. Maybe 3 options:
> 
> - drop_context
> - link_context
> - parent_context
> 
> With these options being arrays of keyvalue pairs. If span attribute key + value exists in the array use the corresponding context propagation method.

Were you thinking it would be just comparison of values or regex for some values? Like if they had `parent_context` set to something like `[["http.request.header.x_custom_header","some-value"]]` it would just check for that attribute that would be set on the span before its created and make the link decision that way?

One other note while looking into how this might work is that if the application wants to use this option they also need to configure `allowed_request_headers` which will be recorded on the span. I could also see it tricky to configure the public request option with key-value pairs if they're using the `baggage` header and want to check for some specific information in that header

thompson-tomo:
> Were you thinking it would be just comparison of values or regex for some values?

It would be a comparison of values as defined in the spec already for other use cases ie no regex but wildcards ie * would ideally be supported.

> would just check for that attribute that would be set on the span before its created and make the link decision that way?

Yes

> One other note while looking into how this might work is that if the application wants to use this option they also need to configure allowed_request_headers which will be recorded on the span.

Correct that would be the case if wanting to use headers

> I could also see it tricky to configure the public request option with key-value pairs if they're using the baggage header and want to check for some specific information in that header

That would be solvable and could complement attributes.

In an otel manner i see the config potentially end up being

```yaml
tracer_provider:
  context_analyzer: # new option to configure it at the span level
    default:
      propagation_style: parent
    spans:
      - name: ab*
        attributes:
          - key: abc.def
            value: 123
        baggage:
          - key: abc.def
            value: 123
        config:
          propagation_style: link
  tracer_configurator/development:
    tracers:
      - name: ab*
        config:
          enabled: true
          propagation_style: link # new option to configure it at the tracer level
```

Issue #2488, opened 2026-08-01
Redis tests failing with v6 client


thompson-tomo:
## Description of the bug

Prior to the release of v6 of the redis client, our tests were passing see https://github.com/open-telemetry/opentelemetry-ruby-contrib/actions/runs/30594332813/job/91043297290 however they are now failing https://github.com/open-telemetry/opentelemetry-ruby-contrib/actions/runs/30676209952/job/91303887202.

When i look at the logs that is where the issue becomes clear that v6 has been released hence cause of issue.

![image](https://github.com/user-attachments/assets/b98d153d-25c9-424c-a509-1b69c203907b)

![image](https://github.com/user-attachments/assets/39ae7b5f-3ff9-4746-a042-491d54469470)

## Share details about your runtime

Ruby contrib ci

## Share a simplified reproduction if possible

<sub>**Tip**: [React](https://github.blog/news-insights/product-news/add-reactions-to-pull-requests-issues-and-comments/) with 👍 to help prioritize this issue. Please use comments to provide useful context, avoiding `+1` or `me too`, to help us triage it. Learn more [in our end user docs](https://opentelemetry.io/community/end-user/issue-participation/).</sub>

Issue #2053, opened 2026-03-06
Update array versions using renovate


thompson-tomo:
We should explore how we can get renovate to update ruby version arrays in the matrix as well the package versions in the  appraisals.

It doesn't need to consider removing versions but we do need to decide versioning strategy ie only minors etc.

PR #2498, opened 2026-08-07
test: Enable renovate to manage appraisal dependencies

thompson-tomo:
Progresses #2053
Progresses #2488

This enables renovate to update appraisal array of versions so that the need to use unpinned is decreased hence avoiding major bumps breaking our ci.

This only cleans up those gems which would already be managed, a future cleanup should update other instances.

The renovate rules will result in only 1 pr for all minor updates, patch updates will be added to the existing patches pr group. Major are kept seperate as there could be breaking & possibly the need to do manual steps.

### DIFF
--- a/instrumentation/grape/Appraisals
+++ b/instrumentation/grape/Appraisals
@@ -4,13 +4,8 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 
-%w[2.0.0 2.1.0 2.2.0 2.3.0 2.4.0 3.0.0 3.1.0].each do |version|
+%w[1.8.0 2.0.0 2.1.0 2.2.0 2.3.0 2.4.0 3.0.0 3.1.0].each do |version|
   appraise "grape-#{version}" do
     gem 'grape', "~> #{version}"
   end
-end
-
-appraise 'grape-1.x' do
-  gem 'rack', '~> 2.0'
-  gem 'grape', '~> 1.8'
 end
--- a/instrumentation/grape/Appraisals
+++ b/instrumentation/grape/Appraisals
@@ -4,7 +4,7 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 
-%w[2.0.0 2.1.0 2.2.0 2.3.0 2.4.0 3.0.0].each do |version|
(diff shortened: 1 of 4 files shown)


thompson-tomo:
@kaylareopelle here is a fix for the ci issue & will stop it from happening again.

instrumentation/grape/Appraisals
```diff
@@ -4,7 +4,7 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 
-%w[2.0.0 2.1.0 2.2.0 2.3.0 2.4.0 3.0.0].each do |version|
+%w[2.0.0 2.1.0 2.2.0 2.3.0 2.4.0 3.0.0 3.1.0].each do |version|
```
thompson-tomo: This is done so that the minor updates ie 3.3.0 don't remove testing on v3.0.0

instrumentation/net_ldap/Appraisals
```diff
@@ -9,12 +10,8 @@
   end
 end
 
-%w[0.20.0].each do |version|
+%w[0.20.0 0.20.0].each do |version|
```
thompson-tomo, line 13: Done so that we keep testing 0.20.0 after updates

instrumentation/redis/Appraisals
```diff
@@ -7,10 +7,8 @@ if Gem::Version.new(RUBY_VERSION) < Gem::Version.new('4')
   end
 end
 
-appraise 'redis-5.x' do
-  gem 'redis', '~> 5.0'
-end
-
-appraise 'redis-latest' do
-  gem 'redis'
+%w[5.0.0 5.4.1].each do |version|
```
thompson-tomo, line 10: The troublesome v6 will be raised as renovate update.

kaylareopelle (approved):
Thank you, @thompson-tomo! I really like this approach. It's great to know that the new releases will still get added to tests and we'll have a chance to review and fix any issues before it impacts the rest of the CI. 

To get the CI passing, I'll merge this as-is. At some point, I think it would be helpful to update the Appraisals file in the instrumentation generator to include this new format and perhaps a note about why we do it this way.

thompson-tomo:
Totally agree @kaylareopelle my plan is to get ci passing via this small pr then rollout the pattern to more gems including instrumentation generator etc.

PR #2499, opened 2026-08-07
fix(action_pack): prepend engine mount path to http.route for mounted Rails engines

mikhail-maltcev:
## Problem

When a Rails application mounts an engine (e.g. mount AdminEngine, at: '/admin'), the route_uri_pattern returned by Rails is engine-relative — it does not include the mount prefix. As a result, the http.route attribute recorded in telemetry was /items/:id instead of /admin/items/:id

## Example

```ruby
get '/items', to: 'items#index'  # GET /items
mount EngineA, at: '/foo'        # GET /foo/items
mount EngineB, at: '/bar'        # GET /bar/items 
# all of them get instrumented with http.route = /items
```

This violates the [HTTP spans specification](https://opentelemetry.io/docs/specs/semconv/http/http-spans/#http-server-span):

> http.route - The matched route template for the request. This MUST be low-cardinality and include all static path segments, with dynamic path segments represented with placeholders.


thompson-tomo (approved):

simi (approved):

kaylareopelle (approved):

kaylareopelle:
Great catch, @mikhail-maltcev! Thank you for your contribution!

PR #2504, opened 2026-08-10
chore(Docker): Alpine's MariaDB Connector/C 3.4.x requires SSL unconditionally

scub:
First off, cool project! As it is documented, the Docker setup was unable to run the test suite for the mysql2 instrumentation. This was largely driven by Alpine's MariaDB Connector/C 3.4.x requiring SSL unconditionally, completing ignoring attempts disable verification on the client or TLS on the server side.

Below is a summary of what was encountered in various configurations in an attempt to get it working with the Alpine based image:

| Attempt | Result |
|---------|--------|
| No changes |  Fails with "TLS/SSL error: self-signed certificate in certificate chain (Mysql2::Error::ConnectionError)" |
| ssl_mode: :disabled on client | Fails with "TLS/SSL error: self-signed certificate in certificate chain (Mysql2::Error::ConnectionError)" |
| Zero SSL options (fully default) + server TLS disabled | Fails with SSL is required, but the server does not support it |
| Upgraded mariadb-connector-c from 3.4.6 → 3.4.9 using Alpine's edge repo matching Debian/CI | Fails with the same self-signed cert error |
| Trust the real self-signed CA (sslca: + ssl_mode: :verify_ca) | Gets us past the cert-chain error, but fails on hostname   verification (auto-generated cert's CN doesn't match mysql the Docker service name) |

In the end, switching off of `ruby:3.3.12-alpine3.23` and onto the Debian based `3.3.12-slim` was sufficient in enabling the test suite to run "out of the box", and let me sidestep committing any sort of PKI here.

I've included some additional steps in the [contributing docs](https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/702f52ee2cacfdd74940d7ad6ae08d97ef35244c/CONTRIBUTING.md#docker-setup) that highlight the use of the `appraisal` gem to ensure all dependencies are installed before invoking the test suite, this mimics what is seen in the [test_gem workflow](https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/702f52ee2cacfdd74940d7ad6ae08d97ef35244c/.github/actions/test_gem/action.yml#L98-L106).

Anyways, Cheers!

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -110,8 +110,23 @@ To test using Docker:
    - This makes the image available locally
 4. Install dependencies for the service you want to interact with
    - `docker-compose run <service-name> bundle install`
-5. Run the tests
-   - `docker-compose run <service-name> bundle exec rake test`
+5. If the instrumentation has an `Appraisals` file (most do), generate and install
+   the appraisal gemfiles
+   - `docker-compose run <service-name> bundle exec appraisal generate`
+   - `docker-compose run <service-name> env BUNDLE_GEMFILE=gemfiles/<appraisal-name>.gemfile bundle install`
+6. Run the tests
+   - If there's no `Appraisals` file: `docker-compose run <service-name> bundle exec rake test`
+   - If there is one: `docker-compose run <service-name> env BUNDLE_GEMFILE=gemfiles/<appraisal-name>.gemfile bundle exec rake test`
+
+For example, for `opentelemetry-instrumentation-mysql2` (service `ex-instrumentation-mysql2-test`,
+appraisal name `mysql2-latest` → gemfile `gemfiles/mysql2_latest.gemfile`):
+
+```bash
+docker-compose run ex-instrumentation-mysql2-test bundle install
+docker-compose run ex-instrumentation-mysql2-test bundle exec appraisal generate
+docker-compose run ex-instrumentation-mysql2-test env BUNDLE_GEMFILE=gemfiles/mysql2_latest.gemfile bundle install
+docker-compose run ex-instrumentation-mysql2-test env BUNDLE_GEMFILE=gemfiles/mysql2_latest.gemfile bundle exec rake test
+```
 
 ## Processing and visualizing traces locally
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM ruby:3.3.12-alpine3.23@sha256:11da101dfad607c6193a921abc815c989bc9f19b43f5f686bbcc7d424298d596 as ruby
+FROM ruby:3.3.12-slim@sha256:65bce1aa8dfc8758b8fadc4629a91d4cf621784b8dfb5fed265d6ed52b0a0fb3 as ruby
 
 # Metadata
 LABEL maintainer="open-telemetry/opentelemetry-ruby-contrib"
@@ -17,29 +17,32 @@ ARG PACKAGES="\
     automake \
     bash \
     binutils \
-    build-base \
+    build-essential \
+    ca-certificates \
     coreutils  \
     execline \
     findutils \
     git \
     grep \
     less \
-    libstdc++ \
+    libstdc++6 \
     libtool \
     libxml2-dev \
-    libxslt-dev \
-    mariadb-dev \
-    sqlite-dev \
+    libxslt1-dev \
+    default-libmysqlclient-dev \
+    libsqlite3-dev \
     openssl \
-    postgresql-dev \
+    libpq-dev \
+    pkg-config \
     tzdata \
     util-linux \
     imagemagick \
     "
 # Install packages
-RUN apk update && \
-    apk upgrade && \
-    apk add --no-cache ${PACKAGES}
+RUN apt-get update && \
+    apt-get upgrade -y && \
+    apt-get install -y --no-install-recommends ${PACKAGES} && \
+    rm -rf /var/lib/apt/lists/*
 
 # Configure Bundler and PATH
 ENV LANG=C.UTF-8 \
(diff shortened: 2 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -110,8 +110,23 @@ To test using Docker:
    - This makes the image available locally
 4. Install dependencies for the service you want to interact with
    - `docker-compose run <service-name> bundle install`
-5. Run the tests
-   - `docker-compose run <service-name> bundle exec rake test`
+5. If the instrumentation has an `Appraisals` file (most do), generate and install
```
kaylareopelle, line 113: Ooh this is interesting. I generally run `docker-compose run <service-name>` and then do my bundling/appraisals, etc. in the shell that opens once I `cd` out of the example and into the root of the instrumentation. 

Is there a benefit to running the commands like this instead of in the shell? Or is it just a matter of preference?

kaylareopelle (commented):
Thanks for opening this PR, @scub! I wonder if the problem is a side-effect of #2147. Out of curiosity, if you try to use the compose file defined in `.docker/infra/mysql` do you hit the same error? 

I tested it by running the following in one terminal:
```
cd .docker/infra/mysql
docker-compose up
```

And in a second terminal:
```
cd instrumentation/mysql2
bundle
bundle exec appraisal install
bundle exec appraisal rake test
```

That infra compose file allowed me to run the tests locally. I'm not using MariaDB AFAIK, so your environment might be different.

thompson-tomo:
@kaylareopelle i don't see it as connected to #2147, the combined services were the same.

The difference is in your case you are connecting from your local machine, our tests run on ubuntu but no where do we check that it works on alpine which is what the development container uses which this would fix.

I am more inclined to see if there is another package we can need to alpine to address it & ideally adding a ci suite which runs the tests also on alpine to avoid regressions.

My thought would be maybe this was caused by the mysql version bump

thompson-tomo:
Also https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/2315 would be the basis of which we add tests. In that case we are checking the check commands work in the development container (they didn't ootb) but do now.

CONTRIBUTING.md
scub: While there's probably an argument that it's easier to script running the tests in the way this doc presents it, it definitely seems to be a matter of preference.

In this case, my aim was to keep to the existing conventions in the doc rather than supplement them by suggesting my own workflow.

scub:
> if you try to use the compose file defined in `.docker/infra/mysql` do you hit the same error?

I am able to successfully run the tests locally by directly using the MySQL instance in`.docker/infra/mysql`, though that is kind of expected. 

The trap here is that on OSX, i'm relying on `mysql-client@8.0` for the underlying connector libraries, rather than those provided by the `ruby:3.3.12-alpine3.23` image which is where the issue exists.

CONTRIBUTING.md
thompson-tomo: I also do & like @kaylareopelle approach as it enables the commands to be consistent regardless of my environment.

What i have done is https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/2522 which adjusts the structure to facilitate a common testing process. Note with appraisals2 this process becomes even simplier

scub:
> I am more inclined to see if there is another package we can need to alpine to address it 

Fair, this was the route I had initially went down. Going so far as to upgrade `mariadb-connector-c` from `3.4.6` → `3.4.9` through Alpine's edge repo to make it consistent with CI. Though the result was the same. 

Other than it's what is currently here, is there some requirement driving the use of Alpine?

> My thought would be maybe this was caused by the mysql version bump

I'm not sold this is the case, the same `mysql` version works in other environments without changes (ie: `ruby:3.3.12-slim`, and locally).

> Also https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/2315 would be the basis of which we add tests. In that case we are checking the check commands work in the development container (they didn't ootb) but do now.

I'm probably overlooking something here... Maybe you can spot it?

As of 08/17/26 1200 EST, I don't actually see the test suite being run as part of the new CI harness ([runs/31822536355/job/94838953780](https://github.com/open-telemetry/opentelemetry-ruby-contrib/actions/runs/31822536355/job/94838953780)).

All results are from the latest commit (cb58afb9a25d2071a801bf6fbea2da345834f34c) on branch `devContainers` from the fork https://github.com/thompson-tomo/opentelemetry-ruby-contrib

```bash
npm install -g @devcontainers/cli

devcontainer build --workspace-folder .; devcontainer up --workspace-folder .; devcontainer exec --workspace-folder . sh -c 'export BUNDLE_PATH=/app/vendor/bundle && npm ci --ignore-scripts && bundle install && lsconfig="-c .github/.linkspector-local_check.yml" npm run check'
...
```

I do see results from the static analysis tools, but nothing is running the test suite. In trying to do so directly, i'm encountering the same `TLS/SSL error: self-signed certificate in certificate chain` failures:

```bash
devcontainer up --workspace-folder .; devcontainer exec --workspace-folder . sh -c "cd instrumentation/mysql2; export BUNDLE_PATH=/app/vendor/bundle; bundle install; bundle exec appraisal install; bundle exec appraisal rake test"
...

31 runs, 30 assertions, 0 failures, 26 errors, 0 skips
```

thompson-tomo:
> Other than it's what is currently here, is there some requirement driving the use of Alpine?

Main one for me it being it is a lightweight environment which differs to the usual development environment.

> I'm not sold this is the case, the same mysql version works in other environments without changes (ie: ruby:3.3.12-slim, and locally).

That's because i had to make changes to the code to get it to work without adding certs as it was failing when renovate tried to upgrade it.

> As of 08/17/26 1200 EST, I don't actually see the test suite being run as part of the new CI harness (runs/31822536355/job/94838953780).

That is correct, it's not there yet but that pr introduces checks that the devcontainer can be used for dev. Once that is merged we can add the dev container as a target for the everything suite i am working on & we can also add it to the gem scoped tests.

I agree it will probs fail as it currently stand. Those additions are designed such that once it is working it remains working.

thompson-tomo:
@scub after giving it some thought, what i would like to see is:

- A new ubuntu target added to the existing Dockerfile so that we are multi targeting. I have already restructured the file in my pr #2315
- additional docker compose service for ubuntu added which mirrors what i have done for Alpine
- 2x additional devcontainers to mirror alpine
- the gem/s which do not work on Alpine have their devcontainer switched to Ubuntu and readme updated to make clear testing on Alpine is not supported. Potentially even add a guard in the test Helper to provide better guidance.
- An issue created for the limitation so it can be revisited

With this approach we are providing options to users & we have a solution which we can easily test going forward.

For me it would be great if we instead used a plain ubuntu image & instead used rvm to install/manage the ruby sdk. This way we can easily test on different runtimes ie jruby.

PR #2511, opened 2026-08-14
chore: Add rubocop extensions to libs

kaylareopelle:
The main .rubocop.yml expects these plugins but they are not in the Gemfiles. This prevented users from running `rake each:rubocop`.

In addition, remove the `rubocop-rspec` plugin. We do not use RSpec for our tests. The only reference in this repo is
for RSpec instrumentation, which is tested using Minitest.


thompson-tomo:
@kaylareopelle i had noticed that to but my intention was to update too appraisals2 and simply import the base into all the gem gemfiles. That way it makes smaller diffs & eliminates the chance of it occurring going forward.

I had been waiting for the install commands so that i had a way to test it.

kaylareopelle:
@thompson-tomo sounds good, let's try that instead. I'll close this out.

thompson-tomo:
@kaylareopelle i have opened #2512 with the linked pr's to make the switch to appraisal2 including testing of the workspace tools/rakefile.

PR #2528, opened 2026-08-23
chore: update last of appraisals

thompson-tomo:
Move to renovate managable array versions to reduce chance of ci breakages.

Note this is part 4 of the switch with this the final pr excluding an aws sdk review.

Renovate rules are:

- patches get added to the patches pr with any other patch
- Minor appraisal updates are all grouped together in 1 pr
- Major updates are raised as individual pr's as they may contain breaking changes.

This also tightens the renovate detection rule to only look at explicit gem dependencies which specify a specific version ie =0.22.0. Also max versions are restricted using < as opposed ~> x.y.z to make intent clear.

### DIFF
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -175,8 +175,8 @@
       description: "Update dependencies in Appraisals",
       managerFilePatterns: ["**/Appraisals"],
       matchStrings: [
-        "gem\\s'(?<packageName>.+?)',\\s'=\\s(?<depVersion>[0-9.]+)'",
-        "\\n\\s%w\\[([0-9.]+\\s)+(?<depVersion>[0-9.]+)\\].+\\n.+\\n\\s+gem\\s'(?<packageName>.+?)', \"~> #{version}\"",
+        "gem\\s'(?<packageName>.+?)',\\s'(?<depVersion>=\\s[0-9.]+)'",
+        "\\n\\s*?%w\\[([0-9.]+\\s)+(?<depVersion>[0-9.]+)\\].+\\n.+\\n\\s+gem\\s'(?<packageName>.+?)', \"~> #{version}\"",
       ],
       datasourceTemplate: "rubygems",
       versioningTemplate: "ruby",
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -175,7 +175,7 @@
       description: "Update dependencies in Appraisals",
       managerFilePatterns: ["**/Appraisals"],
       matchStrings: [
-        "gem\\s'(?<packageName>.+?)',\\s'(?<depVersion>.+\\s[0-9.]+)'",
+        "gem\\s'(?<packageName>.+?)',\\s'=\\s(?<depVersion>[0-9.]+)'",
         "\\n\\s%w\\[([0-9.]+\\s)+(?<depVersion>[0-9.]+)\\].+\\n.+\\n\\s+gem\\s'(?<packageName>.+?)', \"~> #{version}\"",
       ],
       datasourceTemplate: "rubygems",
--- a/instrumentation/active_storage/Appraisals
+++ b/instrumentation/active_storage/Appraisals
@@ -4,17 +4,16 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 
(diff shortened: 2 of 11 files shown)


simi:
CI failure is related to new httpx release I'll fix in other PR

simi:
https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/2535

.github/renovate.json5
```diff
@@ -175,8 +175,8 @@
       description: "Update dependencies in Appraisals",
       managerFilePatterns: ["**/Appraisals"],
       matchStrings: [
-        "gem\\s'(?<packageName>.+?)',\\s'(?<depVersion>.+\\s[0-9.]+)'",
-        "\\n\\s%w\\[([0-9.]+\\s)+(?<depVersion>[0-9.]+)\\].+\\n.+\\n\\s+gem\\s'(?<packageName>.+?)', \"~> #{version}\"",
+        "gem\\s'(?<packageName>.+?)',\\s'(?<depVersion>=\\s[0-9.]+)'",
+        "\\n\\s*?%w\\[([0-9.]+\\s)+(?<depVersion>[0-9.]+)\\].+\\n.+\\n\\s+gem\\s'(?<packageName>.+?)', \"~> #{version}\"",
```
kaylareopelle, line 179: Can you tell me a bit about which packages were being ignored with the previous regex and why?

instrumentation/gruf/Appraisals
```diff
@@ -5,23 +5,25 @@
...
+    gem 'gruf', '< 2.22.0'
   end
 end
 
-if Gem::Version.new(RUBY_VERSION) < Gem::Version.new('5')
-  appraise 'gruf-latest' do
-    gem 'gruf'
+# needs to be updated like above when ruby v5 is released
+
+%w[2.22.0 2.22.0].each do |version|
```
simi, line 25: Is it the same twice?

instrumentation/active_storage/Appraisals
```diff
@@ -18,20 +18,21 @@ if RUBY_ENGINE == 'ruby'
...
+      gem 'activerecord-jdbcsqlite3-adapter', '>= 80.0.pre1', platform: :jruby
       gem 'image_processing', '= 2.0.3'
       gem 'mini_magick', '= 5.3.3'
     end
   end
 end
 
 if RUBY_ENGINE == 'jruby'
-  %w[80.0.pre1].each do |version|
+  %w[80.0.pre1 80.0.pre1].each do |version|
```
simi: same here, seems same version twice

instrumentation/gruf/Appraisals
thompson-tomo: Yes so that when 2.22.1 or 2.23.0 is released it replaces the last item in the array.

instrumentation/active_storage/Appraisals
thompson-tomo: Same response, it is so it can be updated when new version is released.

.github/renovate.json5
thompson-tomo: It was ignoring packages which had stability tests ie old, dup & stable. Reason was the regex would only match when the gem line was indented using 1 space not 2 ie the below didn't match

```
semconv_stability.each do |mode|
  %w[2.11.0 2.12.6].each do |version|
    appraise "trilogy-#{version}-#{mode}" do
      gem 'trilogy', "~> #{version}"
    end
  end
end
```

The first change (they are independent regexes/scans) is just tightening it so that only lines such as sqlite3 below are updated

```

%w[8.0.0 8.1.0].each do |version|
  appraise "activestorage-#{version}" do
    gem 'rails', "~> #{version}"
    gem 'sqlite3', '= 2.9.6'
    gem 'image_processing', '= 2.0.3'
    gem 'mini_magick', '= 5.3.3'
  end
end
```

Lines such as 
```
gem 'sqlite3', '~> 2.9.6'
```
are ignored as renovate can't update it for custom managers.

kaylareopelle (approved):

.github/renovate.json5
kaylareopelle: Got it, thank you!

PR #2534, opened 2026-08-25
test: Improve aws service gems being tested

thompson-tomo:
This updates the aws testing targets to ensure we are testing against a range. Previously multiple appraisals were generating to the same.

With this change the aws service tests are only performed in the core or dedicated service appraisal. This means we can explicitly target specific versions of the client library.

In the process of adjusting targets coverage for Ruby 4.0 has gone from

> Line coverage: 203 / 235 (86.38%)

Up to

> Line coverage: 224 / 235 (95.31%)

Which can be attributed to now testing a version without telemetry plugin.

### DIFF
--- a/instrumentation/aws_sdk/test/test_helper.rb
+++ b/instrumentation/aws_sdk/test/test_helper.rb
@@ -36,9 +36,17 @@ end
 class TestHelper
   class << self
     def telemetry_plugin?(service)
-      m = ::Aws.const_get(service).const_get(:Client)
+      return false unless Aws.const_defined?(service)
+
+      svc = ::Aws.const_get(service)
+      return false unless svc.const_defined?(:Client)
+
+      m = svc.const_get(:Client)
+
       Aws.const_defined?('Plugins::Telemetry') &&
         m.plugins.include?(Aws::Plugins::Telemetry)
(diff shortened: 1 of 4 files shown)


instrumentation/aws_sdk/test/test_helper.rb
```diff
@@ -59,5 +67,13 @@ def semconv_stable?
     def semconv_dup?
       ENV.fetch('BUNDLE_GEMFILE', '').include?('dup')
     end
+
+    def testing_service?(service)
+      target_service?(service) || target_service?('core') || target_service?('2')
+    end
+
+    def target_service?(service)
+      ENV.fetch('BUNDLE_GEMFILE', '').include?("aws_sdk_#{service}")
```
simi: Should this be comparing with lowercased service, since that's what appraisal generates?

thompson-tomo: Have changed it to be extra safe

simi (approved):

kaylareopelle (approved):
This looks good to me. @jterapin @alextwoods do you have any feedback as component owners?

jterapin (approved):
LGTM. Thank you!

PR #2552, opened 2026-08-31
feat: Update OpenAI instrumentation for logs and CI

kaylareopelle:
This PR now makes a few changes:
1. It removes `opentelemetry-logs-api` as a dependency to the library. It was added because the instrumentation emits GenAI events using its methods. Since the Logs API is not stable, we need to avoid mixing stability within the same package. Now, if the instrumentation detects the Logs API and Logs SDK are already installed, the user will receive GenAI events. 

2. OpenAI changes won't initiate a full CI run because the gem is missing from the ci-instrumentation-full.yml workflow.

I first observed this with #2545 which only updates the openai gem, but didn't run the tests for it. 
Screenshot 2026-08-31 at 3 37 07 PM

3. JRuby is excluded from testing. OpenAI makes a "best effort" to support JRuby, but it is not included in their CI. There is currently a require issue within the gem that prevents JRuby from running on it.

### DIFF
--- a/instrumentation/openai/Appraisals
+++ b/instrumentation/openai/Appraisals
@@ -6,14 +6,16 @@
 
 %w[0.66.0 0.80.0].each do |version|
   appraise "openai-#{version}" do
-    gem 'openai', "~> #{version}"
+    # JRuby and TruffleRuby support is best effort for the openai gem
+    # as of 2026-09-02. This may change in the future and we can test it.
+    gem 'openai', "~> #{version}", platform: :mri
   end
 
   # Logs support is opt-in, based on the presence of the
   # opentelemetry-logs-api and opentelemetry-logs-sdk gems.
   # This Appraisal tests behavior with them present.
   appraise "openai-#{version}-logs" do
-    gem 'openai', "~> #{version}"
+    gem 'openai', "~> #{version}", platform: :mri
     gem 'opentelemetry-logs-api', '= 0.4.1'
     gem 'opentelemetry-logs-sdk', '= 0.6.1'
   end
--- a/instrumentation/openai/Appraisals
+++ b/instrumentation/openai/Appraisals
@@ -11,7 +11,7 @@
 
   # Logs support is opt-in, based on the presence of the
   # opentelemetry-logs-api and opentelemetry-logs-sdk gems.
-  # Run some tests with these gems present.
+  # This Appraisal tests behavior with them present.
   appraise "openai-#{version}-logs" do
     gem 'openai', "~> #{version}"
     gem 'opentelemetry-logs-api', '= 0.4.1'
--- a/instrumentation/openai/Appraisals
+++ b/instrumentation/openai/Appraisals
@@ -8,4 +8,13 @@
   appraise "openai-#{version}" do
     gem 'openai', "~> #{version}"
   end
+
+  # Logs support is opt-in, based on the presence of the
+  # opentelemetry-logs-api and opentelemetry-logs-sdk gems.
+  # Run some tests with these gems present.
+  appraise "openai-#{version}-logs" do
+    gem 'openai', "~> #{version}"
+    gem 'opentelemetry-logs-api', '= 0.4.1'
+    gem 'opentelemetry-logs-sdk', '= 0.6.1'
+  end
 end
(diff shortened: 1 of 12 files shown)


thompson-tomo (dismissed):

kaylareopelle:
Since the tests failing were logs-related, it made me look a little closer at the logs implementation. We may not want to ship this gem with the logs API as a dependency. I'll discuss more during the SIG today and post a recap here after.

instrumentation/openai/Appraisals
```diff
@@ -8,4 +8,13 @@
...
   end
+
+  # Logs support is opt-in, based on the presence of the
+  # opentelemetry-logs-api and opentelemetry-logs-sdk gems.
+  # This Appraisal tests behavior with them present.
+  appraise "openai-#{version}-logs" do
+    gem 'openai', "~> #{version}"
+    gem 'opentelemetry-logs-api', '= 0.4.1'
+    gem 'opentelemetry-logs-sdk', '= 0.6.1'
+  end
```
kaylareopelle, line 21: There is a condition in instrumentation.rb that installs the NOOP_LOGGER from the Logs API if the Logs SDK is missing. This condition is not tested because we don't have any appraisals that look at the gem when just the API is installed. 

I wasn't sure if we'd want another set of appraisals just for that case. Happy to add one or make other adjustments.

instrumentation/openai/test/opentelemetry/instrumentation/openai/patches/client_test.rb
```diff
@@ -117,34 +117,64 @@
...
-      _(user_message.body[:content]).must_equal 'Hello!'
+          _(client_span).wont_be_nil
+          _(client_span.attributes['gen_ai.operation.name']).must_equal 'chat'
+          _(client_span.attributes['gen_ai.response.finish_reasons']).must_equal ['stop']
 
-      choice = log_records.find { |r| r.event_name == 'gen_ai.choice' }
-      _(choice.attributes['gen_ai.provider.name']).must_equal 'openai'
-      _(choice.body[:message][:content]).must_equal 'Hello! How can I assist you today?'
-      _(choice.body[:finish_reason]).must_equal 'stop'
+          _(LOG_EXPORTER.emitted_log_records).must_be_empty
```
kaylareopelle, line 141: I don't know if this is a worthwhile assertion since this is hard-coded in the test helper.

xuan-cao-swi: I'd like to keep it for testing `choice_to_log_event`

instrumentation/openai/test/test_helper.rb
```diff
@@ -4,6 +4,11 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 
+if RUBY_ENGINE == 'jruby'
+  warn 'Skipping tests on JRuby: Runtime not supported'
+  exit 0
+end
```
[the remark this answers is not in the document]
kaylareopelle: This is how we're doing it in the rest of the repo.

instrumentation/openai/lib/opentelemetry/instrumentation/openai/instrumentation.rb
```diff
@@ -62,7 +62,18 @@ def require_dependencies
...
+          # Traces API, so we require users to install it themselves.
+          # If the libraries are not detected, Instrumentation.instance.logger
+          # will be nil.
+          if defined?(OpenTelemetry::Logs) && defined?(OpenTelemetry::SDK::Logs)
+            @logger = OpenTelemetry.logger_provider.logger(name: NAME, version: VERSION)
+          elsif defined?(OpenTelemetry::Logs)
+            # If the Logs SDK is missing, we can still run a NOOP_LOGGER
+            # if the API is present.
+            @logger = OpenTelemetry::Logs::LoggerProvider::NOOP_LOGGER
+          end
```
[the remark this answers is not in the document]
kaylareopelle: Removed the case.

simi (approved):

xuan-cao-swi (approved):

mwear (commented):
I have draft PRs that show how we can enable the metrics and logging APIs for instrumentation without exposing them to users. I'd be interested to get your thoughts on it as it could a low-friction approach for us to be able to use our unstable APIs in instrumentation. The overall goal would be to help us build confidence in our APIs by being to dogfood them in our instrumentation. Take a look at:

#2569 and open-telemetry/opentelemetry-ruby/pull/2384

mwear (commented):
This can ship this before open-telemetry/opentelemetry-ruby/issues/2413 is completed. The one change we need before release is updating the scope name from `opentelemetry-instrumentation-openai` to `OpenTelemetry::Instrumentation::OpenAI`. Should we do that as part of this PR?

PR #2558, opened 2026-09-02
fix(net_http): omit nil attributes on spans

### LINKED ISSUE #2541: net_http: `connect` span sets `server.address` to `nil`, emitting an SDK error on every affected connection
### Description

The `connect` span created by the Net::HTTP instrumentation sets `server.address` from `conn_address` (`Net::HTTP#address` / `proxy_address`), which is `nil` for some connections. The SDK rejects nil attribute values, so `OpenTelemetry.handle_error` fires **once per affected connection**:

```
OpenTelemetry error: invalid span attribute value type NilClass for key 'server.address' on span 'connect'
```

The span still records, so it's benign — but on HTTP-heavy paths it floods the log and buries real errors. Observed in production against `google-cloud-*` clients during a storage upload.

**Version:** `opentelemetry-instrumentation-net_http` 0.29.0 (latest); present on `main`.

**Location:** `lib/opentelemetry/instrumentation/net/http/patches/{stable,old,dup}/instrumentation.rb`, `#connect` — the attributes hash is built with `'server.address' => conn_address` and no nil guard. (The same pattern on the `request` span's `server.address => @address` is a latent case of the same thing.)

**Expected:** per the semantic conventions, an attribute whose value can't be determined should be omitted, not set to `nil` — and no SDK error.

**Suggested fix:** `compact` the attribute hash before span creation (covers all three variants, and a nil `server.port` too):

```ruby
attributes = {
  'server.address' => conn_address,
  'server.port' => conn_port
}.compact.merge!(OpenTelemetry::Common::HTTP::ClientContext.attributes)
```

hannahramadan:
The SDK rejects `nil` attribute values and logs an error message. This PR compacts attributes and quiets the errors. 

closes #2541

### DIFF
--- a/instrumentation/net_http/test/opentelemetry/instrumentation/net/http/dup/instrumentation_test.rb
+++ b/instrumentation/net_http/test/opentelemetry/instrumentation/net/http/dup/instrumentation_test.rb
@@ -389,6 +389,7 @@ describe OpenTelemetry::Instrumentation::Net::HTTP::Instrumentation do
       allow(TCPSocket).to receive(:open).and_return(fake_socket)
       Net::HTTP.new(nil, 80).send(:connect)
 
+      _(span.name).must_equal 'connect'
       _(span.attributes.key?('net.peer.name')).must_equal false
       _(span.attributes.key?('server.address')).must_equal false
     end
--- a/instrumentation/net_http/test/opentelemetry/instrumentation/net/http/dup/instrumentation_test.rb
+++ b/instrumentation/net_http/test/opentelemetry/instrumentation/net/http/dup/instrumentation_test.rb
@@ -380,6 +380,20 @@ describe OpenTelemetry::Instrumentation::Net::HTTP::Instrumentation do
       WebMock.disable_net_connect!
     end
 
+    it 'does not record a nil net.peer.name and server.address' do
+      fake_socket = Object.new
+      def fake_socket.setsockopt(*args); end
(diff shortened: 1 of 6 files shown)


xuan-cao-swi (approved):

instrumentation/net_http/test/opentelemetry/instrumentation/net/http/dup/instrumentation_test.rb
```diff
@@ -380,6 +380,19 @@ def fake_socket.close; end
...
+      fake_socket = Object.new
+      def fake_socket.setsockopt(*args); end
+      def fake_socket.close; end
+
+      # Replace the TCP socket creation with our fake socket
+      allow(TCPSocket).to receive(:open).and_return(fake_socket)
+      Net::HTTP.new(nil, 80).send(:connect)
+
+      _(span.attributes.key?('net.peer.name')).must_equal false
+      _(span.attributes.key?('server.address')).must_equal false
```
kaylareopelle, line 394: I think this assertion may have been accidentally removed in the last pass.

```suggestion
      _(span.name).must_equal 'connect'
      _(span.attributes.key?('net.peer.name')).must_equal false
      _(span.attributes.key?('server.address')).must_equal false
```

kaylareopelle (approved):

simi (approved):

thompson-tomo (approved):

PR #2562, opened 2026-09-04
ci: consolidate ruby version defs

thompson-tomo:
This simplifies the way we manage runtimes.

All CI uses the min version we support except releases. The min version is now set as the default in the project root via mise.toml & updated using renovate following existing rules.
The version should only be explicitly defined in workflows when we want to trigger tests each time it changes & we are interested the minor version.

Mise.toml was choose over .ruby-versions to avoid issues for users using rvm etc which could lead to users accidentally changing the ci ruby version.

To cater for new minor/major we ae also tracking this via a dedicated dependency file updated by a custom package manager.

Gemfile.lock has been added for when setup uses cache to improve perfomance/reproducability which will be updated weekly as part of existing pr's raised.


.ruby-version
thompson-tomo: Convert to mise.toml

.gitignore
```diff
@@ -23,9 +23,6 @@ instrumentation/active_record/db
 # IDE Settings
 **/.idea/
 
-# rbenv configuration
-.ruby-version
-
```
thompson-tomo: Add back but allow toys through

.toys/.ruby-version
thompson-tomo: Duplicate as .mise/config.ruby.toml

.toys/release/Gemfile.lock
thompson-tomo: Add to system generated files

Issue #1650, opened 2025-08-25
Remove old semantic conventions from HTTP libraries - 2/26/2026


hannahramadan:
To facilitate the migration to stable HTTP semantic conventions, issue #1116 tracked the introduction of the `OTEL_SEMCONV_STABILITY_OPT_IN` environment variable for HTTP libraries. This environment variable allowed users to opt-in to the new stable conventions, choosing if they wanted to emit the old semantic conventions, the stable one, or both. 

The agent is required to maintain this bridge for 6 months, after which it may drop the environment variable and emit only the stable HTTP and networking conventions. The final HTTP library, Net:Http, was released on August 26, 2025. **We are able to remove this environment variable and only emit stable HTTP conventions for all libraries on February 26, 2026**.

We should remove the dup and old files and folders from the following instrumentations, leaving only the stable ones. These are the PR's to 'undo':

- HTTP.rb: https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/1547
  - Related: https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/1570
- Rack: https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/1594
  - Note changes to action_pack and sinatra
- Net::Http: https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/1572
- HTTPClient: https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/1588
- Rest::Client: https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/1568
- Excon: https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/1569
- Faraday: https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/1592
- Ethon: https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/1561
- HTTPX: https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/1589

PR #2566, opened 2026-09-07
feat: only emit stable http semconv

thompson-tomo:
Progresses #1650

As over 12 months has elapsed since stable semconv was introduced behind feature flag #1547 both the old & dup options have now been removed, leaving only the stable option.


hannahramadan:
Thank you for putting all this together! It will be nice not to have these duplicated files anymore.

Before we remove the options for old and dup, we had previously discussed having schema urls in place so that users of the old semantic conventions have a way to continue reporting old attributes. The collector has a [schema processor](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/processor/schemaprocessor/README.md) that can translate stable conventions back to older ones that is based on schema url. The Ruby SDK doesn't emit schema url yet, but I'm looking at the work for it now, although I don't yet have a solid timeline. What are your thoughts on holding this PR until schema url is available?

PR #2569, opened 2026-09-09
refactor!: make metrics and logs instrumentation-only

mwear:
Draft for discussion. Contrib half of open-telemetry/opentelemetry-ruby#2384.

## Point

Instrumentation should be able to emit metrics and log records the same way it emits spans, without the unstable metrics and logs APIs becoming part of a user's public surface. Users opt in by requiring the SDKs. That lets us exercise and evolve these APIs against our instrumentation without committing users to them, and gives us much more confidence before we stabilize.

## Change

`Instrumentation::Base` gains `#meter` and `#logger` alongside `#tracer`, resolved at install
time:

```ruby
@tracer = OpenTelemetry.tracer_provider.tracer(name, version)
@meter  = OpenTelemetry::Internal.meter_provider.meter(name, version: version)
@logger = OpenTelemetry::Internal.logger_provider.logger(name: name, version: version)
```

With no SDK loaded these are no-ops, so instrumentation can create instruments and emit records and the user sees nothing until they opt in.

`instrumentation-openai`, `instrumentation-logger`, switch over to using `Base#logger.

## Notes
- Includes bugfix that has a separate PR: #2568
- `instrumentation-logger` has breaking bug-fix: its log records change instrumentation scope name from
  `opentelemetry-instrumentation-logger` to `OpenTelemetry::Instrumentation::Logger`
- This is will require careful release sequencing along with the core counterpart PR.
- Tests are expected to fail on this branch because instrumentation updates will require the changes to `Instrumentation::Base` to be released.

### DIFF
--- a/instrumentation/aws_lambda/lib/opentelemetry/instrumentation/aws_lambda/wrap.rb
+++ b/instrumentation/aws_lambda/lib/opentelemetry/instrumentation/aws_lambda/wrap.rb
@@ -60,7 +60,10 @@ module OpenTelemetry
           end
 
           OpenTelemetry.tracer_provider.force_flush(timeout: flush_timeout)
-          OpenTelemetry.meter_provider.force_flush(timeout: flush_timeout) if OpenTelemetry.respond_to?(:meter_provider)
+
+          # Only the metrics SDK supplies a meter provider that can be flushed.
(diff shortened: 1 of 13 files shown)


mwear:
Superseded by #2586

